PR #64, opened 2015-03-20
Regex parsing

anjdreas:


Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
```diff
@@ -278,11 +280,30 @@ namespace UnitsNet
         public static $className Parse(string str, IFormatProvider formatProvider = null)
         {
             if (str == null) throw new ArgumentNullException("str");
-            string[] words = str.Split(new[] {" "}, StringSplitOptions.RemoveEmptyEntries);
-            if (words.Length < 2)
+
+            var numFormat = formatProvider != null ?
+                (NumberFormatInfo) formatProvider.GetFormat(typeof (NumberFormatInfo)) :
+                (NumberFormatInfo) CultureInfo.CurrentCulture.NumberFormat.Clone();
```
anjdreas: I don't know if you considered CurrentCulture vs CurrentUICulture here? I'm not sure which one is right.

http://stackoverflow.com/a/329041/134761
> CurrentCulture is the .NET representation of the default user locale of the system. This controls default number and date formatting and the like.
> 
> CurrentUICulture refers to the default user interface language, a setting introduced in Windows 2000. This is primarily regarding the UI localization/translation part of your app.

I've used CurrentUICulture as default in the unit classes' ToString() and in UnitSystem for outputting and parsing unit abbreviations, since I figured that was the most closely related to localization and text presentation in an app. 

I don't know if CurrentCulture or CurrentUICulture is best suited for parsing quantities in this context? Normally the two are similar, but there is a reason they are separate. The scenario I can best relate to is if an app allows me to choose language, then I usually choose English. I haven't done much localization myself, but I assume it then sets the Thread.CurrentUICulture to "en-US" so those text resources are loaded etc. But being Norwegian, I prefer reading Norwegian dates and numbers so I have configured my Windows to show that. So I'm thinking Thread.CurrentCulture might remain "nb-NO" in a properly localized app, so it shows and parses numbers and dates in my preferred format. If that is the case, then CurrentCulture should be used in this line.

Thoughts?
Sorry for the big post, but I'm fuzzy on what's best practice here.

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
```diff
@@ -278,11 +280,30 @@ namespace UnitsNet
...
+
+            var numRegex = @"[\d., "                        // allows digits, dots, commas, and spaces in the number by default
+                         + numFormat.NumberGroupSeparator   // adds provided (or current) culture's group separator
+                         + numFormat.NumberDecimalSeparator // adds provided (or current) culture's group separator
+                         + @"]*\d";                         // ensures quantity ends in digit
+            var regexString = @"\s*"                                                             // ignore leading whitespace
```
anjdreas: I'd just Trim() instead of regexing this.

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
```diff
@@ -278,11 +280,30 @@ namespace UnitsNet
...
+                            + @"(?<value>[-+]?" + numRegex + @"(?:[eE][-+]?" + numRegex + @")?)" // capture Quantity input
+                            + @"\s?"                                                             // ignore whitespace (if any)
+                            + @"(?<unit>\S+)"                                                    // capture Unit (non-whitespace) input
+                            + @"\s*";                                                            // ignore trailing whitespace
```
anjdreas: Trim() here as well.

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
```diff
@@ -278,11 +280,30 @@ namespace UnitsNet
...
             if (str == null) throw new ArgumentNullException("str");
-            string[] words = str.Split(new[] {" "}, StringSplitOptions.RemoveEmptyEntries);
-            if (words.Length < 2)
+
+            var numFormat = formatProvider != null ?
+                (NumberFormatInfo) formatProvider.GetFormat(typeof (NumberFormatInfo)) :
+                (NumberFormatInfo) CultureInfo.CurrentCulture.NumberFormat.Clone();
+
+            var numRegex = @"[\d., "                        // allows digits, dots, commas, and spaces in the number by default
+                         + numFormat.NumberGroupSeparator   // adds provided (or current) culture's group separator
```
anjdreas: Good thinking. I was about to comment this on the first commit.

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
```diff
@@ -278,11 +280,30 @@ namespace UnitsNet
...
+                         + numFormat.NumberDecimalSeparator // adds provided (or current) culture's group separator
+                         + @"]*\d";                         // ensures quantity ends in digit
+            var regexString = @"\s*"                                                             // ignore leading whitespace
+                            + @"(?<value>[-+]?" + numRegex + @"(?:[eE][-+]?" + numRegex + @")?)" // capture Quantity input
```
anjdreas: I think that second numRegex is overkill? To my knowledge scientific representation only has integers, so \d+ or something should cover it.

anjdreas: I like the tidy lines with comments, good. One minor improvement of readability could be to bake everything with a string.Format() for the total string or string.Replace() placeholders like "{{numRegex}}".

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
```diff
@@ -278,11 +280,30 @@ namespace UnitsNet
...
+                            + @"\s?"                                                             // ignore whitespace (if any)
```
anjdreas: Comment should say something in the line of how this allows both "1kg" and "1 kg".

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
```diff
@@ -278,11 +280,30 @@ namespace UnitsNet
...
+            GroupCollection groups = regex.Match(str).Groups;
+
+            var valueString = groups["value"].Value;
+            var unitString = groups["unit"].Value;
+
+            if (valueString == "" || unitString == "")
             {
                 var ex = new ArgumentException(
-                    "Expected two or more words. Input string needs to be in the format \"<quantity> <unit>\".", "str");
+                    "Expected valid quantity and unit. Input string needs to be in the format \"<quantity> <unit>\".", "str");
```
anjdreas: Add info about second format option with no whitespace between quantity and unit.

Src/UnitsNet/GeneratedCode/UnitClasses/Acceleration.g.cs
```diff
@@ -21,10 +21,12 @@
 
 using System;
 using System.Globalization;
+using System.Text.RegularExpressions;
```
anjdreas: Generated code should be a separate commit.

Tests/CustomCode/ParseTests.cs
```diff
@@ -34,24 +35,47 @@ namespace UnitsNet.Tests.CustomCode
...
+        [TestCase("1km", Result=1000)]
+        [TestCase("1 km", Result = 1000)]
+        [TestCase("1e-3 km", Result = 1)]
+        [TestCase("5.5 m", Result = 5.5)]
+        [TestCase("500,005 m", Result = 500005)]
+        [TestCase(null, ExpectedExceptionName = "System.ArgumentNullException")]
+        [TestCase("1", ExpectedExceptionName = "System.ArgumentException")]
+        [TestCase("km", ExpectedExceptionName = "System.ArgumentException")]
+        [TestCase("1 kg", ExpectedExceptionName = "UnitsNet.UnitsNetException")]
+        public double ParseLengthToMetersUsEnglish(string s)
```
anjdreas: Good job on this test.

Tests/CustomCode/ParseTests.cs
```diff
@@ -34,24 +35,47 @@ namespace UnitsNet.Tests.CustomCode
...
-        public void ParseWithCultureUsingSpaceAsThousandSeparators()
+        [TestCase("5.5 m", Result = 5.5)]
+        [TestCase("500 005 m", Result = 500005)]
+        [TestCase("500.005.050,001 m", ExpectedExceptionName = "UnitsNet.UnitsNetException")] // quantity doesn't match number format
+        public double ParseWithCultureUsingSpaceAsThousandSeparators(string s)
         {
-            var numberFormat = (NumberFormatInfo) CultureInfo.InvariantCulture.NumberFormat.Clone();
+            var numberFormat = (NumberFormatInfo)CultureInfo.InvariantCulture.NumberFormat.Clone();
             numberFormat.NumberGroupSeparator = " ";
+            numberFormat.NumberDecimalSeparator = ".";
```
anjdreas: This is already the default for invariant culture, isn't it? Did you put it here for clarity?

Tests/CustomCode/ParseTests.cs
```diff
@@ -34,24 +35,47 @@ namespace UnitsNet.Tests.CustomCode
...
+
+            return Length.Parse(s, numberFormat).Meters;
+        }
+
+        /// <exception cref="UnitsNetException">Error parsing string.</exception>
+        [TestCase("5,5 m", Result = 5.5)]
+        [TestCase("500.005.050,001 m", Result = 500005050.001)]
+        [TestCase("5.5 m", Result = 55)] // dot is group separator not decimal
```
anjdreas: This case is not intuitive to me. I thought groups were only for thousands, at least for invariant culture?

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
```diff
@@ -278,11 +280,30 @@ namespace UnitsNet
...
+            var numRegex = @"[\d., "                        // allows digits, dots, commas, and spaces in the number by default
+                         + numFormat.NumberGroupSeparator   // adds provided (or current) culture's group separator
+                         + numFormat.NumberDecimalSeparator // adds provided (or current) culture's group separator
+                         + @"]*\d";                         // ensures quantity ends in digit
+            var regexString = @"\s*"                                                             // ignore leading whitespace
+                            + @"(?<value>[-+]?" + numRegex + @"(?:[eE][-+]?" + numRegex + @")?)" // capture Quantity input
+                            + @"\s?"                                                             // ignore whitespace (if any)
+                            + @"(?<unit>\S+)"                                                    // capture Unit (non-whitespace) input
+                            + @"\s*";                                                            // ignore trailing whitespace
+            var regex = new Regex(regexString);
```
anjdreas: I think we should consider caching this regex as a static field and construct it with the Compile flag for performance.

anjdreas:
Awesome work, huge improvement! Look forward to your comments on my input.

anjdreas:
Hope I didn't hog your github stats in any way by creating the pull request for you. Create your own PR if you want.

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
maherkassim: I am actually not sure about this myself, but I think we should look at what double.Parse uses by default (considering the quantity is just forwarded there). Looking at [the docs](https://msdn.microsoft.com/en-us/library/fd84bdyt(v=vs.110).aspx) it looks like [NumberFormatInfo.CurrentInfo](https://msdn.microsoft.com/en-us/library/system.globalization.numberformatinfo.currentinfo(v=vs.110).aspx) is used which is essentially using CultureInfo.CurrentCulture.NumberFormat.

Do you think it would be better to switch to NumberFormatInfo.CurrentInfo?

maherkassim:
I'm surprised that I didn't get a notification about the pull request. Just decided to take a look here after I saw your email. Also, I'm not concerned with the github stats :) This way is good too.

I didn't create the pull request intially in case you had any major comments, like if you were already working on something similar (maybe on a local branch). I will not hesitate to create it in the future :)

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
maherkassim: You're right, looking at [the docs](https://msdn.microsoft.com/en-us/library/fd84bdyt(v=vs.110).aspx) it looks like double.Parse() only accepts digits for the exponential. I will switch it to /d+ as you mentioned.

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
maherkassim: I was actually contemplating whether this should be /s? or /s*.
What would you suggest?

Tests/CustomCode/ParseTests.cs
maherkassim: True, I will switch this to a clearer test case

Tests/CustomCode/ParseTests.cs
maherkassim: Yup, added it for clarity. Should I remove it?

anjdreas: Right. Just keep it.

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
anjdreas: Then we agree. It was probably not clear from my post, but after the train of thoughts I wound up with CurrentCulture too. That would also support the scenario I described, which is neat.

anjdreas: CurrentInfo might be an improvement, I like it. Sort of makes it clear what the intention is.

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
anjdreas: I'd say 0 or 1 whitespace: /s?

Src/Scripts/Include-GenerateUnitClassSourceCode.ps1
maherkassim: Will this be affected by a change to the given formatProvider?
For example, running tests with different cultures.

maherkassim:
6662d95 updated the format strings as suggested. Is there anything that still needs to be done on the parsing?

anjdreas:
I'll take a look at this tomorrow night.

anjdreas:
Merged in to develop on de05aad.
Had to rebase onto latest develop changes, so closing manually.

anjdreas:
Nuget 3.7.1 is out.

PR #66, opened 2015-03-27
Add density unit

Tirael:
Add density unit


anjdreas:
Neat. 
I think you need to implement the base test class. It's not compiling.

Tirael:
I will correct the tests and create a new pull-request.

PR #67, opened 2015-03-27
add density unit (fix tests)

Tirael:


anjdreas:
Thanks. 
For the next time, it would be better if you updated the previous PR though, to avoid duplicate PRs.

In case you're not familiar with how to do that, this might be helpful:
http://stackoverflow.com/a/15055649/134761

anjdreas:
Nuget 3.7.1 is out.

PR #82, opened 2015-05-30
Introduce a new marker for quantities

tongbong:
Introduce a marker to differentiate IComparable (double, string...) from quantities.

That allows to type a little stronger generic APIs working on quantities.

### DIFF
--- a/UnitsNet.Tests/CustomCode/PressureFlowRateTests.cs
+++ b/UnitsNet.Tests/CustomCode/PressureFlowRateTests.cs
@@ -1,0 +1,40 @@
+﻿// Copyright © 2007 by Initial Force AS.  All rights reserved.
+// https://github.com/InitialForce/UnitsNet
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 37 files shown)


anjdreas:
We can do this, but could you list a couple of usecases where this is helpful?
I have yet not had the need to work with any type of unit.

tongbong:
I have encapsulated the UnitsNet units into some project classes to make sure that the system types could be changed somedays without breaking everything. Just a carefull approach when integrating external libs.

So there is one class for Force, one class for Length etc...
Below I removed most of implementation and just kept the BaseUnitName method.
At the moment there is no common types for Length and Force which forces the definition of  BaseUnitName in every subclass.
If Length and Force would inherit from IQuantity, then BaseUnitName could be pull up into the baseclass assuming of course that the ToString method be defined in IQuantity interface.

```
public class ForceDataType : SIDataType
    {
        private ForceUnit _baseUnit;

        public ForceDataType()
            : this(ForceUnit.Newton) {}

        public ForceDataType(ForceUnit baseUnit) 
        {
            _baseUnit = baseUnit;
        }

        public override string BaseUnitName()
        {
            return _baseUnit.ToString();
        }
    }

    public class LengthDataType : SIDataType
    {
        private LengthUnit _baseUnit;

        public LengthDataType()
            : this(LengthUnit.Millimeter) {}

        public LengthDataType(LengthUnit baseUnit) 
        {
            _baseUnit = baseUnit;
        }

        public override string BaseUnitName()
        {
            return _baseUnit.ToString();
        }
    }
```

anjdreas:
I understand your careful approach, especially for serializing I would not trust a 3rd party library not changing the unit enum values, although I try to avoid these types of breaking changes.

Your usecase seems a bit specific though, not sure if this is something anyone else will benefit from?
If I understand correctly, you basically need a common interface to expose the `string BaseUnitName()` method, which would return "Millimeter" for Length and "Newton" for Force unit.

1) Just curious, is there a reason you use Millimeter for Length? Since Meter is Units.NET's base unit for Length. Do you need to choose your own base unit name, or can you use Unit.NET's choice of base unit?

2) What do you use BaseUnitName() string for? Just trying to understand your usecase better. Is it combined with a double value anywhere, or do you only care about units?

tongbong:
For 1)
Yes, the device I am communicating with understands millimeter and below. So I really want to manipulate millimeters. Choosing the base unit would be great. It would probably mean a tradeoff between precision of unit and loss of precision. As an exemple: choosing millimeter as base unit would give a high precision when converting to centimers and microns, but when converting to killometers, well I guess it is ok to loose some precision. 
The base unit from Unit.NET are ok for us.

For 2) Some values are stored during a process. So ideally data structure would look like this:
```
class Record 
{
   IQuantity
   List<IComparable>
}
```
For each sensor, there would be a record. So in the end the system ends up with a list of records.
One feature is to push those data into a report (csv file in this case) like this:
"s","mm"
0, 10
1, 15
2, 10

anjdreas:
I still don't fully understand why you need IQuantity in the Record class though. Are you using it for anything else than to call BaseUnitName()? Can't you just store a `string Unit` property/field in Record instead?

anjdreas:
Just checking in on this, are you still pursuing this idea? I currently don't have a good enough understanding of what you are trying to accomplish to say whether this is a good idea or not.

anjdreas:
Just a heads up, I am about to delete develop and stable branches, and returning to using a master branch. It seems pull requests can't be edited to change target branch, so you will have to create a new PR targeting master instead.

PR #83, opened 2015-06-03
Develop massflowrate

tongbong:
Add new quantity MassFLowRate with 2 units: kg/s and N/s.

See http://en.wikipedia.org/wiki/Mass_flow_rate


UnitsNet/Scripts/UnitDefinitions/MassFlowRate.json
```diff
@@ -0,0 +1,31 @@
...
+            "FromBaseToUnitFunc": "x",
+            "Localization": [
+                {
+                    "Culture": "en-US",
+                    "Abbreviations": ["kg/s"]
+                }
+            ]
+        },        
+        {
+            "SingularName": "NewtonPerSecond",
```
anjdreas: Are you sure NewtonPerSecond is a commonly used unit?
Newton is not a mass unit. I also don't see this unit here:

http://www.onlineconversion.com/flow_rate_mass.htm

My recommendation is to convert from newtons to kilograms first, then use kilogram per second as the representation. We already have Force.FromKilogramForce() or similar, not sure if we have the other way around, Mass.FromNewtons(), but should be trivial to add.

tongbong:
Yes, my mistake. Newton is a force and as such the quantity should be named ForceFlowRate.
I'll push the fix in a moment.

tongbong:
Okay, I pushed the changes.
But as you mentioned earlier, I am wondering if it is the best place to store those units as they are listed in SI.
What do you think?

UnitsNet/Scripts/UnitDefinitions/ForceFlowRate.json
```diff
@@ -0,0 +1,19 @@
+﻿{
+    "Name": "ForceFlowRate",
```
anjdreas: I'm sorry, I really do appreciate your contributions, but this seems a lot like the PressureFlowRate case.

I can't seem to find much info on this unit on google and I'm neither familiar with it nor have an intuitive understanding of what force flow rate means.

Do you consider this unit more widely used or accepted than pressure flow rate?

However, MassFlowRate, which you originally added, does seem like a thing. So KilogramPerSecond, PoundPerSecond, TonnePerSecond and so on all seem like widely used units and I have an intuitive understanding of what it means. 

Did you intentionally remove MassFlowRate or is that something you still want to add?

tongbong:
The MassFlowRate was a misunderstanding from me.
I do not need it right now but yes it does seam like it is used.

The ForceFlowRate is clearly not widely used. As you, I did not find any literature on this topic.
So I would understand that it is not a contribution to the public.

anjdreas:
Ok, closing this for now then, by same reasoning as in #84 .

PR #84, opened 2015-06-03
Add new quantity PressureFlowRate and 2 units: Pa/s and Pa/m

tongbong:


anjdreas:
This looks good, except I'm not familiar with this unit and I don't really find any sources on it. Google search doesn't return much with that exact naming, and no wiki entry.

What does even pressure flow or rate of pressure flow mean? It is not intuitive to me.

I'm just trying to make sure any units that go into Units.NET are widely used/accepted units.

tongbong:
Indeed it does not seam to be known at all and I reacted the same way as you.
I am working with a machine that takes some Pa/s as a unit...
However it still remains a quantity with units.
Now I would understand if it is a no go to push it in the mainstream.

anjdreas:
Thanks. Then I propose not adding this to the main library, at least until a good usecase for it is presented. 

You should still be able to use this class for your own needs, and if you need to parse strings or call ToString() you can extend UnitSystem with the new enum type and the abbreviation strings.

PR #86, opened 2015-06-10
This closes anjdreas/UnitsNet#85 - add more units for energy.

FabioCZ:


UnitsNet.Tests/CustomCode/EnergyTests.cs
```diff
@@ -30,5 +30,55 @@ public class EnergyTests : EnergyTestsBase
         protected override double JoulesInOneJoule { get { return 1; } }
         protected override double KilojoulesInOneJoule { get { return 1E-3; } }
         protected override double MegajoulesInOneJoule { get { return 1E-6; } }
+
+        protected override double BritishThermalUnitsInOneJoule
+        {
+            get { return 1/1055.05585262; }
```
anjdreas: I think you're kind of cheating here, since you basically just copied in the implementation into the test. The idea is to double check that you got it right, by looking up the correct answer for these particular conversions.

For example, I type this in google:
`british termal units in 1 joule` I get `1 joule = 0.00094781712 british thermal units`, so I would put that number in here.

anjdreas:
Other than that, looks good to me. If you can just update the test values and make sure all tests run, I'll merge this in. Leave a comment after you are done so I am notified.

Sorry for not getting to this sooner, it slipped my attention.

Cheers

FabioCZ:
Thanks for your remarks, I pushed another commit with hopefully more appropriate values for the unit tests.

anjdreas:
I just noticed, it says 'unknown' on author of your commits. Not a big deal, but you probably want to check your git config and make sure you have name and email configured.

https://git-scm.com/book/en/v2/Getting-Started-First-Time-Git-Setup

anjdreas:
Actually, email is set, but it's missing your name, or you have explicitly set it to 'unknown'.

FabioCZ:
I had to switch to a Windows machine (I am normally on OS X) to work on this project, and looks like I didn't configure it properly. Anyways, thanks for the merge, and thanks again for a great library.

anjdreas:
You welcome. I released the nuget yesterday in case you didn't notice.

PR #88, opened 2015-07-07
Some units added

dtsvetkov:
Added ValvePosition, ValveDriveSpeed, RotationalSpeedChangeRate, PressureChangeRate. Changed abbreviations with prefixes for Pressure unit (ru-ru culture).


dtsvetkov:
Hello!

Let's start from the end of the list.

ValveDriveSpeed.json
We discussed that point in our team once again and decided we don't need it actually. So, I can remove it at all.

ValvePosition.json
I agree, it looks like the Ratio, but we need also "complement" units. And I'm confused about their abbreviations. Could you advise me?

RotationalSpeedChangeRate.json
We think RotationalAcceleration sounds better. I can rename the unit. Ok?

PressureChangeRate.json
As I understand right, it's OK for you and no changes needed in that unit. Right?

anjdreas:
ValveDriveSpeed
Ok by me to delete this from the PR.

ValvePosition
It seems the term "complement" is used to represent the "other" part of a fraction/ratio/percentage, so lets go with that. If you see in Ratio.json, I specify empty string for `DecimalFraction` as abbreviation. I would expect the same for `ComplementDecimalFraction`, if we are to name it that. The only conflict in that, is that we can't unambiguously parse the two with `Ratio.Parse("0.5")`, but that makes sense anyway I think. For that case, I find it reasonable to do a `double.Parse()` first, then create it with `Ratio.FromDecimalFraction()` or similar for complement units.

So to summarize
* Add `Ratio.ComplementDecimalFraction` with empty string as abbreviation
* Add `Ratio.ComplementPercent` with `%` as abbreviation.

Rename to RotationalAcceleration is fine by me.

PressureChangeRate is fine as it is, however there are many other pressure units in Pressure.json. I leave it up to you to decide whether to add more or not.

anjdreas:
Leave a comment when you have updated the PR so I am notified. Thanks!

anjdreas:
Just checking in on this, still working on it?

dtsvetkov:
Unfortunately, we got some more prioritized tasks, so we had to postpone working on this.
We plan to resume working on this soon.

anjdreas:
Thanks, no problem. Just checking in.

dtsvetkov:
Hello Andreas!
I have updated the PR. 
I added Ratio.ComplementDecimalFraction and Ratio.ComplementPercent units as you advised.
Also after some investigation we decided we don't need RotationalSpeedChangeRate (or RotationalAcceleration) unit any more. So it was deleted as well as ValveDriveSpeed and ValvePositionUnit.

Please take a look. Thanks!

UnitsNet/Scripts/UnitDefinitions/Ratio.json
```diff
@@ -24,6 +36,26 @@
...
+        },
+	{
+            "SingularName": "ComplementPercent",
+            "PluralName": "ComplementPercent",
+            "FromUnitToBaseFunc": "1-x/1e2",
+            "FromBaseToUnitFunc": "100-x*1e2",
+            "Localization": [
+                {
+                    "Culture": "en-US",
+                    "Abbreviations": ["%"]
```
anjdreas: Hm, I wonder how `Ratio.Parse("5 %")` would work. The case with multiple units sharing the exact same abbreviation is not explored, at least not by me.

I would expect the above statement to equal `Ratio.FromPercent(5)`, so how do we ensure that?
The easiest fix is to simply change ComplementPercent abbreviation to `% comp.` or similar. 

Thoughts?

UnitsNet/Scripts/UnitDefinitions/Ratio.json
```diff
@@ -15,6 +15,18 @@
...
         },
+	{
+            "SingularName": "ComplementDecimalFraction",
+            "PluralName": "ComplementDecimalFractions",
+            "FromUnitToBaseFunc": "1-x",
+            "FromBaseToUnitFunc": "1-x",
+            "Localization": [
+                {
+                    "Culture": "en-US",
+                    "Abbreviations": [""]
```
anjdreas: Same here, as below. Duplicate abbreviation. Need a strategy here and add some tests to verify the right unit is actually parsed if we do need to share abbreviations.

anjdreas:
It seems we don't auto-generate any `Parse()` tests, which would probably reveal that either `Percent` or `ComplementPercent` does not parse into the right amount, same for `DecimalFraction` and `ComplementDecimalFraction`.

anjdreas:
So I see from previous discussion that I argued `Parse("5%")` should always parse as Percent, not ComplementPercent, which I still think is fair. Same for DecimalFraction. The problem is just to ensure that this in fact happens, and not the other way around.

Scratch my previous comment about changing abbreviation to "% comp.", I don't think that makes sense. However, I just came to realize something. Do we really need a complementary unit, or would it be sufficient to have a `FromComplement` and `ToComplement` methods instead? So the unit of representation is still just `Percent` or `DecimalFraction`, but you have convenience methods/properties for converting to the complement.

Example:
```
These would be equal:
Ratio.FromPercent(30);
Ratio.FromComplementPercent(70);
Ratio.FromPercent(70).ToComplement();
Ratio.FromDecimalFraction(0.3);
Ratio.FromComplementDecimalFraction(0.7);
Ratio.FromDecimalFraction(0.7).ToComplement();
```

dtsvetkov:
Hello Andreas!
After some thinking we decided to perform all conversions with complement units on our side.
So we don't need to include them to the library.
If no objections, I will remove Ratio.ComplementDecimalFraction and Ratio.ComplementPercent from the PR.
Sorry for bothering you.

anjdreas:
Ok, sounds good to me.

dtsvetkov:
I've updated the PR, please take a look. Thanks!

anjdreas:
This looks good to me now, thanks!

anjdreas:
Nuget 3.21 is now out.

Issue #92, opened 2015-07-30
Json.NET conversion


neutmute:
Hello - I tried persisting the `Area` type with Newtonsoft's Json.NET but by default it makes a big nested object with all the internal types like this 

	{
		SquareCentimeters: 200000.0,
		SquareDecimeters: 2000.0,
		SquareFeet: 215.27830102364831,
		SquareInches: 31000.062000124002,
		SquareKilometers: 2E-05,
		SquareMeters: 20.0,
		SquareMiles: 7.7220077220077222E-06,
		SquareMillimeters: 20000000.0,
		SquareYards: 23.919811224849816
	}

when what I need is just the simple m^2 value of 20.0 (Neo4j database can't persist the nested object).

I wrote a custom converter and it works nicely
https://github.com/neutmute/Neo4jClient.Extension/blob/cskardon-T-3/Neo4jClient.Extension.Test/CustomConverters/AreaJsonConverter.cs

This looks like it could be generated with some magical powershell scripts.
I'm not sure if my problem is an edge case or of wider value.

Would you consider a PR with an extra project added: `UnitsNet.JsonConverters` say?

neutmute:
Actually now I'm wondering if this could be done by a single `UnitJsonConverter<T>` that takes an expression tree as a parameter. Users could configure it at app startup to customise persistence to their needs. It would be a separate project with only one class though it that idea works (to avoid introducing Json.NET dependency into main)

anjdreas:
From the start I did not want to take on serialization. There are so many pitfalls and easy to make breaking changes, so I felt safer leaving this up to client code. I'm open to discuss a solution here though, as I'm sure many will stumble across the same problem.

The most robust method would be to serialize both value and unit, so that in case the default/base unit changes at any point, we have a better chance of still parsing it. Unit abbreviations like "m" and "kg" are pretty fixed, but can be ambiguous. For instance, several Ratio units have no abbreviation. Unit enum names are unique, but it is hard to guarantee they will never be renamed. So I'm not sure what would be a good choice for serializing the unit.

We could add a JsonConverter, but as you say, I would not want to introduce a nuget dependency. With the approach in your second comment, we still would introduce that dependency, won't we? By inheriting from JsonConverter.

neutmute:
Agree on all points and the answer turned out to be quite easy. Using a generic, I define my own converter like this:

    public class AreaJsonConverter : UnitsNetJsonConverter<Area>
    {
        public AreaJsonConverter()
[...]

PR #96, opened 2015-08-03
Make the units serializable via the use of data contract

tongbong:
Datacontract is part of dotnet portable, so it gives the ability to serialize in xml unitsnet objects.

It may be the answer to json serialization because datacontract offers a datacontractjsonserializer class.

Allthough I have tested it locally, the actual unit tests do not cover this use case.
We would need to create integration unit tests for that purpose.

### DIFF
--- a/UnitsNet/GeneratedCode/UnitClasses/Acceleration.g.cs
+++ b/UnitsNet/GeneratedCode/UnitClasses/Acceleration.g.cs
@@ -24,6 +24,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Text.RegularExpressions;
 using System.Linq;
+using System.Runtime.Serialization;
 using JetBrains.Annotations;
 using UnitsNet.Units;
 
@@ -35,11 +36,13 @@ namespace UnitsNet
     ///     Acceleration, in physics, is the rate at which the velocity of an object changes over time. An object's acceleration is the net result of any and all forces acting on the object, as described by Newton's Second Law. The SI unit for acceleration is the Meter per second squared (m/s2). Accelerations are vector quantities (they have magnitude and direction) and add according to the parallelogram law. As a vector, the calculated net force is equal to the product of the object's mass (a scalar quantity) and the acceleration.
     /// </summary>
     // ReSharper disable once PartialTypeWithSinglePart
+	[DataContract]
     public partial struct Acceleration : IComparable, IComparable<Acceleration>
     {
         /// <summary>
         ///     Base unit of Acceleration.
         /// </summary>
+		[DataMember]
         private readonly double _meterPerSecondSquared;
 
         public Acceleration(double meterpersecondsquared) : this()
--- a/UnitsNet/GeneratedCode/UnitClasses/AmplitudeRatio.g.cs
+++ b/UnitsNet/GeneratedCode/UnitClasses/AmplitudeRatio.g.cs
@@ -24,6 +24,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Text.RegularExpressions;
 using System.Linq;
+using System.Runtime.Serialization;
 using JetBrains.Annotations;
 using UnitsNet.Units;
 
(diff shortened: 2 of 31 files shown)


tongbong:
Might be a solution for anjdreas/master#92

tongbong:
```    [Test]
        public void Foo2()
        {
            Length l1 = Length.From(10, LengthUnit.Centimeter);
            DataContractSerializer s = new DataContractSerializer(typeof(Length));
            string filePath = Environment.CurrentDirectory + "Length.xml";
            using (var stream = File.OpenWrite(filePath))
            {
                s.WriteObject(stream, l1);
            }
            Length l2;
            using (var stream = File.OpenRead(filePath))
            {
                l2 = (Length)s.ReadObject(stream);
            }
            Assert.AreEqual(l1, l2);
        }
```

The resulting xml file looks like this:
<Length xmlns="http://schemas.datacontract.org/2004/07/UnitsNet" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"><_meters>0.1</_meters></Length>


Same test but with json:
```
        [Test]
        public void Foo3()
        {
            Length l1 = Length.From(10, LengthUnit.Centimeter);
            DataContractJsonSerializer s = new DataContractJsonSerializer(typeof(Length));
            string filePath = Environment.CurrentDirectory + "Length.json";
            using (var stream = File.OpenWrite(filePath))
            {
                s.WriteObject(stream, l1);
            }
            Length l2;
            using (var stream = File.OpenRead(filePath))
            {
                l2 = (Length)s.ReadObject(stream);
            }
            Assert.AreEqual(l1, l2);
        }
```
Resulting in a json file looking like this:
{"_meters":0.1}

It does not seem to be good enough for json...

anjdreas:
Yes, this should work, but as per the discussion in #92 I am still a bit worried about breaking data compatibility at some point, say we decide to rename the private field to singular, or change the base unit to a different unit, or whatever. We could simply decide that, any breaking changes to units are simply not allowed, unless we bump the major version of the nuget or similar. I'd love some thoughts and feedback from more people on this though, I think it's easy to make a mistake here that will shoot someone horribly in the foot later.

anjdreas:
If we do add support for serialization, we at least need to document what guarantees we put in that implementation, so anyone can make an informed choice on whether to use the built-in method or their own.

anjdreas:
What are your thoughts on this? Should we leave it at the examples or implement data contract?

tongbong:
I checked out #92, and I find it very elegant but to me it does not answer the question.
See I believe the serialized data should be auto descriptive meaning that the serialized data should contain everything needed for its deserialization. As a result two clients could serialize/deserialize despite using different libs and without any configuration.

If the above serialization data are not good enough, how about those ones:
- XML from data contract
```<Length xmlns="http://schemas.datacontract.org/2004/07/UnitsNet" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"><Value>0.1m</Value></Length>```
- JSon from data contract
```
{"Value":"0.1m"}
```

I think I am not wrong saying that an abbreviation is unique among the abbreviations from a unit type.
So in the Value property setter, we could imagine having that kind of parsing code:
```
public string Value
       {
            set 
           {
                Length l = Length.Parse(valueAsString, CultureInfo.InvariantCulture);
                _meters = l.Meters;
           }
      }
```

And the power of data contract makes possible to set the visibility of such property to internal, protected or private.

The client can choose wether he wants to use the xml data contract serializer or the json data contract serializer.

I will try later today or this week to propose such an exemple as in #92.

anjdreas:
See #100 and #99 .

anjdreas:
For your info, #100 and #101 are now closed and merged. Nugets UnitsNet.3.17.0 and UnitsNet.Serialization.JsonNet.1.0.0-alpha2 are out and support JSON serialization. 

Are you still looking to pursue datacontract serialization or will this satisfy your needs?

tongbong:
Sorry I could not answer earlier, yes I will keep looking into it because my use case it to save quantities into some configuration files.

anjdreas:
Just checking up on this, still planning to move forward with this PR?

tongbong:
No plan to move forward anymore.
PR can be closed.

PR #97, opened 2015-08-03
This commit adds two more units of frequency: Beats Per Minute and Be…

FabioCZ:
…ats per Hour. Unit tested and passed.

Needed some more units. Added and tested them just like in my last contribution. 

Thanks again for a great library :)

### DIFF
--- a/UnitsNet.Tests/CustomCode/FrequencyTests.cs
+++ b/UnitsNet.Tests/CustomCode/FrequencyTests.cs
@@ -19,6 +19,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+using System;
+
 namespace UnitsNet.Tests.CustomCode
 {
     public class FrequencyTests : FrequencyTestsBase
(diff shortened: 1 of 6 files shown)


anjdreas:
Thanks yourself, almost happy to pull this in, but I just realized there are probably many variations of beats per minute and the like. 

https://en.wikipedia.org/wiki/Category:Units_of_frequency

According to this, I'm thinking maybe beat is too domain specific. How about counts per minute? Or ticks?

FabioCZ:
I think "ticks per minute" might be easy to confuse with .NET's TimeSpan.Tick (where a tick == 100 ns). 

"Counts per minute" sounds rather specific to radiation: https://en.wikipedia.org/wiki/Counts_per_minute

I think "Cycles per minute" would be good naming.

anjdreas:
I agree with ticks, I had the same association. I didn't have any association with counts per minute as being specific to radiation, but I'm not familiar with that domain. Cycles per minute is good by me.

FabioCZ:
Pushed a new commit with the units renamed, hopefully everything should be good now :)

anjdreas:
Cool, thanks. Nuget 3.16 is now out.

PR #100, opened 2015-08-14
Json.NET serialization converters

bretthysuik:
- Uses PowerShell scripts to auto-generate the source code for each `JsonConverter` class.
- Uses slightly modified code from #92 for the base class for all converters.
- JSON is output in the following format:
```js
{
  "Value": <value>
  "Unit": <unit_enum>
}
```
- For example:
```js
{
  "Value": 33.33,
  "Unit": "MassUnit.Kilogram"
}
```

Work-in-progress outstanding items:
- Update build scripts to auto-update the library version for both `UnitsNet` and `UnitsNet.Serialization.JsonNet`. I don't know how this works though...
- Potentially add projects for .NET 4.5.1 and PCL 4.0 and PCL 4.5? Do we actually need more projects targetting different frameworks? I don't even understand why there's a package for .NET 4.5.1 since that package doesn't use any v4.5.1 features as far as I know...


anjdreas:
Other than that, I think this is shaping up pretty well.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverterBase.cs
```diff
@@ -0,0 +1,130 @@
...
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            object value = null;
+            string enumStr = null;
+            while (reader.Read())
+            {
+                if (value != null && enumStr != null)
+                    break;
+
+                switch (reader.TokenType)
```
anjdreas: This deserializer seems fragile to me. If I understand correctly, it assumes any JSON float/int is the value and any JSON string is the unit. I believe it would be best to check that the JSON property names are "Value" and "Unit" too?

How is the JSON deserializer supposed to work for Data Transfer Objects like this?
```
{
"Name": "Andreas",
"Height": {
  "Value": 1.84,
  "Unit": "LengthUnit.Meter"
  }
}
```

Do we need to tag the `Height` C# property with a LengthAttribute or similar?

anjdreas: Or is it simply sufficient that the type we are deserializing to has a `Height` property with type `Mass`?

bretthysuik: The Json.NET framework is smart enough to choose the correct converter based on the return value from the `CanConvert()` method. The deserializer iterates through the collection of converters to try and find one that will correctly deserialize each property in the JSON.

So in your above example, there is no `DtoJsonConverter` so the individual properties of the JSON are inspected. The `Name` property can be auto deserialized as a string, and the `Height` property is deserialized using `LengthJsonConverter`. There's no need for `LengthAttribute`.

bretthysuik: > I believe it would be best to check that the JSON property names are "Value" and "Unit" too?

I don't think this is strictly necessary, but it wouldn't hurt to include it.

bretthysuik:
OK, I've updated `UnitsNetJsonConverterBase` to use reflection to build the <Unit>.From(<value>, <unit_enum>) call.

bretthysuik:
At this point, you should be able to execute `\Build\build.bat` to generate a NuGet package.

The only item remaining is updating the build scripts to automate the library versions for both `UnitsNet` and `UnitsNet.Serialization.JsonNet`. I don't know how this works so I haven't made any changes.

anjdreas:
Thanks, I'll take a closer look at this soon, this weekend at the latest.

UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonConverterBaseTests.cs
```diff
@@ -0,0 +1,113 @@
...
+        private UnitsNetJsonConverterBase<Ratio> _ratioJsonConverter;
+
+        [SetUp]
+        public void Setup()
+        {
+            _massJsonConverter = new MassJsonConverter();
+            _ratioJsonConverter = new RatioJsonConverter();
+
+            _jsonSerializerSettings = new JsonSerializerSettings() { Formatting = Formatting.Indented };
+            _jsonSerializerSettings.Converters.Add(_massJsonConverter);
```
anjdreas: It would be ideal if we could do with just a single unitsnet converter here, but I'm fine with this for now. We can always take a stab at that in a future PR.

anjdreas:
This looks good to me, great work. A couple of minor things, but I'll just fix them after the merge. 

As for automating library versions, the steps for releasing a nuget involves manually editing the .nuspec files and the version.txt at the root. The buildserver will then patch all the AssemblyInfo.cs files by the version in version.txt, so it's easiest if we use the same version on both libraries. I think it makes sense that the two versions are synchronized anyhow, since it depends on the main lib and any new units added to it.

PR #101, opened 2015-08-23
Single JsonConverter

anjdreas:
Continuation of #100, replacing all generated json converters with a single converter using reflection and a new `BaseUnit` property per unit type to serialize/deserialize.

Request for comments:
* We might run into issues where very large/small values are being serialized and then deserialized using the `From()` method, which takes a double. We already know from test cases that huge values like `Information.Exabyte(100)`, using `Bit` as its base type, cannot use `double` as its base value type due to precision errors in conversions. This issue is already present in #100, so nothing new in this PR. Should we add a third JSON property to hold the type of the base value? Such as a property "ValueType"  with any of the values "double", "long" or "decimal" ? This way we have the possibility of constructing `Information` with its base type `decimal`, although that also requires us to either overload or change the `From()` method to also take a `decimal`.
* I decided not to introduce an interface with units having a `BaseUnit` property returning type `Enum`, since I don't see it adding much value.


anjdreas:
@bretthysuik would like some input from you since you authored #100.

anjdreas:
Added two tests for large Information values, it worked for surprisingly large values. Up to 1E+9 exabytes, represented as bits in double. It seems the issue with precision is mostly during roundtrip conversions between units as seen in existing test cases.

anjdreas:
From this I'm leaning towards calling it "good enough" and not bother with serializing the underlying type of the unit, but would like some input on this.

bretthysuik:
Looks good to me. The only issue I have is setting the NuGet package version to 3.16.0 in 979fa89 (I think it should be 1.0.0 since it's an initial release of a brand new package).

anjdreas:
I see your point. In fact, since we now use reflection, this package is no longer bound to any specific version of the main lib, as long as the naming conventions and namespaces still match. That is a good reason to not sync the two versions, since it leaves us the ability to do breaking changes in the serialization separately from changes in the main lib. I'll see if I can massage the build server to do separate versioning based on .nuspec and drop the version.txt file.

bretthysuik:
[FAKE](http://fsharp.github.io/FAKE/index.html) may be of interest to you in this situation since it can do everything being done in the current builds scripts, but in a simpler and more standardized way. It would be a pretty big change though. Maybe something to consider in the future.

Here's a [getting started tutorial](http://fsharp.github.io/FAKE/gettingstarted.html) if you're interested.

anjdreas:
Thanks, I'm kind of pressed on time as it is, but I've heard good things about FAKE and intend to check it out at some point.

anjdreas:
I think I got it working now. Versioning of AssemblyInfo.cs files based on respective .nuspec files.

anjdreas:
1.0.0-alpha nuget is now out and *should* work with current main lib, but it would be great if you guys could give it a spin.

bretthysuik:
I'll give it a try and report back with results.

bretthysuik:
A new version of the main package needs to be released before the serialization package can be used (the current version, 3.16.0, doesn't contain the `BaseUnit` property required by the JSON converter).. The serialization nuspec needs to be updated to require UnitsNET.dll v3.17.0 or later

anjdreas:
Ah, that's right. My bad.

anjdreas:
Will redeploy later today.

anjdreas:
Both nugets redeployed, with updated dependency.
3.17.0 and 1.0.0-alpha.

bretthysuik:
I've confirmed the new packages work in a fairly complex serialization scenario. Very much appreciated @anjdreas!

anjdreas:
Very cool, no thank you @bretthysuik. You did the legwork.

PR #102, opened 2015-08-28
Add more acceleration units through the use of prefixes

tongbong:


anjdreas:
Thanks, nuget 3.18.0 released.

PR #106, opened 2015-09-22
Fixed bugs in UnitsNetJsonConverter related to nullable types

### LINKED ISSUE #105: Serializing/deserializing nullable types fails in certain situations
The following scenarios result in incorrectly serialized data or failures when deserializing:

- Serializing a non-null nullable value, but *only when nested within another object*
- Deserializing a non-null nullable value as a nullable value

I know this is quite confusing so here are some unit tests that fail:

```csharp
// Serialization
[Test]
public void NonNullNullableValueNestedInObject_ExpectJsonUnaffected()
{
    var testObj = new TestObj()
        {
            NullableFrequency = Frequency.FromHertz(10),
            NonNullableFrequency = Frequency.FromHertz(10)
        };
    string expectedJson = "{\r\n  \"NullableFrequency\": {\r\n  \"Unit\": \"FrequencyUnit.Hertz\",\r\n  \"Value\": 10.0\r\r}," +
                            "{\r\n  \"NonNullableFrequency\": {\r\n  \"Unit\": \"FrequencyUnit.Hertz\",\r\n  \"Value\": 10.0\r\r}\r\n}";

    string json = SerializeObject(testObj);

    Assert.That(json, Is.EqualTo(expectedJson));
}

internal class TestObj
{
    public Frequency? NullableFrequency { get; set; }
    public Frequency NonNullableFrequency { get; set; }
}


// Deserialization
[Test]
public void NonNullNullableValue_ExpectValueDeserializedCorrectly()
{
    Mass? nullableMass = Mass.FromKilograms(10);
    string json = SerializeObject(nullableMass);

    var deserializedNullableMass = DeserializeObject<Mass?>(json);

    // Fails because the corresponding scenario on the serializer side fails
    Assert.That(deserializedNullableMass.Value, Is.EqualTo(nullableMass.Value));
}

[Test]
public void NonNullNullableValueNestedInObject_ExpectValueDeserializedCorrectly()
{
    var testObj = new TestObj()
        {
            NullableFrequency = Frequency.FromHertz(10),
            NonNullableFrequency = Frequency.FromHertz(10)
        };
    string json = SerializeObject(testObj);

    var deserializedTestObj = DeserializeObject<TestObj>(json);

    Assert.That(deserializedTestObj.NullableFrequency, Is.EqualTo(testObj.NullableFrequency));
}
```

bretthysuik:
Fixes #105

### DIFF
--- a/UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonConverterTests.cs
+++ b/UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonConverterTests.cs
@@ -19,6 +19,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+using System;
 using Newtonsoft.Json;
 using NUnit.Framework;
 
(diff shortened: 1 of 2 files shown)


anjdreas:
Other than that, awesome work. Looks good to me, just answer my comment and modify if you agree, then we can merge this in.

anjdreas:
It's reasons like this that I still think we should keep the nuget in prerelease state a little while longer. I'm not actively using this feature myself so I won't be contributing a lot to the testing.

bretthysuik:
I'm fine with keeping it in pre-release for a while longer.

Issue #89, opened 2015-07-07
Represent rate of change more generically


anjdreas:
Similar to discussion on dimensions in #73, I speculate if there is a need for a more generic way of expressing rate of change for any type of unit. Currently we have the following rate of change units:

Flow (volume over time)
Speed (distance change over time)
Acceleration (speed change over time)
RotationalSpeed (angle change over time)

In #88 there are some further suggestions to units, such as 
PressureChangeRate (pressure change over time)
RotationalSpeedChangeRate (angular speed change over time)
ValveDriveSpeed (fraction change over time)

eriove:
Not sure if I should put this in it's own issue...

Have you considered operator overloading to convert between different units (i.e. dividing with a `Speed` with a `TimeSpan` should return an acceleration)? It opens up a large amount of new work with with defining all the different relations but it would give you compile time dimensional analysis.

InitialForce-TeamCity:
I have not personally considered operator overloading so far, mostly because I have simply not had the need for my own purposes, but I'm leaning towards *why not*? I don't see this easily automated with the code generation powershell scripts though, but all generated unit classes are partial so adding operator overloads to say `Custom Code\Length.extra.cs` should be straight forward.

If you want to take a stab at this, please feel free to start a PR to move that discussion there and start implementing some typical usecases.

anjdreas:
I have not personally considered operator overloading so far, mostly because I have simply not had the need for my own purposes, but I'm leaning towards why not? I don't see this easily automated with the code generation powershell scripts though, but all generated unit classes are partial so adding operator overloads to say Custom Code\Length.extra.cs should be straight forward.

If you want to take a stab at this, please feel free to start a PR to move that discussion there and start implementing some typical usecases.

anjdreas:
I still think this issue of representing rate of change in a generic way has value in its own, so I'm leaving this open.

anjdreas:
Closing this as I don't really have a good plan here, and no intention to follow up on it myself.

PR #113, opened 2015-12-04
Feature/additional overloading

eriove:
As discussed in #89:

I finally got code at a stage were its worth sharing. I've added code generation for overloads. This is what it looks like for Length (sorry about the re-formatting of the Length.json file, VS did that automatically).
´´´
"OperatorOverloads": [
    {
      "Operator": "/",
      "OtherUnit": "TimeSpan",
      "ReturnUnit": "Speed",
      "ReturnUnitBasePluralName": "MetersPerSecond",
      "OtherUnitBasePluralName": "TotalSeconds"
    },
´´´
It would be nice if it could automatically create the corresponding overload in Speed (I think all the needed information is there), but since this is the first time I'm working with PowerShell scripts it would be faster for me to do this manually for the 32 quantities...

At the moment I've only added the operator overloading for Length. If you like it I can do the same for the rest of the units.

I've added some tests in InterUnitConversionTests.cs to show the end result but I can't think of good auto-generated test. All tests I can think of would just test the generation code and not the actual conversion. If you have a good idea I'm happy to do the work.

### DIFF
--- a/UnitsNet.Tests/InterUnitConversionTests.cs
+++ b/UnitsNet.Tests/InterUnitConversionTests.cs
@@ -44,7 +44,7 @@ namespace UnitsNet.Tests
         }
 
         [Test]
-        public void LengthAndTimeStampToSpeed()
+        public void LengthAndTimepanToSpeed()
         {
             Length length = Length.FromMeters(10);
             Speed speed = length / TimeSpan.FromSeconds(1);
--- a/UnitsNet.Tests/InterUnitConversionTests.cs
+++ b/UnitsNet.Tests/InterUnitConversionTests.cs
@@ -20,6 +20,7 @@
 // THE SOFTWARE.
 
 using NUnit.Framework;
+using System;
 
 namespace UnitsNet.Tests
 {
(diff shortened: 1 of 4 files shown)


anjdreas:
Thanks for the proposed changes, Erik.

I do think, however, that using code generation is unnecessarily complicated to achieve what you want. There is very little code generated from the JSON so you don't really save much time on implementing new changes and I'd argue it is harder to maintain and learn compared to C# code.

How about implementing this directly in [CustomCode\Length.extra.cs](https://github.com/anjdreas/UnitsNet/blob/master/UnitsNet/CustomCode/UnitClasses/Length.extra.cs) instead? And similar for any other unit types where these types of overloads make sense. I would also recommend a test case for each implemented overload.

eriove:
When I started with the code generation my idea was that it would be enough to define / or *. Then the other operator (for the other unit) could be automatically defined. Then I got stuck at the PowerShell scripts... I'll give it some more thought, but just typing it in the .extra.cs make sense as it is right now.

eriove:
I agree with you that simply adding the overloads in the .extra.cs files is easier. You can close this pull request. Will open a new one with the operator overloading in the extra files.

anjdreas:
Ok cool, thanks.

anjdreas:
I'm fine with large pull requests as long as it address a logical grouping of changes, such as adding overload operators to units.

PR #116, opened 2015-12-14
Mass flow, specific energy and some new units

eriove:
Added mass flow and the following units:
Radians per second
Nautical miles
Knots
Meters per hour
[Second of arc](https://en.wikipedia.org/wiki/Minute_and_second_of_arc)
[Minute of arc](https://en.wikipedia.org/wiki/Minute_and_second_of_arc)

### DIFF
--- a/UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj
+++ b/UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj
@@ -67,9 +67,6 @@
       <Name>UnitsNet.Net35</Name>
     </ProjectReference>
   </ItemGroup>
-  <ItemGroup>
-    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
-  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.
--- a/UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj
+++ b/UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj
@@ -67,6 +67,9 @@
       <Name>UnitsNet.Net35</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.
--- a/UnitsNet/Scripts/UnitDefinitions/Frequency.json
+++ b/UnitsNet/Scripts/UnitDefinitions/Frequency.json
@@ -16,7 +16,19 @@
                 }
             ]
         },
-        {
+		{
+            "SingularName": "RadianPerSecond",
+            "PluralName": "RadiansPerSecond",
+            "FromUnitToBaseFunc": "x/6.2831853072",
+            "FromBaseToUnitFunc": "x*6.2831853072",
+            "Localization": [
+                {
+                    "Culture": "en-US",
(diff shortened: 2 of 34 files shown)


eriove:
Sorry for the delay. Made changes according to your comments. I'll add some comments/questions to the commit.

anjdreas:
Abbreviations min and sec ahould be fine as it does not conflict with any other Angle units. It is mainly for Angle.Parse() and .ToString () these are used. ToString () uses the first one in the list.

anjdreas:
This looks good to me at first glance. Browsing on the phone.

UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj
```diff
@@ -67,6 +67,9 @@
       <Name>UnitsNet.Net35</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+  </ItemGroup>
```
anjdreas: This change should be omitted.

UnitsNet/Scripts/UnitDefinitions/Frequency.json
```diff
@@ -16,7 +16,19 @@
                 }
             ]
         },
-        {
+		{
+            "SingularName": "RadianPerSecond",
```
anjdreas: This should be put in RotationalSpeed.json, shouldn't it?

From google:
> The radian per second (symbolized rad/s or rad/sec) is the Standard International ( SI ) unit of angular (rotational) speed. This quantity can be defined in either of two senses: average or instantaneous.

Wiki: https://en.wikipedia.org/wiki/Radian_per_second

UnitsNet/Scripts/UnitDefinitions/MassFlow.json
```diff
@@ -0,0 +1,31 @@
...
+                    "Culture": "en-US",
+                    "Abbreviations": ["kg/s"]
+                }
+            ]
+        },
+		{
+            "SingularName": "TonnePerDay",
+            "PluralName": "TonnesPerDay",
+            "FromUnitToBaseFunc": "x*86.4000",
+            "FromBaseToUnitFunc": "x/86.4000",
```
anjdreas: These two functions should be swapped I think?

1 kg/s = 86.4 t/d
https://www.google.no/search?q=Tonnes+Per+Day+to+kilograms+per+second&oq=Tonnes+Per+Day+to+kilograms+per+second&aqs=chrome..69i57.6551j0j7&sourceid=chrome&es_sm=93&ie=UTF-8#q=kilograms+per+second+to+Tonnes+Per+Day+

UnitsNet/Scripts/UnitDefinitions/Frequency.json
anjdreas: I see it's already there, so this is a duplicate and should be removed?

eriove: In naval architecture and (sometimes) oceanography rad/s is used for wave frequency.

My guess is that the historic reason is that it makes it easier to go between wave slope (in rad) and frequency.

anjdreas:
I've gone through the entire diff now, if you go over these this should be ready to merge in.

UnitsNet/Scripts/UnitDefinitions/MassFlow.json
eriove: I'll fix this tonight or tomorrow morning when I have access to VS.

There are an awful lot of constants. I'm considering checking the complete diff once more, it's so easy to invert numbers like this. Do you think I should do that?

anjdreas: I know, the only "failsafe" is that the unit tests asks you to implement each case and the wording of the tests were supposed to be more or less googlable so you'd get it right, but I've even accidentally swapped units myself. Easy to do. Not sure how we could make it more fool proof. I've double-checked all your conversions, but it doesn't hurt if you do too.

anjdreas:
On a side note, I recommend doing smaller pull requests, such as one per class of units. This way I'll be able to merge in the good parts faster.

eriove:
Fixed the project file and the inverted conversion in MassFlow, couldn't find more errors (using Google to check was certainly easier than doing it manually).

We will try to do smaller pull requests in the future. We (mainly @ulflind ) have been working with the overloading without code generation (see discussion in #113), that will be quite big too I'm afraid.

anjdreas:
Looks good. Thanks!

anjdreas:
Nuget 3.22.0 out.

PR #123, opened 2016-01-21
Feature/add rotational speed with correction

tongbong:


anjdreas:
This looks good to me, the only concern is changing the base unit. This has not been done before and it could potentially break backwards compatibility with serialization, but I don't believe it should in practice at least not for the Json serialization in this repo, since it serializes both value and unit. Can you think of any other scenario? If not I'll merge this in.

anjdreas:
After some thought I am fine with changing the base unit, and realized we have in fact done similar before. Units.NET does not guarantee this will not change and this is mentioned in the README section on serialization. I'll add a text to the README explicitly stating that the base unit may change and should not be used for serialization.

tongbong:
Yes, at the time I wasn't too happy either to change the base unit, but thought that precision is important and tests were green. I agree that serializing both value and unit will avoid any trouble with change base units.

PR #124, opened 2016-01-26
Feature/operator overloading

eriove:
Operator overloading as discussed in #89 and #113.

We've added the most common operations and the ones we know we will use. It is probably easiest to add more overloads when they are needed.

All code is cross-checked twice by me and @ulflind but I wouldn't be surprised if you find something that needs fixing.

### DIFF
--- a/UnitsNet.Tests/CustomCode/PressureTests.cs
+++ b/UnitsNet.Tests/CustomCode/PressureTests.cs
@@ -18,11 +18,22 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
+using NUnit.Framework;
 
 namespace UnitsNet.Tests.CustomCode
 {
     public class PressureTests : PressureTestsBase
     {
+
+
+        [Test]
+        public void PressureTimesAreaEqualsForce()
+        {
+            var force = Pressure.FromPascals(20) * Area.FromSquareMeters(3);
+            Assert.AreEqual(force, Force.FromNewtons(60));
+        }
+
+
         protected override double AtmospheresInOnePascal
         {
             get { return 9.8692*1E-6; }
--- a/UnitsNet.Tests/CustomCode/PressureTests.cs
+++ b/UnitsNet.Tests/CustomCode/PressureTests.cs
@@ -32,7 +32,7 @@ namespace UnitsNet.Tests.CustomCode
         }
 
         [Test]
-        public void AreaTimePressureEqualsForce()
+        public void AreaTimesPressureEqualsForce()
         {
             var force = Area.FromSquareMeters(3) * Pressure.FromPascals(20);
             Assert.AreEqual(force, Force.FromNewtons(60));
--- a/UnitsNet.Tests/CustomCode/PressureTests.cs
+++ b/UnitsNet.Tests/CustomCode/PressureTests.cs
@@ -24,8 +24,6 @@ namespace UnitsNet.Tests.CustomCode
 {
     public class PressureTests : PressureTestsBase
     {
-
-
         [Test]
         public void PressureTimesAreaEqualsForce()
         {
@@ -33,6 +31,12 @@ namespace UnitsNet.Tests.CustomCode
             Assert.AreEqual(force, Force.FromNewtons(60));
         }
 
+        [Test]
+        public void AreaTimePressureEqualsForce()
+        {
+            var force = Area.FromSquareMeters(3) * Pressure.FromPascals(20);
+            Assert.AreEqual(force, Force.FromNewtons(60));
+        }
 
         protected override double AtmospheresInOnePascal
         {
--- a/UnitsNet.Tests/CustomCode/SpeedTests.cs
+++ b/UnitsNet.Tests/CustomCode/SpeedTests.cs
@@ -20,28 +20,39 @@
 // THE SOFTWARE.
 
 using System;
-
 using NUnit.Framework;
-using System;
 
 namespace UnitsNet.Tests.CustomCode
 {
     public class SpeedTests : SpeedTestsBase
     {
-
         [Test]
         public void SpeedDevidedByTimespanEqualsAcceleration()
         {
             var acceleration = Speed.FromMetersPerSecond(20) / TimeSpan.FromSeconds(2);
             Assert.AreEqual(acceleration, Acceleration.FromMeterPerSecondSquared(10));
         }
+
         [Test]
-        public void SpeedTimesTimespanEqualsLength()
+        public void SpeedTimesTimeSpanEqualsLength()
         {
             var length= Speed.FromMetersPerSecond(20) * TimeSpan.FromSeconds(2);
             Assert.AreEqual(length, Length.FromMeters(40));
         }
 
+        [Test]
+        public void TimeSpanSpeedTimesEqualsLength()
+        {
+            var length = TimeSpan.FromSeconds(2) * Speed.FromMetersPerSecond(20);
+            Assert.AreEqual(length, Length.FromMeters(40));
+        }
+
+        [Test]
+        public void LengthDividedByTimeSpanEqualsSpeed()
+        {
+            var speed = Length.FromMeters(20) / TimeSpan.FromSeconds(2);
+            Assert.AreEqual(speed, Speed.FromMetersPerSecond(10));
+        }
 
         protected override double FeetPerSecondInOneMeterPerSecond
         {
--- a/UnitsNet.Tests/CustomCode/SpeedTests.cs
+++ b/UnitsNet.Tests/CustomCode/SpeedTests.cs
@@ -27,7 +27,7 @@ namespace UnitsNet.Tests.CustomCode
     public class SpeedTests : SpeedTestsBase
     {
         [Test]
-        public void SpeedDevidedByTimeSpanEqualsAcceleration()
+        public void SpeedDividedByTimeSpanEqualsAcceleration()
         {
             var acceleration = Speed.FromMetersPerSecond(20) / TimeSpan.FromSeconds(2);
             Assert.AreEqual(acceleration, Acceleration.FromMeterPerSecondSquared(10));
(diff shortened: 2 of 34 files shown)


anjdreas:
Thanks, I will try to take a look at this soon. A bit occupied with selling my house these days.

eriove:
No problem. Good luck with selling the house.

UnitsNet.Tests/CustomCode/SpeedTests.cs
```diff
@@ -20,11 +20,67 @@
...
+
+        [Test]
+        public void SpeedTimesDurationEqualsLength()
+        {
+            var length = Speed.FromMetersPerSecond(20) * Duration.FromSeconds(2);
+            Assert.AreEqual(length, Length.FromMeters(40));
+        }
+
+        [Test]
+        public void TimeSpanSpeedTimesEqualsLength()
```
anjdreas: Typo

UnitsNet.Tests/CustomCode/SpeedTests.cs
```diff
@@ -20,11 +20,67 @@
...
+
+        [Test]
+        public void LengthDividedByTimeSpanEqualsSpeed()
+        {
+            var speed = Length.FromMeters(20) / TimeSpan.FromSeconds(2);
+            Assert.AreEqual(speed, Speed.FromMetersPerSecond(10));
+        }
+
+        [Test]
+        public void DurationSpeedTimesEqualsLength()
```
anjdreas: Typo

UnitsNet.Tests/CustomCode/TorqueTests.cs
```diff
@@ -18,11 +18,25 @@
...
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
+using NUnit.Framework;
 
 namespace UnitsNet.Tests.CustomCode
 {
     public class TorqueTests : TorqueTestsBase
     {
+        [Test]
+        public void TorqueDevidedByLengthEqualsForce()
```
anjdreas: Typo

UnitsNet.Tests/CustomCode/TorqueTests.cs
```diff
@@ -18,11 +18,25 @@
...
+        {
+            var force = Torque.FromNewtonMeters(4) / Length.FromMeters(2);
+            Assert.AreEqual(force, Force.FromNewtons(2));
+        }
+        [Test]
+        public void TorqueDevidedByForceEqualsLength()
```
anjdreas: Typo

UnitsNet/CustomCode/UnitClasses/Angle.extra.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright © 2007 by Initial Force AS.  All rights reserved.
```
anjdreas: Not your fault, but these headers are outdated. I'll address that myself after merging this in.

UnitsNet/Scripts/UnitDefinitions/SpecificEnergy.json
```diff
@@ -22,7 +22,7 @@
             "PluralName": "CaloriesPerGram",
             "FromUnitToBaseFunc": "x*4.184e3",
             "FromBaseToUnitFunc": "x/4.184e3",
-            "Prefixes": ["Kilo"],
+			      "Prefixes": ["Kilo"],
```
anjdreas: Fix whitespace

UnitsNet/CustomCode/UnitClasses/Density.extra.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace UnitsNet
+{
+    public partial struct Density
+    {
+        public static Mass operator *(Density density, Volume volume)
+        {
+            return Mass.FromKilograms(density.KilogramsPerCubicMeter * volume.CubicMeters);
+        }
+
+        public static Mass operator *(Volume volume, Density density)
```
anjdreas: This applies to all units, but I think we should establish a consistent pattern for where operator overloading is added. The same applies to unit tests of cross-unit conversions.

```public static Mass operator *(Volume volume, Density density)```

Should this reside in Volume, Density or Mass?

For me, I find the mest intuitive that the left-hand operator determines this, in this case `Volume`, since that is what we are "acting" on. We have `Volume`, and we want to turn it into something else. The main drawback is that for multiplication, where one often adds an overload of the same calculation, but with `Volume` as right-hand operator will be moved to `Density.exta.cs` instead, so it is not clear when reading whether both have been implemented or not, but I find that drawback less bad than violating the pattern by grouping them both in `Volume.extra.cs` or both in `Density.extra.cs`.

Thoughts?

I'm fine cleaning up this in a separate PR in order to get this merged in sooner.

File: UnitsNet/CustomCode/UnitClasses/Length.extra.cs [hunk 8]
```diff
@@ -37,6 +35,39 @@ public static Length FromFeetInches(double feet, double inches)
         {
             return FromInches((FeetToInches * feet) + inches);
         }
+
+        public static Speed operator/ (Length length, TimeSpan timeSpan)
+        {
+            return Speed.FromMetersPerSecond(length.Meters / timeSpan.TotalSeconds);
+        }
+
```
anjdreas: A bit inconsistent code formatting, sometimes empty lines between methods, sometimes not. But don't spend time on this, I'll just do a commit doing a ReSharper code format of the entire solution.

UnitsNet/CustomCode/UnitClasses/RotationalSpeed.extra.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            return Angle.FromRadians(rotationalSpeed.RadiansPerSecond * duration.Seconds);
+        }
+
+        public static Angle operator *(Duration duration, RotationalSpeed rotationalSpeed)
+        {
+            return Angle.FromRadians(rotationalSpeed.RadiansPerSecond * duration.Seconds);
+        }
+
+        public static Power operator *(RotationalSpeed rotationalSpeed, Force force)
```
anjdreas: Is this right? Not familiar with this, but thought it might be Torque not Force?

UnitsNet/CustomCode/UnitClasses/RotationalSpeed.extra.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            return Power.FromWatts(rotationalSpeed.RadiansPerSecond * force.Newtons);
+        }
+
+        public static Power operator *(Force force, RotationalSpeed rotationalSpeed)
```
anjdreas: Same here.

UnitsNet.Tests/CustomCode/PressureTests.cs
```diff
@@ -18,11 +18,26 @@
...
     {
+        [Test]
+        public void PressureTimesAreaEqualsForce()
+        {
+            var force = Pressure.FromPascals(20) * Area.FromSquareMeters(3);
+            Assert.AreEqual(force, Force.FromNewtons(60));
+        }
+
+        [Test]
+        public void AreaTimePressureEqualsForce()
```
anjdreas: Typo

UnitsNet/CustomCode/UnitClasses/Density.extra.cs
anjdreas: TimeSpan is the exception, since we can't add operator overloads to TimeSpan. Those should probably just stay where they are, maybe with small one-liner comment on why.

anjdreas:
What a massive piece of work, great job.
I only found two operator overloads that might be off, everything else looks good to me and with good test coverage. Awesome!

I'm awaiting your feedback on my comments, then let me know when you think it is ready for merge and I'll take a final look at it.

#### Future improvement
1. ```UnitA / UnitA = Ratio ?```

2. ```UnitA / int|long|double|decimal = UnitA```

This one is tricky, should we preserve .NET round behavior so that `Length.FromMeters(3) / (int)2 = 1, 1.5 or 2`? Internally we use double representations for most units, so it would by probably default to 1.5d internally, and I think that is the most intuitive, but one could argue dividing by integer should round down but I don't think I like that behavior. Rounding up is even less intuitive as only Convert.ToInt32() and Math.Round() methods achieve that.

UnitsNet/CustomCode/UnitClasses/RotationalSpeed.extra.cs
eriove: The first ones are correct
rad/s * s = rad
RotationalSpeed * Force is wrong. I'll remove it. I'm guessing we are missing some other overload. We'll create a new pull request if it is needed.

UnitsNet/CustomCode/UnitClasses/Density.extra.cs
eriove: I agree with you. It's intuitive to have it in the file file with the left hand operator.

We'll make another pull-request with this change.

It's quite confusing with both `Duration` and `TimeSpan`. What was the idea behind `Duration`? Did you need extremely long or short times that worked better with a floating point numbers?

UnitsNet/CustomCode/UnitClasses/Length.extra.cs
eriove: Thanks! I think I will try ReSharper again, was a couple of years since I tried it and these are things were it really shines (which I didn't thought I needed back then).

eriove:
It should be ready to merge now (see comments on your comment, typos should be fixed).

1. `UnitA / UnitA` already returns a `double` which is very intuitive to me
2. Dividing by a `double`is also implemented, it's quite logical that it behaves as it does right now. I would prefer if the compiler didn't implicitly cast long, int and decimal to double and used the existing overload but the end result makes more sense than the alternatives. Since the returned `Unit` would always be a double it would be quite strange if it was rounded up or down.

UnitsNet/CustomCode/UnitClasses/Density.extra.cs
anjdreas: I was thinking the exact same thing. It was introduced early on and will break compatibility by removing it, so I'm considering making a list of things to do on the next major version bump. 

One advantage over TimeSpan is that it is compatible with our code generation tools as well as operator overloads, but the biggest drawback is duplicate code for both TimeSpan  and Duration cases as well as not understanding why both exists.

It's possible that implementing implicit cast from TimeSpan to Duration could help, since that would remove the need for implementing both cases in tests and operator overloads, but it might not be intuitive that you can pass a TimeSpan to a method that takes Duration.

I'll have to think a little about that one, not sure what is the best approach. Keeping it or removing it.

anjdreas:
You are right of course, I forgot we already had some operator overloads.
I agree with your reasoning and your latest change looks good. Merging in now.

anjdreas:
Nuget 3.25 out.

PR #125, opened 2016-01-28
Add more Area units

tongbong:
Replace hardcoded units in the json with the use of prefixes

### DIFF
--- a/UnitsNet.Tests/CustomCode/AreaTests.cs
+++ b/UnitsNet.Tests/CustomCode/AreaTests.cs
@@ -19,18 +19,41 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+using UnitsNet.Units;
+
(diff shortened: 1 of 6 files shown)


anjdreas:
Not sure if you noticed, but there is a reason I didn't use prefixes on this unit.

MicrosquareMeter sounds pretty funky to me, no? :smile:

anjdreas:
I suppose a custome scheme must be implemented in powershell to support something like this:
<NamePrefix><UnitPrefix><BaseUnit> to get <Square><Micro><meter>

It could be worth it if we had many units with the same problem, but right now it's only Area and Volume I can think of. I'd be happy to accept a PR on supporting such a scheme though.

tongbong:
You are correct, I did not check the naming... I will be more careful next time.
I guess your scheme proposition is a good idea.
My company is working on a machine dealing with square micro meter unit.
So let's reject this PR (I let you close the PR), and I will add this unit the hard way ^^
Anyway let's keep in mind the scheme thing as a possible feature.

anjdreas:
Alright, sounds good.

Issue #126, opened 2016-01-30
International Unit (Pharmacology)


neutmute:
Hello - I'm writing an app that needs to process nutrient info. Vitamin A is stored in International Units - [see here](https://developer.calorieking.com/food-api/documentation#Nutrients)

[International Units](https://en.wikipedia.org/wiki/International_unit) are widely used in pharmacology it appears (I'm not an expert in this domain) but it is specific to each thing being measured. ie: 10 IU of Vitamin A is not the same mass or volume for Vitamin D. 

Any suggestions on how to deal with this? 

I guess I should make my own custom type since 1 IU of Vitamin A cannot be added to 1 IU of Vitamin D. 
For for all the other nutrient dimensions that have kJ, mg I'd use the standard Units.NET types.

anjdreas:
Hi, not familiar with those units myself, but from your description I agree with your approach. Create new units like VitaminAQuantity and VitaminDQuantity, with different calculations in their construction methods e.g. FromInternationalUnit() and their conversion properties e.g. InternationalUnit.
This way you have different types to represent these incompatible quantities even though they have similar units names. 

This domain seems widely used to me so if you want to do a pull request I'd be happy to merge it in.
See the readme for how to add new units and creating a PR.

anjdreas:
If you want conversions from VitaminA and D to kJ, mg and other units, you add those other units to both A and D with different conversion functions. If that makes sense. 

Just ask if anything is unclear,  but read the readme for instructions.

anjdreas:
You can also add conversions to Volume, Mass and such unit classes. Add a VitaminAQuantity.extra.cs file with a manually written partial class. See ither .extra.cs files for inspiration.

neutmute:
Excellent tips. I had only thought of making an 'International Unit', not 'VitaminA' unit. 
PR submitted :)

PR #127, opened 2016-01-30
Add VitaminA international unit

neutmute:
Hi Andreas - As discussed in #126, this PR adds VitaminA International Unit type.

I didn't add other Vitamin's or conversions to micrograms yet - keeping things simple to start off with.

Let me know how this looks!

### DIFF
--- a/UnitsNet.Tests/CustomCode/VitaminATests.cs
+++ b/UnitsNet.Tests/CustomCode/VitaminATests.cs
@@ -1,0 +1,31 @@
+﻿// Copyright © 2007 by Initial Force AS.  All rights reserved.
+// https://github.com/InitialForce/UnitsNet
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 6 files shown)


neutmute:
One thought - are these properties correct in the json:

            "SingularName": "InternationalUnit",
            "PluralName": "InternationalUnits",

Or should the words be separated by spaces?

            "SingularName": "International Unit",
            "PluralName": "International Units",

anjdreas:
Looks good to me!
I'll await merging in assuming you want to add more units to VitaminA. You decide if you want to do VitaminD in a separate PR or in the same one, does not matter to me just let me know when things are ready to merge in and I'll review it.

Singular/PluralName cannot have space, as the string is used to generate C# enum type and C# method/property names.

neutmute:
Excellent. 

re: more additions, sorry - I gave the wrong impression in my comment that I would submit further commits for conversions/Vitamin D but this should suffice for my projected needs and the PR is ready to merge if satisfactory.

TLDR:
* Vitamin D isn't part of the nutrition info I need to store but more an example to show that IU's from different molecules cannot be mingled.
* I won't need the microgram conversion either as far as I know. Using the IU type and being able to divide/multiply/add is sufficient for my project needs.

anjdreas:
Nuget 3.24 out.

neutmute:
awesome - thanks

PR #128, opened 2016-02-01
Feature/cls compliant

eriove:
I added the [assembly: CLSCompliant(true)], since the assembly is CLS compliant this make sense.


anjdreas:
Please rebase this PR onto the latest upstream/master instead of a merge commit. A PR should only contain commits that contribute something so it's clear what it changes.

Also strip out the bump of version in AssemblyVersion("3.25.0")]. In latest master this has been set back to 1.0.0 as it should only be bumped as part of building and publishing nugets.

eriove:
I've tried rebase-ing (`git rebase upstream/master`). It's the first time I've done a rebase so I might have done something wrong but the end result made no changes to the code (which was expected since the merge essentially did the same thing in terms of code changes). When I merged I changed the version number to 1.0.0 so the actual code changes should be correct in this pull request.

Will rebase instead of merging the next time.

anjdreas:
Alright, you should do an interactive rebase so you can remove the merge commit from the rebased commits (only rebasing the CLSCompliant commit on top of master). No worries though.

eriove:
Thanks!

Issue #130, opened 2016-02-05
Cup as VolumeUnit


neutmute:
Hi again! 
I'm working on a food nutrition system and we need to be able to reference the volume unit of cup (see [wiki def](https://en.wikipedia.org/wiki/Cup_%28unit%29)).

[UnitsNet.Unit.VolumeUnit](https://github.com/anjdreas/UnitsNet/blob/master/UnitsNet/GeneratedCode/Enums/VolumeUnit.g.cs) has `Teaspoon `and `Tablespoon` but no _Cup_.

I propose adding:
* MetricCup (250mL)

I don't need US Cups for now but would be happy to add to help clarify the distinction. There are two(!) US cups. Their names might be
* UsCustomaryCup
* UsLegalCup

Would a PR for `MetricCup` be accepted?

anjdreas:
Absolutely :-) sounds good.

neutmute:
#131 submitted

PR #131, opened 2016-02-05
Add metric, us legal and us customary cups to volume

neutmute:
As per #130 adds the three cup units to volume.

I didn't add an abbreviation for the three as it is too ambiguous IMHO.
I tried "metric cup" but that broke `Volume.Parse` which didn't like the two word combination


anjdreas:
Nuget 3.25 out.

PR #133, opened 2016-02-07
Added ability to change the default unit for .ToString()

eriove:
For many measures the unit that makes most sense varies between applications. I added the static property `ToStringDefaultUnit` which can be used to set the default unit to use for a ToString() without arguments. This primarily makes debugging easier. Existing behaviour will be kept as long as `ToStringDefaultUnit` isn't set.

The tests generated aren't thread safe but as far as I know NUnit doesn't support multi-threaded tests.

I imagined this property would be used by setting it once in a static constructor for a solution, or that it is set manually when someone starts debugging. This makes `ToString()` even more dangerous than usual since the result may vary if some other part of the code may set `ToStringDefaultUnit` but since the same thing will happen if someone sets the current culture most developers _should_ know that it's bad practice to not add input arguments to `ToString()`... (Yes I see that bug at least once a month)


eriove:
Don't feel any rush to merge this/release a new version. It's just a "nice to have" thing I created for fun.

anjdreas:
I think this is a fair change. I've had the same thought a couple of times how to let people choose their preferred unit in ToString(), but it was never a big enough concern for me to deal with it. Looks good to me, just fix a bit of whitspace and this should be ready to merge.

eriove:
I've replaced all tabs with white space now.

anjdreas:
Great!  I'm unavailable this whole week but will get back to you.

UnitsNet.Tests/GeneratedCode/VolumeTestsBase.g.cs
```diff
@@ -295,5 +295,291 @@ public void EqualsReturnsFalseOnNull()
...
+        public void ToStringReturnsCorrectNumberAndUnitWithDefaultUnit()
+        {
+            Volume.ToStringDefaultUnit = VolumeUnit.CubicMeter;
+            Volume cubicmeter = Volume.FromCubicMeters(1);
+            string cubicmeterString = cubicmeter.ToString();
+            Assert.AreEqual("1 " + UnitSystem.GetCached(null).GetDefaultAbbreviation(VolumeUnit.CubicMeter), cubicmeterString);
+        }
+
+        [Test]
+        public void ToStringReturnsCorrectNumberAndUnitWithCentiliterAsDefualtUnit()
```
anjdreas: I like you went the whole 9 yards here, but I do think we should avoid adding tons of tests that basically test the same thing. I really just want to test the default behavior and one or two custom ToStringDefaultUnit values to ensure the implementation is working. Testing every unit seems to me to add little benefit since all units are generated the same way.

I propose you remove the powershell modification that generates these tests per unit, and instead manually write some test cases you find sufficient to verify the implementation.

eriove:
Just noticed that there is a bug/design limitation regarding structs and `ToString()` in Visual Studio 2015. If I understand [this answer](http://stackoverflow.com/a/31569625/1144943) correctly `ToString()` on a struct can't access static properties. I'll try some workarounds before I commit with all the generated tests removed.

anjdreas:
Interesting link, never knew about that.

eriove:
It seems like the behavior changes depending on where you've put breakpoints in the code and if you step or run between said breakpoints. I've seen {2 cm}, {UnitsNet.Length} and {2 (no abbreviation for LengtUnit.Centimeter)} so far...

eriove:
The only way to work around it (that I can think of) would be to store a field in each struct with the unit to use. But that would increase the size of the struct and isn't something I would like to do.

I just remove some tests and commit. It sounded like this is something that will be fixed in a later version of Visual Studio.

anjdreas:
Did you try `[DebuggerDisplay("{ToString()}")]` ?

eriove:
Yes, it made Visual Studio lock up completely for a few seconds.

eriove:
It looks like I failed with the rebase upstream (again)...

anjdreas:
Remember you can rebase your branch onto origin/master and force push to fix mistakes in your git commit history.

anjdreas:
This looks good to merge to me, is it done?

eriove:
It's done.

anjdreas:
Awesome! Will release soon, waiting for another PR.

anjdreas:
Nuget 3.26 is now out.

Issue #132, opened 2016-02-05
How to handle ambiguous unit abbreviations


anjdreas:
There are several units with ambiguous unit abbreviations, and there is currently no good way to handle this. This issues is created as a place to discuss solutions.

Example: #131
Metric cup, US legal cup and customary cups might be natural to assign "cup" or "metric cup" as abbreviations, but "cup" is ambiguous and "metric cup" is currently not supported due to whitespace.

neutmute:
Is the use case for <Unit>.Parse(string input) to be able to take natural language input and turn it into a quantifiable unit? 
If so then users are most likely to just type "1 cup" - rather than "1 metric cup" or "1 legal cup". Which exact cup they mean is driven by context that only the application can know - ie: An Australian would expect "1 cup" to imply "1 metric cup" and may not even be aware of other definitions of cups (I wasn't until today :)). An American nutritionist would perhaps expect "1 cup"  to be interpreted as "1 legal cup".

That is to say I don't think the problem is solvable from a parsing perspective. Support for whitespaces could help but I don't think users would know/like needing to type "1 metric cup". 

Idea: Optionally allow the application to supply configurable mappings to the Parse method.
eg: "cup" => VolumeUnit.MetricCup.
These would override / supplement the built in abbreviations.

re: ToString(): In the absence of an abreviation it could fall back to using the enum unit for in these cases? ie: `$"{Quantity} {EnumUnit.ToString()";`  
Useful for debug/watch windows and logs but not so much for UI output for an end user. Again in that case I think the application could just 'know the context' and emit the appropriate text instead.

anjdreas:
Agree, there is probably no way for Parse() to handle ambiguous units. I don't even know how the current implementation handles this, but I imagine it chooses an arbitrary one (the one defined last in json, or possibly the last in lexical ordering?)

1. I propose Parse() throws an exception if there are ambiguous units defined, it's better than getting a calculation error due to two very similar units. I can imagine most programmers are not aware of this problem in most cases, such as the cups already mentioned and metric ton vs long and short tons. 

2. I propose units who naturally wants to share the same abbreviation should do so. In the cups example, this means ToString() will output "1 cup" for all of them, but Volume.Parse("1 cup") will throw an exception due to the ambiguity.

anjdreas:
[...]

PR #140, opened 2016-02-16
Handle ambiguous abbreviation parsing by throwing exception

neutmute:
As discussed on #132, this PR supports having abbreviations shared across units and throws when ambiguous.

I'll add some notes to the diff


neutmute:
This should be good now - I haven't placed all abbreviations onto the units yet - just noticed you were busy reformatting all json files ;)

anjdreas:
Sweet, looks good to me.

anjdreas:
Nuget 3.26 is now out. 

I'm curious, did you get notified when I mentioned you in the github release?
https://github.com/anjdreas/UnitsNet/releases

neutmute:
Awesome - thank you. 
Checking my emails - no, I only received notifications from your comments - not the release notes.

anjdreas:
Ok good to know, I've always wondered the best way to notify contributors of a new release. Was hoping tagging in the release notes would do it, but guess not.

neutmute:
If it helps, personally, I don't need the notification as I can see you release quickly after a merge so I just keep an eye out. On other library projects I have a personal myget feed that builds from my fork and I use that until the PR becomes official. Your fast releases and the nature of the library - not needing dogfooding before PR - save me that extra work.

It does give you more work though in manually balling up a release though, so you could automatically release on merge via nuget/teamcity. You lose the custom change log in the nuspec but it saves you from the manual work and contributors know a merge = a release.
If you want to bundle a few merges, perhaps have the default branch as `develop` with merges going into `develop`. A manual merge by yourself to `master` would trigger the CI build and release to nuget.

Just food for thought!

anjdreas:
Thanks for the input. I did have the nuget publish automated in TeamCity a while back, but after changing branch style from gitflow back to github flow I didn't bother fixing it. It really is just bumping the .nuspec version and running two bat files to deploy, so the effort and frequency are too low to make me want to spend time automating it again.

I do want to keep a changelog though, so I'm thinking instead of putting it in the .nuspec, maybe just add it as annotation to the git tags instead. That should make it available on github too.

PR #143, opened 2016-02-25
Added DynamicViscosity to base units

gsokoll:
I've added in Dynamic Density with a few of the more common units.  I've run the unit tests via ReSharper and they all passed.  I have only included EN-US localisations.  Let me know if anything isn't right.


anjdreas:
Great, looks good!

anjdreas:
Nuget 3.27 out.

gsokoll:
You're welcome - glad you liked it :)

PR #144, opened 2016-02-29
Add degree per second unit

tongbong:


UnitsNet/Scripts/UnitDefinitions/RotationalSpeed.json
```diff
@@ -20,6 +20,23 @@
...
+      {
+        "SingularName": "DegreePerSecond",
+        "PluralName": "DegreesPerSecond",
+        "FromUnitToBaseFunc": "(Math.PI/180)*x",
+        "FromBaseToUnitFunc": "(180/Math.PI)*x",
+        "Prefixes": [ "Nano", "Micro", "Milli" ],
+        "Localization": [
+          {
+            "Culture": "en-US",
+            "Abbreviations": [ "r/s" ]
```
anjdreas: en-US abbreviation is wrong, isn't it?

UnitsNet.Tests/CustomCode/RotationalSpeedTests.cs
```diff
@@ -42,6 +42,15 @@ public class RotationalSpeedTests : RotationalSpeedTestsBase
 
         protected override double RevolutionsPerSecondInOneRadianPerSecond => 1/(2*Math.PI);
 
+        protected override double DegreesPerSecondInOneRadianPerSecond => (180 / Math.PI) * 1;
+
+        protected override double MillidegreesPerSecondInOneRadianPerSecond => ((180 / Math.PI) * 1)* 1E3;
+
```
anjdreas: I prefer tests having pre-calculated values here, as it's too easy to fall in the trap of just using the reverse conversion expression from the implementation. I recommend you use an external conversion calculator as a sanity check.

tongbong: Should be fixed in commit  e6ba20a.
You are right, I found myself wondering why I was repeating the formula...
Thanks for the tip.

UnitsNet/Scripts/UnitDefinitions/RotationalSpeed.json
tongbong: Should be fixed  in commit 27a01f7.
Thanks for the feedback.

anjdreas:
Perfect!

PR #147, opened 2016-02-29
add micro liters per minute

rdelhommer:


rdelhommer:
Alrighty.  It's done.  I went ahead and added some other prefixes because why not :P

UnitsNet/Scripts/UnitDefinitions/Flow.json
```diff
@@ -64,10 +64,12 @@
       "PluralName": "LitersPerMinute",
       "FromUnitToBaseFunc": "x/60000.00000",
       "FromBaseToUnitFunc": "x*60000.00000",
+	  "Prefixes": [ "Nano", "Micro", "Milli", "Centi", "Deci", "Kilo"],
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "LPM" ]
+          "Abbreviations": [ "LPM" ],
+		  "AbbreviationsWithPrefixes": ["nLPM", "μLPM", "mLPM", "cLPM", "DLPM", "KLPM"]
```
anjdreas: It should be "dLPM" shouldn't it? It's probably not even a widely used abbreviation, but we should at least get the prefixes right.

rdelhommer:
You're totally right.  I noticed that kilo was wrong as well.

anjdreas:
Nuget 3.28 is out.

PR #154, opened 2016-03-31
Added overload for length divided by speed

eriove:
I just discovered that I manged to miss the overload for `Duration=Length/Speed`


anjdreas:
Nuget 3.30 out

PR #156, opened 2016-04-05
Added BrakeSpecificFuelConsumption

eriove:
No more planned pull requests after this one...

### DIFF
--- a/UnitsNet/Scripts/UnitDefinitions/BrakeSpecificFuelConsumption.json
+++ b/UnitsNet/Scripts/UnitDefinitions/BrakeSpecificFuelConsumption.json
@@ -28,10 +28,11 @@
       ]
     },
     {
-      "SingularName": "PoundPerHorsepowerHour",
-      "PluralName": "PoundsPerHorsepowerHour",
+      "SingularName": "PoundPerMechanicalHorsepowerHour",
+      "PluralName": "PoundsPerMechanicalHorsepowerHour",
       "FromUnitToBaseFunc": "x*1.689659410672e-7",
       "FromBaseToUnitFunc": "x/1.689659410672e-7",
+      "XmlDocSummary": "The pound per horse power hour uses mechanical horse power and the imperial pound",
       "Localization": [
         {
           "Culture": "en-US",
--- a/UnitsNet/Scripts/UnitDefinitions/BrakeSpecificFuelConsumption.json
+++ b/UnitsNet/Scripts/UnitDefinitions/BrakeSpecificFuelConsumption.json
@@ -23,7 +23,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "kg/W" ]
+          "Abbreviations": [ "kg/J" ]
         }
       ]
     },
--- a/UnitsNet/Scripts/UnitDefinitions/BrakeSpecificFuelConsumption.json
+++ b/UnitsNet/Scripts/UnitDefinitions/BrakeSpecificFuelConsumption.json
@@ -1,0 +1,43 @@
+﻿{
+  "Name": "BrakeSpecificFuelConsumption",
+  "BaseUnit": "KilogramPerJoule",
+  "XmlDoc": "Brake specific fuel consumption (BSFC) is a measure of the fuel efficiency of any prime mover that burns fuel and produces rotational, or shaft, power. It is typically used for comparing the efficiency of internal combustion engines with a shaft output.",
+  "Units": [
+    {
+      "SingularName": "GramPerKiloWattHour",
+      "PluralName": "GramsPerKiloWattHour",
+      "FromUnitToBaseFunc": "x/3.6e9",
+      "FromBaseToUnitFunc": "x*3.6e9",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "g/kWh" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "KilogramPerJoule",
+      "PluralName": "KilogramsPerJoule",
+      "FromUnitToBaseFunc": "x",
+      "FromBaseToUnitFunc": "x",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "kg/W" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "PoundPerHorsepowerHour",
+      "PluralName": "PoundsPerHorsepowerHour",
+      "FromUnitToBaseFunc": "x*1.689659410672e-7",
+      "FromBaseToUnitFunc": "x/1.689659410672e-7",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "lb/hph" ]
+        }
+      ]
+    }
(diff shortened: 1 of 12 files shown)


anjdreas:
I'm not familiar with these units and a quick Google on Kilogram Per Joule give me little.
My initial thought is question if this belongs in the core lib, or if we need to start addressing the need for domain specific libs.
Thoughts?

eriove:
Almost everyone is using gram/kWh (I've never seen anything else except on Wikipedia, but I've never worked with old engineers from a country with imperial units). That gives round and easy numbers (the specific consumption of an internal combustion engine is typically between 150 and 240 g/kWh. If someone used kg/J it would be microgram/Joule (200 g/kWh = 55 μg/J). I only added kg/J because it felt odd not to include the SI unit.

We didn't included this unit in the initial (big) pull request we did since we weren't sure if it would fit in the scope of the library. If your'e not working with internal combustion engines you won't use this unit, but then they're quite common in many industries (everything from energy, as in power plants and back up generators, to automotive and shipping industry).

It's up to you to decide. We have set up an internal version of the scripts to generate units that doesn't fit in the general library (so far only including spectral density for translative motions) so it will be easy for us to just move these changes to that library,

anjdreas:
I honestly don't know what is a good way to categorize whether a family of units belong in core lib or in an additional lib. Intuitively I want to achieve a sort of 80/20 ratio, to avoid the core lib growing fat.

I've thought of trying to split into engineering domains, such as Chemical, Medical, Construction, Aerospace, Astrophysics, Electrical and so on, but there are many overlaps.
If we were to classify this particular PR in a domain, what would you suggest?
Seeing it is related to combustion engines, I would think either CombustionEngines if that is a large enough domain to warrant its own category, or something broader like Mechanical.

Relevant: https://en.wikipedia.org/wiki/Engineering#Main_branches_of_engineering

What do you think?

eriove:
In general I like the idea of multiple libraries by engineering domains. This particular pull request probably fits best into mechanical engineering (that's where you start at university if you want to study combustion engines).

But as you say there are many overlaps. Perhaps the best way of defining the common library is by the units used by more than one of the sub libraries. I find it hard to make this change without breaking changes though in the existing libraries though.

eriove:
I've generally created desktop or server applications and never really thought of the library size as a problem. My main reasons for creating multiple libraries is to decrease coupling (namespaces with too many types might be a usability problem too, but that doesn't req). In this case the only coupling is the operator overloads and they would work even if the library was split into multiple DLLs.

How constrained are IoT devices and phones really?

anjdreas:
> How constrained are IoT devices and phones really?
Not sure, and it is a fair point, but I also think that many apps tend to have anything between 5-100 nugets, and many small streams make a river.

What we could do, however, is postpone this until it becomes an actual problem, meaning someone creates an issue addressing the size of the lib. Then when that time comes, we do a decision on the units currently in our lib, and try to split up into engineering domains or whatever, then we bump the nuget major version to make a clean break.

If you agree with this, then I can merge this in and we continue business as usual until said time.

anjdreas:
The decision is also nudged by the fact that I have limited time to put into this project, such as building a build system and contribution guideline for core vs extra libs.

eriove:
I can totally understand that. This sounds like the easiest approach for now.

Another way to decrease the library size is to refactor common code (like `ParseWithRegex`) into static generic classes. But that's something that really has to wait until there is a problem (since the run time size will depend on how generics are handled in a specific platform and has to be measured).

UnitsNet/Scripts/UnitDefinitions/BrakeSpecificFuelConsumption.json
```diff
@@ -0,0 +1,43 @@
+﻿{
+  "Name": "BrakeSpecificFuelConsumption",
+  "BaseUnit": "KilogramPerJoule",
+  "XmlDoc": "Brake specific fuel consumption (BSFC) is a measure of the fuel efficiency of any prime mover that burns fuel and produces rotational, or shaft, power. It is typically used for comparing the efficiency of internal combustion engines with a shaft output.",
+  "Units": [
+    {
+      "SingularName": "GramPerKiloWattHour",
+      "PluralName": "GramsPerKiloWattHour",
+      "FromUnitToBaseFunc": "x/3.6e9",
```
anjdreas: Is this right? 
I found `1 gram/kilowatt-hour = 2.777777777778 x 10-10 kilogram/joule` on this [calculator](http://www.translatorscafe.com/cafe/units-converter/fuel-efficiency-mass/gram-per-kilowatt/hour-5bg/kw*h5d-to-pound-per-horsepower/hour/)

UnitsNet/Scripts/UnitDefinitions/BrakeSpecificFuelConsumption.json
```diff
@@ -0,0 +1,43 @@
...
+      "FromBaseToUnitFunc": "x",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "kg/J" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "PoundPerHorsepowerHour",
```
anjdreas: Should this be distinguished from PoundPerMetricHorspowerHour, or is that already not ambiguous in the engineering field?

UnitsNet/Scripts/UnitDefinitions/BrakeSpecificFuelConsumption.json
eriove: That's the same value, isn't it?
1/3.6e9 = 2.7778e-10

eriove: `1 g/kWh /1000 => kg/kWh /1000 => kg/Wh /3600 => kg/Ns = kg/J
	1/1000		1/1 000 000	1/3600 000 000	
1 g/kWh => 1/3600 000 000 kg/J`

UnitsNet/Scripts/UnitDefinitions/BrakeSpecificFuelConsumption.json
eriove: I honestly don't know. I tend to avoid horse power at all costs, since it's so hard to know which type you are talking about (and shipping has used MW for quite awhile now).

Since there is pounds involved I would guess that it is [mechanical horse power](https://en.wikipedia.org/wiki/Horsepower#Mechanical_horsepower). Let me cross check and update the pull request.

eriove: It's `MechanicalHorsePower` and `Punds` (as they are defined in the Power and Mass units). Changing the name of the unit.

anjdreas: > That's the same value, isn't it?
1/3.6e9 = 2.7778e-10

Right you are, my bad!

anjdreas:
Is this ready to merge now?

anjdreas:
Never mind. All points are addressed.

anjdreas:
Nuget 3.32 out

Issue #167, opened 2016-06-02
Feature Request: Add MassFlowUnit.KilogramPerHour


enterprisebug:
in MassFlowUnit enum add support for KilogramPerHour.

MassFlowUnit.KilogramPerSecond is already present.

MarcDrexler:
Please read  https://github.com/anjdreas/UnitsNet/wiki/Adding-a-New-Unit

enterprisebug:
Thanks for clarification. Didn't see that. This is a quite detailed explanation! Will do a contrib.

anjdreas:
Thanks, let's continue in PR.

PR #168, opened 2016-06-02
Added MassFlowUnit.KilogramPerHour

enterprisebug:
solves issue #167

### DIFF
--- a/UnitsNet.Tests/CustomCode/MassFlowTests.cs
+++ b/UnitsNet.Tests/CustomCode/MassFlowTests.cs
@@ -31,7 +31,7 @@ namespace UnitsNet.Tests.CustomCode
         protected override double DecagramsPerSecondInOneGramPerSecond => 1E-1;
 
         protected override double HectogramsPerSecondInOneGramPerSecond => 1E-2;
-        protected override double KilogramsPerHourInOneGramPerSecond => 3.6;
+        protected override double KilogramsPerHourInOneGramPerSecond => 0.277778;
 
         protected override double KilogramsPerSecondInOneGramPerSecond => 1E-3;
 
--- a/UnitsNet.Tests/CustomCode/MassFlowTests.cs
+++ b/UnitsNet.Tests/CustomCode/MassFlowTests.cs
@@ -31,7 +31,8 @@ namespace UnitsNet.Tests.CustomCode
         protected override double DecagramsPerSecondInOneGramPerSecond => 1E-1;
 
         protected override double HectogramsPerSecondInOneGramPerSecond => 1E-2;
-        protected override double KilogramsPerHourInOneGramPerSecond => 0.277778;
+
+        protected override double KilogramsPerHourInOneGramPerSecond => 3.6;
 
         protected override double KilogramsPerSecondInOneGramPerSecond => 1E-3;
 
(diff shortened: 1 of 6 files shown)


UnitsNet/Scripts/UnitDefinitions/MassFlow.json
```diff
@@ -17,6 +17,22 @@
       ]
     },
     {
+      "SingularName": "KilogramPerHour",
+      "PluralName": "KilogramsPerHour",
+      "FromUnitToBaseFunc": "x*0.277778",
+      "FromBaseToUnitFunc": "x/0.277778",
```
anjdreas: I would prefer `x/3.6` and `x*3.6` here.

anjdreas:
Other than that this looks good, let me know when fixed and I will merge it in.

enterprisebug:
Fixed it.

UnitsNet/Scripts/UnitDefinitions/MassFlow.json
```diff
@@ -19,8 +19,8 @@
     {
       "SingularName": "KilogramPerHour",
       "PluralName": "KilogramsPerHour",
-      "FromUnitToBaseFunc": "x*0.277778",
-      "FromBaseToUnitFunc": "x/0.277778",
+      "FromUnitToBaseFunc": "x*3.6",
+      "FromBaseToUnitFunc": "x/3.6",
```
anjdreas: Swap these functions

UnitsNet.Tests/CustomCode/MassFlowTests.cs
```diff
@@ -31,7 +31,7 @@ public class MassFlowTests : MassFlowTestsBase
         protected override double DecagramsPerSecondInOneGramPerSecond => 1E-1;
 
         protected override double HectogramsPerSecondInOneGramPerSecond => 1E-2;
-        protected override double KilogramsPerHourInOneGramPerSecond => 3.6;
+        protected override double KilogramsPerHourInOneGramPerSecond => 0.277778;
```
anjdreas: Uh-oh :-) You just flipped the implementation upside down so it is no longer correct. This is very easy to do, so my go-to sanity check is to google `KilogramsPerHour In OneGramPerSecond` and that tells me this should read `3.6`.

anjdreas:
I've added some examples to the Add a New Unit steps, to highlight getting this part right. Let me know if that might have helped you avoid this.

Step 4 in https://github.com/anjdreas/UnitsNet/wiki/Adding-a-New-Unit

4. **Fix generated test stubs** to resolve compile errors tests
  * Override the missing abstract properties in the unit test class (ex: [LengthTests.cs](https://github.com/anjdreas/UnitsNet/blob/master/UnitsNet.Tests/CustomCode/LengthTests.cs#L35))
  * Use a good source for choice of conversion factors, with at least **7+ [significant figures](https://en.wikipedia.org/wiki/Significant_figures)**
  * Triple-check the number you write here, this is **the most important piece**.
  * Example: `InchesInOneMeter` in [LengthTests.cs](https://github.com/anjdreas/UnitsNet/blob/master/UnitsNet.Tests/CustomCode/LengthTests.cs#L35)
    * I find the conversion factor to be `39.37007874` from an online unit conversion calculator, it has 10 significant figures so that is plenty
    * I add the code: `protected override double InchesInOneMeter => 39.37007874;`
    * I Google to double-check: `Inches In OneMeter` and it tells me `1 Meter = 39.3701 Inches`
    * I check again by intuition, is there really around 40 inches in a meter? Yes, sounds about right.

enterprisebug:
double and triple checking is getting me headache :)

Explanation of adding a new unit is better now. Thanks.

Next unit will get it's own branch (as you stated)

anjdreas:
Thanks! Nuget 3.35 is out.

PR #179, opened 2016-08-10
Added Volumes: Drum and Tote; Added Areas: Acre and Hectare

BMcFerrin:
I need these for a project I am working on; thought they may be useful for others as well.

### DIFF
--- a/UnitsNet.Tests/CustomCode/VolumeTests.cs
+++ b/UnitsNet.Tests/CustomCode/VolumeTests.cs
@@ -49,6 +49,8 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double DecilitersInOneCubicMeter => 1E4;
 
+        protected override double DrumsInOneCubicMeter => 4.80312818132;
+
         protected override double HectolitersInOneCubicMeter => 1E1;
 
         protected override double ImperialGallonsInOneCubicMeter => 219.96924;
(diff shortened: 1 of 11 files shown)


UnitsNet.Tests/CustomCode/VolumeTests.cs
```diff
@@ -75,6 +77,8 @@ public class VolumeTests : VolumeTestsBase
 
         protected override double TeaspoonsInOneCubicMeter => 202884.13621105801;
 
+        protected override double TotesInOneCubicMeter => 0.9606256824;
```
anjdreas: I don't think I consider tote to be a unit of measurement. It is a physical container that has some common sizes widely used, but this definition belongs in an application, not in Units.NET.

Source, wiki:
> IBCs can be manufactured to a customer's exact requirements in terms of capacity, dimensions, and material. The length and width of an IBC are usually dependent on the pallet dimension standard of a given country

and

> The most common sizes are 1,040 liters or 275 U.S. gallons or 229 imperial gallons and 1,250 liters or 330 U.S. gallons or 275 imperial gallons (the 1040 liter IBCs are often listed as being 1000 liters)

https://en.wikipedia.org/wiki/Intermediate_bulk_container

The same reasoning on drums.
If you agree, please remove these two definitions from the PR.

anjdreas:
The Acre and Hectare units look good and are ready to merge.

anjdreas:
Just checking up on this, what are your thoughts on my comments?

BMcFerrin:
I agree with your comments, I just haven't taken the time to make the
changes you suggested. I'll do that soon.

anjdreas:
Alright, no problem. Leave a comment when you push changes, I am not always notified by github.

BMcFerrin:
Alright, I just quickly removed the Tote and Drum units.

anjdreas:
Great, thanks. Nuget 3.41.0 is out.

PR #187, opened 2016-09-29
Add VolumeUnit.OilBarrel and FlowUnit.OilBarrelsPerDay

BrandonLWhite:
https://en.wikipedia.org/wiki/Barrel_(unit)#Oil_barrel

### DIFF
--- a/UnitsNet.Tests/CustomCode/FlowTests.cs
+++ b/UnitsNet.Tests/CustomCode/FlowTests.cs
@@ -50,5 +50,7 @@ namespace UnitsNet.Tests.CustomCode
         protected override double DecilitersPerMinuteInOneCubicMeterPerSecond => 600000.00000;
 
         protected override double KilolitersPerMinuteInOneCubicMeterPerSecond => 60.00000;
+
+        protected override double OilBarrelsPerDayInOneCubicMeterPerSecond => 543439.65056533388306722269588172;
     }
 }
--- a/UnitsNet.Tests/CustomCode/VolumeTests.cs
+++ b/UnitsNet.Tests/CustomCode/VolumeTests.cs
@@ -89,6 +89,8 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double UsLegalCupsInOneCubicMeter => 4166.666666667;
 
+        protected override double OilBarrelsInOneCubicMeter => 6.2898107704321051280928552764086;
+
         [Test]
         public void VolumeDividedByAreaEqualsLength()
         {
--- a/UnitsNet.Tests/GeneratedCode/FlowTestsBase.g.cs
+++ b/UnitsNet.Tests/GeneratedCode/FlowTestsBase.g.cs
@@ -49,6 +49,7 @@ namespace UnitsNet.Tests
         protected abstract double MillilitersPerMinuteInOneCubicMeterPerSecond { get; }
         protected abstract double MillionUsGallonsPerDayInOneCubicMeterPerSecond { get; }
         protected abstract double NanolitersPerMinuteInOneCubicMeterPerSecond { get; }
+        protected abstract double OilBarrelsPerDayInOneCubicMeterPerSecond { get; }
         protected abstract double UsGallonsPerMinuteInOneCubicMeterPerSecond { get; }
 
 // ReSharper disable VirtualMemberNeverOverriden.Global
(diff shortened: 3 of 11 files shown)


anjdreas:
First off, thanks for the PR.
I have a couple of concerns with this unit, though.

I am not familiar with this unit myself, but OilBarrel seems maybe a bit ambiguous? it seems similar to #179, where the PR had units Tote and Drum. 
From the wiki link you posted, emphasis mine:

> An oil barrel (abbreviated as bbl) is a unit of volume whose definition has **not been universally standardized**. In the United States, an oil barrel is defined as 42 US gallons, which is about 159 litres[10] or 35 imperial gallons.[11] In Canada, oil companies measure oil in cubic metres but convert to barrels on export, since most of Canada's oil production is exported to the US. The nominal conversion factor is 1 cubic metre = 6.2898 oil barrels,[12] but conversion is generally done by custody transfer meters on the border since the **exact conversion factor depends on oil density and temperature**.
>
>**Outside the United States, volumes of oil are usually reported in cubic metres (m3) instead of oil barrels**. Cubic metre is the basic volume unit in the International System. Canadian companies operate internally and report to Canadian governments in cubic metres, but often convert to US barrels for the benefit of American investors and oil marketers. They will generally quote prices in Canadian dollars per cubic metre to other Canadian companies, but use US dollars per barrel in financial reports and press statements, making it appear to the outside world that they operate in barrels.

A couple of examples showing a variety of conversions with barrels:
http://www.metric-conversions.org/volume/barrels-to-liters.htm

From this, I could be swayed to let OilBarrel be defined per-application instead of in Units.NET, however, I have not found any conflicting definitions on Oil Barrel volume, so it might be that if we call this `UsOilBarrel` instead of `OilBarrel`, that would be sufficiently clear. What do you think?

On a sidenote, I saw some Drum and Tote unit code were still left in source code although that should have been removed in #179, but this PR will clean that up. The build server always regenerates the code before creating the nuget anyway, so it was never shipped.

BrandonLWhite:
Seeing the Drum and Tote diffs show up after I ran GenerateUnits.bat confused me for a bit, but I eventually figured out what was going on.

I can assure you that bbl is very much a standard in a market that generates countless billions $USD.

Yes, many of our international customers use m3 which our software certainly supports.  However, much of the academic equations that are involved in various scientific analysis use bbl and bbl/d.

Some more examples:
https://www.statista.com/topics/1783/global-oil-industry-and-market/
http://www.bloomberg.com/energy

Even this javascript library supports it: http://mathjs.org/docs/datatypes/units.html  We use this library in our UI.  (Note though, they have made the unfortunate mistake of labeling beer barrels as 'bbl'.  That should be 'bl' and oil barrels 'bbl'.  There are many references out there to confirm this).

I'm happy to change the enum to UsOilBarrel, but in my opinion, that would be redundant.  As you indicated, there is no other definition of OilBarrel in the energy industry.

Regarding the conversion depending upon density and temperature.  This comes into play when you are paying money for a particular amount of energy.  When you start moving farther up stream (trading, financials), they like to try to normalize things so that different oil supplies with different energy potentials can look the "same" on the books.  So, for example, oil with less energy potential gets scaled down on its volume, contrary to physical reality.  But these industries are not actually dealing with real-world volumes.  Truthfully, they should be using a unit like BTU or joules.  FWIW natural gas delivery amounts are often expressed in BTU.  True to form though there are many conversions that attempt to convert a volume of natural gas into a BTU equivalent.  But I digress -- It really is irrelevant when talking about units of measure.  A physical volume of oil always has an unambiguous conversion to any other volume of unit.

The unit conversions are unambiguous and universally adopted in the industry.  Companies would literally be sued (at worst) or lose business (at best) if they deviated from such.  And yeah, it used to happen all the time -- but that is ancient history.

BrandonLWhite:
Google agrees: https://www.google.com/search?btnG=1&pws=0&q=1+oil+barrel+to+m3

There is only one "Oil Barrel"

anjdreas:
Nice read :-) :+1: 
I trust your judgment here, you seem to know this domain very well. If you think OilBarrel is sufficiently distinct, I'm happy to merge this in.

BrandonLWhite:
My pleasure, and thanks for providing this library.

anjdreas:
Nuget 3.42.0 is out.

Issue #186, opened 2016-09-22
Support for runtime conversion configuration


BrandonLWhite:
In our industrial/scientific application, we have a need for allowing a user to configure an arbitrary floating point value acquired via digital acquisition (DAQ) into a unit of measure that they specify.  Thus, the UnitClass must be specified in addition to the *Unit type.

We must then be able to convert this value to some other arbitrary *Unit within the same unit class in various calculations, views, reports, logging/trending and what not, again dictated by user configuration.

It would be awesome if UnitsNet could support this directly.  What is needed:
1. Enum for UnitClasses to support runtime configuration.
2. A runtime conversion function such as `double Convert(UnitClass class, Unit from, Unit to, double value)'

anjdreas:
This is a topic that regularly comes up, and I have wanted it myself at various times, but then I also realized that converting between arbitrary, possibly incompatible, units is a slippery slope. There are very few inter-unit conversions, such as `Mass.FromGravitationalForce(Force f)` and operator overloads for common arithmetic, such as `Speed * TimeSpan = Length`. These are the exceptions.

If we were to represent units generically, I fear we would promote use where type safety is out the door. Type safety is one of the most valuable benefits to me in this library. Sure, giving it up would be optional, but it's something about helping people fall into the pit of success, so to speak. 

These problems can typically be solved while staying type-safe. I don't know your application, but I would assume your user would have to choose what type of unit is coming in, so I would think it was natural to first select Force category, then Newton or Kilogram next. This allows you to work on Force measurements generically, while also knowing what specific ForceUnit the user chose, so you can easily present texts of readings using `myForce.ToString(userSelectedForceUnit)`.
If you need to convert it, you know it is a Force measurement, so you list all the other ForceUnit enum values for him to convert to, then use `myForce.As(newForceUnit)`. It would not make sense to list Volume units in this case anyway.

My philosophy is that an application doesn't work with universal units if there are conversions involved, you have your set of types of units (Length, Mass, Volume) you care about, and you conditionally handle these separately. For reports/graphs, you usually fall back to numbers and unit abbreviation strings, but you don't really need to know more than that just to present it. If you need to convert them, you need to know a lot more.

[...]

PR #190, opened 2016-09-30
UnitClass enum

BrandonLWhite:
Hi, this is an incremental step for #186.  This simply adds the PowerShell scripting for generating a UnitClass enum which would then be available for use in external code.

Please review for naming, file locations, etc.

Merging in this change would be helpful to me as I could then remove my hand-rolled enum in my application and adopt the UnitsNet.UnitClass enum instead.


anjdreas:
Awesome, looks good to me.

PR #194, opened 2016-10-18
Force from mass and acceleration overloads

JKSnd:
Previously there was only Force.FromMassByAcceleration(Mass, double)
This pull request adds Force.FromMassByAcceleration(Mass, Acceleration)

### DIFF
--- a/UnitsNet.Tests/CustomCode/ForceTests.cs
+++ b/UnitsNet.Tests/CustomCode/ForceTests.cs
@@ -68,6 +68,13 @@ namespace UnitsNet.Tests.CustomCode
             Acceleration acceleration = Force.FromNewtons(27)/Mass.FromKilograms(9);
             Assert.AreEqual(acceleration, Acceleration.FromMeterPerSecondSquared(3));
         }
+    
+        [Test]
+        public void ForceDividedByAccelerationEqualsMass()
+        {
+          Mass acceleration = Force.FromNewtons(200)/Acceleration.FromMeterPerSecondSquared(50);
+          Assert.AreEqual(acceleration, Mass.FromKilograms(4));
+        }
 
         [Test]
         public void MassByAccelerationEqualsForceUsingDouble()
--- a/UnitsNet.Tests/CustomCode/ForceTests.cs
+++ b/UnitsNet.Tests/CustomCode/ForceTests.cs
@@ -61,8 +61,7 @@ namespace UnitsNet.Tests.CustomCode
             var force = Force.FromPressureByArea(Pressure.FromNewtonsPerSquareMeter(6), Length2d.FromMeters(5, 2));
             Assert.AreEqual(force, Force.FromNewtons(60));
         }
-
-
+    
         [Test]
         public void ForceDividedByMassEqualsAcceleration()
         {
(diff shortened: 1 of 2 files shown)


UnitsNet/CustomCode/UnitClasses/Force.extra.cs
```diff
@@ -70,5 +70,10 @@ public static Force FromMassByAcceleration(Mass mass, double metersPerSecondSqua
         {
             return new Force(mass.Kilograms*metersPerSecondSquared);
         }
+
+        public static Force FromMassByAcceleration(Mass mass, Acceleration acceleration)
```
anjdreas: Not sure you noticed, but the buildserver failed on this overload when compiling the WinRT solution. If you are not familiar with WinRT Components (I wasn't until I was asked to support it), it has very strict requirements on public members in order to support WinJS and C++ UWP apps.

Anyhow. we can't have overloads in WinRT, so I suggest we wrap the `double` overload in a `#if !WINDOWS_UWP  ... #endif` to exclude it from WinRT. It would be a breaking change for that target, but I know for a fact that the number of WinRT users is very low and the chance they are using this particular method is even lower. That one guy, if he exists, can deal with the breaking change. YOLO!

anjdreas: Output from build server:
```
[17:46:48][Step 1/2] "C:\TeamCity\buildAgent\work\6c2ac6daa7a037ba\UnitsNet\UnitsNet.WindowsRuntimeComponent.csproj" (Build target) (4) ->
[17:46:48][Step 1/2] (ExportWindowsMDFile target) -> 
[17:46:48][Step 1/2]   C:\TeamCity\buildAgent\work\6c2ac6daa7a037ba\UnitsNet\CustomCode\UnitClasses\Force.extra.cs(25,33,25,38): error WME1085: The 2-parameter overloads of UnitsNet.Force.FromMassByAcceleration must have exactly one method specified as the default overload by decorating it with Windows.Foundation.Metadata.DefaultOverloadAttribute. [C:\TeamCity\buildAgent\work\6c2ac6daa7a037ba\UnitsNet\UnitsNet.WindowsRuntimeComponent.csproj]
[17:46:48][Step 1/2] 
[17:46:48][Step 1/2]     28 Warning(s)
[17:46:48][Step 1/2]     1 Error(s)
```

http://teamcity.chump.work/viewLog.html?guest=1&buildId=735&buildTypeId=UnitsNet_BuildTest&tab=buildLog#_

UnitsNet.Tests/CustomCode/ForceTests.cs
```diff
@@ -49,13 +49,34 @@ public void ForceDividedByAreaEqualsPressure()
         }
 
         [Test]
+        public void PressureByAreaEqualsForce()
+        {
+            Force force = Force.FromPressureByArea(Pressure.FromNewtonsPerSquareMeter(9), Area.FromSquareMeters(9));
```
anjdreas: I would prefer using different numbers as arguments, since in theory you are open to bugs where only one of the arguments are used. It's highly theoretical, but I'd still change it just for the better practice.

UnitsNet.Tests/CustomCode/ForceTests.cs
```diff
@@ -49,13 +49,34 @@ public void ForceDividedByAreaEqualsPressure()
...
+            Assert.AreEqual(force, Force.FromNewtons(81));
+
+            force = Force.FromPressureByArea(Pressure.FromNewtonsPerSquareMeter(9), Length2d.FromMeters(3, 3));
```
anjdreas: It is not very important to me, but I would consider breaking this up into two tests. One with Area and one with Length2d. Just add a suffix to the test method name, such as `PressureByAreaEqualsForce_Length2d`. Same for the other test `MassByAccelerationEqualsForce`.

JKSnd:
Thanks for the feedback!

I wasn't aware of the WinRT restriction, so thank you for pointing that out. The failure notification I was emailed wasn't quite detailed enough for me to discover it was that particular overload. It just told me not enough tests were run.

For the tests, I'm totally cool with being as strict as you like about arguments and assertions in each test method. Happy to make those edits!

UnitsNet.Tests/CustomCode/ForceTests.cs
```diff
@@ -49,13 +49,38 @@ public void ForceDividedByAreaEqualsPressure()
...
             Assert.AreEqual(acceleration, Acceleration.FromMeterPerSecondSquared(3));
         }
 
         [Test]
+        public void MassByAccelerationEqualsForce()
+        {
+            Force force = Force.FromMassByAcceleration(Mass.FromKilograms(9), Acceleration.FromMeterPerSecondSquared(3));
+            Assert.AreEqual(force, Force.FromNewtons(27));
+
+            force = Force.FromMassByAcceleration(Mass.FromKilograms(9), 3);
```
anjdreas: Could you split up this one too, while at it?

anjdreas:
Nuget 3.43 is out.

PR #197, opened 2016-10-25
Created AcPotential unit

JKSnd:
I need to represent electrical signals in such a way that I can quickly determine their peak, peak to peak, and RMS values quickly. Integration with AmplitudeRatio is crutial. Assumes signal is a sinusoid.

I didn't believe it was appropriate to extend ElectricPotential since further development (aka adding a Frequency property) would muddy the intent of ElectricPotential. So far EP feels very much like a DC representation of the same.

It could be argued that DC potential is a special case of AC, however that requires polymorphic RMS calculations and is honestly overkill for what I need at the moment.

### DIFF
--- a/UnitsNet.Tests/CustomCode/AcPotentialTests.cs
+++ b/UnitsNet.Tests/CustomCode/AcPotentialTests.cs
@@ -33,20 +33,20 @@ namespace UnitsNet.Tests.CustomCode
     [Test]
     public void FromElectricPotential()
     {
-      Assert.AreEqual(2.5, AcPotential.FromPeakElectricalPotential(ElectricPotential.FromMillivolts(2500)).VoltsPeak);
-      Assert.AreEqual(102.5, AcPotential.FromRMSElectricalPotential(ElectricPotential.FromVolts(102.5)).VoltsRMS);
+      Assert.AreEqual(2.5, AcPotential.FromPeakElectricalPotential(ElectricPotential.FromMillivolts(2500)).VoltsPeak, VoltsPeakTolerance);
+      Assert.AreEqual(102.5, AcPotential.FromRMSElectricalPotential(ElectricPotential.FromVolts(102.5)).VoltsRMS, VoltsRMSTolerance);
     }
 
     [Test]
     public void TakesAbsoluteValueOfInput()
     {
-      Assert.AreEqual(1.4, new AcPotential((decimal) -1.4).VoltsPeak);
-      Assert.AreEqual(12,new AcPotential((long) -12).VoltsPeak);
-      Assert.AreEqual(1.2,new AcPotential(-1.2).VoltsPeak);
-      Assert.AreEqual(15,new AcPotential(-15).VoltsPeak);
-      Assert.AreEqual(2,AcPotential.FromVoltsPeak(-2).VoltsPeak);
-      Assert.AreEqual(2.45,AcPotential.FromVoltsPeakToPeak(-2.45).VoltsPeakToPeak);
-      Assert.AreEqual(2.78,AcPotential.FromVoltsRMS(-2.78).VoltsRMS);
+      Assert.AreEqual(1.4, new AcPotential((decimal) -1.4).VoltsPeak, VoltsPeakTolerance);
+      Assert.AreEqual(12,new AcPotential((long) -12).VoltsPeak, VoltsPeakTolerance);
+      Assert.AreEqual(1.2,new AcPotential(-1.2).VoltsPeak, VoltsPeakTolerance);
+      Assert.AreEqual(15,new AcPotential(-15).VoltsPeak, VoltsPeakTolerance);
+      Assert.AreEqual(2,AcPotential.FromVoltsPeak(-2).VoltsPeak, VoltsPeakTolerance);
+      Assert.AreEqual(2.45,AcPotential.FromVoltsPeakToPeak(-2.45).VoltsPeakToPeak, VoltsPeakToPeakTolerance);
+      Assert.AreEqual(2.78, AcPotential.FromVoltsRMS(-2.78).VoltsRMS, VoltsRMSTolerance);
     }
   }
 }
--- a/UnitsNet.Tests/CustomCode/AcPotentialTests.cs
+++ b/UnitsNet.Tests/CustomCode/AcPotentialTests.cs
@@ -41,11 +41,11 @@ namespace UnitsNet.Tests.CustomCode
     public void TakesAbsoluteValueOfInput()
     {
       Assert.AreEqual(1.4, new AcPotential((decimal) -1.4).VoltsPeak, VoltsPeakTolerance);
-      Assert.AreEqual(12,new AcPotential((long) -12).VoltsPeak, VoltsPeakTolerance);
-      Assert.AreEqual(1.2,new AcPotential(-1.2).VoltsPeak, VoltsPeakTolerance);
-      Assert.AreEqual(15,new AcPotential(-15).VoltsPeak, VoltsPeakTolerance);
-      Assert.AreEqual(2,AcPotential.FromVoltsPeak(-2).VoltsPeak, VoltsPeakTolerance);
-      Assert.AreEqual(2.45,AcPotential.FromVoltsPeakToPeak(-2.45).VoltsPeakToPeak, VoltsPeakToPeakTolerance);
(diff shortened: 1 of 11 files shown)


anjdreas:
Cool, I'll take a look at this tomorrow.

JKSnd:
Hmmm. I'm getting a "works on my machine" situation. The test AcPotentialTests.TakesAbsoluteValueOfInput fails on the server but passes locally for me. I'm very confused by this.

Any place you would suggest for me to start looking?

anjdreas:
Build server says:

http://teamcity.chump.work/viewLog.html?buildId=817&buildTypeId=UnitsNet_BuildTest

```
Expected: 1.3999999999999999d +/- 1.0000000000000001E-05d
  But was:  -1.3999999999999999d
at UnitsNet.Tests.CustomCode.AcPotentialTests.TakesAbsoluteValueOfInput() in C:\TeamCity\buildAgent\work\6c2ac6daa7a037ba\UnitsNet.Tests\CustomCode\AcPotentialTests.cs:line 44
```

Sure you committed and pushed all your local changes?

JKSnd:
I can navigate the source on my fork and see the calls to Math.Abs() that I would expect.

Does the build server run the generation scripts each time instead of using the checked in g.cs files? If that's the case then it makes sense that the Math.Abs() calls I added to the constructors are being blown away. Would it be possible to add an "Absolute" property to the json file and include logic for it in its parser? Or would it make more sense to just crystalize this special case by defining the unit class purely as "custom code"?

anjdreas:
Ah, yes buildserver regenerates the code, and we can't have custom code in the generated files anyhow. The json function can be anything you want, so add System.Math.Abs(x) or similar and I believe that should work. 

On the phone here so limited ability to look into the details.

JKSnd:
I suppose I could wrap each "From___to____Func" definition in a Math.Abs(), but doing so doesn't really capture the intent that this unit is meant to be unsigned. I'll go with it for this one, and keep an eye out for the next time such a unit pops up. Maybe we can figure out an unsigned unit generation convention then?

anjdreas:
Absolutely, let me know what you think would work best and we'll discuss
options.

JKSnd:
Awesome!

I like what you did with the json files for logarithmic units. Perhaps something similar, like "Absolute"?

So the definition for AcPotential would read

```
{
  "Name": "AcPotential",
  "BaseUnit": "VoltPeak",
  "Absolute": "True",
  "XmlDoc": "A sinusoidally time-varying representation of ElectricPotential."
  "Units": [ ...........]
}
```

And this would tell the generation script to wrap the assignment logic in a Math.Abs call. Something like

 ```
public AcPotential(double voltspeak)
{
    _voltsPeak =Math.Abs(Convert.ToDouble(voltspeak));
}
```

It could be used to add a line to the summary to let developers know that the absolute value of whatever is passed in will be taken. Like...

 ```
    /// <summary>
    ///     A sinusoidally time-varying representation of ElectricPotential.
    ///     This is an absolute-valued unit. 
    /// </summary>
    public partial struct AcPotential : IComparable, IComparable<AcPotential>
```

JKSnd:
Also how do you feel about removing the generated files from the repository and adding a pre-build step to the projects to run the generator?

When I started contributing to this project I assumed the generation was to make the initial creation of units easy and nothing more, since the g.cs files were checked in. I feel like I would have been more likely to "fall into the pit of success" if each build re-generated those files because I would have had that test fail locally before pushing.

anjdreas:
Thanks, your proposal looks good to me. 
A couple of questions:

1. Do I understand correctly AcPotential would never make sense to have a negative value, regardless of unit?
2. Would it make the most sense to throw an exception or to take the absolute in the ctor?
3. Do you know of any other unit classes or units, new or existing, we might want to implement the same behavior for?
4. Would it be best to define the `Absolute` decorator per unit class or per unit? Possibly both?

See [Remove generated code from repo #199](https://github.com/anjdreas/UnitsNet/issues/199)

JKSnd:
> Do I understand correctly AcPotential would never make sense to have a negative value, regardless of unit?

Short answer: Yep, you've got it.

Longer answer: I would eventually like to represent AcPotential with three properties: Amplitude [Volts], Frequency [Hertz], and Phase [Angle] (I skimmed the units and didn't see a unit yet that was comprised of existing units, so I compromised and just offered up the Amplitude part of it). In the math world, you represent sinusoidal signals (roughly) as A*cos(f*t + θ) (where A is amplitude, f is frequency, t is time, and θ is phase). Expressing a signal as -A*cos(f*t + θ) is equivalent to A*cos(f*t + Θ) where Θ is (θ +/- 180 degrees). So, having sign information in the amplitude of a signal is really shorthand for a 180 degree phase shift.
BONUS: an AcPotential doesn't need to only be a sinusoid. It could be a triangle wave, a square wave, etc. Each of those would calculate their RMS values differently.

> Would it make the most sense to throw an exception or to take the absolute in the ctor?
If we go full blown Amplitude, Frequency, Phase then I think the ctor should take the absolute of the ampiltude and apply the sign as a phase shift. In the extremely simplified case right now, I don't have a strong opinion either way. Although taking the absolute would set a precedent for later when phase is added.

> Do you know of any other unit classes or units, new or existing, we might want to implement the same behavior for?

- AlternatingCurrent (the brother to AcPotential)
- Anything that can be represented by a [phasor](https://en.wikipedia.org/wiki/Phasor) (any time-varying quantity)

> Would it be best to define the Absolute decorator per unit class or per unit? Possibly both?
I think per unit class, if we go this route. In my mind it unifies the behaviour of the units, whereas if it were per unit then some could be signed and others not. Its not that this would be undesirable, but no use case immediately comes to mind for it.

anjdreas:
Ok, so a few things here. How do you propose we proceed? Merge in the Math.Abs() for now and find an alternative way to represent it later, or do that already now?

>  I would eventually like to represent AcPotential with three properties: Amplitude [Volts], Frequency [Hertz], and Phase Angle. 

This would probably be best implemented by hand in UnitsNet, instead of trying to generate it. 
The problem is, this approach does not mix well with the generated code approach, where we are only able to add methods/properties to the generated unit class - not modify things like storage fields.

How about this:
* `AcPotential` is generated with JSON, as per the current PR status.
* You add `AcPotentialAfp` with three additional readonly fields/props:
  * `ElectricPotential Amplitude { get; }`
  * `Frequency Frequency { get; }`
  * `Angle Phase { get; }`
* It derives from `AcPotential`, so all its generated goodness is reused, and it constructs the `AcPotential` base class with a value calculated from its AFP fields.
* Users of `AcPotentialAfp` can access the AFP values, the base value, and all the conversion methods
* Users of `AcPotential` can only access the base value and the conversion methods

Where do we put `AcPotentialAfp`?
----
For your reference, the CustomCode folder is intended for custom overrides of code in GeneratedCode folder. See a pretty good [discussion](https://github.com/anjdreas/UnitsNet/issues/46) about namespaces and nuget structures. We decided to keep everything in Core for now, and we did find that moving things into engineering/scientific domain namespaces is not trivial due to overlaps.

Following that thought, we could place `AcPotentialAfp` in the core lib and the core namespace for now - or consider putting it into `ElectricalEngineering` folder/namespace. I am really not sure.

JKSnd:
Hmmmmm maybe this is actually stretching the scope of Units.Net a little too far. AC Potential and Alternating current are concepts that are composed of properties that have units, rather than being a unit.

Does it feel like it would be more appropriate to create as a part of a physics toolkit project that's separate to Units.Net?

anjdreas:
I agree with that. It's something about keeping the core lean and have a single purpose - working with units and conversions. Compositions of units that make sense in a particular scientific/business domain might be best put into separate libraries. We can still host that code here if that makes the most sense, especially if it becomes very coupled to UnitsNet and wants to share the `UnitsNet.` prefix in the nuget name etc. It would be nice to have a one-stop-shop for units in .NET, beyond the core units.

So, merge this in?

JKSnd:
Awesome! I feel good about this discussion.

I actually think we should hold off on merging this one in. I wouldn't want to pollute the core since this is really more of an abstraction over ElectricPotential. I think something of this sort would be best hosted in a separate signals project.

On my end there is no major rush for this requirement, so I'm happy to wait and make sure that the concept lives in the correct place.

Ok to close this PR without merging?

anjdreas:
Ok by me. Let me know when you make some progress.

PR #202, opened 2016-10-30
Generate code in pre-build step

### LINKED ISSUE #199: Remove generated code from repo
Created from [comment](https://github.com/anjdreas/UnitsNet/pull/197#issuecomment-256794748) by @JKSnd.

> Also how do you feel about removing the generated files from the repository and adding a pre-build step to the projects to run the generator?
> 
> When I started contributing to this project I assumed the generation was to make the initial creation of units easy and nothing more, since the g.cs files were checked in. I feel like I would have been more likely to "fall into the pit of success" if each build re-generated those files because I would have had that test fail locally before pushing.

anjdreas:
To make it more clear that users should not edit these files directly. Will reset their changes on each local build, similar to what build server does.

Fixes #199.


anjdreas:
@JKSnd for your information. I'll just merge this in right away.

Issue #135, opened 2016-02-08
Strong named assembly


tongbong:
One of our client only integrate strong named assemblies...
https://msdn.microsoft.com/en-us/library/wd40t7ad%28v=vs.110%29.aspx

It seams easy to do:
https://msdn.microsoft.com/en-us/library/xc31ft41%28v=vs.110%29.aspx

Any thought about this?

anjdreas:
I don't have any experience with that,  but know what it is about and open to the idea.
I recall some pain points discussed about nuget and strong naming a long time ago, but I don't recall exactly what it was all about. Something about assembly versioning is the only thing that comes to mind.

anjdreas:
FYI. I am away all this week and will not be able to follow up before next week.

anjdreas:
Do you want to pursue this and add a PR? I don't have much time to put into this project these days unfortunately.

anjdreas:
As I thought, there is a lot of debate on whether to strong name sign open source nuget libs.

https://github.com/octokit/octokit.net/issues/405#issuecomment-41640976
https://github.com/reactiveui/ReactiveUI/issues/43#issuecomment-37471892

Based on these, I am inclined to NOT sign Units.NET. Open to discuss, naturally.

bplett:
I understand your hesitancy in strong-naming, but what has me quite confused is that it appears you have a certificate in each of the source projects, and are essentially strong-naming the source, but not the NuGet package.  I would think this should be done the other way around:  don't strong-name the source, so that other developers can build without having the certificate problems, but strong-name the NuGet package so that those who don't need to compile the source can simply use the NuGet package in their own strongly named solutions.

bplett:
Oops - ignore that cryptic reference to how you've already strongly named the source.  I was looking at my fork.

Still, the point remains.  I agree - don't sign the source.  However, please sign the NuGet so that people can use it without having to compile it.  :-)

anjdreas:
@bplett I'm not sure I follow. what do you mean by signing the nuget but not signing the projects?

Is this what you are referring to?
https://github.com/aspnet/Signing/blob/dev/Spec.md

WGVDNSoftware:
[...]

PR #206, opened 2016-11-11
Add strong name signed core nuget

anjdreas:
Draft for strong name signing as discussed in #135.


anjdreas:
Heads up on strong name signing key now being published and also changed to a new key:
https://github.com/anjdreas/UnitsNet/issues/262

PR #208, opened 2016-11-21
Added new unit ForcePerLenght

VBP8501:
Hey Andreas, thanks for your splendid lib UnitsNet! I hereby wanted to contribute with the units that we are missing in your lib for our project...


JKSnd:
Hi @VBP8501, is there a more specific name that could be applied to this measure? I know that surface tension is expressed in force per unit length, but I suspect that other concepts may use it as well.

Also, please consider adding operator overloads that would produce ForcePerLength values, or vice versa. Like a division operator in the Force unit so Force/Length =FpL and a multiplication in FpL such that FpL*Length = Force. I imagine those would be convenient

VBP8501:
I was thinking of 'distributed load' as an alternative name but that might be too specific for my application. For example the weight of a cable is expressed as per unit length. 

I was not aware of the possibility to add the operators. I will look into that tomorrow. Thx for the suggestion

JKSnd:
I suppose it would depend on the direction of the forces, wouldn't it? A distributed load, to me (I'm an EE, not a civil engineer), that the forces act perpendicularly to the plane that the member lies on. Surface tension would be tangential, I think. It maybe not matter, but I thought it was worth bringing up.

Is the directionality of the forces a requirement for your project? If so, please see if [Pk.Spatial](https://github.com/PkSound/Spatial/) could be useful to you.

anjdreas:
@VBP8501 Thanks, I'm not familiar with these units and would not know what naming convention to use. ForcePerLength is descriptive enough to me and makes sense for a multiplication overload to get Force.
DistributedLoad does not tell me as much, such as whether it is 1D or 2D (or possibly 3D?).

Some Google results of "force per length"

Magnetic force between wires:
http://hyperphysics.phy-astr.gsu.edu/hbase/magnetic/wirfor.html#c1

Force per unit length or area:
https://www.comsol.com/community/forums/general/thread/94621/

Surface tension:
http://www.translatorscafe.com/cafe/EN/units-converter/surface-tension/8-1/pound-force%2Finch-newton%2Fmeter/

So far, ForcePerLength seems a good, generic choice to me. What do you think?

VBP8501:
@JKSnd: thanks for the suggestion. Directionality is a requirement but not a problem that needs to be solved with a specific library. I'll keep it in mind though for other projects.

@anjdreas: I agree, I also did a google search for a standard naming but could not come up with a better one.

I made the operator overload and will do a new pull request

anjdreas:
@VBP8501 You can just update this PR, no need to do a second PR. There are many resources and guides on this, here is one: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

Basically just push new commits to the same branch on your fork, and rebase/force push if you want to rewrite the history of the branch, such as squashing commits together.

VBP8501:
Yes but I thought it better to keep the history (extra commit for the operator overloads)

anjdreas:
Sure, just add another commit to this PR then, if that is what you meant?

PR #214, opened 2016-12-09
Created USSurveyFoot unit

mludlum:
Equal to exactly 1200/3937 meters, a US Survey foot differentiates from the International foot because of existing survey data based on an old definition from 1893.  

1 ft = 0.999998000004 ft (survey)

The difference is important in large scale mapping applications.

https://en.wikipedia.org/wiki/Foot_(unit)#Survey_foot

### DIFF
--- a/UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj
+++ b/UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj
@@ -67,9 +67,6 @@
       <Name>UnitsNet.Net35</Name>
     </ProjectReference>
   </ItemGroup>
-  <ItemGroup>
-    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
-  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.
--- a/UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj
+++ b/UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj
@@ -67,6 +67,9 @@
       <Name>UnitsNet.Net35</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
+  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
        Other similar extension points exist, see Microsoft.Common.targets.
--- a/UnitsNet.Tests/CustomCode/LengthTests.cs
+++ b/UnitsNet.Tests/CustomCode/LengthTests.cs
@@ -33,7 +33,7 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double FeetInOneMeter => 3.28084;
 
-        protected override double USSurveyFeetInOneMeter => 3.2808333333465;
+        protected override double USSurveyFeetInOneMeter => 3937/1200;
 
         protected override double InchesInOneMeter => 39.37007874;
 
--- a/UnitsNet.Tests/CustomCode/LengthTests.cs
+++ b/UnitsNet.Tests/CustomCode/LengthTests.cs
@@ -19,6 +19,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+using System;
 using NUnit.Framework;
 using UnitsNet.Units;
 
@@ -31,6 +32,8 @@ namespace UnitsNet.Tests.CustomCode
         protected override double DecimetersInOneMeter => 10;
 
         protected override double FeetInOneMeter => 3.28084;
+
+        protected override double USSurveyFeetInOneMeter => 3.2808333333465;
 
         protected override double InchesInOneMeter => 39.37007874;
 
--- a/UnitsNet.Tests/CustomCode/LengthTests.cs
+++ b/UnitsNet.Tests/CustomCode/LengthTests.cs
@@ -33,7 +33,7 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double FeetInOneMeter => 3.28084;
 
-        protected override double USSurveyFeetInOneMeter => 3937.0/1200.0;
+        protected override double UsSurveyFeetInOneMeter => 3.280833333333333;
 
         protected override double InchesInOneMeter => 39.37007874;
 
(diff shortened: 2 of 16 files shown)


JKSnd:
@anjdreas  This looks good to me, but would you please double check?

UnitsNet.Tests/CustomCode/LengthTests.cs
```diff
@@ -33,7 +33,7 @@ public class LengthTests : LengthTestsBase
 
         protected override double FeetInOneMeter => 3.28084;
 
-        protected override double USSurveyFeetInOneMeter => 3.2808333333465;
+        protected override double USSurveyFeetInOneMeter => 3.2808333333;
```
anjdreas: Why would we not want the most accurate factor? What do you mean by "eliminate truncation error"?

UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj
```diff
@@ -67,6 +67,9 @@
       <Name>UnitsNet.Net35</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
anjdreas: This was probably added without intention? If so, remove it.

UnitsNet/Scripts/UnitDefinitions/Length.json
```diff
@@ -70,6 +70,18 @@
       ]
     },
     {
+      "SingularName": "USSurveyFoot",
+      "PluralName": "USSurveyFeet",
+      "FromUnitToBaseFunc": "x*0.3048006096",
+      "FromBaseToUnitFunc": "x/0.3048006096",
```
anjdreas: As per your comment
> Equal to exactly 1200/3937 meters

and also swapping the functions (see my previous comment about your test value being wrong), I would change these conversions to:
```
"FromUnitToBaseFunc": "x*3937/1200",
"FromBaseToUnitFunc": "x*1200/3937",
```

Now this reads as expected; to convert one meter (base unit) to USS feet, multiply by 1200 and divide by 3937, a factor of about 0.3.

This way we use the most accurate definition instead of an artificially limited factor. If we later go from double to decimal, it would automatically get a better conversion precision.

UnitsNet.Tests/CustomCode/LengthTests.cs
```diff
@@ -32,6 +33,8 @@ public class LengthTests : LengthTestsBase
 
         protected override double FeetInOneMeter => 3.28084;
 
+        protected override double USSurveyFeetInOneMeter => 3.2808333333;
```
anjdreas: I think you got this backwards. My sanity check says USS Feet should be about 0.3 meters. Check your conversion funtions. It's very easy to get this backwards.

UnitsNet/Scripts/UnitDefinitions/Length.json
```diff
@@ -70,6 +70,18 @@
...
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "ftUS" ]
```
anjdreas: This link says `ft US`: http://georeference.org/doc/units.htm

What do you think, both or one of them?

anjdreas:
@JKSnd Found a couple of things, see my review. It's particularly important to verify the test values are correct. It was not in this case and would give wrong values in conversions, even if all your tests are green. 

Most often the problem is that the conversion functions are defined "backwards", as in this case, and should be swapped. Maybe the naming could be improved to avoid that, not sure.

Pro tip: Often you can just copy and paste the property name `USSurveyFeetInOneMeter` into google and it will give you the conversion you need, but if not I usually go find an online converter.

UnitsNet.Tests/CustomCode/LengthTests.cs
mludlum: Probably a bad commit comment but one of the tests failed because I used a truncated value to get the above value.  Your idea of putting the fractions in is better.   So, it should be 3937/1200

UnitsNet.Tests/CustomCode/LengthTests.cs
mludlum: I'm confused.  If mine is backwards, isn't the previous property "FeetInOneMeter" also backwards?  I just followed suit from FeetInOneMeter.  A U.S. Survey Foot is very close to an International Foot.

UnitsNet/Scripts/UnitDefinitions/Length.json
mludlum: I propose both or just "ftUS". I can't find an independent authority on the abbreviation.   The source you sited is a forum for geo software called the Manifold System which I've never heard of.   

ARCGis is more prominent software and uses ftUS. 

http://webhelp.esri.com/arcgisdesktop/9.3/index.cfm?TopicName=About_setting_distance_units

mludlum:
@anjdreas please remove my recent changes.

JKSnd:
@anjdreas thanks! I'll definitely do a more thorough job of checking conversion factors in the future. I foolishly assumed that since the tests were passing that all was ok. Bad, bad me.

mludlum:
Oops!  I meant to write "review" my changes and not "remove".  I commit changes that incorporated most of his points and on others I need responses.  With respect to using fractions vs. decimal, I read in the instructions that "10 significant digits" was plenty.  It might help to add a comment that using fractions for numbers is preferable whenever possible (especially in the case of irrational numbers).

anjdreas:
Oh God, I think my brain farted on the last review. @mludlum you are absolutely correct that it was me who got it backwards. Sorry!

UnitsNet/Scripts/UnitDefinitions/Length.json
mludlum: I committed changes per this request.

mludlum:
Gentlemen, let me know what I need to do to keep this moving forward.  Thanks.

UnitsNet.Tests/CustomCode/LengthTests.cs
```diff
@@ -32,6 +33,8 @@ public class LengthTests : LengthTestsBase
 
         protected override double FeetInOneMeter => 3.28084;
 
+        protected override double USSurveyFeetInOneMeter => 3937.0/1200.0;
```
anjdreas: I'm not sure if this point comes across in the wiki guide, but it is kind of important that the test values are not functions or calculations, but values confirmed by some other means - to double check that the conversion function is in fact correct. My go to solution is to copy and paste the property name into google, and it usually comes up with the answer directly or points me to some online conversion tool. This value should be a number with a sufficient amount of digits.

UnitsNet/GeneratedCode/Enums/LengthUnit.g.cs
```diff
@@ -55,6 +55,7 @@ public enum LengthUnit
         Millimeter,
         Nanometer,
         NauticalMile,
+        USSurveyFoot,
```
anjdreas: Change to `UsSurveyFoot` to be consistent with existing namings.

anjdreas: Make sure to regenerate the code after changing this.

mludlum:
Requested changes finished.

anjdreas:
Looks good to me, @JKSnd ?

JKSnd:
I'm good with it. I'll do a release.

JKSnd:
I'm good with it, I'll handle the release on this one.

PR #219, opened 2017-01-22
Fix arithmetic for Temperature

### LINKED ISSUE #218: Temperature arithmetic is wrong
According to https://en.wikipedia.org/wiki/Celsius#Temperatures_and_intervals 

> What is often confusing about the Celsius measurement is that it follows an interval system but not a ratio system; that it follows a relative scale not an absolute scale. This is put simply by illustrating that while 10 °C and 20 °C have the same interval difference as 20 °C and 30 °C the temperature 20 °C is not twice the air heat energy as 10 °C. As this example shows, degrees Celsius is a useful interval measurement but does not possess the characteristics of ratio measures like weight or distance.

and further more at https://de.wikipedia.org/wiki/Grad_Celsius#Temperaturdifferenz (german)
<pre>
delta t = 2°C - 1°C  is equivalent to delta T = 2K - 1K
delta t = 1K (= 1°C) is equivalent to delta T = 1K (not equal -272.15°C)
</pre>

And that will fail here

    var delta_t = 2.DegreesCelsius( ) - 1.DegreesCelsius( );
    var delta_T = 2.Kelvins( ) - 1.Kelvins( );

    System.Diagnostics.Debug.Assert( delta_t == delta_T );
    System.Diagnostics.Debug.Assert( delta_t.Kelvins == delta_T.Kelvins );
    System.Diagnostics.Debug.Assert( delta_t.DegreesCelsius == delta_T.DegreesCelsius );
    System.Diagnostics.Debug.Assert( delta_t.Kelvins == 1 );
    // next will fail because delta_t.DegreeCelsius is -272.15
    System.Diagnostics.Debug.Assert( delta_t.DegreesCelsius == 1 );

anjdreas:
Fixes #218 

- [x] Default arithmetic with `Temperature` removed, they were plain wrong
- [x] Custom arithmetic for `Temperature` and `TemperatureDelta` is added
- [x] Add `Temperature.Multiply()` and `.Divide()`, easy to get wrong
- [x] `TemperatureDelta` added with default arithmetic, which should cover the examples you gave above


anjdreas:
@SirRufo Can you look over and comment? 

I made a stab at this, but having problem with arithmetic doing things like `Temperature.FromDegreesCelsius(10) * 2 == Temperature.FromDegreesCelsius(20)`.

See `TemperatureTests.cs` for commented tests. I think these should be doable, but my brain gave up and wants to sleep :-)

anjdreas:
On a side note, the whole inherit units stuff was a side track, as I wound up not using it when I realized the conversions were different. I am keeping it, as it's been a topic of discussion before. Might come in handy later.

SirRufo:
We cannot add two temperatures or multiply a temperature by a n (which is just adding for n times), because it does not make sense at all. 

If we measure two cubes with each have a side length of 1 m, putting them together will indeed double the volume (2 * 1 cbm = 2 cbm), but the temperature of both (20°C measured for each) will not double - it will remain the same at 20°C. If the temperature is different (10°C and 20°C) then putting both together we will have a total temperature of 15°C (assuming both are of the same homogene material and ideal conditions). 

The whole temperature arithmetic has to be done with the delta values.

BTW: I would also remove the - operator from temperature type.

UnitsNet/CustomCode/UnitClasses/Temperature.extra.cs
```diff
@@ -0,0 +1,77 @@
...
+
+#if !WINDOWS_UWP
+// Operator overloads not supported in Universal Windows Platform (WinRT Components)
+using System;
+
+namespace UnitsNet
+{
+    public partial struct Temperature
+    {
+        public static Temperature operator -(Temperature right)
```
SirRufo: I would remove this

anjdreas:
Thanks, yes I think you are right. As much as 2 * 20 degrees Celsius should be 40 degrees Celsius, this must be multiplied by a temperature delta of 2 degrees Celsius. After a good night's sleep I finally realized that the different temperature scales have different zero points in addition to different scaling factors, so you can't simply multiply by two, but you can multiply or divide by some amount within each scale - the `TemperatureDelta`.

I'll remove the number arithmetic and only keep the Delta arithmetic with regards to Temperature, including the negate `-` operator. Same thing there.

So, the next thing that comes to mind. Are there other unit classes beyond `Temperature` that also has this problem of arithmetic with numbers not making sense?
Logarithmic units come to mind, but we already have a special case for these.

anjdreas:
@SirRufo Please review again. I found that `*` and `/` operators would not work with `TemperatureDelta`. I realized that you need to perform those operations within the context of a particular unit, due to different zero points of different units. 

```csharp
Temperature.FromDegreesCelsius(20) * TemperatureDelta.FromDegreesCelsiusDelta(3) // NOT 60 degrees C..
Temperature.FromDegreesCelsius(20).Multiply(3, TemperatureUnit.DegreeCelsius) // 60 degrees C
```

Am I mistaken, or is this the path we must take for multiplication and division?

I've added some more tests now and all arithmetic seem to work as I expect now. Fahenheit and Celsius have different scales and different zero points, so they should cover most cases.

By the way, I did not mean to hijack your PR, but I went down a rabbit hole and have yet to climb out.

SirRufo:
I only have a look at the units ... 
> x°C * y°C = x*y°C<sup>2</sup>. 

Sure we need this?

That are the ones I see at first:

```
double operator /( TemperatureDelta a, TemperatureDelta b )
TemperatureDelta operator /( TemperatureDelta a, double b )
TemperatureDelta operator *( TemperatureDelta a, double b )
TemperatureDelta operator *( double a, TemperatureDelta b )
```

Useful units (which are new to your lib) dealing with `TemperatureDelta` would be:
- heat capacity, with unit [J K<sup>−1</sup>]
- specific heat capacity, with unit [J kg<sup>−1</sup> K<sup>−1</sup>]

```
HeatCapacity = Energy / TemperatureDelta
SpecificHeatCapacity = HeatCapacity / Mass
SpecificEnergy = Energy / TemperatureDelta
SpecificHeatCapacity = SpecificEnergy / Mass
```

but that is another story

anjdreas:
> x°C * y°C = x*y°C

Where did you see this? I thought I removed all `Temp * Temp` calculations and tests.
What I did add, was `Temperature Multiply(double factor, TemperatureUnit unit)` and `Temperature Divide(double divisor, TemperatureUnit unit)`, as in my previous example.

The `TemperatureDelta` operator overloads you mention should also be implemented, if you were asking to implement them?

Good point about the new units `HeatCapacity` and `SpecificHeatCapacity`, you are welcome to add these units and operator overloads in a PR.

anjdreas:
So from what I can tell, this PR is now ready to merge.
* Default arithmetic with `Temperature` removed
* Custom arithmetic for `Temperature` and `TemperatureDelta` is added, also a custom `Multiply()` and `Divide()` method since this is easy to get wrong for the consumers of the lib.
* `TemperatureDelta` added with default arithmetic, which should cover the examples you gave above

PR #224, opened 2017-02-01
Replace all occurences of the Greek lower case letter 'mu' with the c…

JenTechSystems:
…orrect unicode Micro symbol.

Add tests to ensure all units that use the Micro scale are using the correct unicode character.

### DIFF
--- a/UnitsNet.Tests/UnitSystemTests.cs
+++ b/UnitsNet.Tests/UnitSystemTests.cs
@@ -257,6 +257,64 @@ namespace UnitsNet.Tests
         }
 
         [Test]
+        public void ShouldUseCorrectMicroSign()
+        {
+            // "\u00b5" = Micro sign
+            Assert.AreEqual(AccelerationUnit.MicrometerPerSecondSquared, Acceleration.ParseUnit("\u00b5m/s²"));
+            Assert.AreEqual(AmplitudeRatioUnit.DecibelMicrovolt, AmplitudeRatio.ParseUnit("dB\u00b5V"));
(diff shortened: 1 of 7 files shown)


JenTechSystems:
What exactly is this build message telling me? 
Is it expecting a new unit test class in the UnitsNet.Tests project?

UnitsNet.Tests/UnitSystemTests.cs
```diff
@@ -257,6 +257,64 @@ public string ScientificNotationUpperIntervalFormatting(double value)
         }
 
         [Test]
+        public void ShouldUseCorrectMicroSign()
+        {
```
anjdreas: Nice set of tests this, even though it will get out of date at some point it is a nice sanity check.

anjdreas:
Just a notice that I have a severe case of the manflu and might need some extra time to respond the next few days. As for the message, my reverse proxy to make the TeamCity build server available on the Internet broke, so you won't be able to access it.

UnitsNet.Tests/UnitSystemTests.cs
JenTechSystems: Yea, it would have been ideal to tap into the code generation and have this test generated for each unit that uses the Micro symbol. The amount I have to contribute to this is dictated by the return on investment for my client though...so that extra level didn't quite make the cut (yet) unfortunately.

anjdreas:
I found the problem, the UnitsNetException type is internal in WINDOWS_UWP profile (Windows Runtime Component target). I added an assembly attribute to make it visible to tests.

anjdreas:
For your reference: d2f0c0825ed2bae4270e6b094f9946d357eda07c

PR #228, opened 2017-02-20
Implemented support for runtime conversions via UnitSystem.Convert and TryConvert

BrandonLWhite:
This is my first stab at implementing #186.  Please review and advise on needed changes.

### DIFF
--- a/UnitsNet.Tests/UnitSystemTests.cs
+++ b/UnitsNet.Tests/UnitSystemTests.cs
@@ -512,5 +512,29 @@ namespace UnitsNet.Tests
             return new CultureInfo(cultureName);
 #endif
         }
+
+        [Test]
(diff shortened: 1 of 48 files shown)


BrandonLWhite:
It looks like there are failing builds due to my use of CreateDelegate.  I didn't realize I needed to test out the build using build.bat or build-all-release.bat.

I think I'll rewrite this to not use reflection and instead a powershell script to generate a switch() or if/else.

You can close this PR and I will submit another in the near future.

anjdreas:
Ok, thanks for investigating the build error on your own.
I agree with the powershell vs reflection approach.

On a side note, the build error was for the UnitsNet.WindowsRuntime project only. You could have worked around that by excluding the delegate with `#if WINDOWS_UWP`, since that target is very picky about the types it exposes in the public API.

PR #229, opened 2017-02-22
Add cycles unit

JenTechSystems:
I need to add a unit to represent 'cycles' in an mains electrical power system.
I have made the assumption that the Frequency in the system is 50Hz i.e. 50 cycles in a second therefore CyclesInOneSecond=1/50=0.02

This will be true for Europe, most of Africa, most of Asia, most of South America and Australia but as I understand it, in most of North America and some parts of Asia, the Frequency of systems is 60Hz.

So we need to accommodate for this frequency as well but I'm not sure how.

Perhaps we actually need to create a new unit, called UsCycles (?), similar to UsTeaspoons, with the definition:

    {
      "SingularName": "UsCycle",
      "PluralName": "UsCycles",
      "FromUnitToBaseFunc": "x*0.0167",
      "FromBaseToUnitFunc": "x/0.0167",
      "Localization": [
        {
          "Culture": "en-US",
          "Abbreviations": [ "cycle", "cycles" ]
        }
      ]
    }

And subsequently in DurationTests:

UsCyclesInOneSecond => 60;


UnitsNet/Scripts/GenerateUnits.ps1
```diff
@@ -173,7 +173,7 @@ function Add-PrefixUnits {
                     "Deci" { "d", "1e-1d"; break; }
                     "Centi" { "c", "1e-2d"; break; }
                     "Milli" { "m", "1e-3d"; break; }
-                    "Micro" { "μ", "1e-6d"; break; }
+                    "Micro" { "µ", "1e-6d"; break; }
```
anjdreas: I think you brought in some old changes to this PR.

UnitsNet/UnitDefinitions/Duration.json
```diff
@@ -162,6 +162,18 @@
           "Abbreviations": [ "нс" ]
         }
       ]
+    },
+    {
+      "SingularName": "Cycle",
```
anjdreas: I think maybe you are mixing domains here. Duration, to me, is in the time domain, while the factor 0.02 from 1/50 Hz, is from the electrical engineering domain. 

So we could add this as `UtilityFrequency` with units `CyclesPerSecond50` and `CyclesPerSecond60` instead, but I don't see it makes sense to convert between the two units.

I think instead a better approach is to use `Frequency`, which is already in the lib.
So in the US, they have a utilty frequency of 60 Hertz, and you could obtain the time period by something like `TimeSpan period = Frequency.FromHertz(60).Period`.

The `.Period` property does not currently exist, but that could easily be added to `Frequency.extra.cs` partial class.
I chose `TimeSpan` because `Duration` kind of just duplicates `TimeSpan` and is a candidate to be removed from the lib.

anjdreas: Another thing that could be convenient, is to have a way to obtain the `Frequency` or the time period of the utility frequency at a particular country, something like `Frequency usUtilityFrequency = Frequency.GetUtilityFrequencyForCountry("US");`

JenTechSystems: OK, so this has all spawned from the requirement to convert between 'seconds' and 'cycles' in our equations
=>where, given a system with a frequency of 50Hz 1 cycle = 1/frequency = 1/50 = 0.02 seconds.

So if we have the variable:

A = 20 cycles

I need to convert this variables units and end up with:

A = 20 * 0.02 = 0.4 seconds

So our equations literally have the unit 'cycle' and 'cycles' in them which I need to fit into the Units.Net framework somehow...

To put it in Frequency implies you can scale between a 'cycle' and 'hertz' but you cant do this without knowing how many seconds one cycle takes

Your suggestion of the Period extension method on the Frequency class makes sense - this will provide us the time taken for one cycle, thus from that we can convert between Time and Cycles by:

`TimeSpan seconds = Frequency.FromHertz(60).Period * NumberOfCycles`

The way I've been running my conversions is like this e.g. between a source frequency to a target frequency:

```
	Frequency sourceFrequency;
            if (Frequency.TryParse(sourceQuantity, out sourceFrequency))
            {
                FrequencyUnit targetFrequencyUnit;
                if (FrequencyWrapper.TryParseUnit(targetUom, out targetFrequencyUnit))
                {
                    switch (targetFrequencyUnit)
                    {
                        case FrequencyUnit.Undefined:
                            break;
                        case FrequencyUnit.Hertz:
                            result = sourceFrequency.Hertz;
                            return true;
			// and so on...
```

So which Unit definition should hold the 'cycle' unit?

And how do we structure the conversion between Cycles and Time?

I like your suggestion of getting the Frequency per country - but where in the Units.Net architecture would this reside? Would you want to see the Frequencies per country setup in a .config file somewhere - or a constants file?

anjdreas:
> And how do we structure the conversion between Cycles and Time?

How would this work?
```
Cycles c = Frequency.FromHertz(50).GetCycles(20);
TimeSpan t = cycles.Time; // 0.4 s
double count = cycles.Count; // 20
```

This could be achieved by:
*  Add a new type `struct Cycles { TimeSpan Time; double Count; }`
* Add `Cycles GetCycles(double)` to `Frequency.extra.cs` or as an extension method

This could then be done just as easily in the lib as in your own code, but it seems to me the concept of cycles is something others may benefit from too when working with frequencies.

JKSnd:
I'd argue that we shouldn't introduce cycles as its own unit. Will we ever convert between cycles50 and cycles60? I don't see a use case for that in power, and I don't see a use case for it in audio either.

I'm wondering why a cycles quantity needs to be strongly typed, especially since introducing this makes UnitsNet a little less general and a little more power-grid specific. Arguably the audio world could take advantage of this too, but do we really want to add overloads for every frequency of importance between 20Hz and 20kHz? That's not very maintainable in my mind.

So the real issue I see here isn't that we're missing a unit for cycles, but rather that you can't overload the division operator for doubles. In a perfect world you would be able to do something like:
`double cycles = 3.0;`
`var duration = cycles/Frequency.FromHertz(50);`

I would suggest adding a method to frequency to accomplish this. something like:
`var duration= Frequency.FromHertz(50).GetDurationGivenCycles(20)`

anjdreas:
@JKSnd I'm with you, but the advantage of having a `Cycles` type is that it can encode information about the duration of each cycle, and you can also obtain `Frequency` from a `Cycles` type.

anjdreas:
By the way the idea of Cycles50 and Cycles60 were already dismissed, just a part of my train of thoughts.

JKSnd:
For the sake of argument, not to be argumentative, frequency itself is already the encoding of the duration of a cycle. I think adding a cycles type is muddying the waters. At the end of the day, its your call though, Andreas.

anjdreas:
Ah, I miswrote. I meant it encodes the count of cycles in addition to duration/frequency information. I'm not favoring either way strongly, just trying to think of options that could benefit the use case in this PR. It would be easy to add this outside the lib, but if Cycles is something many would find useful then maybe that is worth adding. So, not decided on this end, I don't have a personal use for this. I'd listen to what you guys say who work in this domain of units.

UnitsNet/UnitDefinitions/Duration.json
JKSnd: Small suggestion toward the above code snippet. I would use the .As() method on Frequency instead of that switch statement, since its more readable.

```
Frequency sourceFrequency;
if (Frequency.TryParse(sourceQuantity, out sourceFrequency))
{
  FrequencyUnit targetFrequencyUnit;
  if (FrequencyWrapper.TryParseUnit(targetUom, out targetFrequencyUnit))
  {
    result = sourceFrequency.As(targetFrequencyUnit);
    return true;
  }
}
```

Feel free to ignore this suggestion, of course.

anjdreas:
Any follow up thoughts here @JKSnd , @JenTechSystems ?

anjdreas:
Soo.. is there an interest to bring this forward?
@JKSnd did my last correction affect your opinion about a `Cycle` type?

I think it could be handy, but also something that is trivial to implement in the application as well. It just occurs to me that the concept of `Cycles` seems generic enough to have a place in the lib.

JKSnd:
If this gets added to the lib, I think the Feb 23 suggestion by @anjdreas is the best way to go about it. I can get behind that.

JenTechSystems:
Apologies for going dark for a while there...other priorities took over. I will be looking back into this in the next week, so will get back to shortly.

anjdreas:
Great! No worries, it's the life of open source. Just trying to prevent issues from going stale.

JenTechSystems:
> And how do we structure the conversion between Cycles and Time?
> How would this work?

> Cycles c = Frequency.FromHertz(50).GetCycles(20);
> TimeSpan t = cycles.Time; // 0.4 s
> double count = cycles.Count; // 20

The issue I'm faced with here is that the above suggestion implies we're starting with a Frequency.

But I am starting with a Duration unit - And I need to get the Cycles out of that.

Im struggling to see wow would I do that with the way you're thinking of implementing Cycles.

So if we have something like:

```
public double GetDuration(double cyclesValue){

}
```

How would you implement this function?

anjdreas:
I'm not sure I follow when you say you are starting with a Duration, not a Frequency.
I skimmed through the previous discussions above, but I don't think I have a good picture of what exactly you are trying to accomplish. Could you please post a brief example?

As for your method, the only way I see that implemented is if you toss in the frequency as an argument:
```
public TimeSpan GetDuration(double cycles, Frequency f)
{
    return cycles * f.Period;
}

// Alternate version
public TimeSpan GetDuration(double cycles, TimeSpan cyclePeriod)
{
    return TimeSpan.FromTicks(cycles * cyclePeriod.Ticks);
}
```

JKSnd:
I agree with passing in the frequency (unless the get duration method is a part of the frequency class...).
So, hertz would be convenient since it is in units of cycles per second.

```
public TimeSpan GetDuration(double cyclesValue, Frequency f){
    return  TimeSpan.FromSeconds(cyclesValue * f.Hertz);
}
```

If it were rolled into the frequency class...

```
public TimeSpan GetDurationGivenCycles(double cycles){
    return TimeSpan.FromSeconds(cycles*this._hertz);
}
```

JenTechSystems:
Apologies for all the confusion here guys!

Lets start from scratch...

I have one entity, A, it has a variable, X, and it is expressed in 'cycles'.

e.g.

A.X = 20 cycles

I have another entity, B, it has the same variable, X, but it is expressed in 'seconds' and I need to take the value of A.X and put it in B.X, converting the A.X value to seconds along the way.

The way I'm doing conversions between units is like this:

```
    public class DurationUnitConverter : IUnitConverter
    {
        public bool TryConvert(string sourceQuantity, string targetUom, out double? result)
         {
            Duration sourceDuration;
            if (Duration.TryParse(sourceQuantity, out sourceDuration))
            {
                DurationUnit targetDurationUnit;
                if (DurationWrapper.TryParseUnit(targetUom, out targetDurationUnit))
                {
                    switch (targetDurationUnit)
                    {
                        case DurationUnit.Day:
                            result = sourceDuration.Days;
                            return true;
// all other duration units omitted for brevity
                        case DurationUnit.Hour:
                            result = sourceDuration.Hours;
                            return true;
                        case DurationUnit.Cycle:
                            result =  Frequency.FromHertz(50).Period * sourceDuration.Cycles;
                            return true;
                    }
                }
            }
         }
}
```

But I know this isn't what you want because we dont want 'Cycle' as a duration unit.

So I need a way to tell that Im dealing with Cycles...once I know that, I need to get the 'seconds' value for a given variable expressed in cycles.

Does that make sense?

JKSnd:
My sticking point boils down to the Single Responsibility Principle. A Cycle is a count, whereas a Duration is a time span. Coupling them together in this way breaks SRP for the Duration struct.

This converter seems to be duplicating functionality that's already built into the Duration struct. Regardless, I think the calling code is more telling than the converter implementation. Maybe you can send us a little more information on that?

anjdreas:
I kind of lean with @JKSnd on this, converting between seconds and cycles (assuming 50 Hz) may very well make sense in the application you are currently doing, but I don't think it works on a general basis in Units.NET. This would be better glued together in the application, just like you did.

We can still do the `Cycle` type as discussed above and the `TimeSpan Period` property on `Frequency`, and helper methods to calculate cycles or durations between these, but no generic conversion between seconds (time) and cycles or frequency.

Or am I missing something?

anjdreas:
Just checking up on this to get a reply to our comments.

angularsen:
Closing this due to inactivity. Please reopen if you intend to purse this.

PR #241, opened 2017-03-16
Added short and long hundredweight mass unit of measure.

cumminsjp:


UnitsNet/UnitDefinitions/Mass.json
```diff
@@ -115,6 +115,34 @@
...
+      "SingularName": "LongHundredweight",
+      "PluralName": "LongHundredweight",
+      "FromUnitToBaseFunc": "x/0.01968413055222121",
+      "FromBaseToUnitFunc": "x*0.01968413055222121",
+      "XmlDocSummary": "The long or imperial hundredweight (abbreviation cwt) is a unit of mass equal to 112 pounds in US and Canada.",
+      "XmlDocRemarks": "https://en.wikipedia.org/wiki/Hundredweight",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "cwt" ]
```
anjdreas: Hm, are there no unambiguous abbreviations for long and short? This would cause problems on Parse("1 cwt")  string.

cumminsjp: I couldn't find any. From everything I could find, the same abbreviation is used for both.  I'll take a mental note that the abbreviations should be unique per Unit class?  

I don't need the long version, so I could remove it. I added it because it was listed on the same wiki page and thought it might be useful.

anjdreas: http://www.traditionaloven.com/culinary-arts/weight/convert-gram-g-to-hundredweight-cwt-short-us.html

Someone uses `sh cwt` for short, did not find any for long.

anjdreas: No you should keep both, the worst that happens is that you get short instead long when trying to parse, or vice versa.

anjdreas: Or maybe you get an ambiguous exception, I don't recall actually. The latter would probably make the most sense, so I suppose you lose Parse()-ability on these two units. Which makes sense, I guess.

cumminsjp: Confirmed the UnitsNet.AmbiguousUnitParseException:

`UnitsNet.AmbiguousUnitParseException : Cannot parse 'cwt' since it could be either of these: 8, 16
   at UnitsNet.UnitSystem.Parse(String unitAbbreviation, Type unitType) in F:\git\UnitsNet\UnitsNet\CustomCode\UnitSystem.cs:line 227
   at UnitsNet.UnitSystem.Parse[TUnit](String unitAbbreviation) in F:\git\UnitsNet\UnitsNet\CustomCode\UnitSystem.cs:line 200
   at UnitsNet.Mass.ParseUnit(String str, IFormatProvider formatProvider) in F:\git\UnitsNet\UnitsNet\GeneratedCode\UnitClasses\Mass.g.cs:line 1171
   at UnitsNet.Mass.<>c.<Parse>b__117_0(String value, String unit, IFormatProvider formatProvider2) in F:\git\UnitsNet\UnitsNet\GeneratedCode\UnitClasses\Mass.g.cs:line 1086
   at UnitsNet.UnitParser.ParseWithRegex[TUnit](String regexString, String str, ParseUnit`1 parseUnit, IFormatProvider formatProvider) in F:\git\UnitsNet\UnitsNet\CustomCode\UnitParser.cs:line 109
   at UnitsNet.UnitParser.ParseUnit[TUnit](String str, IFormatProvider formatProvider, ParseUnit`1 parseUnit, Func`3 add) in F:\git\UnitsNet\UnitsNet\CustomCode\UnitParser.cs:line 65
   at UnitsNet.Mass.Parse(String str, IFormatProvider culture) in F:\git\UnitsNet\UnitsNet\GeneratedCode\UnitClasses\Mass.g.cs:line 1082
   at UnitsNet.Mass.Parse(String str) in F:\git\UnitsNet\UnitsNet\GeneratedCode\UnitClasses\Mass.g.cs:line 1047
   at UnitsNet.Tests.CustomCode.MassTests.ParseHundredWeightString() in F:\git\UnitsNet\UnitsNet.Tests\CustomCode\MassTests.cs:line 107`

anjdreas: Ok then, this PR should be good to go.

cumminsjp: cwt is always the short hundredweight in US / Canada, so using `sh cwt` would be confusing here.  

Maybe someone from the UK or Australia can chime in on what `cwt` typically means (if anything).

I know the Hundredweight can't be used anymore for trade:

image
http://www.legislation.gov.uk/ukpga/1985/72/pdfs/ukpga_19850072_en.pdf

anjdreas:
Nuget 3.57 out.

anjdreas:
I'm fine with it, as-is. We only lose parsing for thes two units, and if there are no widely established conventions to disambiguate their abbreviations, then we shouldn't try to come up with some. It can always be changed if someone provides the correct ones later.

PR #246, opened 2017-05-03
Added cm/h and mm/h

eriove:
cm/h is used to measure vessel ice aggregation and mm/h is used to measure precipitation (rainfall)

### DIFF
--- a/UnitsNet.Tests/CustomCode/SpeedTests.cs
+++ b/UnitsNet.Tests/CustomCode/SpeedTests.cs
@@ -64,9 +64,9 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double KilometersPerMinutesInOneMeterPerSecond => 0.06;
 
-        protected override double CentimetersPerHourInOneMeterPerSecond => 3600*100;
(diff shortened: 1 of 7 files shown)


UnitsNet.Tests/CustomCode/SpeedTests.cs
```diff
@@ -64,6 +64,10 @@ public class SpeedTests : SpeedTestsBase
 
         protected override double KilometersPerMinutesInOneMeterPerSecond => 0.06;
 
+        protected override double CentimetersPerHourInOneMeterPerSecond => 3600*100;
```
anjdreas: I would prefer if the tests did not perform calculations, but rather write the constant value directly. 

I think it is an important exercise that the values here are not derived or calculated, but rather looked up in a source when possible, at least as a rule of thumb.

eriove: I agree. Updated the code.

anjdreas:
Great, thanks!

anjdreas:
Nuget 3.59 out.

PR #248, opened 2017-05-03
Adding decanewtonspersecond

ngeslin:
Adding decanewtonspersecond


anjdreas:
Hi, it seems you are maybe struggling a bit here. Can I help?

ngeslin:
Thanks but I found why was going wrong with this request. I tried to add a unit but the branch I was using had some issues. I will fix it and make another request later.

PR #254, opened 2017-05-12
Fix UnitParser.ParseUnit for multiple-worded units

gojanpaolo:
https://github.com/anjdreas/UnitsNet/issues/253

1. Changed regex to accept multiple worded non-numeric units
2. Used string.Replace to remove all known separators. This way, we
don't need to include it in our regex. Also the separator "," is already
defined in NumberFormatInfo numFormat
3. Removed code block for <invalid> group since it will not go into this
anymore.
4. Added Unit test for Parsing Mass "Short Tons" and "Long Tons"

### DIFF
--- a/UnitsNet.Tests/CustomCode/ParseTests.cs
+++ b/UnitsNet.Tests/CustomCode/ParseTests.cs
@@ -42,7 +42,7 @@ namespace UnitsNet.Tests.CustomCode
         [TestCase("500,005 m", Result = 500005)]
         [TestCase(null, ExpectedExceptionName = "System.ArgumentNullException")]
         [TestCase("1", ExpectedExceptionName = "System.ArgumentException")]
-        [TestCase("km", ExpectedExceptionName = "UnitsNet.UnitsNetException")]
+        [TestCase("km", ExpectedExceptionName = "System.ArgumentException")]
         [TestCase("1 kg", ExpectedExceptionName = "UnitsNet.UnitsNetException")]
         public double ParseLengthToMetersUsEnglish(string s)
         {
@@ -115,6 +115,17 @@ namespace UnitsNet.Tests.CustomCode
             CultureInfo usEnglish = CultureInfo.GetCultureInfo("en-US");
             Length result;
             return Length.TryParse(s, usEnglish, out result);
+        }
+
+        [TestCase("333 short tn", Result = 333)]
+        public double ParseMassShortTon(string s)
+        {
+            return Mass.Parse(s).ShortTons;
+        }
+        [TestCase("333 long tn", Result = 333)]
+        public double ParseMassLongTon(string s)
+        {
+            return Mass.Parse(s).LongTons;
         }
     }
 }
--- a/UnitsNet.Tests/CustomCode/SpeedTests.cs
+++ b/UnitsNet.Tests/CustomCode/SpeedTests.cs
@@ -68,6 +68,12 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double MillimetersPerHourInOneMeterPerSecond => 3600000;
 
+        protected override double InchesPerSecondInOneMeterPerSecond => 39.3700787;
+
+        protected override double YardsPerMinuteInOneMeterPerSecond => 65.6167979;
+
+        protected override double YardsPerSecondInOneMeterPerSecond => 1.09361330;
+
         [Test]
         public void DurationSpeedTimesEqualsLength()
         {
--- a/UnitsNet/CustomCode/UnitParser.cs
+++ b/UnitsNet/CustomCode/UnitParser.cs
@@ -54,13 +54,14 @@ namespace UnitsNet
 
             const string exponentialRegex = @"(?:[eE][-+]?\d+)?)";
 
-            string regexString = string.Format(@"(?:\s*(?<value>[-+]?{0}{1}{2}{3})?{4}{5}",
+            string regexString = string.Format(@"(?:\s*(?<value>[-+]?{0}{1}{2}{3})?",
                 numRegex, // capture base (integral) Quantity value
                 exponentialRegex, // capture exponential (if any), end of Quantity capturing
                 @"\s?", // ignore whitespace (allows both "1kg", "1 kg")
-                @"(?<unit>[^\s\d,]+)", // capture Unit (non-whitespace) input
-                @"(and)?,?", // allow "and" & "," separators between quantities
-                @"(?<invalid>[a-z]*)?"); // capture invalid input
+                @"(?<unit>[^\d]+)"); // capture Unit (non-numeric)
+
+            //remove separators
+            str = str.Replace("and", ""); 
 
             List<TUnit> quantities = ParseWithRegex(regexString, str, parseUnit, formatProvider);
             if (quantities.Count == 0)
@@ -89,15 +90,7 @@ namespace UnitsNet
 
                 string valueString = groups["value"].Value;
                 string unitString = groups["unit"].Value;
-                if (groups["invalid"].Value != "")
-                {
-                    var newEx = new UnitsNetException("Invalid string detected: " + groups["invalid"].Value);
-                    newEx.Data["input"] = str;
-                    newEx.Data["matched value"] = valueString;
-                    newEx.Data["matched unit"] = unitString;
-                    newEx.Data["formatprovider"] = formatProvider?.ToString();
-                    throw newEx;
-                }
+
                 if ((valueString == "") && (unitString == "")) continue;
 
                 try
--- a/UnitsNet.Tests/GeneratedCode/SpeedTestsBase.g.cs
+++ b/UnitsNet.Tests/GeneratedCode/SpeedTestsBase.g.cs
@@ -59,6 +59,7 @@ namespace UnitsNet.Tests
         protected abstract double DecimetersPerMinutesInOneMeterPerSecond { get; }
         protected abstract double DecimetersPerSecondInOneMeterPerSecond { get; }
         protected abstract double FeetPerSecondInOneMeterPerSecond { get; }
+        protected abstract double InchesPerSecondInOneMeterPerSecond { get; }
         protected abstract double KilometersPerHourInOneMeterPerSecond { get; }
         protected abstract double KilometersPerMinutesInOneMeterPerSecond { get; }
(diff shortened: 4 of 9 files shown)


UnitsNet/CustomCode/UnitParser.cs
```diff
@@ -54,13 +54,14 @@ internal static class UnitParser
...
 
-            string regexString = string.Format(@"(?:\s*(?<value>[-+]?{0}{1}{2}{3})?{4}{5}",
+            string regexString = string.Format(@"(?:\s*(?<value>[-+]?{0}{1}{2}{3})?",
                 numRegex, // capture base (integral) Quantity value
                 exponentialRegex, // capture exponential (if any), end of Quantity capturing
                 @"\s?", // ignore whitespace (allows both "1kg", "1 kg")
-                @"(?<unit>[^\s\d,]+)", // capture Unit (non-whitespace) input
-                @"(and)?,?", // allow "and" & "," separators between quantities
-                @"(?<invalid>[a-z]*)?"); // capture invalid input
+                @"(?<unit>[^\d]+)"); // capture Unit (non-numeric)
```
anjdreas: This works with our current cases, but we might stumble onto problems here later if we get abbreviations with numbers in them. As an example, we currently only list `m²` in `Area.Squaremeter`, but for convenience sake we might want to also add `m^2` so that they can parse both variants. The latter would not work with this implementation.
I'm fine with postponing that until we cross that bridge, but one option is to define the regex as a number immediately followed by any of the known unit abbreviations for the specified culture (`IFormatProvider`).

Something like this should work:
```
string valueRegex = $"{numRegex}{exponentialRegex}";
string unitRegex = "(m²|m\^2|cm²|cm\^2)";
string valueUnitRegex = $"(?<value>{valueRegex})\s?(?<unit>{unitRegex})"`;
```

The `unitRegex` string values can be enumerated by something like this:
```
var unitSystem = UnitSystem.GetCached(formatProvider);
string[] abbrevsForAllUnits = 
    Enum.GetValues(typeof(TUnit))
    .Cast<TUnit>()
    .SelectMany(unitSystem.GetAllAbbreviations)
    .Select(Regex.Escape)
    .ToArray();
```

And you'd need to refactor the `UnitParser.ParseUnit()` method a bit:
* Add `where TUnit : struct, IComparable, IFormattable` (should have been added to begin with)

UnitsNet.Tests/CustomCode/ParseTests.cs
```diff
@@ -116,5 +116,16 @@ public bool TryParseLengthUnitUsEnglish(string s)
...
+
+        [TestCase("333 short tn", Result = 333)]
+        public double ParseMassShortTon(string s)
+        {
+            return Mass.Parse(s).ShortTons;
+        }
+        [TestCase("333 long tn", Result = 333)]
+        public double ParseMassLongTon(string s)
+        {
+            return Mass.Parse(s).LongTons;
```
anjdreas: Nice!

UnitsNet.Tests/CustomCode/SpeedTests.cs
```diff
@@ -68,6 +68,12 @@ public class SpeedTests : SpeedTestsBase
 
         protected override double MillimetersPerHourInOneMeterPerSecond => 3600000;
 
+        protected override double InchesPerSecondInOneMeterPerSecond => 39.3700787;
```
anjdreas: I think you've mixed two pieces of work here. I saw you were working on the master branch, but it would probably be better for you to work on a feature branch to separate the two bulks of changes and create separate pull requests.
We can proceed like this if you are not familiar with how to do it, no biggie, but just a heads up for next time.

UnitsNet.Tests/CustomCode/SpeedTests.cs
```diff
@@ -68,6 +68,12 @@ public class SpeedTests : SpeedTestsBase
...
+
+        protected override double YardsPerMinuteInOneMeterPerSecond => 65.6167979;
+
+        protected override double YardsPerSecondInOneMeterPerSecond => 1.09361330;
```
anjdreas: I double checked these values though, looks good!

UnitsNet.Tests/CustomCode/SpeedTests.cs
gojanpaolo: Yes I messed up with the branches. I'm not familiar using Git yet.

UnitsNet/CustomCode/UnitParser.cs
gojanpaolo: Whenever I add `where TUnit : struct, IComparable, IFormattable`, I get a bunch of errors.
I think you're mistaking TUnit in UnitParser.ParseUnit<TUnit>() as the Unit Enums (i.e. LengthUnit, MassUnit, etc) but TUnit is actually the UnitClasses (Length, Mass).

gojanpaolo: ok.. so I tried hacking into the UnitSystem class to get all the abbreviation given a UnitClasses type.
The problem is, using the format of regex you suggested. It will fail on test "1ft 1invalid" since it recognizes the "INvalid" as inches. It also fails the test "1ft monkey 1in" since it recognizes "FT monkey".
I thought about restricting it to match only exact abbreviation but how will it handle something like "m" and "mm"?

I think having a single regex or parse algorithm for all the Unit types will have its limitations.
If we want the parsing to handle special cases then maybe it will be easier to implement one parse method for each Unit type.

UnitsNet.Tests/CustomCode/SpeedTests.cs
gojanpaolo: Oh yeah, I might as well ask you how should I have done the two pieces of work?
Should I have created two forks? one for the bugfix and one for the addfeature?

UnitsNet/CustomCode/UnitParser.cs
anjdreas: Ah, you raise some good points here.

> I think you're mistaking TUnit in UnitParser.ParseUnit() as the Unit Enums

You are right, I was probably too quick here, but `UnitSystem` should be able to provide this information as you found out.

>  It will fail on test "1ft 1invalid" since it recognizes the "INvalid" as inches. It also fails the test "1ft monkey 1in" since it recognizes "FT monkey". It also fails the test "1ft monkey 1in" since it recognizes "FT monkey".
I thought about restricting it to match only exact abbreviation but how will it handle something like "m" and "mm"?

I think your suggestion should work, we just need ensure we don't match on text that is not part of a valid pair of `{value}{unit}` or the accepted delimiters between pairs or between value and unit.  I tried a few things and wound up with this: `^((?<value>\d+)\s?(?<unit>m²|m\^2|cm²|cm\^2|ft|in|m|mm)\s?)+$` and I added unit tests for the cases you listed above that you can view here: https://regex101.com/r/EMLIwt/1

With the start/end boundaries, a `+` repeater to match on at least one pair and an optional whitespace `\s?` between each pair, the regex now seems to properly match on full abbreviations, and it should not incorrectly match `m` on `1mm`.

> I think having a single regex or parse algorithm for all the Unit types will have its limitations.
If we want the parsing to handle special cases then maybe it will be easier to implement one parse method for each Unit type.

I hear you and I agree that might be something we eventually have to reach for, but I also do think my updated regex might just work. The upside of the complexity is that it allows for parsing any number of valid pairs of value+unit and that the implementation is uniform across units, so it is consistent and easier to reason about. The downside is the complexity of the regex itself, performance and the fact that it might not make sense or at least not be commonly used to parse all kinds of pairs of quantities from a string, like we are used to from `1ft 2in` and `1' 2''`.

What do you think?

UnitsNet.Tests/CustomCode/SpeedTests.cs
anjdreas: I would recommend googling for some guides on git, github, forks and pull requests. I haven't read this one, but at first glance it seems to cover the basics: https://gist.github.com/Chaser324/ce0505fbed06b947d962

The basic gist of it is:
* Add official repo as remote so you can sync in changes, I prefer to call this `origin` and if I have a fork I call that `my`, but that is just my preference: `git remote set-url origin https://github.com/anjdreas/unitsnet` and `git remote add my https://github.com/gojanpaolo/unitsnet`
* Make sure your local `master` branch points to `origin/master` (official one), since I don't think it makes sense to use your fork's master for much: `git branch --set-upstream-to=origin/master`
* Checkout `master` and pull in the latest changes from `origin`: `git checkout master && git pull --rebase`
* Create a local branch (based on `origin/master`) for any new work you want to create a pull request for, `fix-unit-parsing` or `add-pascal-unit`: `git checkout -b fix-unit-parsing origin/master`
* Do your work as one or more commits
* When you are ready for review, `rebase` your branch on top of any new commits on the official repo, then push that branch up to your github fork: `git rebase origin/master && git push -u my`
* Visit github.com to create a pull request from that branch

It's a bit boilerplate the first time, especially setting up the remotes and stuff, but it gets easier with practice. You might also want to use some GUI clients to make things easier, but I personally prefer the command line.

anjdreas: And, for the second pull request, repeat all but the first steps, this way both branches are based on origin/master and don't have any commits from the other branch.

UnitsNet/CustomCode/UnitParser.cs
gojanpaolo: Hi, I have been working on another project last week thus the late response. I'll get back to this once I finished the said project (maybe a couple weeks more). :)

anjdreas: No worries, thanks for posting an update! Did my proposal make sense to you?

gojanpaolo: > it allows for parsing any number of valid pairs of value+unit and that the implementation is uniform across units, so it is consistent and easier to reason about. 
Yes! especially that sentence. I'll try using your regex and add unit tests for it when I get the time. :+1:

anjdreas:
Still waiting for you to address my feedback. Also, it needs to be rebased on top of latest master as there is a merge conflict right now.

gojanpaolo:
I think I will still be busy on projects at work for the next couple of months. :/

anjdreas:
Fixed by #265.

Issue #258, opened 2017-05-29
Add naval length units Fathom and Shackle


OsirisTerje:
These two are pretty well defined.  The last one missing is the cable, but there exist multiple different definitions for it, dependent upon country.

PR #259, opened 2017-05-29
Added fathom and shackle length units  #258

OsirisTerje:
Added fathom and shackle definitions.  See there are some russian abbreviatons in the defintions, but didnt know what those should be for russian.
Also left out "cable", since there seems to be too many different country specific variations.

### DIFF
--- a/UnitsNet.Tests/CustomCode/LengthTests.cs
+++ b/UnitsNet.Tests/CustomCode/LengthTests.cs
@@ -55,11 +55,11 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double YardsInOneMeter => 1.09361;
 
-        private const double fathomsInOneMeter = 1.0 / 1.8288;
-        protected override double FathomsInOneMeter => fathomsInOneMeter;
+        
+        protected override double FathomsInOneMeter => 0.546806649;
 
-        private const double shacklesInOneMeter = 1.0 / 27.432;
-        protected override double ShacklesInOneMeter => shacklesInOneMeter;
+        
+        protected override double ShacklesInOneMeter => 0.0364538;
 
 
         protected override double NauticalMilesInOneMeter => 1.0/1852.0;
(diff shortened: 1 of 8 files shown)


UnitsNet.Tests/CustomCode/LengthTests.cs
```diff
@@ -55,6 +55,12 @@ public class LengthTests : LengthTestsBase
 
         protected override double YardsInOneMeter => 1.09361;
 
+        private const double fathomsInOneMeter = 1.0 / 1.8288;
+        protected override double FathomsInOneMeter => fathomsInOneMeter;
+
+        private const double shacklesInOneMeter = 1.0 / 27.432;
+        protected override double ShacklesInOneMeter => shacklesInOneMeter;
```
anjdreas: Please remove the fields to follow the code style.
Also it is important that these values are not calculations, but constant values looked up in some source of truth (google, wikipedia, online converters, books etc.). This is to serve as a double check of the calculations in the `.json` file.
I usually just google the name of the property and the value often shows up directly in the google calculator:
https://www.google.no/search?q=FathomsInOneMeter&oq=FathomsInOneMeter&aqs=chrome..69i57.808j0j9&sourceid=chrome&ie=UTF-8

UnitsNet/UnitDefinitions/Length.json
```diff
@@ -130,6 +130,38 @@
...
+      "PluralName": "Fathoms",
+      "FromUnitToBaseFunc": "x*1.8288",
+      "FromBaseToUnitFunc": "x/1.8288",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "fathom" ]
+        },
+        {
+          "Culture": "ru-RU",
```
anjdreas: You should omit the entire array element for the Russian abbreviation if you don't have one, which is fine.
It will always fallback to the US English for cultures that don't have an abbreviation defined.

OsirisTerje: Ok, will fix

UnitsNet.Tests/CustomCode/LengthTests.cs
OsirisTerje: Sure

OsirisTerje:
Fixes done

anjdreas:
Nuget 3.64 is out.

PR #257, opened 2017-05-28
Migrate UnitsNet + Serialization projects to "VS2017-style" projects targeting netstandard1.0

supermihi:
## Overview
This PR resolves #255:
1. Adds a .NET core / "VS2017-style" project for `UnitsNet.Serialization.JsonNet`
2. Migrates `UnitsNet.NetStandard10.csproj` to current .NET core / "VS2017-style" toolset format (was necessary to get 1. running)
3. Adds a `netcoreapp1.1` project for the serialization unit tests
4. Adds a solution `UnitsNet.Core.sln` that references only the abovementioned project that are compatible with dotnet core (and hence support cross-platform build).

## Notes
The new `csproj`-Format is much more compact than the other and can be used for various target frameworks at once: I haven't included this yet because I only have a linux machine at hand right now so I cannot test, but using VS 2017 it should be possbile to add [net35 and uap10 target frameworks](https://docs.microsoft.com/en-us/nuget/schema/target-frameworks) to the respective `.NetStandard10.csproj`-files which would allow to remove all the other projects.

Further, since one now can build NuGet packages directly from a new-style `csproj` (using `dotnet pack` on the CLI or the "pack" option in VS 2017) the "boilerplate" of this repository could probably be further simplified.

## Known Limitation
`UnitsNet.Tests` does not yet work with .NET core because it needs to be migraed to NUnit 3 first. I could not figure out (without a windows machine) how the code generation works, so I did not touch that test project yet. I'll be happy to help with the migration but wanted to await the feedback to this PR first.

PR #261, opened 2017-06-03
Migrate to VS2017 /MSBuild15 / .NET Standard / .NET Core

### LINKED ISSUE #255: Serialization with .NET core / .NET standard
While the main nuget package appears to work well with dotnet core, UnitsNet.Serialization.JsonNet refuses to install in a dotnet core app because it targets only the net35 framework. Is it possible to build a nuget package targeting dotnet core or dotnet standard, or are there any dependencies on incompatible packages or system APIs?

supermihi:
This PR resolves #255. See #257 for initial discussion.

TODO:
- [x] Verify Net35 project can be removed
- [x] Verify Portable40 project can be removed
- [ ] Verify signed packages "just work" when upgrading to new package (key file was converted)
- [ ] Set up build and tests for `UnitsNet.WindowsRuntimeComponent` (not ~sure yet if~ dotnet CLI compatible)
- [ ] See if `UnitsNet.sln` can be removed or renamed to `UnitsNet.WindowsRuntimeComponent.sln` if that is the only project that is not compatible with dotnet CLI.
- [ ] Set up .xml output for tests to `\Artifacts\Logs\<TestProjectName>.xml` files, currently [blocked by NUnit support in dotnet CLI](https://github.com/nunit/nunit3-vs-adapter/issues/323). XUnit seems to have found a workaround, so either wait for NUnit to do the same or switch to XUnit.
- [ ] Update README on new build targets (net35, netstandard1.0, WindowsRuntimeComponent)


anjdreas:
Fixes #262 when merged.

anjdreas:
Slowly moving towards completion. I spent some time redoing all the build scripts, but it now properly builds, tests, packs and publishes nugets via Appveyor CI.

https://www.nuget.org/packages/UnitsNet/3.65.0-alpha1
https://www.nuget.org/packages/UnitsNet.Signed/3.65.0-alpha1

What remains that I can think of:
- [ ] Verify Net35 project can be removed
- [ ] Verify Portable40 project can be removed
- [ ] Verify signed packages "just work" when upgrading to new package (key file was converted)
- [ ] Set up build and tests for `UnitsNet.WindowsRuntimeComponent` (not sure yet if dotnet CLI compatible)
- [ ] Set up .xml output for tests to `\Artifacts\Logs\<TestProjectName>.xml` files, currently [blocked by NUnit support in dotnet CLI](https://github.com/nunit/nunit3-vs-adapter/issues/323). XUnit seems to have found a workaround, so either wait for NUnit to do the same or switch to XUnit.

supermihi:
Wow, now you put some effort into this! :) Would you mind pushing the current alpha version of the serialization package to nuget, because there's currently no version on nuget that works with dotnet core?

anjdreas:
I had some issues building, but just pushed some manual nugets out now. They are prerelease versions `-alpha3`.

I must say, this whole vs2017 journey has not been entirely pleasant. It took a lot more time plow through than I'd like to admit, but I do like a lot of the concepts that `dotnet` CLI brings.

anjdreas:
Currently running into this issue: https://github.com/Microsoft/vstest/issues/445#issuecomment-277592096

Tests refuse to run when the output dir is changed using `<BaseIntermediateOutputPath>` in the test project, and it is needed due to having multiple test projects in the same folder to share the source code.
Sigh.

supermihi:
Sorry that I sort of "tricked" you into this effort. ;-) I naively assumed that it would be sufficient to remove a few lines in the `*.csproj` files and be happy with it. But I think it's worth the effort, as you say the direction Microsoft is heading with the `dotnet` CLI and the .NET Standard is a good one, though at the moment tooling and documentation ase pure chaos.

If there's any specific task you need help with please let me know; unless it's fixing the code-gerenating/build/pack/sign scripts, since I don't know Powershell and don't want to learn it ATM. :)

anjdreas:
Hah, no worries :-) It would have to be done at some point anyway, and your effort really helped get over those initial roadblocks, so thanks for that!
Did the nugets work out as expected?

supermihi:
Yes, I'm using the alpha version including JsonNet serialization in a dotnetcore app, works like a charm. Thanks!! :)

anjdreas:
Cool! :-)

anjdreas:
Merged! Phew.. a lot of hurdles on the windows runtime component stuff, as well as nunit and xunit with the new `dotnet test` and `dotnet xunit` commands. Did not like having multiple test projects in the same folder for one thing.

Thanks a lot for putting in the effort here. Hopefully this is now fairly future proof and not too many things change when dotnet CLI 2.0 goes RTM.

supermihi:
Thank you as well for finishing this! I can only hope that Microsoft's approach with .NET standard will help to simplify multi-target libraries in the future ...

Issue #264, opened 2017-07-08
Parsing fails when mapped abbreviations contain numbers


plubbr:
Setting up by mapping an abbreviation to an unit:
`UnitSystem unitSystem = UnitSystem.GetCached("en-US");
            unitSystem.MapUnitToAbbreviation(UnitsNet.Units.AccelerationUnit.MeterPerSecondSquared, "m2");`

The then following works as expected, printing "MeterPerSecondSquared":
`Console.WriteLine(unitSystem.Parse<UnitsNet.Units.AccelerationUnit>("m2").ToString());`

But parsing from Acceleration , I expect "12 m/s²" but it throws exception 

> UnitsNetException: Error parsing string. The unit is not a recognized AccelerationUnit.

`Console.WriteLine(Acceleration.Parse("12 m2").ToString());`

It likely lies with the regular expression pattern in [UnitParser.ParseUnit](https://github.com/anjdreas/UnitsNet/blob/master/UnitsNet/CustomCode/UnitParser.cs#L50).

This is important for me because I am expecting to parse user-typed input. It is not reasonable for the user to enter the pretty input of m/s², but rather m/s^2 which has the same meaning.

I will fork and try to work it out.

plubbr:
So I am a bit confused now... I wrote a [test](https://github.com/plubbr/UnitsNet/commit/2e78d2031a81dc69a4c79b0ff76f277c283e69bc) for the above conditions and it is passing in xunit. However, it fails with symbols (like ^). Will get a sample repo up tomorrow to compare to test results.

anjdreas:
Huh, that is odd. Only thing I noticed was in your example you wrote `Acceleration.Parse("12 m2")` instead of `m/s^2`.

Also, UnitSystem is static and can cause problems if tests are run in parallel. I don't know how xunit is setup as I just recently switched to it. 

Look forward to see what you find.

plubbr:
Yes, I removed the "^" while trying to troubleshoot and find the root cause. In my tests, I removed the `UnitSystem.ClearCache();` and it seems to work as it does in [this sample app](https://github.com/plubbr/UnitsNetAbbreviations).

Also, I found the issue is with this line in the regex parser:
[`@"(?<unit>[^\s\d,]+)", // capture Unit (non-whitespace) input`](https://github.com/anjdreas/UnitsNet/blob/master/UnitsNet/CustomCode/UnitParser.cs#L61).
[...]

Issue #260, opened 2017-05-30
Naming conventions and terminology


anjdreas:
One thing I have repeatedly found tricky to name or describe, is how to refer to the concepts that `Length` and `Force` represent. 

It is natural to refer to centimeters and newtons as units, or units of measure, but I always lacked a good name for their grouping/family. After some googling, it seems *physical quantity* is the correct term.

[wiki/Units_of_measurement](https://en.wikipedia.org/wiki/Units_of_measurement) spells it out pretty well:
> A unit of measurement is a definite magnitude of a quantity, defined and adopted by convention or by law, that is used as a standard for measurement of the same quantity.[1] Any other value of that quantity can be expressed as a simple multiple of the unit of measurement.
> 
> For example, length is a physical quantity. The metre is a unit of length that represents a definite predetermined length. When we say 10 metres (or 10 m), we actually mean 10 times the definite predetermined length called "metre".

Example :
> You can convert between *units* of the same *physical quantity*.

Proposal (final, edited 2017-08-18)
---
- *Quantities* are represented by struct types like `Length` and `Force` and corresponding `enum` types `LengthUnit` and `ForceUnit`
- *Units* are represented by `enum` values like `LengthUnit.Meter` and `ForceUnit.Newton`
- *Unit conversions* are represented by instance properties, such as `Length.Meters` and `Force.Newtons` and custom conversion methods like `Force.FromPressureByArea()`
- Generic type parameters are named
  - `TQuantity` for unit struct types, such as `Length`
  - `TUnitType` for unit enum types, such as `LengthUnit`
  - `TUnit` for unit enum values, such as `LengthUnit.Meter`

Any thoughts on the matter?

I would like to go through the docs and source to ensure the wording is used consistently, as well as on future PR reviews.

Sources:
https://en.wikipedia.org/wiki/Physical_quantity
https://en.wikipedia.org/wiki/Units_of_measurement
https://en.wikipedia.org/wiki/Quantity
https://en.wikipedia.org/wiki/Metrology#Definition_of_units

JKSnd:
Cool! I really like this idea. I took a quick look through the quantities that are currently defined to see if any wouldn't fit the "physical" descriptor. Most are, with the exception of Level, which I suppose is arguable.

I wonder if simply calling these top-level classes "quantities" would suffice?

anjdreas:
That makes sense to me, I can go with `quantity`.

anjdreas:
[...]

Issue #253, opened 2017-05-11
Mass.Parse Error on ToString(MassUnit.ShortTon) and ToString(MassUnit.LongTon)


gojanpaolo:
These lines throws an error:

`Mass.Parse(new Mass().ToString(MassUnit.ShortTon));`
`Mass.Parse(new Mass().ToString(MassUnit.LongTon));`

anjdreas:
Hi, thanks for pointing that out, I was not aware.
The culprit seems the regex that tries to extract value and unit from the string: https://github.com/anjdreas/UnitsNet/blob/master/UnitsNet/CustomCode/UnitParser.cs#L61

It currently reads any non-whitespace after the value as the unit, which means it only reads a single word. The units you posted have two word unit abbrevations; "short tn" and "long tn".

If you want to take a stab at fixing the regex, please do. If not I might find some time this weekend to look into this.

gojanpaolo:
Sure, I'll try doing it tomorrow.

anjdreas:
I didn't see you added the new commit, will review next week as I'm away.

plubbr:
I verified the issue @gojanpaolo was seeing has been resolved from #265, which is in release 3.68.

`Mass.Parse(new Mass().ToString(MassUnit.ShortTon));`
`Mass.Parse(new Mass().ToString(MassUnit.LongTon));`
no longer throw errors, but properly return the parsed Mass.

anjdreas:
Thanks, I also verified this just now.

PR #265, opened 2017-07-09
Bugfix/parse mapped abbrev

plubbr:
So the direction I went was to change the regex to in the UnitParser.ParseUnit to match one of any of the abbreviations for the unit type. This seems more robust as it uses UnitSystem abbreviations and allows for more flexible abbreviations (either mapped at run time or generated).

Related issue: #264

### DIFF
--- a/UnitsNet.Tests/CustomCode/ParseTests.cs
+++ b/UnitsNet.Tests/CustomCode/ParseTests.cs
@@ -177,6 +177,7 @@ namespace UnitsNet.Tests.CustomCode
         [InlineData("_")]
         [InlineData("?")]
         [InlineData("123")]
+        [InlineData("od to")]
         public void TryParseLengthUnitAbbreviation(string s)
         {
             UnitSystem unitSystem = UnitSystem.GetCached("en-US");
@@ -201,6 +202,7 @@ namespace UnitsNet.Tests.CustomCode
         [InlineData("_")]
         [InlineData("?")]
         [InlineData("123")]
+        [InlineData("od to")]
         public void TryParseLengthUnitAbbreviationWithNumbers(string s)
         {
             UnitSystem unitSystem = UnitSystem.GetCached("en-US");
--- a/UnitsNet.Tests/CustomCode/ParseTests.cs
+++ b/UnitsNet.Tests/CustomCode/ParseTests.cs
@@ -165,26 +165,50 @@ namespace UnitsNet.Tests.CustomCode
         }
 
         [Theory]
-        [InlineData("10 m^2", true)]
-        public void TryParseLengthUnitAbbreviationWithNumbers(string s, bool expected)
+        [InlineData("!")]
+        [InlineData("@")]
+        [InlineData("#")]
+        [InlineData("$")]
+        [InlineData("%")]
+        [InlineData("^")]
+        [InlineData("&")]
+        [InlineData("*")]
+        [InlineData("-")]
+        [InlineData("_")]
+        [InlineData("?")]
+        [InlineData("123")]
+        public void TryParseLengthUnitAbbreviation(string s)
         {
-            UnitSystem.ClearCache();
             UnitSystem unitSystem = UnitSystem.GetCached("en-US");
-            unitSystem.MapUnitToAbbreviation(UnitsNet.Units.LengthUnit.Meter, "m^2");
-            Length result;
-            bool actual = Length.TryParse(s, out result);
-            Assert.Equal(expected, actual);
+            string abbrev = $"m{s}s";
+            unitSystem.MapUnitToAbbreviation(UnitsNet.Units.LengthUnit.Meter, abbrev);
+            UnitsNet.Units.LengthUnit result;
+            bool actual = unitSystem.TryParse<UnitsNet.Units.LengthUnit>(abbrev, out result);
+            Assert.Equal(true, actual);
+            Assert.Equal(UnitsNet.Units.LengthUnit.Meter, result);
         }
 
         [Theory]
-        [InlineData("10 m2")]
-        public void ParseLengthUnitAbbreviationWithNumbers(string s)
+        [InlineData("!")]
+        [InlineData("@")]
+        [InlineData("#")]
+        [InlineData("$")]
+        [InlineData("%")]
+        [InlineData("^")]
+        [InlineData("&")]
+        [InlineData("*")]
+        [InlineData("-")]
+        [InlineData("_")]
+        [InlineData("?")]
+        [InlineData("123")]
+        public void TryParseLengthUnitAbbreviationWithNumbers(string s)
         {
-            UnitSystem.ClearCache();
             UnitSystem unitSystem = UnitSystem.GetCached("en-US");
-            unitSystem.MapUnitToAbbreviation(UnitsNet.Units.LengthUnit.Meter, "m2");
-            Length actual = Length.Parse(s);
-            Assert.Equal(new Length(10), actual);
+            string abbrev = $"m{s}s";
+            unitSystem.MapUnitToAbbreviation(UnitsNet.Units.LengthUnit.Meter, abbrev);
+            bool actual = Length.TryParse($"10 {abbrev}", out Length result);
+            Assert.Equal(true, actual);
+            Assert.Equal(Length.FromMeters(10d), result);
         }
 
         private static string AssertExceptionAndGetFullTypeName(Action code)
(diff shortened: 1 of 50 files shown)


plubbr:
I want to add a few more tests, so this should not be merged quite yet.
I did not realize that the solution to this was basically the same as the solution to #254 and #253 -- I am going to add tests to verify those issues are resolved as well.

anjdreas:
Been a bit busy, will get to this soon.

UnitsNet/CustomCode/UnitParser.cs
```diff
@@ -54,11 +55,27 @@ internal static class UnitParser
...
             const string exponentialRegex = @"(?:[eE][-+]?\d+)?)";
 
+            // Special regex characters that need escaping
+            string[] regexSpecialCharacters = {"$","^","*","?"};
+            string[] unitAbbreviations = UnitSystem.GetCached(formatProvider)
+                .GetAllAbbreviations(unitType)
+                .OrderByDescending(s => s.Length) // Important to order by length -- if "m" is before "mm" and the input is "mm", it will match just "m" and throw invalid string error
+                .ToArray();
+            // Escape special regex characters
+            for (int i = 0; i < unitAbbreviations.Length; i++)
```
anjdreas: How about using `Regex.Escape()` instead:
https://msdn.microsoft.com/en-us/library/system.text.regularexpressions.regex.escape.aspx

UnitsNet/CustomCode/UnitParser.cs
```diff
@@ -54,11 +55,27 @@ internal static class UnitParser
...
+                }
+            }
+            string unitsRegex = $"({String.Join("|", unitAbbreviations)})";
+
             string regexString = string.Format(@"(?:\s*(?<value>[-+]?{0}{1}{2}{3})?{4}{5}",
                 numRegex, // capture base (integral) Quantity value
                 exponentialRegex, // capture exponential (if any), end of Quantity capturing
                 @"\s?", // ignore whitespace (allows both "1kg", "1 kg")
-                @"(?<unit>[^\s\d,]+)", // capture Unit (non-whitespace) input
+                $@"(?<unit>{unitsRegex})", // capture Unit (non-whitespace) input
```
anjdreas: The `@` is no longer needed. Update comment to something like `// capture Unit by a list of abbreviations`

UnitsNet/CustomCode/UnitSystem.cs
```diff
@@ -411,6 +411,33 @@ public string[] GetAllAbbreviations(Type unitType, int unitValue)
...
+        /// <returns>Unit abbreviations associated with unit.</returns>
+        [PublicAPI]
+        public string[] GetAllAbbreviations(Type unitType)
+        {
+            Dictionary<int, List<string>> unitValueToAbbrevs;
+            List<string> abbrevs = new List<string>();
+
+            if (_unitTypeToUnitValueToAbbrevs.TryGetValue(unitType, out unitValueToAbbrevs))
+            {
+                foreach (var item in unitValueToAbbrevs)
```
anjdreas: Can simplify this to `return unitValueToAbbrevs.Values.ToArray();`

UnitsNet/CustomCode/UnitSystem.cs
```diff
@@ -411,6 +411,33 @@ public string[] GetAllAbbreviations(Type unitType, int unitValue)
...
         }
 
+        /// <summary>
+        ///     Get all abbreviations for unit.
+        /// </summary>
+        /// <param name="unitType">Enum type for unit.</param>
+        /// <param name="unitValue">Enum value for unit.</param>
+        /// <returns>Unit abbreviations associated with unit.</returns>
+        [PublicAPI]
+        public string[] GetAllAbbreviations(Type unitType)
```
anjdreas: Neat, this method has been requested before.

UnitsNet/CustomCode/UnitParser.cs
```diff
@@ -37,7 +37,8 @@ internal static class UnitParser
         internal static TUnit ParseUnit<TUnit>([NotNull] string str,
             [CanBeNull] IFormatProvider formatProvider,
             [NotNull] ParseUnit<TUnit> parseUnit,
-            [NotNull] Func<TUnit, TUnit, TUnit> add)
+            [NotNull] Func<TUnit, TUnit, TUnit> add,
+            Type unitType)
```
anjdreas: How about adding another generic parameter instead? And renaming `TUnit` to `TQuantity`, as per recent discussion on naming conventions in #260.
Suggestion: `ParseUnit<TQuantity, TUnit>()`

UnitsNet.Tests/CustomCode/ParseTests.cs
```diff
@@ -164,6 +164,55 @@ public void TryParseLengthUnitUsEnglish(string s, bool expected)
...
+        [InlineData("$")]
+        [InlineData("%")]
+        [InlineData("^")]
+        [InlineData("&")]
+        [InlineData("*")]
+        [InlineData("-")]
+        [InlineData("_")]
+        [InlineData("?")]
+        [InlineData("123")]
+        [InlineData("od to")]
```
plubbr: Ah, indeed you are right. `od to` was simply to test a space and the name was confusing. I will change to "SpecialCharacters".

UnitsNet/CustomCode/UnitParser.cs
plubbr: That looks like a much better approach. By the looks of it, I will still need a for loop though? If the project is ever updated to .NET Standard 2.0, the for loop can be replaced by:
`string unitsRegex = $"({String.Join("|", unitAbbreviations.ForEach(s => Regex.Replace(s)))})";`

UnitsNet/CustomCode/UnitParser.cs
plubbr: This foreach loop goes away with Regex.Escape. Will also update comment.

UnitsNet/CustomCode/UnitSystem.cs
plubbr: It is giving me a compile time error trying to do this. I think it is because `unitValueToAbbrevs.Values` is a collection of lists. Is there another way with LINQ that I have not found?

UnitsNet/CustomCode/UnitParser.cs
plubbr: I think I follow... see if the next commits are what you are expecting. [Another idea](https://github.com/plubbr/UnitsNet/commit/d20713528ef0d514db28cb61b0eab1c326a91ced) is to use an interface for quantity, then the `TUnit` argument can be used to determine the type with no need for a redundant parameter.

UnitsNet/CustomCode/UnitSystem.cs
anjdreas: Oh, right. Try `unitValueToAbbrevs.Values.SelectMany(x => x).ToArray()`

plubbr: Nice, I like it. A good snippet to remember.

anjdreas:
Let me know when it's ready for another review

plubbr:
It should be ready to review again.

plubbr:
@anjdreas This is ready for review.

anjdreas:
Hi, I'm on vacation trip until July 30. Will review when I get back.

anjdreas:
I meant July 25

plubbr:
No worries; thanks for the update. Have a great trip!

UnitsNet/CustomCode/UnitParser.cs
```diff
@@ -29,15 +29,15 @@
...
 namespace UnitsNet
 {
-    internal delegate TUnit ParseUnit<out TUnit>(string value, string unit, IFormatProvider formatProvider = null);
+    internal delegate TQuantity ParseUnit<out TQuantity>(string value, string unit, IFormatProvider formatProvider = null);
 
     internal static class UnitParser
     {
         [SuppressMessage("ReSharper", "UseStringInterpolation")]
-        internal static TUnit ParseUnit<TUnit>([NotNull] string str,
+        internal static TQuantity ParseUnit<TQuantityEnum, TQuantity>([NotNull] string str,
```
anjdreas: I renamed the type parameters to match a [naming convention](https://github.com/anjdreas/UnitsNet/issues/260).
Instead of `TQuantity, TUnit` it is now `TQuantityEnum, TQuantity`.

PR #275, opened 2017-08-24
[WIP] Automatic overload generator

eriove:
As discussed in #260 here is a proof of concept where I generate operator overloads from a vector of SI quantity exponents.

Right now I'm adding the exponents in the tests but they would be in the JSON files once everything is complete.

Also have some issues with paths, new project formats and nCrunch to sort out before this builds out of the box (no GUI support for symbolic links)

Have not even started to think about the PowerShell scripts but I don't think that will be a problem.


anjdreas:
Nice! This looks exciting, I think I see where you are going with this.
You got the unit conversion going, but I don't yet see how the multiplier/value will be included.

Some initial thoughts:
* How do we represent 5 meters?
* How do we represent 5 kilometers?
* Will this replace our current approach of `FromBaseUnitFunc` and `ToBaseUnitFunc` in JSON? If so, will we now define SI array values per unit + a multiplier (5 centi/ 10 kilo/1 mega etc)? Will we still define a base unit to store internally or can we now generate conversions between all units based on their SI arrays? That would be awesome!
* Can we bundle this information in `Length`, so that `Length.Centimeters` has an SI array that denotes centimeters, not its base unit meters? On that note, the unit it was created with could also be stored so that say ToString() favors the value and unit it was created with rather than the base unit.

eriove:
This is only for adding operator overloads. The existing JSON-files would need the new array and also a property that points out the base SI-unit when it isn't the base unit (most units that has the automatic milli, centi, deci etc.)

Once I can generate the operator overloads from with the DLL my idea is to create a new Powershell script that calls the DLL and generates C# code with the operator overloads. The end result would be exactly what me and Ulf did manually more than a year ago for some of the units. The main advantage is that it would extend to all units and is less error prone. 

It also opens up the possibility to add more intermediate units in an easy way (such as m^-1 or s/m). 

But that is for the future. My main reason for doing this is that I really wanted to do it this way back when we did it manually, but I couldn't figure out how before I saw @phatcher's post (in #260) and understood how it could be done.

With some clever work it _might_ be possible to replace the `FromUnitToBaseFunc` and `FromBaseToUnitFunc` but I hadn't thought of that before I read your reply

anjdreas:
Alright, I'll be watching this eagerly :-)

eriove:
@anjdreas and @aignas this doesn't build but the code generation works and it should show approximately where I'm going

The PowerShell code will require heavy refactoring once the generated code looks good, learning PowerShell by doing this...

UnitsNet.OperatorOverloads/ArrayExtensions.cs
```diff
@@ -22,6 +22,21 @@ public static int[] ElementwiseSubtract(this int[] left, int[] right)
             return result;
         }
 
+        public static int[] ElementwiseAdd(this int[] left, int[] right)
```
anjdreas: You can use LINQ's `.Zip()`, like so

```
	public static int[] ElementwiseAdd(this int[] left, int[] right)
	{
		return left.Zip(right, (l,r) => l + r).ToArray();
	}
```

However, it just picks the shortest length of the two, so you may want to keep the length check.

UnitsNet.OperatorOverloads/OverloadGenerator.cs
```diff
@@ -26,10 +26,42 @@ public IEnumerable<Quantity> GetDivisionOverloads(Quantity quantity)
...
+            }
+        }
+
+        public IEnumerable<Overload> GetOverloads(Quantity quantity)
+        {
+            var multiplications = GetMultiplicationOverloads(quantity);
+            foreach (var multiplication in multiplications)
+            {
+                yield return multiplication;
+                yield return new Overload(multiplication.Result, multiplication.Right, multiplication.Left);
```
anjdreas: Is this an intended duplicate?

UnitsNet/UnitDefinitions/Speed.json
```diff
@@ -78,5 +78,6 @@
         }
       ]
     }
-  ]
+  ],
+  "SiArray": [ 1, 0, -1, 0, 0, 0, 0 ]
```
anjdreas: I'm just trying to understand still, so you're adding the SI array to each quantity, which represents the base unit?
In the case of speed, it's `m/s`. And you use this to automatically generate all possible `*` and `/` operators across quantities, such as `Length / Time = Speed`. 

Neat!

Thinking further, can we use this to 
1. Convert to other speed units? Use `FromBaseToUnitFunc` or do we add an `SiArray` to each unit?

UnitsNet.OperatorOverloads/OverloadGenerator.cs
eriove: It was intended, but is wrong. My thought was that multiplication is communicative, but the other multiplication will end up in the other type.

UnitsNet/UnitDefinitions/Speed.json
eriove: I'm considering creating the SI-arrays by parsing the unit names (or abbreviations), but I'm not sure if that is actually faster than just creating the arrays by hand.

It would be possible to create a similar double array with conversion factors for other unit systems (like Imperial). That could be used to automatically generate the code for conversion to that unit system. Since we already have Imperial for most quantities I'm not sure if it is really necessary. But it would be nice when adding new quantities.

angularsen:
I don't mean to nag, but curious if you plan to make progress on this? :-)

eriove:
I plan to continue but it has been too much at work lately so I haven't had time

angularsen:
@eriove I'll close this PR to keep the issue list tidy, but please just reopen it whenever you pick it up again.

PR #281, opened 2017-09-05
Added 'lt' to liter abbreviations

gmantaos:


anjdreas:
Are you sure this is a widely accepted abbreviation? Can you provide some sources, because I'm not finding much to back this abbreviation.

> The litre (SI spelling) or liter (American spelling) (SI symbols L or l,[1] commonly, but incorrectly, abbreviated as ltr)

https://en.wikipedia.org/wiki/Litre

Google's converter does accept it though.
The reason I am hesitant to add non-official abbreviations is to avoid conflicts where it becomes ambiguous what "1 lt" means when parsing strings into quantities.

gmantaos:
Ye as I see it there may be a conflict with this one: https://en.wikipedia.org/wiki/Long_ton

Why I encountered the issue with liter is because I'm currently working with product lists and almost all of them use "lt" for liter. Might also be a localization issue since the only possible conflicts are outside the metric system.

PR #284, opened 2017-09-08
Volume from Mass and Density

ThomasN-Saab:


ThomasN-Saab:
There's no description of the Pull Request because the comment field was greyed out so I couldn't write anything.

Anyway, this is the missing operator which I brought up in this topic: https://github.com/anjdreas/UnitsNet/issues/283

anjdreas:
Nuget 3.73 is out with your changes.
Thanks for putting in the effort!

ThomasN-Saab:
Thanks! That's just what I needed.

PR #286, opened 2017-09-11
Add US and UK Beer barrel volumes

sheam:
Barrels are extremely confusing.

US Oil Barel was alright in Units.Net. It is 42 US Gal, with abbreviation "bbl".

I added UK Beer Barrel which is 36 UK Gal, with abbreviation "bl (imp.)".

I added US Beer Barrel which is 31 US Gal, with abbreviation "bl (US)".

All three use the "bbl" as their abbreviation, so I wasn't sure what to do there... It seems BBL is more associated with oil, and I have seen 'bl' for beer.  My pref would be to just use "bbl" for all 3, as they are unlikely be used in the same application. But I don't think you can have them the same.

### DIFF
--- a/UnitsNet.Tests/CustomCode/VolumeTests.cs
+++ b/UnitsNet.Tests/CustomCode/VolumeTests.cs
@@ -75,7 +75,11 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double TeaspoonsInOneCubicMeter => 202884.13621105801;
 
+        protected override double UkBeerBarrelsInOneCubicMeter => 6.1102568972;
+
         protected override double UkTablespoonsInOneCubicMeter => 66666.6666667;
+
+        protected override double UsBeerBarrelsInOneCubicMeter => 8.5216790723083;
 
         protected override double TeaspoonsTolerance => 1E-3;
 
--- a/UnitsNet.Tests/CustomCode/VolumeTests.cs
+++ b/UnitsNet.Tests/CustomCode/VolumeTests.cs
@@ -75,7 +75,7 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double TeaspoonsInOneCubicMeter => 202884.13621105801;
 
-        protected override double UkBeerBarrelsInOneCubicMeter => 6.1102568972;
+        protected override double ImperialBeerBarrelsInOneCubicMeter => 6.1102568972;
 
         protected override double UkTablespoonsInOneCubicMeter => 66666.6666667;
 
--- a/UnitsNet/UnitDefinitions/Volume.json
+++ b/UnitsNet/UnitDefinitions/Volume.json
@@ -295,15 +295,15 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ ]
+          "Abbreviations": []
         },
         {
           "Culture": "ru-RU",
-          "Abbreviations": [ ]
+          "Abbreviations": []
         },
         {
           "Culture": "nb-NO",
-          "Abbreviations": [ ]
+          "Abbreviations": []
         }
       ]
     },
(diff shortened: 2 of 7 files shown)


sheam:
Maybe "bbl (oil)", "bbl (us)", and "bbl (imp.)" ?

anjdreas:
Hi, thanks for adding this.
We are already familiar with some ambiguous and confusing units in the business of volume :-) I don't have any prior experience with these particular units and even just skimming through some details in the [Barrel unit wiki](https://en.wikipedia.org/wiki/Barrel_(unit)#Definitions_and_units) gave me a headache on the ambiguity involved.

I guess it comes down to trusting your judgement as the domain expert here, and your proposal from a layman's perspective seems fine to me.

One comment on consistency, we do use these already (`oz (imp.)`, `gal (U.S.)`) and I would like you to consider whether those imperial and US abbreviation formats would work for you. We can always add multiple abbreviations per unit, but the first one is the main one used in `.ToString()`. The others are considered when using `.Parse()`:

sheam:
Alright, I think I have the inconsistencies cleaned up.

sheam:
What is the next step here. Is there something you need me to do?

anjdreas:
Next step is on me :-) I'll get to this today.

UnitsNet.Tests/CustomCode/VolumeTests.cs
```diff
@@ -75,8 +75,12 @@ public class VolumeTests : VolumeTestsBase
 
         protected override double TeaspoonsInOneCubicMeter => 202884.13621105801;
 
+        protected override double ImperialBeerBarrelsInOneCubicMeter => 6.1102568972;
+
         protected override double UkTablespoonsInOneCubicMeter => 66666.6666667;
 
+        protected override double UsBeerBarrelsInOneCubicMeter => 8.5216790723083;
```
anjdreas: What is the source of this constant?

From [wiki](https://en.wikipedia.org/wiki/Barrel_(unit)#Fluid_barrel_in_the_US_and_UK) I found:
> In the US most fluid barrels (apart from oil) are 31.5 US gallons (26 imp gal; 119 L) (half a hogshead), but a beer barrel is 31 US gallons (26 imp gal; 117 L).

I read that as `1 US beer barrel = 117 L = 0.117 m^3 => 1 m^3 = 8.54700854700.. US beer barrels`.

UnitsNet/UnitDefinitions/Volume.json
```diff
@@ -434,6 +434,30 @@
...
+    },
+    {
+      "SingularName": "UsBeerBarrel",
+      "PluralName": "UsBeerBarrels",
+      "FromUnitToBaseFunc": "x*0.1173477658",
+      "FromBaseToUnitFunc": "x/0.1173477658",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "bl (U.S.)" ]
```
anjdreas: Won't this abbreviation conflict with fluid and dry barrels? If we have ambiguous units, then Volume.Parse("1 bl (U.S)") would fail with an exception. It is very possible that some units actually do have ambiguous abbreviations and in that case it is correct to not allow to parse it. Just checking in as I can't easily find the abbreviations for fluid and dry barrels. I just see there are a lot of different kinds of barrels mentioned everywhere.

UnitsNet.Tests/CustomCode/VolumeTests.cs
sheam: Hmmm, google and https://www.convertunits.com/from/US+gallons/to/barrel+[US,+beer] say 31 US Gal. Convert-to.com also says 31: http://convert-to.com/conversion/volume/convert-us-barrel-beers-to-gal-us.html.

So I then converted 31 US Gal to m^3 (8.5 216 791)

UnitsNet/UnitDefinitions/Volume.json
sheam: http://convert-to.com/conversion/volume/volume_multimenu.html

I have usually seem fluid barrels as fl bl

UnitsNet.Tests/CustomCode/VolumeTests.cs
anjdreas: Aha, 31 US gallons = 117.348 L.
It seems 31 US gallons is the truth then and 117 L was just a rounded value. Should be good as-is then.

anjdreas:
Ok, I think all points are addressed then. Looks good to me, thanks!

anjdreas:
Nuget 3.74 is on its way out with these changes.

PR #287, opened 2017-09-14
Add US Therm, EU Therm and UK Therm

DInozemtsev:


DInozemtsev:
Hello there. 
1. I have used this Wikipedia article for reference https://en.wikipedia.org/wiki/Therm.
2. Google Unit converter provides very close value on 1 Therm to 1 Joule conversion. Check this out https://www.google.com/search?q=1+therm+in+joule

DInozemtsev:
Sirs, is there something else that I could do to help you with this Pull Request? If so just let me know. :)

UnitsNet.Tests/CustomCode/EnergyTests.cs
```diff
@@ -34,10 +36,18 @@ public class EnergyTests : EnergyTestsBase
...
+
+        protected override double DecaimperialThermsInOneJoule => 9.478171203551087813109937767482e-15;
+
+        protected override double DecausThermsInOneJoule => 9.4804342797334860315281322406817e-10;
+
         protected override double ElectronVoltsInOneJoule => 6.241509343260179e18;
 
         protected override double ErgsInOneJoule => 10000000;
 
+        protected override double EuThermsInOneJoule => 9.4781339449889105832843629746176e-9;
```
eriove: I get this to 9.47816988e-9 when doing 1/105505600 in [Google](https://www.google.se/search?dcr=0&source=hp&q=1%2F105505600&oq=1%2F105505600)

UnitsNet.Tests/GeneratedCode/EnergyTestsBase.g.cs
```diff
@@ -54,31 +54,43 @@ public abstract partial class EnergyTestsBase
     {
         protected abstract double BritishThermalUnitsInOneJoule { get; }
         protected abstract double CaloriesInOneJoule { get; }
+        protected abstract double DecaeuThermsInOneJoule { get; }
```
eriove: @anjdreas I would say that this should have a capital E since it is an abbreviation. What do you think? (Same with the others)

UnitsNet/UnitDefinitions/Energy.json
```diff
@@ -89,6 +89,57 @@
           "Abbreviations": [ "Wh" ]
         }
       ]
+    },
+    {
+      "SingularName": "EuTherm",
```
eriove: This is called EC Therms in the Wikipedia article, probably because the unit is from the time when EU was called EC. Is EU Therm the most common name right now?

UnitsNet.Tests/GeneratedCode/EnergyTestsBase.g.cs
anjdreas: I agree, this doesn't look quite right. These are probably generated by powershell and the `deca` prefix in JSON. `Deca` + `camelcase(EuTherms)`. This works for things like `Decigram` and `Millinewtonmeters`, but not for `DecieuTherm`. Not sure how we would fix this other than not using prefix and instead adding `DecaEuTherm` manually in JSON. Probably the easiest fix.

anjdreas: On second thought, `DecaEuTherm` is probably not consistent with our existing naming either. Maybe `Decaeutherm`, or `Decaeurotherm` as `eurotherm` seems to be a wording I find when googling this.
I might prefer `Eurotherm` and `Decaeurotherm` after chewing on it a bit.

DInozemtsev: @anjdreas Please also note that we also have `DecaimperialThermsInOneJoule`, `DecausThermsInOneJoule`. So to keep things simple at the moment, I would use ReSharper to rename this properties to `DecaEuThermsInOneJoule `, `DecaImperialThermsInOneJoule` and `DecaUsThermsInOneJoule`. What would say about it?

UnitsNet.Tests/CustomCode/EnergyTests.cs
DInozemtsev: Yeah, make sense, I'll change this value.

UnitsNet/UnitDefinitions/Energy.json
DInozemtsev: Well, I wish I could tell. From my experience people just use **therm** implicitly meaning a US or EU unit, that they solely use everywhere in documentation across a project. So looks like it is up to us how to call it.

UnitsNet.Tests/GeneratedCode/EnergyTestsBase.g.cs
eriove: The C# files get replaced when building so that would unfortunately not work. The "simplest" solution would be to add the deca units manually to the json file.

UnitsNet/UnitDefinitions/Energy.json
eriove: I would go with calling it EC therm then to match Wikipedia. Then we are consistent with at least one other source. I'll leave the last word to @anjdreas

anjdreas:
A couple of things:
* `DecaEcTherm` would break capitalization consistency with the rest of our units (`Kilogram`, `Millimeter`), unless we define that abbreviations like `Ec` or `Us` should keep their first character uppercase for prefixed units (deca, milli etc)
* A possibly better naming scheme, is to put such notations as a suffix `DecathermEc` and `ThermEc`. This has the advantage of circumventing the whole special case for abbreviations since they won't be next to a unit prefix. It also better matches the textual representation in the wiki*
A final, minor point, is that you group together similar units, like `BarrelImperial` next to `BarrelUs`. If so, this should be made a code convention for future units - and add to our [wishlist](https://github.com/anjdreas/UnitsNet/issues/180) to change older units as part of a bigger breaking change.
* EC vs EU: If I read it right, Wikipedia seems to prefer EC and I would need to hear a counter-argument from a domain expert to not go with the wikipedia definition.

https://en.wikipedia.org/wiki/Therm#Definitions   *

> Ten therms are known as a decatherm. Common abbreviations are Dth for decatherm (sometimes dekatherm[4]), MDth for a thousand decatherms and MMDth for a million decatherms.

DInozemtsev:
As for me `DecathermEc` looks nice and neat. However it is not completely clear how would we name abbreviations. Should it be like: 
`th (EC)`, `th (US)`, `th (UK)`,
`Dth (EC)`, `Dth (US)`, `Dth (UK)`,
`MDth (EC)`, `MDth (US)`, `MDth (UK)`, 
`MMDth (EC)`, `MMDth (US)`, `MMDth (UK)`,

anjdreas:
We have some sort of convention going with previous units, but it is not set in stone and probably not entirely consistent either. I'm losing a bit track of all the various units and their abbreviations.

Example: Volume unit abbreviations `gal (U.S.)`, `bl (U.S.)`, `bl (imp.)` etc.

I don't know what is the better choice here. I personally think `UK` and `US` reads better than `U.S.` and `imp.`, but maybe it simply depends on the context and the conventions for each unit.
I guess I would prefer to use the same abbreviations that are used in the domain, and if there are none maybe fall back to our U.S. and imp. convention for consistency?

cc @eriove @JKSnd

DInozemtsev:
I would stay with already existing format, i mean like `gal (U.S.)` and sorry for delay, sirs, I was traveling at the end of last week. Hope will have enough time to implement this changes in a next few days.

DInozemtsev:
Ok, that seems to be it. Sirs, could you plz review? I was in a bit rush, so maybe had missed something.

eriove:
Looks good to me, @angularsen could you check to so that we didn't misunderstand each other.

UnitsNet.Tests/CustomCode/EnergyTests.cs
```diff
@@ -36,17 +36,17 @@ public class EnergyTests : EnergyTestsBase
...
 
-        protected override double DecausThermsInOneJoule => 9.4804342797334860315281322406817e-10;
+        protected override double DecathermsUsInOneJoule => 9.4804342797334860315281322406817e-10;
 
         protected override double ElectronVoltsInOneJoule => 6.241509343260179e18;
 
         protected override double ErgsInOneJoule => 10000000;
 
-        protected override double EuThermsInOneJoule => 9.4781339449889105832843629746176e-9;
+        protected override double ThermsEcInOneJoule => 9.47816988e-9;
```
angularsen: Any reason you shortened this?
Also I get a slightly different number [here](http://www.metric-conversions.org/energy-and-power/joules-to-therms-ec.htm): `1J= 9.4781712e-9Therm(EC)`

DInozemtsev: @eriove Mentioned this is probably more correct value. Please see the screenshot below.

angularsen: Oh, okay. I have no familiarity with these units, why is there a different number in the calculator? Are there different standards for EC therms?

Second converter here:
http://www.kylesconverter.com/energy,-work,-and-heat/joules-to-therms-(e.c.)
`9.478171203e-9`

I think @eriove possibly found a value that was not entirely accurate.

From the last link:
> 1 Therm (E.C.):
Exactly 100,000 BTUIT. Approximately 1.05505585262 x 108 Joules (SI). Therm (E.C.) is based on the international table BTU, see Therm (U.S.) for Therms based on 59 °F BTU.

eriove: I took the number from EC Therms on the [Wikipedia page](https://en.wikipedia.org/wiki/Therm#Definitions) which links to the ISO BTU, but I have no familiarity either so @DInozemtsev have to decide.

DInozemtsev: Ok, I can explain how I got a previous value.
So according Wikipedia:
`1 Th(E.C.) ≡ 100,000 BTU = 105,505,600 joules`

To get 1 Joule in Therms (E.C.):

`1/105,505,600 = 9.4781698791343777012784155532976e-9`
 (this is what I get at Windows 10 Calculator.)

I think we should go with this value. @angularsen what do you think about this?

angularsen: I think your value 105,505,600 is a rounded value, please read the quote from wiki:
> 1 Therm (E.C.):
Exactly 100,000 BTUIT. Approximately 1.05505585262 x 108 Joules (SI). Therm (E.C.) is based on the international table BTU, see Therm (U.S.) for Therms based on 59 °F BTU

`1 Th EC = 1e5 BTU` exactly and approx `105,505,852.62 joules`. Then doing `1/x` gives me `9.4781712031331720001278504447561e-9`, which I believe is the correct value as it also matches the two online converters in above comments. This value should be used in the test case. Also since this is an infinite value, I think we should use the expression `x/105505585.262` in the JSON for `FromBaseToUnitFunc`, and `x*105505585.262` for `FromUnitToBaseFunc`.

DInozemtsev:
Ok, I've just pushed changes. Could you sirs, @angularsen @eriove, please review them?

angularsen:
Looks good to me!

eriove:
Looks good to me!

eriove:
@angularsen the [build](https://ci.appveyor.com/project/anjdreas/unitsnet) does not seem to get triggered. Did I do something wrong with version stepping?

angularsen:
@eriove Ah, I had to update AppVeyor to point to new github username, it's now building, but nuget push failed so I'll have to look at that in the evening.

angularsen:
Nuget 3.75 is out. I had to build and push manually from my own computer. There seems to be something wrong with appveyor, I'm piggybacking on someone elses' forum post here:
http://help.appveyor.com/discussions/problems/8215-cannot-publish-to-nuget-get-400-bad-request-error

PR #295, opened 2017-10-04
Added a few new units and new prefixes for existing units.

DInozemtsev:


DInozemtsev:
Sir @angularsen 
1st. I've fixed test values for **kBTU** and **mBTU** 
2nd. I can ask my colleagues from the States to make some screenshots of ENERGY STAR Portfolio Manager as a proof that they actually using such strange things as **kcf (thousand cubic feet)**, **Kcm (Thousand Cubic meters)** and **KGal (thousand gallons)** in theirs systems. 
Also please let me know if you still don't want those in your project and I'll get rid of them.

angularsen:
Ok, it's not that I don't trust you that someone is actually using such units, they do make sense, but my main concern is adding esoteric / rare units that no one else benefits from. I did google a bit more this time and I do find some references to usages of kilocubic feet, so I guess it's used enough in the wild to warrant being included in the library.

Thanks for the swift creation of the new pull request, the changes look good!

UnitsNet.Tests/CustomCode/EnergyTests.cs
```diff
@@ -50,6 +54,8 @@ public class EnergyTests : EnergyTestsBase
 
         protected override double FootPoundsInOneJoule => 0.737562149;
 
+        protected override double GigabritishThermalUnitsInOneJoule => 9.4781339449889105832843629746176e-13;
```
angularsen: Actually, this one seems to still use the old constant?

UnitsNet.Tests/CustomCode/FlowTests.cs
```diff
@@ -27,6 +27,8 @@ public class FlowTests : FlowTestsBase
 
         protected override double CubicDecimetersPerMinuteInOneCubicMeterPerSecond => 60000.00000;
 
+        protected override double CubicFeetPerHourInOneCubicMeterPerSecond => 1.2713300937959091422759633856933e+5;
```
angularsen: This value is slightly different than CubicFeetPerSecond multiplied by 3600.

`35.314666213 * 3600 = 1.271327983668e+5`

Which one should be corrected?

angularsen:
Found a couple more minor things, if you could address those then this should be ready to merge.

UnitsNet.Tests/CustomCode/FlowTests.cs
DInozemtsev: Since CubicFeetPerSecond was already there, I would use `1.271327983668e+5` as a test value for CubicFeetPerHour

UnitsNet.Tests/CustomCode/EnergyTests.cs
DInozemtsev: Oops... my bad.

UnitsNet.Tests/CustomCode/FlowTests.cs
angularsen: I confirmed with two online converters cubic feet per second:

`35.314666721489 ft3/s` from [converter1](https://www.unitjuggler.com/convert-flowrate-from-m3s-to-cfm.html)
`35.314 666 721 ft3/s` from [converter2](http://www.traditionaloven.com/tutorials/flow-rate/convert-m3-cubic-meter-per-second-to-ft3-cubic-foot-per-second.html)

The value is not perfectly identical, but pretty much good.

So, using the first value above times 3600, the per hour value should be: `127132.8001973604`.

If you could update both foot per second and foot per hour accordingly, then this should be good.

angularsen:
Commenting here so it is visible, see comment above: https://github.com/angularsen/UnitsNet/pull/295#discussion_r142793720

DInozemtsev:
Made it. ;)

angularsen:
Hooray :-) Thanks a bunch! Nuget 3.77 is now out.

PR #297, opened 2017-10-10
Fix missing prefixes for localized abbreviations

DiSnider:
$prefixIndex variable shouldn't be incremented while enumerating by $unit.Localization because in that case for second and for all the next cultures abbreviation will be taken from the subsequent prefix index that is wrong.


angularsen:
Great, thanks!
Next time, it would be good if you could run `generate-code.bat` and push the regenerated code files too. No biggie, just helps get an overview of the complete changes. I did verify locally and will push to master separately.

Issue #282, opened 2017-09-05
*.From*(decimal) overload is not present.


aignas:
Hello,

I noticed a small discrepancy in the API. Imagine that I have a `decimal`, which may be `nullable`, but that is not the main thing. In order to convert it to a Length type, for example, I have one of the following options:

```
decimal length = 3.14;
Length length = From.Meters((double)length);
```

or using the extension methods:

```
decimal length = 3.14;
Length length = length.Meters();
```

Am I missing something or is it indeed the case that the overloads for decimals are missing? Would you be willing to have this in the library?

anjdreas:
Hi, yes decimal overloads are missing. In retrospect, if I had written this today I would probably have picked decimal as the backing type to begin with, instead of double. But I'm all for adding `From` method overloads at least. It would still be represented by double internally and you would still get double in all the conversion properties, like `double m = myLength.Meters`, but it would be more convenient for code that already works with decimal. Also, I'm not sure how `double` to `decimal` assignment works and whether you need/want to use `decimal d = Convert.ToDecimal(myLength.Meters)` to retain more precision.

A couple of options:
1) Generate `From` method overloads taking `decimal` and `decimal?` instead of `double` and `double?`. 
2) Generate a duplicate of all quantity types, using `decimal` as internal representation and either call it `LengthD` vs `Length` or put it in a different namespace `UnitsNet.Decimal.Length` vs `UnitsNet.Length`. We could also consider releasing these new types in a separate nuget to avoid doubling the existing nuget size, such as a new nuget `UnitsNet.Decimal`.

Which option would you consider the best for your case?

aignas:
Hi, thanks for the reply. I think that adding the overloads would be a nice addition. In my case I need to consume decimals and convert everything into `Length`s and I am not terribly concerned with the backing type, however, I appreciate that there may be a case where this distinction may be important.

If it was up to me, I would do the following:
#. Add extra overloads taking `decimal` and `decimal?` and remove the `double` and `double?` overloads only if necessary.
#. Add an extra NuGet called `UnitsNet.Decimal` if we need the `decimal` as the backing type. However, I see a challenge for existing code bases where converting all `using` statements to a different `namespace` may pose a challenge, but at the same time I am not aware if `NuGet` provides a mechanism to define conflicting packages.

As the second step may require much more effort, I would open a separate issue for this. I would be quite happy to say that the issue is solved for me once the option 1 is implemented. :)

eriove:
[...]

PR #298, opened 2017-10-14
Adding decimal, int and long overloads

eriove:
As discussed in #282 I've added decimal overloads. If decimal is added, int and long has to be added too for overload resolution to work in a decent way (otherwise a cast or a suffix is needed for integers).

This increases the size of the binary with 134 KB from 709 to 843 KB.

I had extremely bad timing with your renaming of variables in the scripts so the rebase was quite tricky. Be on the lookout for errors related to this. I'll review this (again) tomorrow when I've slept on it for a night so please don't merge before I've reviewed it once more.

### DIFF
--- a/UnitsNet/GeneratedCode/Quantities/AmplitudeRatio.g.cs
+++ b/UnitsNet/GeneratedCode/Quantities/AmplitudeRatio.g.cs
@@ -178,6 +178,33 @@ namespace UnitsNet
             return new AmplitudeRatio(decibelmicrovolts - 120);
         }
 
+		/// <summary>
+        ///     Get AmplitudeRatio from DecibelMicrovolts.
+        /// </summary>
+        public static AmplitudeRatio FromDecibelMicrovolts(int decibelmicrovolts)
+        {
+            return new AmplitudeRatio(decibelmicrovolts - 120);
+        }
+
+		/// <summary>
+        ///     Get AmplitudeRatio from DecibelMicrovolts.
+        /// </summary>
+        public static AmplitudeRatio FromDecibelMicrovolts(long decibelmicrovolts)
+        {
+            return new AmplitudeRatio(decibelmicrovolts - 120);
+        }
+
+		// Windows Runtime Component does not support decimal type
+#if !WINDOWS_UWP
+		/// <summary>
+        ///     Get AmplitudeRatio from DecibelMicrovolts of type decimal.
+        /// </summary>
+        public static AmplitudeRatio FromDecibelMicrovolts(decimal decibelmicrovolts)
+        {
+	        return new AmplitudeRatio(Convert.ToDouble(decibelmicrovolts) - 120);
+        }
+#endif
+
         /// <summary>
         ///     Get AmplitudeRatio from DecibelMillivolts.
         /// </summary>
@@ -185,6 +212,33 @@ namespace UnitsNet
         {
             return new AmplitudeRatio(decibelmillivolts - 60);
         }
+
+		/// <summary>
+        ///     Get AmplitudeRatio from DecibelMillivolts.
+        /// </summary>
+        public static AmplitudeRatio FromDecibelMillivolts(int decibelmillivolts)
+        {
+            return new AmplitudeRatio(decibelmillivolts - 60);
+        }
+
+		/// <summary>
+        ///     Get AmplitudeRatio from DecibelMillivolts.
+        /// </summary>
+        public static AmplitudeRatio FromDecibelMillivolts(long decibelmillivolts)
+        {
+            return new AmplitudeRatio(decibelmillivolts - 60);
+        }
+
+		// Windows Runtime Component does not support decimal type
+#if !WINDOWS_UWP
+		/// <summary>
+        ///     Get AmplitudeRatio from DecibelMillivolts of type decimal.
+        /// </summary>
+        public static AmplitudeRatio FromDecibelMillivolts(decimal decibelmillivolts)
+        {
+	        return new AmplitudeRatio(Convert.ToDouble(decibelmillivolts) - 60);
+        }
+#endif
 
         /// <summary>
         ///     Get AmplitudeRatio from DecibelsUnloaded.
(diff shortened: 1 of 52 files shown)


UnitsNet/GeneratedCode/Quantities/AmplitudeRatio.g.cs
```diff
@@ -187,6 +214,33 @@ public static AmplitudeRatio FromDecibelMillivolts(double decibelmillivolts)
...
+
+        // Windows Runtime Component does not support decimal type
+#if !WINDOWS_UWP
+        /// <summary>
+        ///     Get AmplitudeRatio from DecibelMillivolts of type decimal.
+        /// </summary>
+        public static AmplitudeRatio FromDecibelMillivolts(decimal decibelmillivolts)
+        {
+            return new AmplitudeRatio(Convert.ToDouble(decibelmillivolts) - 60);
+        }
```
angularsen: I had an idea, but not sure if I prefer it or not myself:
```
public static double From<TNumber>(TNumber val) where TNumber : struct, IConvertible
{
    return Convert.ToDouble(val) - 60;
}
```

You get int, long, float, double, decimal all in one method - but the method signature is probably a bit weird when consuming it and it may also accept struct types that implement `IConvertible`. I just tried `TimeSpan` and it did not accept that type at least, so not sure how much of a problem that is. `TNumber` should hopefully indicate we are looking for a number type and Xmldoc could also be added for `<typeparam`> to describe what we expect here.
Ideally, c# added type constraints for only matching on number primitives and doing arithmetic on generic numbers, but it doesn't have that although I have seen some open source libs for that.

UnitsNet/Scripts/Include-GenerateQuantitySourceCode.ps1
```diff
@@ -197,6 +198,33 @@ namespace UnitsNet
             return new $quantityName($func);
         }
 
+        /// <summary>
```
angularsen: The build server complains about overloads not supported in Windows Runtime Component.

> [C:\projects\unitsnet\UnitsNet\UnitsNet.WindowsRuntimeComponent.csproj]
C:\projects\unitsnet\UnitsNet\GeneratedCode\Quantities\Information.g.cs(67,33,67,44): error WME1085: The 1-parameter overloads of UnitsNet.Information.FromTerabits must have exactly one method specified as the default overload by decorating it with Windows.Foundation.Metadata.DefaultOverloadAttribute. 

Try adding `Windows.Foundation.Metadata.DefaultOverloadAttribute` to the `double?` methods to tell it which overload to use for Windows UWP.
Alternatively wrap all the other overloads in `#if !WINDOWS_UWP ... #endif`, but I think the attribute should do the trick.

angularsen: Other than that, I don't see any traces of problems with the renamed files and folders, although I'll take a closer look when the AppVeyor build goes through.

eriove: I'm trying to add the trick. First time using UWP, I've got rid of the build errors locally at least.

UnitsNet/GeneratedCode/Quantities/AmplitudeRatio.g.cs
eriove: I didn't know of the `IConvertible` interface. I like the the current implementation better though, since we already do all the code generation I'm less worried about brevity and it has no side effects/unexpected behaviors.

Is there a particular reason we use `Convert` instead of type casting to double?

angularsen: Hm, not sure. I think it was to be on the safe side, as Convert methods do has some additional magic - such as rounding when converting to integers, whereas casting just floors the int.
I don't know if it makes a practical difference for ToDouble() though, I haven't looked into decimal to double cast vs conversion behavior.

UnitsNet/GeneratedCode/Quantities/AmplitudeRatio.g.cs
```diff
@@ -111,14 +111,14 @@ public AmplitudeRatio(double decibelvolts)
 
         #region Properties
 
-		/// <summary>
-		///     The <see cref="QuantityType" /> of this quantity.
-		/// </summary>
+        /// <summary>
```
eriove: Not sure what happened to the indentation here. The new one looks right to me though

angularsen: Might have been a sneaky tab char, or merge conflict when merging in the master changes to your branch.

UnitsNet/GeneratedCode/Quantities/AmplitudeRatio.g.cs
eriove: Checked the [reference source](https://referencesource.microsoft.com/#mscorlib/system/convert.cs,1596) and it casts internally. That method call will most likely be inlined by the JIT compiler so there should be no practical difference.

eriove:
I've reviewed the code and don't have more comments. Let me know if I should make any additional changes.

angularsen:
Looks good to me, my only concern is increasing the size of the binary from ~726kB to ~863kB just to achieve the convenience of not having to cast decimal values - because that is the primary motivation for this change, right? In light of this I'm having second thoughts on the generic type approach, but we can always consider that as an improvement for a future PR.

I realize this library is no toddler anymore as it nears the 1MB mark :-|

eriove:
Yeah, when I had to add long and int too I was worried that it would grow even more. A 19% size increase is considerable. I have some colleagues that work with very size constrained code, I'll check with them tomorrow to see what they say (they are using UnitsNet).

If someone is really size constrained [IL Linker](https://github.com/mono/linker) or [ILStrip](https://github.com/glueckkanja/ILStrip) seems like the right way to go.

angularsen:
Yup, there is also discussion to add this to .NET core: https://github.com/dotnet/announcements/issues/30

There are many other options too though, for this library. We could split out a `UnitsNet.Core` with some definition of most widely used units, then maybe addon packages such as `UnitsNet.ElectricalEngineering` or even smaller packages such as `UnitsNet.Force`. Not sure how this would all work with overlapping engineering domains and sounds like considerable amount of effort, but if someone needs this badly enough ... where there is will there is a way :-)

PR #305, opened 2017-10-27
Feature: Added MolarMass Quantity Type

### LINKED ISSUE #303: Request: Molar Mass
together with my colleague @ferittuncer, we are modelling a thermodynamic library for a several chemical species. To be able to freely change unit frame, we need to have a `MolarMass `quantity type data on hands. 

For this purpose we propose to contribute to the library by including this type. 

Waiting for feedbacks. Best regards

ferittuncer:
Implemented molar mass which we discussed in #303 

Fixes #303


UnitsNet/UnitDefinitions/MolarMass.json
```diff
@@ -0,0 +1,43 @@
...
+        {
+          "Culture": "ru-RU",
+          "Abbreviations": [ "г/моль" ],
+          "AbbreviationsWithPrefixes": [ "нг/моль", "мкг/моль", "мг/моль", "сг/моль", "дг/моль", "даг/моль", "гг/моль", "кг/моль" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "PoundPerMole",
+      "PluralName": "PoundsPerMoles",
```
angularsen: I'm fixing this to `PoundsPerMole` and `GramsPerMole`.

UnitsNet/UnitDefinitions/MolarMass.json
```diff
@@ -0,0 +1,43 @@
...
+      "SingularName": "PoundPerMole",
+      "PluralName": "PoundsPerMole",
+      "FromUnitToBaseFunc": "x*0.45359237",
+      "FromBaseToUnitFunc": "x/0.45359237",
+      "Prefixes": [ "Kilo", "Mega" ],
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "lb/mol" ],
+          "AbbreviationsWithPrefixes": [ "KLbs/mol", "MLbs/mol" ]
```
angularsen: Kilo should be lowercase `klbs`. Also, I think I prefer the singular form when it comes to abbreviations. I know there are varying practices here, but unless `klbs/mol` is clearly preferred over `klb/mol` in this domain then I think we should use singular form to stay consistent in the library.

angularsen:
I took the liberty of editing my suggestions in directly. Please review if you agree, then this should be ready to merge.

angularsen:
I can't push the generated code apparently, so if you could run `generate-code.bat` and push the regenerated code up that would be great.

ferittuncer:
I reviewed your changes and didn't find any problem. I think it's ready to merge.

UnitsNet/UnitDefinitions/MolarMass.json
```diff
@@ -0,0 +1,42 @@
...
+      "Prefixes": [ "Nano", "Micro", "Milli", "Centi", "Deci", "Deca", "Hecto", "Kilo" ],
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "g/mol" ]
+        },
+        {
+          "Culture": "ru-RU",
+          "Abbreviations": [ "г/моль" ],
+          "AbbreviationsWithPrefixes": [ "нг/моль", "мкг/моль", "мг/моль", "сг/моль", "дг/моль", "даг/моль", "гг/моль", "кг/моль" ]
```
angularsen: Can we drop these prefixes and rely on the auto-generated prefixes, like the unit below?
It seems to me they only add the standard prefix letter here.

angularsen: Never mind, we can do that separately. No need to delay the PR over this.

angularsen:
Awesome, looks good to me!

UnitsNet/UnitDefinitions/MolarMass.json
ferittuncer: Yes, I thought those prefixes were necessary. But if it's not, let's drop them.

angularsen: Actually, it turns out only English is supported by the scripts. It should be trivial to extend the script to support other languages too as long as they simply prepend a certain string for each prefix. Russian seems to do this, but it is not a language I am familiar with.

ferittuncer: I am going to open an issue about this to remember later.

PR #306, opened 2017-10-27
Use generic number type to reduce From() method overload count

angularsen:
This is a follow up of #298 where many new overloads were introduced for each of the 550+ units.

This PR tries to mitigate this by:
* Add new `struct QuantityValue` type, with implicit cast overloads from all numeric types and implicit cast back to `double`. The overload hit is done once - instead of 550 times for all units.
* Use `FromValue` namespace alias to simplify compatibility with Windows Runtime Component (reduce the amount of #if sections)

I tried using generic type constraints, but I couldn't get it to infer the numeric types for the nullable overloads.

#### Result

UnitsNet.dll is reduced from **870 to 800 kB** for 50 quantities and 550 units. The reduction is also a proportional to the number of units and will grow better as we add more units. The number of overloads was reduced to 1/4th, while also adding support for three new numeric types: short, float and byte.
We still need to have two overloads for nullability.

### DIFF
--- a/UnitsNet/QuantityValue.cs
+++ b/UnitsNet/QuantityValue.cs
@@ -40,7 +40,7 @@ namespace UnitsNet
 
         // Obsolete is used to communicate how they should use this type, instead of making the constructor private and have them figure it out
         [Obsolete("Do not use this constructor. Instead pass any numeric value such as int, long, float, double, decimal, short or byte directly and it will be implicitly casted to double.")]
-        public QuantityValue(double val)
+        private QuantityValue(double val)
         {
             _value = val;
         }
--- a/UnitsNet/QuantityValue.cs
+++ b/UnitsNet/QuantityValue.cs
@@ -1,0 +1,68 @@
+﻿// Copyright © 2007 Andreas Gullberg Larsen (angularsen@gmail.com).
+// https://github.com/angularsen/UnitsNet
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Operator overloads not supported in Windows Runtime Components, we use 'double' type instead
+#if !WINDOWS_UWP
+using System;
+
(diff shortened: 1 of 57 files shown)


angularsen:
Any thoughts on this approach @eriove @JKSnd ?
Does it make the intellisense around consuming the From methods too weird?

JKSnd:
In theory I like the idea, but I'm not set up to kick it around conveniently at the moment. Could maybe get around to pulling and building and playing with sometime in the next couple days though. Sorry I don't have more bandwidth right now

angularsen:
Thanks, that would be great. There is no rush, I can wait a few days.

UnitsNet/QuantityValue.cs
```diff
@@ -0,0 +1,68 @@
...
+    ///     From 8 (int, long, double, decimal + each nullable) down to 2 (QuantityValue and QuantityValue?).
+    ///     This also adds more numeric types with no extra overhead, such as float, short and byte.
+    /// </remarks>
+    public struct QuantityValue
+    {
+        private readonly double _value;
+
+        // Obsolete is used to communicate how they should use this type, instead of making the constructor private and have them figure it out
+        [Obsolete("Do not use this constructor. Instead pass any numeric value such as int, long, float, double, decimal, short or byte directly and it will be implicitly casted to double.")]
+        public QuantityValue(double val)
```
eriove: This can be made private instead, or have I missed something?

UnitsNet/QuantityValue.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright © 2007 Andreas Gullberg Larsen (angularsen@gmail.com).
```
eriove: Wrong year

UnitsNet/QuantityValue.cs
```diff
@@ -0,0 +1,68 @@
...
+        public static implicit operator QuantityValue(decimal val) => new QuantityValue(Convert.ToDouble(val));
+        public static implicit operator QuantityValue(short val) => new QuantityValue(val);
+        public static implicit operator QuantityValue(byte val) => new QuantityValue(val);
+#pragma warning restore 618
+        
+        #endregion
+
+        #region To double
+
+        public static implicit operator double(QuantityValue number) => number._value;
```
eriove: Would there be a downside to making this explicit? (Except for longer generated code)

It somehow feels safer to keep the implicit conversions to the minimum.

UnitsNet/QuantityValue.cs
angularsen: Fixed in master and in branch, I saw most headers were out of...date... 😎

UnitsNet/QuantityValue.cs
angularsen: Good catch, fixed.

UnitsNet/QuantityValue.cs
angularsen: Another good observation, you are right this should only affect generated code and we should be able to make this `internal` too.

angularsen: Nope, `internal` is not valid.

angularsen: Not yet addressed this, will soon.

angularsen: Fixed.

angularsen:
@JKSnd Any input before we merge this in?

Latest numbers on reducing size are from 917 to 852 kB, around 7% saving.

Also, does anyone know of a good tool to evaluate the size breakdown a .NET .dll ? 

I've googled a bit but can only find one for Win32 .dll files. It would be interesting to learn what pieces of code or data take up the most place to see if we can further reduce the footprint. 

After a brief glance, it seems to me that the library primarily consists of a ton of methods with small method bodies, such as all the number extension methods and all the quantity `From()` methods and unit conversion properties. Very few methods are fat and they are typically not generated N times.

eriove:
I haven't seen such a tool. Would be useful though.

angularsen:
Ok, merged.

eriove:
This broke our build today. I'm not sure if there is anything we can do to avoid that happening to others. The only thing I can think of is adding back the old constructors and increase the size again.

angularsen:
Hm.. Not good, but you did foresee this happening earlier in this thread so I guess it's not entirely unexpected.
I think we will just have to live with it as I don't want to go back to the larger size. Consolidating versions or at least upgrading all versions to get this recent change is not unreasonable, but in hindsight it would probably be better to bump the major version for it.

PR #309, opened 2017-11-07
Feature: MolarEnergy

### LINKED ISSUE #301: Request: MolarEnergy
We are modelling entropy and enthalpy in a scientific software project. Scientific papers we use, refer to molar based measurements such as `Joule per Mole` and `Joule per Mole per Kelvin`.

We couldn't find a good way to implement this using the library as-is. So we are willing to implement and contribute to the library but we thought it would be better if we discuss first.

Perhaps we can introduce a quantity type called MolarEnergy for it. What are your thoughts?

ferittuncer:
Implemented MolarEnergy 
Fixes #301


angularsen:
Looks good!

PR #312, opened 2017-11-07
New Quantity: Entropy

### LINKED ISSUE #311: Proposal: Entropy
We would like to implement `Entropy` quantity according to this page: 

- https://physics.nist.gov/cuu/Units/units.html

ferittuncer:
Implemented Entropy quantity
Fixes #311


angularsen:
Just a tip, no need to add an issue first if you plan to create a pull request immediately anyway. The discussion can then continue in the PR instead. However, it may sometimes be good to create an issue first to discuss whether the feature is desired so you don't go through a lot of work only to get it rejected. Not that I plan to reject this, but just a heads up.

Also, please add "Fixes #000" to the PR description so that when the PR is merged it will auto-close the issue.

UnitsNet.Tests/CustomCode/EntropyTests.cs
```diff
@@ -0,0 +1,55 @@
...
+// THE SOFTWARE.
+
+
+using System;
+
+namespace UnitsNet.Tests.CustomCode
+{
+    public class EntropyTests : EntropyTestsBase
+    {
+        protected override double CaloriesPerKelvinInOneJoulePerKelvin => 1 / 4.184;
```
angularsen: Test constants should not be computations, but rather constants looked up in some reference material. This is an important step to avoid copying the same implementation over to the test. The values look correct though.

F.ex use this converter: http://www.gordonengland.co.uk/conversion/entropy.htm

ferittuncer: Well I think using only one constant and describing manipulations for conversions is better. What do I mean:

1 Calorie equals to 4.184 Joule, but if we start from the other end 1 Joule equals to 0.239006 Calorie. Conversion constant is the same (by value) but reader can't say that they are the same until calculating it.

However if we pick 4.184 or 0.239006 as main constant, we can describe other conversions as manipulations of the main constant and reader can easily verify that they are the same thus avoiding conversion bugs.

For example:
1 Calorie equals to 4.184 Joules and 1 Joule equals to 1/4.184.


But anyway, I'll refactor as you wish.

ferittuncer: So we should show computation steps in implementations, resulting values in tests, right?

For example: 1/4.184 in implementation, 0.239006 in test.

ferittuncer: We can implement Specific Entropy. And you already merged Molar Entropy :)

ferittuncer: I'm going to work on other units you suggested.

angularsen: Correct, the important part is to basically double check that the implementation is correct. You can reuse the constant and multiple it by `1e3` for the kilo prefixes etc. That's fine I think and is easy to confirm for correctness.

ferittuncer:
I see that some units has IT subscript, when I googled I saw that I 1 calorie equals to 4.1840 Joule and 1 calorieIT equals to 4.1868 Joule.

Also I haven't encountered to any calorieIT unit in the library yet. 

Should I proceed and implement these (for example: `calorieIT per degree Celsius`)? Should I implement both calorie and calorieIT versions? What do you think?

angularsen:
I have no domain knowledge here, so I can't say for sure whether both are widely used or not.
However, google's own unit converter does not include CalorieIT, and I don't find that many search results on the IT variants so that indicates to me it is perhaps not very widely used? I really don't know. If we are unsure if anyone will need it, then I vote to skip it until someone asks for it.

ferittuncer:
I agree that we should skip units with IT subscript for now, until someone asks for it. Then I think it's ready to merge, could you please review?

UnitsNet/UnitDefinitions/Entropy.json
```diff
@@ -0,0 +1,46 @@
+﻿{
+  "Name": "Entropy",
+  "BaseUnit": "JoulePerKelvin",
+  "XmlDoc": "",
```
angularsen: Did you forget to add the xmldoc?

I'm adding in this text based on a few sections in the wiki:

> Entropy is an important concept in the branch of science known as thermodynamics. The idea of "irreversibility" is central to the understanding of entropy.  It is often said that entropy is an expression of the disorder, or randomness of a system, or of our lack of information about it. Entropy is an extensive property. It has the dimension of energy divided by temperature, which has a unit of joules per kelvin (J K−1) in the International System of Units.

PR #317, opened 2017-11-10
Hotfix: Deleted Duplicate Quantity: SubstanceAmount

ferittuncer:
Looks like when we changed SubstanceAmount to AmountOfSubstance, generated codes for SubstanceAmount remained.


angularsen:
Ah, good catch. I did look out for that, but somehow missed it.

PR #321, opened 2017-11-10
Hotfix: Fixed LapseRate Units

### LINKED ISSUE #319: LapseRate Unit: KiloDegreesCelsiusPerMeter instead of DegreesCelsiusPerKilometer
I recently sent a pull request for LapseRate quantity which is merged, but I see that I created a unit `KiloDegreesCelsiusPerMeter` instead of `DegreesCelsiusPerKilometer`. I want to fix it but I don't know how to prefix a unit which comes after `Per`. 

I checked other quantities, there is no other quantity which has a prefix after `Per`. So I wonder, is it possible? If not, should I use `MilliDegreesCelsiusPerMeter` instead?

ferittuncer:
Fixes #319

### DIFF
--- a/UnitsNet/UnitDefinitions/LapseRate.json
+++ b/UnitsNet/UnitDefinitions/LapseRate.json
@@ -1,14 +1,13 @@
 ﻿{
   "Name": "LapseRate",
-  "BaseUnit": "DegreeCelsiusPerMeter",
+  "BaseUnit": "DegreeCelsiusPerKilometer",
   "XmlDoc": "Lapse rate is the rate at which Earth's atmospheric temperature decreases with an increase in altitude, or increases with the decrease in altitude.",
   "Units": [
     {
-      "SingularName": "DegreeCelsiusPerMeter",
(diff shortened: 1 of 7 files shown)


UnitsNet/UnitDefinitions/LapseRate.json
```diff
@@ -1,14 +1,13 @@
...
   "XmlDoc": "Lapse rate is the rate at which Earth's atmospheric temperature decreases with an increase in altitude, or increases with the decrease in altitude.",
   "Units": [
     {
-      "SingularName": "DegreeCelsiusPerMeter",
-      "PluralName": "DegreesCelciusPerMeter",
+      "SingularName": "DegreeCelsiusPerKilometer",
+      "PluralName": "DegreesCelciusPerKilometer",
       "FromUnitToBaseFunc": "x",
       "FromBaseToUnitFunc": "x",
-      "Prefixes": [ "Kilo" ],
```
angularsen: I assume you don't want DegreeCelsiusPerMeter anymore? It was included before in addition to the Kilo one.

PR #322, opened 2017-11-16
New Quantity: SpecificEntropy

gratestas:


ferittuncer:
I reverted `bf96253`, which is an irrelevant commit, and fixed conflicts. There is another issue with SpecificEntropy conversions and @gratestas is working on it.

UnitsNet/UnitDefinitions/SpecificEntropy.json
```diff
@@ -0,0 +1,47 @@
+{
+  "Name": "SpecificEntropy",
+  "BaseUnit": "JoulePerKilogramKelvin",
+  "XmlDoc": "The SpecificEntropy",
+  "XmlDocRemarks": "Specific entropy is an amount of energy required to raise temperature of a substance by 1 Kelvin per unit mass",
```
angularsen: XmlDocRemarks is not supported, move this into XmlDoc.

UnitsNet/UnitDefinitions/SpecificEntropy.json
```diff
@@ -0,0 +1,47 @@
...
+          "Culture": "en-US",
+          "Abbreviations": [ "J/(kg*C)" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "CaloriePerGramKelvin",
+      "PluralName": "CaloriesPerGramKelvin",
+      "FromUnitToBaseFunc": "x*4.184e3",
+      "FromBaseToUnitFunc": "x/4.184e3",
```
angularsen: I got ` 2.390057e-4` at: https://www.gordonengland.co.uk/conversion/specific_heat.htm
And `0.0002390057361377` at https://www.translatorscafe.com/unit-converter/en/specific-heat-capacity/1-6/joule%2Fkilogram%2FK-calorie%20(IT)%2Fgram%2F%C2%B0C/

ferittuncer: Yeah, it should be `"FromUnitToBaseFunc": "x*4.184e-3"`  instead of  `"FromUnitToBaseFunc": "x*4.184e3"`.

ferittuncer: But somehow I can't manage to make the tests pass when `x*4.184e3`. Hmm...

ferittuncer: Just asking to prevent any potential misunderstanding: Do you suggest to change the `2.39005e-4` with what you got, `2.390057e-4`, in `SpecificEntropyTests.cs`?

angularsen: I would use the must accurate value found in a references source ` 0.0002390057361377` in the test.
Then these JSON functions must also be updated as they seem wrong, but I don't know if they should be represented as some calculation or as the same constant.

ferittuncer:
I changed the abbreviations according to https://www.princeton.edu/~maelabs/mae324/glos324/entropy.htm

UnitsNet/UnitDefinitions/SpecificEntropy.json
ferittuncer: I couldn't spot the error in JSON function :/

Fixed values in tests.

ferittuncer: BTW, https://www.translatorscafe.com/unit-converter/en/specific-heat-capacity/1-6/joule%2Fkilogram%2FK-calorie%20(IT)%2Fgram%2F%C2%B0C/ gives `0.000238845896627` not `0.0002390057361377`.

I updated values in the test as `0.0002390057`.

I would like to leave this p.r. to you and @gratestas to finalize before merging, as I'm a bit confused.

angularsen: Careful, you linked to the IT calorie unit. It is *slightly* different than `Th` calories. I think we should add both and name them explicitly to avoid ambiguity.

ferittuncer: Oh, yeah, but I copied your link :)

I agree that we should fix this ambiguity in the entire library soon.

angularsen: Sorry, my bad. The JSON functions are probably good, at least they match the constant pretty well.

angularsen:
Nice, @ferittuncer do you want to do the honors of releasing a new nuget?

ferittuncer:
Sure @angularsen, 3.80.0 is on its way! :)

angularsen:
Wohoo! :-)

angularsen:
@ferittuncer I noticed you didn't push the git tag 3.80, so I've updated the [wiki](https://github.com/angularsen/UnitsNet/wiki/Collaborators:-Release-New-Nuget) with a missing parameter `git push --follow-tags` that auto-pushes the tags when you push the code. I use a bash alias `gpu='git push --follow-tags --set-upstream'` that does this for me, so I forgot about it.

If you just want to push tags, and not code, you can run `git push --tags`.

If you could run either of these two commands to push the tag you have locally, that would be great!

ferittuncer:
Thanks for information. I am going to do just as you described next time @angularsen. I see that you released 3.81.0 now, so do I have to take any action for now?

angularsen:
Yes, please push tag 3.80.0 so that the repository has a tag for the release. I could create the tag, but then you and me/others would have different versions of the same tag.

ferittuncer:
@angularsen I did push the tags but it seems I pushed a wrong tag mistakenly too. Sorry, I'm not experienced with tags. What should we do about this? https://github.com/angularsen/UnitsNet/commit/7bdd0f0c8bdd530b5ea9b5ed7e994c8dd8ba3281

angularsen:
No biggie, just delete the tag both locally and on remote.

```
git push --delete origin tagname   #remote 
git tag --delete tagname                #locally
```

The only thing is that if anyone else has pulled from the repo, they will also have the tag locally and if they are a collaborator they will re-push the same tag back up again. Git is a bit messy on these things.

ferittuncer:
Yup, tags corrected. : )

PR #323, opened 2017-11-17
changed expressions in term of algebraic operations with corresponding numeric results

gratestas:
as it is on demand, conversion values of a property in `UnitDefinition/*.json` files and those in `CustomCode/*.cs` files must be expressed differently in order to lead tests to reasonable execution. in case of specific energy, I noticed that there is no such consistency. on both side conversion factors from/to base unit are expressed identically. For instance,

in UnitDefinition : ` "FromBaseToUnitFunc": "x/4.184e3" `
in CustomCode : ` "CaloriesPerGramInOneJoulePerKilogram": "x/4.184e3",`

### DIFF
--- a/UnitsNet.Tests/CustomCode/SpecificEnergyTests.cs
+++ b/UnitsNet.Tests/CustomCode/SpecificEnergyTests.cs
@@ -27,9 +27,9 @@ namespace UnitsNet.Tests.CustomCode
     {
         protected override double JoulesPerKilogramInOneJoulePerKilogram => 1.0;
 
-        protected override double CaloriesPerGramInOneJoulePerKilogram => 1.0/4.184E3;
+        protected override double CaloriesPerGramInOneJoulePerKilogram => 2.410800386e-4;
 
-        protected override double KilocaloriesPerGramInOneJoulePerKilogram => 1.0/4.184E6;
+        protected override double KilocaloriesPerGramInOneJoulePerKilogram => 2.410800386e-7;
 
 
         protected override double KilojoulesPerKilogramInOneJoulePerKilogram => 1.0E-3;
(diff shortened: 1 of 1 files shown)


UnitsNet.Tests/CustomCode/SpecificEnergyTests.cs
```diff
@@ -27,9 +27,9 @@ public class SpecificEnergyTests : SpecificEnergyTestsBase
     {
         protected override double JoulesPerKilogramInOneJoulePerKilogram => 1.0;
 
-        protected override double CaloriesPerGramInOneJoulePerKilogram => 1.0/4.184E3;
+        protected override double CaloriesPerGramInOneJoulePerKilogram => 2.410800386e-4;
```
angularsen: Computing this value gives me: `0.00023900573`
Was the function not correct to begin with or is this `IT` vs `th` units?
In retrospect, we maybe should have added `It` and `Th` suffixes to the unit names to disambiguate them.

UnitsNet.Tests/CustomCode/SpecificEnergyTests.cs
```diff
@@ -27,9 +27,9 @@ public class SpecificEnergyTests : SpecificEnergyTestsBase
...
 
-        protected override double KilocaloriesPerGramInOneJoulePerKilogram => 1.0/4.184E6;
+        protected override double KilocaloriesPerGramInOneJoulePerKilogram => 2.410800386e-7;
```
angularsen: Same here.

UnitsNet.Tests/CustomCode/SpecificEnergyTests.cs
ferittuncer: Looks like he divided with 41**48** instead of 41**84**. I'm fixing this now.

And perhaps we should either say in README that we don't support IT's but only TH's, or we should name them as you said.

ferittuncer:
Please take a look @angularsen

PR #324, opened 2017-11-20
Feature: Add LapseRate extensions

jnyrup:
This PR extends LapseRate with four operator overloads:

* TemperatureDelta  / LapseRate -> Length
* Temperature  / LapseRate -> Length
* Length * LapseRate -> TemperatureDelta
* LapseRate * Length -> TemperatureDelta

### DIFF
--- a/UnitsNet/CustomCode/Quantities/LapseRate.extra.cs
+++ b/UnitsNet/CustomCode/Quantities/LapseRate.extra.cs
@@ -1,0 +1,48 @@
+﻿// Copyright (c) 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com).
+// https://github.com/angularsen/UnitsNet
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+// 
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+// 
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+namespace UnitsNet
+{
+    // Windows Runtime Component has constraints on public types: https://msdn.microsoft.com/en-us/library/br230301.aspx#Declaring types in Windows Runtime Components
+    // Public structures can't have any members other than public fields, and those fields must be value types or strings.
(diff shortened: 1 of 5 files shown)


ferittuncer:
Welcome and thanks for the contribution. I couldn't find any errors but let's wait for @angularsen for merging.

jnyrup:
What are your gut feelings about the `Temperature / LapseRate -> Length`?
Will it cause more harm than good based on previous confusion related to the difference between `Temperature` and `TemperatureDelta`.

And similar, how about conversions to LapseRate?

I think `TemperatureDelta / Length -> LapseRate` makes perfect sense, while `Temperature / Length -> LapseRate` might  too ambiguous?

For consistency I could imagine we wanted to either only use `TemperatureDelta` or use `Temperature` in both conversions from and to `LapseRate`.

Any preferences on whether `TemperatureDelta / Length -> LapseRate` is an extension to `TemperatureDelta` or `Length`?

angularsen:
If I understand correctly, I think `Temperature` cannot be used for the same reason you can't add two temperatures, you can only add a delta (range on scale) to a temperature (point on scale) or add two deltas to get a new larger delta.

`LapseRate [▲°C/m]` describes the gradient for how much the temperature changes (delta) over a distance, so the proposed equation becomes:

```
TemperatureDelta [▲°C] / LapseRate [▲°C/m] = 
TemperatureDelta [▲°C] / (TemperatureDelta [▲°C] / Length [m]) = 
Length [m]
```

In hindsight, LapseRate abbreviation should probably be changed to reflect this? It is currently missing the delta symbol. But not sure if this goes against the convention in the domain.

> Any preferences on whether TemperatureDelta / Length -> LapseRate is an extension to TemperatureDelta or Length?

I would say TemperatureDelta, because it is the leading quantity.

UnitsNet/CustomCode/Quantities/LapseRate.extra.cs
```diff
@@ -0,0 +1,53 @@
...
+#endif
+    {
+        // Windows Runtime Component does not allow operator overloads: https://msdn.microsoft.com/en-us/library/br230301.aspx
+#if !WINDOWS_UWP
+        public static Length operator /(TemperatureDelta left, LapseRate right)
+        {
+            return Length.FromKilometers(left.KelvinsDelta / right.DegreesCelciusPerKilometer);
+        }
+
+        public static Length operator /(Temperature left, LapseRate right)
```
angularsen: I vote to drop this.

angularsen: If this is removed, I think this PR is good to go.

jnyrup:
Thanks for the review. 
I've updated the PR with two changes: 

* removed the `Temperate / LapseRate -> Length` 
* added `TemperatureDelta / Length -> LapseRate`

Regarding the abbreviation:

* Currently it is `°C/m` for `LapseRateUnit.DegreeCelsiusPerKilometer`, which is wrong due to a missing a `k`.
* `▲°C/km` would be fine with me.

ferittuncer:
@jnyrup About missing k, yeah I did implement that and when I changed base unit forgot to change the abbreviation also, my bad. 

And yeah, we should change the abbreviation to `▲°C/km`

jnyrup:
And related, why do we use ▲ instead of ∆?

ferittuncer:
I'm not aware of a specific reason for using ▲. I used ▲ because @angularsen used it in his comment above. Let's wait for his answer.

angularsen:
No preference for me other than it already being used for TemperatureDelta, I suppose ∆ is more correct.

ferittuncer:
Hmm, then I'm opening an issue for replacing every occurance of ▲ with ∆ in the library, for the sake of consistency.

I also see some minor differences in code-style in various places (talking in general, not about this p.r.). Perhaps we can set a standard on this and write it to wiki. Standards make contributing less confusing and reviewing less difficult, I believe.

angularsen:
That would be good! And maybe adding it as a contribution guideline so it pops up whenever someone creates a new issue / PR.

jnyrup:
I've updated the PR so the unit of lapse rate is now `▲°C/km`.

angularsen:
@jnyrup Shouldn't we use the new delta symbol? Per #328 .

jnyrup:
I thought that was the job of #327?

angularsen:
Sure, but if you are already modifying the abbreviation here, why not use the correct symbol to begin with?

jnyrup:
The PR has been updated with the correct delta symbol.
Thanks for the fast reviews!

angularsen:
Thanks! Nuget 3.81 is out.

PR #329, opened 2017-11-25
fix subscripts and superscripts

jnyrup:
I found some places where a number wasn't sub/super-scripted.


angularsen:
Nice! Thanks.

Issue #328, opened 2017-11-24
Contribution Guideline


ferittuncer:
As discussed in #324, we should write up a contribution guideline to make it easier to contribute and review.

Rules that are enforced by test cases are marked with an asterisk `*` and #328 will add the initial set of tests.

Updated 2018-02-04 by Andreas:
### Coding Conventions
* Match the existing code style, we generally stick to "Visual Studio defaults" and [.NET Foundation Coding Guidelines](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md)
* If you use ReSharper there is a [settings file](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet.sln.DotSettings) that will take effect automatically
* There is an [.editorconfig](https://github.com/angularsen/UnitsNet/blob/master/.editorconfig) to configure whitespace and some C# syntax
* Add the file header to new files you create

#### Test Code
* Test class: Use `Tests` suffix for the type you are testing, such as `UnitSystemTests`
* Test method: `<method>_<condition>_<result>` (`Parse_AmbiguousUnits_ThrowsException`)
* If there are many tests for a single method, you can wrap those in an inner class named the same as the method and then you can skip that part of the test method names

### Unit definitions (.JSON)
For a fairly complete summary of the unit definition JSON schema, see [Meter of Length](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/UnitDefinitions/Length.json#L7). It has prefix units and multiple cultures.

#### Conversion functions
Converting from unit A to B is achieved by first converting from unit A to the base unit, then from the base unit to unit B. To achieve this, each unit defines two conversion functions.

* Prefer multiplication for `FromUnitToBaseFunc` (`x*2.54e-2` for `Inch` to `Meter`)
* Prefer division for `FromBaseToUnitFunc` (`x/2.54e-2` for `Meter` to `Inch`)
* Prefer scientific notation `1e3` and `1e-5` instead of `1000` and `0.00001`
* Prefer a constant if the conversion factor is finite (`x*2.54e-2` for `Inch`)
* Prefer a calculation if the conversion factor is infinite (`(x/72.27)*2.54e-2` for `PrinterPoint`)

#### Units
Generally we try to name the units as what is the most widely used.

* Use prefix for country variants, such as `ImperialGallon` and `UsGallon`

[...]

PR #333, opened 2017-12-04
Acceleration units g, ft/s² and in/s² have been added.

DrDroop:


UnitsNet/UnitDefinitions/Acceleration.json
```diff
@@ -1,20 +1,56 @@
...
+          "Culture": "en-US",
+          "Abbreviations": [ "ft/s²" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "StandardGravity",
+      "PluralName": "StandardGravity",
+      "FromUnitToBaseFunc": "x*9.80665",
+      "FromBaseToUnitFunc": "x*0.1019727",
```
angularsen: This constant is infinite, so use `x/9.80665` to preserve the precision.

UnitsNet/UnitDefinitions/Acceleration.json
```diff
@@ -1,20 +1,56 @@
...
+          "Culture": "en-US",
+          "Abbreviations": [ "in/s²" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "FootPerSecondSquared",
+      "PluralName": "FeetPerSecondSquared",
+      "FromUnitToBaseFunc": "x*0.304800",
+      "FromBaseToUnitFunc": "x*3.28084",
```
angularsen: Use `x/0.304800` to preserve precision.

UnitsNet/UnitDefinitions/Acceleration.json
```diff
@@ -1,20 +1,56 @@
...
+          "Culture": "en-US",
+          "Abbreviations": [ "m/s²" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "InchPerSecondSquared",
+      "PluralName": "InchesPerSecondSquared",
+      "FromUnitToBaseFunc": "x/39.3700787",
+      "FromBaseToUnitFunc": "x*39.3700787",
```
angularsen: Based on http://www.kylesconverter.com/acceleration/meters-per-second-squared-to-inches-per-second-squared
FromUnitToBasefunc: `x*0.0254`
FromBaseToUnitFunc: `x/0.0254`

DrDroop: Should we also include abbreviations that do not rely on the superscript?  in/s^2  Do you have a standard way of doing this?

UnitsNet/UnitDefinitions/Acceleration.json
DrDroop: Should we also include abbreviations that do not rely on the superscript?  ft/s^2

UnitsNet/UnitDefinitions/Acceleration.json
angularsen: I'm not sure I follow. Are you saying you would like to add `in/s^2` in addition to `in/s²`, so that both can be parsed?

We recently discussed this in our work in progress contribution guideline, where we favor superscripts: https://github.com/angularsen/UnitsNet/issues/328

However, it is a point that it may be desirable to be able to parse non-superscript variants as well.
I believe you can achieve this by simply adding it: `"Abbreviations": [ "in/s²", "in/s^2" ]`
This will make the superscript one the default, used in ToString(), but also support parsing strings like `Acceleration.Parse("5.3 in/s^2")`. If that is what you meant, I think that is how you would do it in this PR.

We could consider enhancing the code generation script to automatically add `^2` and `^3` alternatives to abbreviations that use ² and ³ superscripts, so we don't have to manually duplicate this in all the .JSON files.

UnitsNet/UnitDefinitions/Acceleration.json
angularsen: On second thought, I think I would say no, because we haven't usually done so. But I get your point and as in my previous comment, I am thinking maybe the library should generally allow parsing these ^2 variants as substitutes for superscripts.

UnitsNet/UnitDefinitions/Acceleration.json
DrDroop: Yes, I'm talking about adding additional abbreviations that might be more commonly used in coding.   Certainly having the code generation scrip automatically add the ^2 and ^3 would be ideal.  I will update the abbreviations.

DrDroop: Have you made the changes that are seen above?  I'm pretty green when it comes to Git and not entirely sure how to get the most recent version of Acceleration.json locally.

angularsen: I haven't made any changes. So, since you are working on your fork's master, just continue to add more commits to that branch and push those (to your fork's master). The new commits will then show up in this PR.

Next time though, it's better to create a new branch such as `add-acceleration-units` so that you can just delete that branch when the pull request is merged. It can be confusing to figure out local vs remote master branch, as well as your fork's master vs the original repo's master. I prefer to point my local master to the original repo's master, so that I can just `git pull` to get the lastest official changes. Many ways to do this though, just ask and I'll try to help.

UnitsNet/UnitDefinitions/Acceleration.json
DrDroop: Fair enough; thanks.

UnitsNet/UnitDefinitions/Acceleration.json
DrDroop: I made the precision changes and did a push.

DrDroop:
Awesome!  Thank you!

PR #335, opened 2017-12-07
Add .NET 4.0 target

### LINKED ISSUE #334: Additional net40 and net45 compiled Dlls
Hi,
your project is very useful and works great in our software, but i have to use only net 4.0 or net 4.5 compiled dlls.
Can you add additional compiled to your package?

regards
Stefan

angularsen:
Consumers targeting 4.0 or 4.5 without the .NET Core SDK will have to load the old 3.5 runtime in order to run this library, which costs performance and memory.

Adding this target allows those consumers to stay on the 4.0 runtime.

NOTE: Do not merge this until it is clear whether netstandard 1.0 covers this usecase or not.

Fixes #334

### DIFF
--- a/UnitsNet.Serialization.JsonNet/UnitsNet.Serialization.JsonNet.Signed.csproj
+++ b/UnitsNet.Serialization.JsonNet/UnitsNet.Serialization.JsonNet.Signed.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="UnitsNet.Serialization.JsonNet.Common.props" />
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;net35</TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;net35;net40</TargetFrameworks>
     <PackageId>UnitsNet.Serialization.JsonNet.Signed</PackageId>
     <Title>Units.NET Serialization with Json.NET (signed)</Title>
   </PropertyGroup>
--- a/UnitsNet.Serialization.JsonNet/UnitsNet.Serialization.JsonNet.csproj
+++ b/UnitsNet.Serialization.JsonNet/UnitsNet.Serialization.JsonNet.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="UnitsNet.Serialization.JsonNet.Common.props" />
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;net35</TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;net35;net40</TargetFrameworks>
   </PropertyGroup>
   <PropertyGroup>
     <!-- Workaround for building with dotnet CLI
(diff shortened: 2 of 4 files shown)


angularsen:
I'd like some input on this to make sure this is not a bad idea. Please see discussion in #334.

jnyrup:
Unfortunately .net standard 1.0 does not cover this.
This is due to the framework precedence, when NuGet has to choose between multiple available framweworks.
[Here](http://nugettoolsdev.azurewebsites.net/4.1.0/framework-precedence?framework=net45&exludedIdentifiers=&excludePortable=true) the list for net45, which is the lower full framework supporting .netstandard1.0

[This](http://nugettoolsdev.azurewebsites.net/4.1.0/get-nearest-framework?project=net45&package=netstandard1.0%0D%0Anet35) shows that with net35 and .netstandard1.0, net35 will be chosen, but adding net40 will [solve](http://nugettoolsdev.azurewebsites.net/4.1.0/get-nearest-framework?project=net45&package=netstandard1.0%0D%0Anet35%0D%0Anet40) it.

I just noticed we even have references to net20 in the [code](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/InternalHelpers/ReflectionBridgeExtensions.cs).

angularsen:
Right, that was me being lazy and copying some MIT licensed code to deal with the different reflection APIs in different .NET versions. I see a lot of the code is still just commented out instead of just keeping the code and the .NET targets we actually need - what a lazy bum :-)

eriove:
Sorry for the late reply. We were doing this for a while in our internal NuGet packages and I see no reason not to do it here. It becomes a bit annoying when there are 90 projects being compiled for an extra framework but for this case I see no downsides.

angularsen:
Thanks, I just wanted a second opinion to be sure.

angularsen:
Nuget 3.85 on its way out with net40 target, please let me know if it works for you @jnyrup

jnyrup:
While I haven't tried 3.85 yet, I tried adding net40 as target to check the difference between loading a net35 lib and a net40 lib. 
When compiling a net45 app referencing the lib, the only difference between the two libs is unification of mscorlib libs from CLRv2 to CLRv4. 
According to https://stackoverflow.com/questions/13804219/mixing-net-3-5-with-4-4-5-assemblies-in-the-same-process and https://stackoverflow.com/questions/6216599/does-using-assembly-compiled-in-older-version-of-net-framework-affects-whole-pe there should be no runtime implications as CLRv4 can execute net35 libs and side-by-side-execution multiple runtime is opt-in.

So now I'm getting curious how the apparent performance impact was measured. 
Any insights @hsddeveloper201505?

hsddeveloper201505:
Hi, first let me thank you for the quick response and change.
@jnyrup Thank you for the Information, we'll look into that.

angularsen:
I am curious too, love to see some details.
I think it's fine to keep the net40 target if that simplifies consuming the nuget in any way, but from the link above it *seems* (not sure) that there should be no performance implications referencing a net35 dll unless you explicitly enable side-by-side loading of both CLR2 and CLR4 with something like this
```
<configuration>  
  <startup >  
    <supportedRuntime version="v2.0.50727" />  
    <supportedRuntime version="v4.0" />  
  </startup>  
</configuration>  
```

https://stackoverflow.com/a/6216650/134761
https://docs.microsoft.com/en-us/dotnet/framework/deployment/in-process-side-by-side-execution

Issue #360, opened 2018-01-09
Add more Flow units


gojanpaolo:
CubicMeterPerMinute
CubicFootPerMinute
CubicYardPerSecond
CubicYardPerMinute
CubicYardPerHour
UsGallonsPerSecond
UsGallonsPerHour

gojanpaolo:
I noticed that some of the unit definitions SingularName are not actually singular. Should I change those while at it?

      "SingularName": "LitersPerMinute",
      "PluralName": "LitersPerMinute",

      "SingularName": "UsGallonsPerMinute",
      "PluralName": "UsGallonsPerMinute",

      "SingularName": "OilBarrelsPerDay",
      "PluralName": "OilBarrelsPerDay",

angularsen:
Hm, we should change this but that would be a breaking change and I don't see a good way to add new units and obsolete these as we have with other wrong namings in the past, since we would get name conflicts. I think we should add this to the wishlist issue (please locate it) and change this on the next major version bump?

angularsen:
A bit busy these days, will try to find time this weekend if no one else picks up the review before then.

PR #361, opened 2018-01-09
Add More Flow Units

gojanpaolo:
Fixes https://github.com/angularsen/UnitsNet/issues/360

### DIFF
--- a/UnitsNet/UnitDefinitions/Flow.json
+++ b/UnitsNet/UnitDefinitions/Flow.json
@@ -19,6 +19,18 @@
         }
       ]
     },
+	{
+      "SingularName": "CubicMeterPerMinute",
+      "PluralName": "CubicMetersPerMinute",
+      "FromUnitToBaseFunc": "x/60",
+      "FromBaseToUnitFunc": "x*60",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "m³/min" ]
+        }
+      ]
+    },
     {
       "SingularName": "CubicMeterPerHour",
       "PluralName": "CubicMetersPerHour",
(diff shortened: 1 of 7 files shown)


UnitsNet/UnitDefinitions/Flow.json
```diff
@@ -101,6 +173,18 @@
...
+    },
+	{
+      "SingularName": "UsGallonsPerSecond",
+      "PluralName": "UsGallonsPerSecond",
+      "FromUnitToBaseFunc": "x/264.1720523581484",
+      "FromBaseToUnitFunc": "x*264.1720523581484",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "galUS/s" ]
```
eriove: In volume we've written `gal (U.S.)`. Then `gal (U.S.)/s` would be consistent here, but that seems awkward. @angularsen what do you think?

UnitsNet/UnitDefinitions/Flow.json
```diff
@@ -113,6 +197,18 @@
...
+    },
+	{
+      "SingularName": "UsGallonsPerHour",
+      "PluralName": "UsGallonsPerHour",
+      "FromUnitToBaseFunc": "x/951019.38848933424",
+      "FromBaseToUnitFunc": "x*951019.38848933424",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "galUS/h" ]
```
eriove: Same comment regarding `gal (U.S.)` here

UnitsNet/UnitDefinitions/Flow.json
```diff
@@ -47,6 +59,66 @@
...
+    },
+    {
+      "SingularName": "CubicFootPerHour",
+      "PluralName": "CubicFeetPerHour",
+      "FromUnitToBaseFunc": "x*7.8657907199999087346816086183876e-6",
+      "FromBaseToUnitFunc": "x/7.8657907199999087346816086183876e-6",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "cf/hr" ]
```
eriove: The official abbreviation of cubic foot is ft³ and cf isn't listed on Wikipedia. I would suggest to change this to ft³/h and perhaps add this as an alternative

gojanpaolo: Will do. :)

UnitsNet/UnitDefinitions/Flow.json
gojanpaolo: We actually display US Gallons as gal (US).
Takes less space and seems to be clear enough.
Also less awkward when used in Flow gal (US)/s

We also display US Survey Feet as ft (US) ... ft (US)/s

UnitsNet/UnitDefinitions/Flow.json
```diff
@@ -47,6 +59,66 @@
...
+    },
+    {
+      "SingularName": "CubicFootPerHour",
+      "PluralName": "CubicFeetPerHour",
+      "FromUnitToBaseFunc": "x*7.8657907199999087346816086183876e-6",
+      "FromBaseToUnitFunc": "x/7.8657907199999087346816086183876e-6",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "ft³/h", "cf/hr" ]
```
angularsen: Is `cf/hr` used? Having it here means you expect to parse it.
If we keep this, should the same be added for cubic yards per hour below?

UnitsNet/UnitDefinitions/Flow.json
angularsen: It is awkward, but at least it is consistent. I think it's fine.

UnitsNet/UnitDefinitions/Flow.json
gojanpaolo: `cf/hr` was already there before this PR so I didn't remove it. I suggest we put that task of removing it in the wishlist for breaking changes.

angularsen: Ah I missed that. Good point, if you could add it to the wishlist that would be good!

gojanpaolo: Done! :)

angularsen:
Alrighty, this should be good to go then

angularsen:
Nuget 3.87 on the way out

gojanpaolo:
Thank you! :100:

Issue #358, opened 2018-01-08
Add MassFlux


gojanpaolo:
Can we add [MassFlux](https://en.wikipedia.org/wiki/Mass_flux) (Mass/Area/Time) to the library?
Let me know if we can then I'll submit a PR when I get some free time.
Thank you! :)

angularsen:
I don't see why not. I have no idea how widely used it is, but seems to be a well established quantity. 😀

gojanpaolo:
Great! Will submit PR a bit later.
We usually encounter it when multiplying Density and Velocity which becomes MassFlux.
The MassFlux can then be used to get MassFlowRate given an area or used as a factor to calculate another Density or Velocity.

PR #362, opened 2018-01-09
Add Mass Flux

gojanpaolo:
Fixes https://github.com/angularsen/UnitsNet/issues/358

Add Mass Flux unit definition
Also added operators with Mass Flux' relation to Density, Speed, Area, and Mass Flow.


UnitsNet/UnitDefinitions/MassFlux.json
```diff
@@ -0,0 +1,20 @@
...
+    {
+      "SingularName": "GramPerSecondPerSquareMeter",
+      "PluralName": "GramsPerSecondPerSquareMeter",
+      "FromUnitToBaseFunc": "x/1e3",
+      "FromBaseToUnitFunc": "x*1e3",
+      "Prefixes": [ "Kilo" ],
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "g/s-m²" ]
```
eriove: It seems like the minus sign got misplaced here. Try changing it to `g·s⁻¹·m²`

I just discovered that http://graphemica.com is a great site for these signs

UnitsNet.Tests/CustomCode/MassFluxTests.cs
```diff
@@ -0,0 +1,73 @@
+﻿//------------------------------------------------------------------------------
```
eriove: Remove the auto generated part of this header and change the year to 2018

UnitsNet/UnitDefinitions/MassFlux.json
eriove: Yes, thanks. I will cross check the numbers once more. This is a sign that that's all the reviewing I should be doing tonight :-)

eriove: Forgot that there are no numbers in the tests when there aren't any units with different bases (imperial, metric etc). I'm done with the review then

gojanpaolo: @eriove Is the naming ok?
GramPerSecondPerSquareMeter
notice two "Per". 

This is because MassFlux is defined as MassFlow per Area (MassFlowPerArea).
Then MassFlow is Mass per Time (MassPerTime).
Thus giving us MassPerTimePerArea.

Though I'm not sure if there is a convention to be consistent on other units when dealing with this.

angularsen:
Thanks, good work on adding the extra test cases.

gojanpaolo:
Thank you too!

PR #364, opened 2018-01-11
Added Altitude, AreaDensity, Capacitance, ElectricCharge, ElectricChargeDensity

baxcoders:
Added the following physical quantities:

Altitude - MetersAboveSeaLevel (masl)
AreaDensity - KilogramsPerMeterSquared (kg/m²)
Capacitance - Farad (F)
ElectricCharge - Coulomb (C)
ElectricChargeDensity - CoulombPerMeterCube (C/m³)


gojanpaolo:
Altitude seems to be redundant to Length in the context of unit conversion.

UnitsNet/UnitDefinitions/Altitude.json
```diff
@@ -0,0 +1,24 @@
+﻿{
```
eriove: As mentioned previously in the pull request I believe this is the same physical quantity as `Length` or am I missing something. I have to admit I get confused by all the usages in aviation of masl.

UnitsNet/UnitDefinitions/ElectricCurrentDensity.json
```diff
@@ -0,0 +1,20 @@
+﻿{
+  "Name": "ElectricCurrentDensity",
+  "BaseUnit": "AmperePerMeterSecondSquared",
+  "XmlDoc": "In electromagnetism, current density is the electric current per unit area of cross section.",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Current_density",
+  "Units": [
+    {
+      "SingularName": "AmperePerMeterSecondSquared",
```
eriove: Please use `SquareMeter` instead of instead of `MeterSecondSquare` to be consistent with other units

UnitsNet/UnitDefinitions/ElectricChargeDensity.json
```diff
@@ -0,0 +1,20 @@
+﻿{
+  "Name": "ElectricChargeDensity",
+  "BaseUnit": "CoulombPerMeterCube",
+  "XmlDoc": "In electromagnetism, charge density is a measure of the amount of electric charge per unit length, surface area, or volume.",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Charge_density",
+  "Units": [
+    {
+      "SingularName": "CoulombPerMeterCube",
```
eriove: Please use `CubicMeter` to be consistent with `Volume`

UnitsNet/UnitDefinitions/AreaDensity.json
```diff
@@ -0,0 +1,19 @@
+﻿{
+  "Name": "AreaDensity",
+  "BaseUnit": "KilogramPerMeterSquared",
+  "XmlDoc": "The area density of a two-dimensional object is calculated as the mass per unit area.",
+  "Units": [
+    {
+      "SingularName": "KilogramPerMeterSquared",
```
eriove: Please use `SquareMeter` instead of instead of `MeterSecondSquare` to be consistent with other units

UnitsNet/UnitDefinitions/Altitude.json
baxcoders: Altitude is not redundant to length because, even if from a dimensional point of view there is no difference, the altitude implicitly contains the information about the reference system.
Altitude is a widespread absolute measure of the height of a point.
In fact, in order to determine the altitude of a point, you need to have the reference datum of the level of sea (please see also: https://en.wikipedia.org/wiki/Altitude). 
If you use only length, you are going to miss the information about the reference, hence it won't be possible to detect an absolute height.

Moreover, the context is important. 
Who uses this measure is used to think with a reference in mind (the sea level) and then expects to work with this kind of measure unit.

UnitsNet/UnitDefinitions/ElectricCurrentDensity.json
baxcoders: AmperePerMeterSecondSquared was clearly a wrong unit (a typo probably).
However you're true about the consistency issue and we will modify as requested.

UnitsNet/UnitDefinitions/AreaDensity.json
baxcoders: Good point, we will modify as requested.

UnitsNet/UnitDefinitions/ElectricChargeDensity.json
baxcoders: Good point, we will modify as requested.

UnitsNet/UnitDefinitions/Altitude.json
eriove: @angularsen Some input from you would be good on this

eriove: In my opinion this would be out of scope of UnitsNet. Wouldn't it would make more sense to add another property with the reference datum level and save these 2 `Length` inside a new struct or class in you project?

baxcoders: We chose to use this library because it provides everything we need out of the box. We don't have to create any kind of wrapper to adapt, thus that solution of yours doesn't really fit our needs.

The concept should be pretty similar to TemperatureDelta. We're always talking about temperature, but the delta means you're setting a reference system and starting your measurements from a specific value.

More, I must admit _Altitude_, in the terms we presented it to you, is very highly used in renewable technology context, much more than the other physical quantities we added, actually.

angularsen: I have only skimmed the wiki and are not very familiar with the domain of altitude and units.

1. Are there other reference datums widely used, something like MASL 1970 or MASL 1990? I have not researched this, but the argument for not using Length only holds water if there are actually multiple reference datums to consider that would be represented as multiple altitude units. If they exist and are widely used, then they should be part of this PR.
2. Are there operator overloads for other quantities that would be useful? Example: `Length / Duration = Speed`

We generally want to avoid single-unit quantities in the library, because I think the value of these are limited and could just as easily be defined per-application. I personally think UnitsNet shines when you have multiple units to convert between, operator overloads that construct other quantities.

baxcoders: I understand your point.
I searched in order to find alternative units definitions to convert to/from masl and wikipedia said: 
_"Feet above sea level is the most common analogue for metres above sea level in the American measurement system."_
Said that, we never had occurrences of that unit but it could be added as well. 

If you keep struggling with the idea, what do you think if we add _masl_ among the abbreviations of Length - Meter? It's a bit odd, but could it work?

angularsen: I definitely do not want to add it to Length, that would be very strange indeed.

We could add the feet variant, but again, this feels like Length to me - just that this particular length/distance has a semantic meaning (distance to mean sea level). I feel this definition belongs in an application rather than in a shared library of units. It doesn't seem to benefit users much except having a separate type for it to help encode the meaning, and you can achieve that yourself with a wrapper type - or use C# type aliasing (`using Altitude = UnitsNet.Length;`) although that doesn't help you distinguish it from Length types. 

Seeing that there are no other scales or "datums" used, I argue that altitude should be kept out of the library to avoid adding too many of these single-unit quantities.
I propose you remove Altitude from this PR so we can merge in the rest of it, and if you still feel strongly about Altitude or have new arguments for adding it we can continue that conversation in another issue. However, I think we have covered most arguments for and against by now.

baxcoders: Ok, we'll proceed as requested.

angularsen:
You will have to fix the merge conflicts, but other than that this should be good to merge.

angularsen:
Thanks! That was the last of the bunch :-) Awesome job adding all these. Nuget on the way out.

PR #367, opened 2018-01-11
Added MagneticFlux, Magnetization, Permeability, Permettivity, Precipitation.

baxcoders:
MagneticFlux - Weber (Wb)
Magnetization - AmperePerMeter (A/m)
Permeability - HenryPerMeter (H/m)
Permettivity - FaradPerMeter (F/m)
Precipitation - MillimeterPerMinute (mm/min)


gojanpaolo:
What will be the difference between Precipitation and Speed?

baxcoders:
Hello,

from a dimensional point of view, there is no difference.
But from the point of view of the context, who is measuring how much rain is falling is used to measure the level of a gauge that is filling with the rain itself.

However, in this case we can avoid to specify a different physical quantity and use (or add) this measure unit under "Speed".

UnitsNet/UnitDefinitions/MagneticFlux.json
```diff
@@ -0,0 +1,20 @@
...
+  "BaseUnit": "Weber",
+  "XmlDoc": "In physics, specifically electromagnetism, the magnetic flux through a surface is the surface integral of the normal component of the magnetic field B passing through that surface.",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Magnetic_flux",
+  "Units": [
+    {
+      "SingularName": "Weber",
+      "PluralName": "Webers",
+      "FromUnitToBaseFunc": "x",
+      "FromBaseToUnitFunc": "x",
+      "Localization": [
```
angularsen: Any prefix units to add? 
Add any of these units?
https://www.translatorscafe.com/unit-converter/en/magnetic-flux/1-10/weber-tesla%20meter%C2%B2/

UnitsNet/UnitDefinitions/Magnetization.json
```diff
@@ -0,0 +1,20 @@
+﻿{
+  "Name": "Magnetization",
+  "BaseUnit": "AmperePerMeter",
+  "XmlDoc": "In classical electromagnetism, magnetization is the vector field that expresses the density of permanent or induced magnetic dipole moments in a magnetic material.",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Magnetization",
+  "Units": [
```
angularsen: Any prefix units?
Add any of these units?
http://www.hkd.co.jp/english/unit_conv/

UnitsNet/UnitDefinitions/Magnetization.json
```diff
@@ -0,0 +1,20 @@
+﻿{
+  "Name": "Magnetization",
```
angularsen: Is Magnetization the correct naming?
I see magnetic field and flux density being referenced [here](http://www.hkd.co.jp/english/unit_conv/), but I don't know anything about these units.

UnitsNet/UnitDefinitions/Permeability.json
```diff
@@ -0,0 +1,20 @@
...
+  "BaseUnit": "HenryPerMeter",
+  "XmlDoc": "In electromagnetism, permeability is the measure of the ability of a material to support the formation of a magnetic field within itself.",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Permeability_(electromagnetism)",
+  "Units": [
+    {
+      "SingularName": "HenryPerMeter",
+      "PluralName": "HenriesPerMeter",
+      "FromUnitToBaseFunc": "x",
+      "FromBaseToUnitFunc": "x",
+      "Localization": [
```
angularsen: Any prefix units?
Any of these units?
https://www.calculand.com/unit-converter/?gruppe=Permeability&einheit=Henry+per+metre+%5BH%2Fm%5D

UnitsNet/UnitDefinitions/Permittivity.json
```diff
@@ -0,0 +1,20 @@
...
+  "BaseUnit": "FaradPerMeter",
+  "XmlDoc": "In electromagnetism, permittivity is the measure of resistance that is encountered when forming an electric field in a particular medium.",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Permittivity",
+  "Units": [
+    {
+      "SingularName": "FaradPerMeter",
+      "PluralName": "FaradsPerMeter",
+      "FromUnitToBaseFunc": "x",
+      "FromBaseToUnitFunc": "x",
+      "Localization": [
```
angularsen: Any prefix units?
Any of these units?
https://www.calculand.com/unit-converter/?gruppe=Permittivity&einheit=Farad+per+kilometre+%5BF%2Fkm%5D

UnitsNet/UnitDefinitions/Magnetization.json
baxcoders: https://en.wikipedia.org/wiki/Magnetization

It's an important parameter for some devices found in electric substations.

UnitsNet/UnitDefinitions/MagneticFlux.json
baxcoders: We listed in this PR only the units we're really using. We know that probably their relative physical quantities could be expanded with more definitions and prefixes but they're out of scope at the moment.
We are committed to adding them in future if their usage will become necessary.

angularsen: Fair enough, other units can be added later when needed.

angularsen:
Fixed the merge conflict in github gui.

PR #369, opened 2018-01-11
Added SolidAngle and ThermalConductivity.

baxcoders:
SolidAngle - Steradian (sr)
ThermalConductivity - WattPerMeterKelvin (W/m·K)


UnitsNet/UnitDefinitions/SolidAngle.json
```diff
@@ -0,0 +1,20 @@
...
+  "Name": "SolidAngle",
+  "BaseUnit": "Steradian",
+  "XmlDoc": "In geometry, a solid angle is the two-dimensional angle in three-dimensional space that an object subtends at a point.",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Solid_angle",
+  "Units": [
+    {
+      "SingularName": "Steradian",
+      "PluralName": "Steradians",
+      "FromUnitToBaseFunc": "x",
+      "FromBaseToUnitFunc": "x",
```
angularsen: Consider adding these units?
http://www.calculator.org/property.aspx?name=solid+angle

UnitsNet/UnitDefinitions/ThermalConductivity.json
```diff
@@ -0,0 +1,20 @@
+﻿{
+  "Name": "ThermalConductivity",
+  "BaseUnit": "WattPerMeterKelvin",
+  "XmlDoc": "Thermal conductivity is the property of a material to conduct heat.",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Thermal_Conductivity",
+  "Units": [
```
angularsen: Any prefix units to add?
Add these units?
http://www.unitconversion.org/unit_converter/thermal-conductivity.html

UnitsNet/UnitDefinitions/SolidAngle.json
baxcoders: We listed in this PR only the units we're really using. We know that probably their relative physical quantities could be expanded with more definitions and prefixes but they're out of scope at the moment.
We are committed to adding them in future if their usage will become necessary.

angularsen: That is fair. They can be added later if someone finds them useful.

Issue #363, opened 2018-01-10
Rename Flow to VolumeFlow


angularsen:
To disambiguate it from MassFlow and more consistent naming.

- [ ] Mark all `Flow` units as obsolete via .json file (example: [TemperatureDelta.KelvinDelta](https://github.com/angularsen/UnitsNet/blob/431527d7ae761f789c101ffda73700b00772ff3b/UnitsNet/UnitDefinitions/TemperatureDelta.json#L105))
- [ ] Add `VolumeFlow` .json based on Flow (copied and renamed) 
- [ ] Run `generate-code.bat`

All obsoleted code will be addressed on next major version bump, as listed in #180 .

gojanpaolo:
I'll start doing this. :)

angularsen:
You can, but before this can be reviewed it should rebased on top of origin/master so that you don't include flow unit changes in the PR.

gojanpaolo:
How do I do that? Sorry, kinda new to GitHub.

angularsen:
In git command line, something like this:
```
$ (someotherbranch): git checkout -b mynewbranch origin/master
$ (mynewbranch): ...do stuff...
$ (mynewbranch): git commit
$ (mynewbranch): git push
```

Then create PR from mynewbranch on your fork.

gojanpaolo:
I just did something like this...
$ (AddMoreFlowUnits): git checkout -b RenameFlowToVolumeFlow origin/master

but it didn't get where I left off from AddMoreFlowUnits

gojanpaolo:
So I did the tasks for this issue but branching from PR https://github.com/angularsen/UnitsNet/pull/361 

I guess I'll just wait for that PR to be accepted before I submit PR for [RenameFlowToVolumeFlow branch](https://github.com/gojanpaolo/UnitsNet/tree/RenameFlowToVolumeFlow) ?

angularsen:
Ah I see. 
Easiest route: Wait until the first branch is merged into master, then branch off master to do the rename.
Harder route: Branch off AddMoreFlowUnits branch now, do the rename, push the branch and create PR with a note saying this should not be merged until the first PR is merged. After it is merged, rebase on top of orgin/master to effectively get to the same state as the easy route, then force push that to update the PR and notify it is ready for review.

Frankly, the latter option does not gain much and adds a lot more friction so I would go with the easy route.

gojanpaolo:
[...]

PR #375, opened 2018-01-14
Rename Flow To Volume Flow

gojanpaolo:
Fixes https://github.com/angularsen/UnitsNet/issues/363


angularsen:
Looks good, I assume you did copied things over and renamed it so I didn't verify the conversion test values.

gojanpaolo:
Yes I just copy/pasted from FlowTests. Thought about rearranging it but didn't to avoid extra manual checking. Thank you for closing!

ferittuncer:
@gojanpaolo I know its kinda late, I had to come up with this earlier, but, shouldn't it be volumetric flow instead of volume flow?

We are developing a tool which can solve pipe networks and we always use the term volumetric flow and I never heard volume flow. My background is CS, so I don't have a deep knowledge about these thermodynamic context, so I may be wrong. Anyway, just wanted to ask if you considered this alternative.

gojanpaolo:
I'm fine with either also. We can add VolumetricFlow and make VolumeFlow obsolete.

ferittuncer:
If both terms are okay let's not create another obsolote and stick with VolumeFlow then.

angularsen:
I second that!

Issue #377, opened 2018-01-14
Add FromDiameter to Area.extra


gojanpaolo:
Can we add something like:
```
public static Area FromDiameter(Length diameter)
{
	return Math.PI * diameter * diameter / 4;
}
```

I was hesitant to suggest this before but I saw similar functions for [PowerRatio](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/CustomCode/Quantities/PowerRatio.extra.cs) which uses some kind of computation to convert from one unit to another.

angularsen:
Sure, we have several of these from before so I'm fine with this.

angularsen:
On second thought, maybe it should be named `FromCircleDiameter()` to be very clear what is about? At least I find that slightly more informative.

angularsen:
...aaand, consider adding the radius variant while at it. :-)

gojanpaolo:
You bet! :)

PR #378, opened 2018-01-14
Added Area FromCircle methods

gojanpaolo:
Fixes https://github.com/angularsen/UnitsNet/issues/377

Concerns:
How to handle negative radius/diameter
1. Should we throw exception?
2. Should we just ignore (or absolute value)?
   - Current implementation is ignored which implicitly gets absolute value since it squares the value

Need to add tests either case.

Note:
[High precision area of circle online calculator](http://keisan.casio.com/exec/system/1228460515)


angularsen:
Hm, since the default behavior is that of absolute value, I see no reason to handle this specifically so I think we should simply allow it.

angularsen:
Ok maybe I merged a bit early if you intended to add test case for the negative values, can add that separately I guess :-) Sorry!

gojanpaolo:
Ahh lol. That's fine. We can add the test case next time. Implementation will not change anyway. :)
Thank you!

Issue #376, opened 2018-01-14
Add Operators for the relationship Volume = VolumeFlow/TimeSpan (or Duration)


gojanpaolo:
e.g.
```
public partial struct Volume
{
        //...
        public static VolumeFlow operator /(Volume volume, TimeSpan timeSpan)
        {
               return VolumeFlow.FromCubicMetersPerSecond(volume.CubicMeters / timeSpan.Seconds);
        }
}
```

gojanpaolo:
Will do this after https://github.com/angularsen/UnitsNet/pull/375 :)

PR #379, opened 2018-01-14
Add Operators for the relationship Volume = VolumeFlow/TimeSpan

gojanpaolo:
Fixes https://github.com/angularsen/UnitsNet/issues/376


angularsen:
Good job, looks flawless.

Issue #371, opened 2018-01-13
Design proposal: Base type (WIP)


angularsen:
NOTE: This is work in progress and early thoughts, but input is most welcome.

An interface or abstract type implemented by all quantity types, such as `Length` and `Mass`.

### Motivation
- **Represent generic quantities in code**, where we don't need to know if it is `Length` or `Mass` but want to communicate it is a quantity - we need to use `object` today
- **Get textual representation of quantity**  with `someQuantity.ToString()` => `"5 kg"`, this works today except we need to treat it as `object` and we don't have access to overloads for things like culture and significant digits after radix
- **Enumerate units for this quantity** as this currently requires reflection, a usecase is to dynamically show GUI for converting a passed in quantity (1 meter) to any of the other length units
- **Present generic quantities visually in a 2D graph** by showing unit selector for the quantities to change unit and getting the values for each quantity in the selected unit, without hardcoding support for specific quantities

### Discussion points
- Should we change quantities from struct to class? As discussed in the double/decimal issue, with class we might benefit from supporting both double and decimal internal representations without doubling the code. With an interface, it will quickly be boxed anyway unless using [generic methods instead](http://1024strongoxen.blogspot.no/2010/02/avoid-boxing-of-structs-implementing.html)
- Should the new members be added to quantity base type or be added to the `UnitSystem` type. One thing to consider is what culture to use for things like abbreviations and number formatting, where `UnitSystem` instances are created explicitly with a given culture (default uses CurrentUICulture in Windows).

### Proposal
- Add interface `IQuantity` since structs can't inherit base types, but users should be wary that casting to interface means boxing the value with those performance implications
- **ToString() should reflect the original value and unit** instead of using the base unit, which may be very different than the units the user is working with (addressed in #389)
- **Add getter property for `QuantityType`** since quantities already expose this value
- **Add getter properties for original value and unit** to present or filter on these values separately
- **Add getter property for unit abbreviations** as current you must figure out to use `UnitSystem.Default.GetAllAbbreviations()`
- **Add getter properties for singular and plural names of quantity** for both `"ForcePerLength"` and `"Force per length"` forms, to make it easier to textually describe the quantity type. Currently you use reflection to obtain `QuantityType` enum value and call `ToString()` on it to get `"ForcePerLength"`.
- **Add methods to obtain singular and plural names of a given unit** to get `"Newton
- **Add getter property for the quantity description in JSON** so this can be displayed next to the quantity selector GUI
[...]

Issue #354, opened 2018-01-04
Getting abbreviation list of a quantity


YektaMirkan:
I am trying to bind abbreviation list of quatities to `combobox` `itemsource` property. But I am not able to do it. Because I couldn't find `abbreviations` property . In c# code I solved this problem with using `LINQ`. But now I need abbreviation list in xaml. So I can't use `LINQ` to get abbreviations from units. There is a way to achive this problem. Using converter class to convert units to abbreviations. But I need a base class otherwise I need to create a converter class for each unit. So these are the only ways that I could find. 
What do you think about base class or abbreviation list?
Or do you have any other solution to get abbreviation list of quatities ?

angularsen:
I don't fully understand your example, but here are some pointers that may help you out:

Example converter app:
https://github.com/angularsen/UnitsNet#example-creating-a-unit-converter-app

Pull request on XAML binding to units using reflection:
https://github.com/angularsen/UnitsNet/pull/353

If you are able to achieve what you want with C#, then you should be able to do the same with XAML by using a view model property that calculates the combobox list content in C# and binding the XAML to it. I personally prefer this approach over complicated/specialized value converters.

Does this help?

YektaMirkan:
I am able to get abbreviation list in xaml but I need to create a converter class for each unit. Because there isn't any base class for units. `Temperature.Units.ToList().Skip(1).Select(Temperature.GetAbbreviation).ToList()` 
I have solved the problem in this way. But I want to create a common converter class which can convert units to abbreviations without knowing unit type. I think this problem can only be solved by using base class otherwise I need to use reflection and it costs me a lot.

angularsen:
Closing this in favor of #371. Help move that forward instead.

Issue #308, opened 2017-11-07
Why don't we have a base class for quantities?


ferittuncer:
Wouldn't it be beneficial to move common things in quantity classes to a quantity base class?

Sometimes we don't know which quantity we are dealing with, but we want to print them, or get abbreviations etc. which is common in all quantities. This use case is important especially in GUI development. 

What are your thoughts?

angularsen:
I agree, I've just been waiting for the right real-world usecase and the right person with the motivation to get it done - solving a real need.
Could you please detail an example/pseudo code where having a base class or interface would simplify things?

ferittuncer:
I am planning to give you a proposal in detail when I have time, but for instance I remember we had to do reflection hack because of lacking a base `Unit` class.

For example, we have a list view control which lists (prints their name, or abbreviation) units of a given quantity. However, as we don't know which quantity we are dealing with, we want a List like this: `List<Unit>` which is not possible.

So instead, we had to do `List<object>` and lots of reflection and casting.

Similarly we could keep a `List` of quantities like: `List<Quantity>` if we had `Quantity` class.

To be able to demonstrate my arguments easily I'm working on a converter app. I'll share it soon.

eriove:
Since the quantities are structs I've been worried by the boxing overhead of using an interface, but if the alternative is using an `object` and reflection the overhead of boxing will still be an improvement to what we have today and there wouldn't be a difference for existing code that uses the structs directly.

angularsen:
@ferittuncer Looking forward to the pseudo code, as generally in these cases I have found that it's unnecessary to work "blindly" with the units. You usually know what quantity the units are for since you typically want to be able to convert to other units at some point. There is also a `QuantityType` enum type you can use to associate the values with a quantity and there is `UnitConverter` that can be used to dynamically convert between units based on unit names/abbreviations and a quantity name, but I'll await an example to comment further.

angularsen:
@eriove Right, and I guess performance is perhaps not the primary focus if you have already resorted to reflection to achieve this :-)

angularsen:
Closing this in favor of #371

PR #380, opened 2018-01-14
Add unit converter WPF sample app

### LINKED ISSUE #274: Request: Create demo unit converter app
I think it would be good to showcase a simple unit converter app to give a practical sample to link to, in particular to demonstrate #181.

* Create Winforms, WPF or UWP app in spirit of screenshot
* Update TODO in README about the example converter app

image

angularsen:
To illustrate how to enumerate quantities, units and to convert between them without hard coding any quantities.

image

The app lists all quantities (Length, Mass etc), and when you select one, it then lists all units for that quantity (if Length was chosen; meters, centimeters etc). Then it converts the numeric value from one selected unit to another.

This sample also highlights a limitation in the library that requires reflection in order to enumerate units for a selected quantity and to get the abbreviation for a unit, since the library does not provide a generic means to retrieve these without referencing types like `Length` and `LengthUnit` directly. There is definitely room for improvement in the library here, as previously discussed in #371 #354 #353 #308.

Fixes #274


angularsen:
@eriove @ferittuncer @JKSnd for input. Also, this sample app pretty much sums up the current limitation of UnitsNet that currently requires reflection to do conversions without hard coded quantity types.

@YektaMirkanhi Since you asked about #354 Getting abbreviation list of a quantity 
@dayewah Since you made a sample app and have been doing this dance already, this may be of interest to you.

gojanpaolo:
Need null checking here?
```
        public string FromHeader => $"Value [{SelectedFromUnit.Abbreviation}]";
        public string ToHeader => $"Result [{SelectedToUnit.Abbreviation}]";
```

ferittuncer:
I just started to examine this sample. I got an exception just after a clicking to a quantity.

`NullPointerException` at line 85 in `MainWindowVm.cs`.

angularsen:
Thanks for testing it, I've added the null checks. I was not able to reproduce this behavior so likely a racing condition.

ferittuncer:
Working flawlessly now. 

I wonder why you didn't make result textbox editable like Google's converter.

angularsen:
Ah, that was just me being lazy and trying to the the least amount of work :-)
I guess I just foresaw some issue with infinite cycles of one field change updating the other, which again updates the first one and so on. It's obviously solvable, I just avoided it the lazy bum that I am 😆

I'm merging in this since the bug is fixed. If anyone wants to improve this app that is awesome.

Issue #350, opened 2017-12-29
Scientific Notation - Precision


ferittuncer:
I changed a test value from `14173.2283465` to `1.41732283465E4`, in `SpeedTest.cs`, and it made tests fail.

When I debugged I saw that there is a small difference in actual and expected. So scientific notation seems less precise.

Do you know anything about this? If this is the case, we should not use scientific notation in test values I think (which was what I'm doing so far in my commits, as it looks nicer).

angularsen:
I guess you are at the very edge of passing the tests with the default max error of 1e-5 for that particular unit vs base unit. I guess I'm not surprised there is a slight difference in precision with and without scientific notation, but I suspect it's more of a random deviation rather than "scientific notation is always less precise" - but that is just speculation on my part.

As for test values, they don't need to be in scientific notation really. The key thing is for them to verify the conversion to some known precision, so either keep the value as-is or override the max error for that unit to 1e-4.

ferittuncer:
I wouldn't prefer increasing tolerance value. I think we should strive for lowest tolerance. So let's keep test values as is, instead of writing in scientific notation.

Also, I wonder, is there a specific reason that we use absolute tolerance instead of relative tolerance in tests?

angularsen:
It's a good question that I don't have a good answer to. We have discussed relative errors in the past, also when discussing the new `Equals()` method that takes a tolerance argument, but the problem is that we don't have any context on what unit the consumer is working with so we can't have provide any good default values that work for all units. 

As for tests, if I recall correctly the tolerance specifies the max error we allow for a round-trip conversion from unit A to base unit B and back to unit A, and the error is measured for the value of unit A. This gives an idea of what precision to expect for all our units, since some units are very different (smaller/larger) than the base unit and thus a lot of precision is lost in the conversion.

Does this answer your question or could you perhaps sketch out some pseudo code for how we could better write these tests using relative values?

gojanpaolo:
@ferittuncer I tried changing the SpeedTests to scientific notation but it didn't fail. I think the test you're talking about is InchesPerHourInOneMeterPerSecond which is 141732.283465 = 1.41732283465E5
Should I create PR for [this](https://github.com/gojanpaolo/UnitsNet/tree/SpeedTestsToScientificNotation)?

ferittuncer:
Well, if you create, we can discuss it. I realized that scientific notation is sometimes less precise, so it sometimes break tests. I don't know why it doesn't fail when you try.

[...]

PR #381, opened 2018-01-16
Speed tests to scientific notation

gojanpaolo:
See https://github.com/angularsen/UnitsNet/issues/350


ferittuncer:
Looks like #350 was my typo, E4 instead of E5, as this doesn't fail. I'll double check everything.

gojanpaolo:
That's what I think so too. I also can't find any reference that talks about precision of scientific notation in .NET

PR #383, opened 2018-01-16
Added physical quantity: BitRate

baxcoders:
Base unit: BitPerSecond (bit/s, bps) + Prefixes: "Kilo", "Mega", "Giga", "Tera", "Peta", "Exa", "Kibi", "Mebi", "Gibi", "Tebi", "Pebi", "Exbi"
Additional unit: BytePerSecond (B/s) + Prefixes: "Kilo", "Mega", "Giga", "Tera", "Peta", "Exa", "Kibi", "Mebi", "Gibi", "Tebi", "Pebi", "Exbi"


UnitsNet.Tests/CustomCode/BitRateTests.cs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        protected override double BitsPerSecondInOneBitPerSecond => 1d;
+
+        protected override double BytesPerSecondInOneBitPerSecond => 0.125d;
+
+        protected override double ExabitsPerSecondInOneBitPerSecond => 1e-18d;
+
+        protected override double ExabytesPerSecondInOneBitPerSecond => 0.125d * 1e-18d;
+
+        protected override double ExbibitsPerSecondInOneBitPerSecond => 1d / Math.Pow(1024, 6);
```
angularsen: Test cases values should not be computations but rather some constant value looked up in a reference source as a double check that we got it right. The most typical mistake is to get the function inverted by mistake both in the implementation and in the test case (by copy & paste), so using a second source of truth is a good way to mitigate that. For infinite fractions, 7 significant digits is usually plenty. We also prefer the scientific `e` notation for these values.

baxcoders: Hello, thank you for your review.

For that physical quantity we took a page from "InformationTests.cs" file where methods are expressed the same way.

` protected override double ExabytesInOneBit => 0.125d*1e-18d;`
` protected override double ExbibitsInOneBit => 1d/Math.Pow(1024, 6);`

However, we got your point and we will proceed as requested.
Those are the modifications that will be made:

`ExabytesPerSecondInOneBitPerSecond => 0.125d * 1e-18d;`
changed to:
`ExabytesPerSecondInOneBitPerSecond => 1.25E-19;`

and 

`protected override double ExbibitsPerSecondInOneBitPerSecond => 1d / Math.Pow(1024, 6);`
changed to:
`protected override double ExbibitsPerSecondInOneBitPerSecond => 8.67361738E-19;`

Are the formats correct?

Do we change also all the others units affected by the same issue?

+ ExbibitsPerSecondInOneBitPerSecond => 1d / Math.Pow(1024, 6);
+ ExbibytesPerSecondInOneBitPerSecond => 8d / Math.Pow(1024, 6);
+ GibibitsPerSecondInOneBitPerSecond => 1d / Math.Pow(1024, 3);
+ GibibytesPerSecondInOneBitPerSecond => 1d / 8 / Math.Pow(1024, 3);
+ GigabytesPerSecondInOneBitPerSecond => 0.125d * 1e-9d;
+ KibibitsPerSecondInOneBitPerSecond => 1d / 1024d;
+ KibibytesPerSecondInOneBitPerSecond => 1d / 8 / 1024d;
+ MebibitsPerSecondInOneBitPerSecond => 1d / Math.Pow(1024, 2);
+ MebibytesPerSecondInOneBitPerSecond => 1d / 8 / Math.Pow(1024, 2);
+ MegabytesPerSecondInOneBitPerSecond => 0.125d * 1e-6d;
+ PebibitsPerSecondInOneBitPerSecond => 1d / Math.Pow(1024, 5);
+ PebibytesPerSecondInOneBitPerSecond => 1d / 8 / Math.Pow(1024, 5);
+ PetabytesPerSecondInOneBitPerSecond => 0.125d * 1e-15d;
+ TebibitsPerSecondInOneBitPerSecond => 1d / Math.Pow(1024, 4);
+ TebibytesPerSecondInOneBitPerSecond => 1d / 8 / Math.Pow(1024, 4);
+ TerabytesPerSecondInOneBitPerSecond => 0.125d * 1e-12d;

angularsen: Yes that looks good to me, and yes please also address all other units.

We probably have some discrepancies in the code base, but have started building a contribution guideline #328  that will try to keep things more consistent moving forward.

baxcoders: Ok, done.

We noticed a possible mistake in ExbibytesInOneBit of InformationTests.cs (and then in ExbibytesPerSecondInOneBitPerSecond of BitRateTests.cs which we've already fixed).
Actually is:

_protected override double ExbibytesInOneBit => 8d/Math.Pow(1024, 6);_

but it should be:

protected override double ExbibytesInOneBit => 1d/8/Math.Pow(1024, 6);_

Do you agree?

angularsen: Good catch, this looks a mistake for sure. These values are so small (`1e-18`) so they are dwarfed by the way-too-large default tolerance of `1e-5`. We are discussing #350  using relative errors instead, and this is probably a good use case for exactly that.

I pushed a fix in beadd3d. I guess ideally the tolerance values should have been updated to be more strict too, but then all values should be updated and not just the exbibytes one, so I didn't bother. Also we are probably rewriting to relative test values soon anyway and will have to revisit the values then.

angularsen:
Thanks, I'll have to recheck all the new values then this should be good to go.

PR #393, opened 2018-01-30
Add symbols nuget packages with source

angularsen:
This produces separate nuget packages that includes symbols (.pdb) and source (/src folder of nuget).
It helps debugging these nugets from the consuming side and should allow you to F11 into its code on-the-fly.

I'm not sure why this wasn't added in the first place, I guess I just forgot about it.


angularsen:
Anyone care to review this before I go ahead and merge it?

angularsen:
Guess not, going in hot! 🔥 😆

PR #398, opened 2018-02-03
Add Area SquareUsSurveyFeet

gojanpaolo:


UnitsNet/UnitDefinitions/Area.json
```diff
@@ -146,6 +146,18 @@
...
+    },
+	{
+      "SingularName": "SquareUsSurveyFoot",
+      "PluralName": "SquareUsSurveyFeet",
+      "FromUnitToBaseFunc": "x*0.09290341161",
+      "FromBaseToUnitFunc": "x/0.09290341161",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "ftUS²" ]
```
angularsen: We have a convention of this format `ft (US)²` for country suffixes. Would that work?

https://github.com/angularsen/UnitsNet/issues/328#issuecomment-357041749

angularsen: Or perhaps `ft² (US)` if that is better..?

gojanpaolo: Ahh. I forgot about the convention again, my bad. I referenced the length which uses ftUS.

gojanpaolo: `ft² (US)` looks better. Should the name be changed to `SquareFeettUsSurvey` ?

angularsen: I think `SquareFeetUsSurvey` best matches how I've read it when googling, but our convention so far has been `<variation>Unit` such as `MetricTeaspoon` vs `UsTeaspoon`. I think it's best to follow our conventions unless it goes against a very widely accepted naming scheme in the domain.
So I vote to keep it `UsSurveySquareFeet`, since in `Length` it is also named `UsSurveyFeet`.

PR #401, opened 2018-02-08
Add acceleration units

gojanpaolo:
KnotsPerSecond
KnotsPerMinute
KnotsPerHour


angularsen:
Never heard of these before, but at least knots per second and per minute showed up on google :-)

gojanpaolo:
Yes these also weird for me personally, but this is what they use in production reports. xD

gojanpaolo:
The default unit of the parameter is actually Knots per Hour lol.

PR #403, opened 2018-02-12
Added Power gradient

baxcoders:


angularsen:
I'm not sure about this one. One one side `W/s` sounds kind of reasonable, but googling power gradient and watts per second leaves me very little results. Is this really widely used? What are some usecases?

baxcoders:
Active power gradient and reactive power gradient are both representing derivative (spatial) functions of their respective powers.
These measures are rare and were used in photovoltaic cells in order to report how power changes in time and along the single submodules.

However, we're keeping them just for legacy purposes (we dismissed recently those devices using those specific signals).
If you're not fine with them we can do without.

angularsen:
Ok, then I propose we don't add these. We are already having issues with the binary size and have to say no sometimes for units that simply are not widely used.

PR #408, opened 2018-02-12
Added TonPerHour (t/h) for MassFlow

baxcoders:


UnitsNet/UnitDefinitions/MassFlow.json
```diff
@@ -32,6 +32,18 @@
         }
       ]
     },
+    {
+      "SingularName": "TonPerHour",
```
angularsen: Rename this to `TonnePerHour` to match the same unit in `Mass`. I believe `ton` is the American naming.

baxcoders: Fixed, thank you.

PR #418, opened 2018-03-19
Features/pressure mega newton per square meter

dschuermans:
Added Mega prefix for NewtonPerSquareMeter in the pressure unit.
Not sure why RotationalStiffnessUnit also has changes, happend after I ran the bat file...

### DIFF
--- a/UnitsNet/UnitDefinitions/Pressure.json
+++ b/UnitsNet/UnitDefinitions/Pressure.json
@@ -107,17 +107,17 @@
       "PluralName": "NewtonsPerSquareMeter",
       "FromUnitToBaseFunc": "x",
       "FromBaseToUnitFunc": "x",
-      "Prefixes": [ "Kilo" ],
+      "Prefixes": [ "Kilo", "Mega" ],
       "Localization": [
         {
           "Culture": "en-US",
           "Abbreviations": [ "N/m²" ],
-          "AbbreviationsWithPrefixes": [ "kN/m²" ]
+          "AbbreviationsWithPrefixes": [ "kN/m²", "MN/m²" ]
         },
         {
           "Culture": "ru-RU",
           "Abbreviations": [ "Н/м²" ],
-          "AbbreviationsWithPrefixes": [ "кН/м²" ]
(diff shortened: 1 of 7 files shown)


angularsen:
Hey, you need to either rebase on top of `upstream/master` in order to lose all the old commits from this PR or manually cherry pick the commits you want.

Try this:
```
git checkout features/PressureMegaNewtonPerSquareMeter
git rebase --interactive upstream/master
```

Then in the editor, remove all commit lines you don't want to include. Type `:x` to save and exit if you are not familiar with vim :-)

As for getting extra changes when running `generate-code.bat`, this can happen if `master` has recently merged several pull requests and so needs to regenerate to get all the latest units. Normally this is safe to ignore, but I've pushed a new commit a66dcdf0 now that fixes that.

dschuermans:
@angularsen Thanks, i'll do it now and update the PR :)

UnitsNet/UnitDefinitions/Pressure.json
```diff
@@ -107,17 +107,17 @@
...
       "FromUnitToBaseFunc": "x",
       "FromBaseToUnitFunc": "x",
-      "Prefixes": [ "Kilo" ],
+      "Prefixes": [ "Kilo", "Mega" ],
       "Localization": [
         {
           "Culture": "en-US",
           "Abbreviations": [ "N/m²" ],
-          "AbbreviationsWithPrefixes": [ "kN/m²" ]
+          "AbbreviationsWithPrefixes": [ "kN/m²", "MN/m²" ]
```
angularsen: `AbbreviationsWithPrefixes` is only needed if prefixing `k` and `M` does not work, so for `en-US` culture it can be removed. I'll remove this in master after merging. Russian still needs to be kept.

angularsen: Fixed in https://github.com/angularsen/UnitsNet/pull/421

angularsen:
Nuget is out https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F3.94.0

Issue #372, opened 2018-01-13
Proposal: Reducing size of library (WIP)


angularsen:
NOTE: This is work in progress and early thoughts, but input is most welcome.

### Motivation
`UnitsNet` nuget is nearing the 1MB mark and this is probably a lot bigger than most consumers would expect, since most users only work with a small set of quantities and units.

### Proposal 1 - Core nuget + quantity nugets + meta nugets
This tries to achieve a kind of extension model, where new quantities are added via nugets - potentially 3rd party quantities.

- Move all shared code into `UnitsNet.Core` (`UnitSystem`, `UnitConverter`, `QuantityParser`, `UnitFormatter`, `QuantityValue` etc..)
- Repurpose our current `UnitsNet` nuget to a meta package that brings in the most common units (think 80/20, this is what we want most people to use)
- Add nugets for all quantities `UnitsNet.Quantities.Length`, `UnitsNet.Quantities.Mass` etc. (around 80 of these now), which has the `Length` struct type (`Length.g.cs` and `Length.extra.cs` files) and any related extension methods (`NumberToLengthExtensions`), with a dependency on `UnitsNet.Core`
- Add metapackage nugets to bring in a set of quantities for an engineering/technical domain, such as `UnitsNet.Meta.ElectricalEngineering`, `UnitsNet.Meta.AudioEngineering`, `UnitsNet.Meta.PetroleumEngineering`, `UnitsNet.Meta.ChemicalEngineering`

Concerns:
- Discoverability is hurt by moving code out into separate packages, so you need to browse packages to see what is possible unless you add a metapackage that brings in all or the most common quantities
- A lot of nugets (80 for quantities plus a few meta packages to bring in groups of quantities)
- Extension model with static types is tricky. How can we provide operator overloads such as `Force = Pressure * Area`, when the user may not have added all three nuget packages? Should there be a nuget package for the Force+Pressure+Area combination? Can we move operator overloads as well as interfaces and enums for all units to the core nuget, while keeping the quantity implementations and the bulk of the quantity code in separate nugets? Need to think of something clever here.

### Proposal 2 - Use extension methods and builder pattern to reduce member count per quantity
Instead of this:
```
Length l = Length.FromMeters(1);
Length l = Length.FromKilometers(1);
Length l = Length.FromMillimeters(1);
```
We could do something like this:
```
Length l = Length.From.Meters(1);
Length l = Length.From.Kilo.Meters(1);
[...]

PR #419, opened 2018-03-19
Features/rotational stiffness per length unit

dschuermans:
Added new rotational stiffness per length unit and conversions to units involved (rotational stiffness and length)


dschuermans:
@angularsen - Any idea why it keeps including my previous commits? :/
I updated my master with your master before I began and based my branches of off that...

angularsen:
Same as in https://github.com/angularsen/UnitsNet/pull/418#issuecomment-374352460

Rebase or cherry pick to only keep the commits you want on top of the official master branch.

dschuermans:
@angularsen - Thanks, updated the PR :-)

UnitsNet/CustomCode/Quantities/RotationalStiffness.extra.cs
```diff
@@ -37,6 +37,16 @@ public partial struct RotationalStiffness
         {
             return Torque.FromNewtonMeters(rotationalStiffness.NewtonMetersPerRadian * angle.Radians);
         }
+
+        public static RotationalStiffnessPerLength operator /(RotationalStiffness rotationalStiffness, Length length)
+        {
+            return RotationalStiffnessPerLength.FromNewtonMetersPerRadianPerMeter(rotationalStiffness.NewtonMetersPerRadian / length.Meters);
+        }
+
+        public static Length operator /(RotationalStiffness rotationalStiffness, RotationalStiffnessPerLength rotationalStiffnessPerLength)
```
angularsen: You could later add `RotStiff = RotStiffPerLength * Length` and `RotStiff = Length * RotStiffPerLength` if you find a need for those.

angularsen: Nevermind, I see you already added one of these below :-)

UnitsNet/UnitDefinitions/RotationalStiffnessPerLength.json
```diff
@@ -0,0 +1,20 @@
...
+    {
+      "SingularName": "NewtonMeterPerRadianPerMeter",
+      "PluralName": "NewtonMetersPerRadianPerMeter",
+      "FromUnitToBaseFunc": "x",
+      "FromBaseToUnitFunc": "x",
+      "Prefixes": [ "Kilo", "Mega" ],
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "N·m/rad/m", "Nm/rad/m" ]
```
angularsen: Dimensionally, this could be simplified to `N/rad`. I can't find much when googling `RotationalStiffnessPerLength` or `NewtonMeterPerRadianPerMeter`. Are you sure these are widely used units?
I'm a bit hesitant to add quantities or units that are not well documented, that are often very specific to certain applications.

dschuermans: @angularsen - All units i need and will be adding in the future are all related to structural engineering.
This particular unit is used when you have a stiffness per length, much like the force per length.
Consider having springs on a surface edge at a certain interval. Instead of having to specify each spring separately in just rotation stiffness, you can specify them in one go using the rotation stiffness per length.
If you want more structural engineering software to use this library, you'll want these units ;)

angularsen: Alright, thanks for the clarification. This absolutely sounds like a widely used domain to me so we can add these, although as discussed in https://github.com/angularsen/UnitsNet/issues/372 proposal 1, we might want to move such quantities out into a separate nugets at some point - although that design comes with some significant challenges we still need to work out

angularsen:
Nuget out https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F3.94.0

PR #421, opened 2018-03-21
Fix prefix abbrevs

angularsen:
Fixes several wrong abbreviations.
Removes redundant prefix abbreviations (`k`, `M` etc is prefixed by default for kilo, mega etc.)

#### Breaking changes
GigabritishThermalUnit: `mmBTU` => `GBTU`
MegabritishThermalUnit: `mBTU` => `MBTU`
Kilopound: `KLbs` => `klb` (neither is commonly used)
Megapound: `MLbs` => `Mlb` (neither is commonly used)

### DIFF
--- a/UnitsNet/GeneratedCode/UnitSystem.Default.g.cs
+++ b/UnitsNet/GeneratedCode/UnitSystem.Default.g.cs
@@ -1272,7 +1272,7 @@ namespace UnitsNet
                         new CulturesForEnumValue((int) EnergyUnit.GigabritishThermalUnit,
                             new[]
                             {
-                                new AbbreviationsForCulture("en-US", "mmBTU"),
+                                new AbbreviationsForCulture("en-US", "GBTU"),
                             }),
                         new CulturesForEnumValue((int) EnergyUnit.GigawattHour,
                             new[]
@@ -1309,7 +1309,7 @@ namespace UnitsNet
                         new CulturesForEnumValue((int) EnergyUnit.MegabritishThermalUnit,
                             new[]
                             {
-                                new AbbreviationsForCulture("en-US", "mBTU"),
+                                new AbbreviationsForCulture("en-US", "MBTU"),
                             }),
                         new CulturesForEnumValue((int) EnergyUnit.Megajoule,
                             new[]
@@ -2231,7 +2231,7 @@ namespace UnitsNet
                         new CulturesForEnumValue((int) MassUnit.Kilopound,
                             new[]
                             {
-                                new AbbreviationsForCulture("en-US", "KLbs"),
+                                new AbbreviationsForCulture("en-US", "klb"),
                                 new AbbreviationsForCulture("ru-RU", "kфунт"),
                             }),
                         new CulturesForEnumValue((int) MassUnit.Kilotonne,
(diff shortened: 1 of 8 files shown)


angularsen:
Nuget is out https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F3.94.0

d-barker:
Interesting change. What is a GBTU?

[Wikipedia](https://en.wikipedia.org/wiki/British_thermal_unit#:~:text=The%20British%20thermal%20unit%20(Btu,water%20by%20one%20degree%20Fahrenheit)
The British thermal unit (Btu or BTU) is a unit of heat; it is defined as the amount of heat required to raise the temperature of one pound of water by one degree Fahrenheit. 

My understanding is that mmBTU is one million BTU's. This is often used as the energy of a large volume of gas for example has a very large BTU value and for us humans to better relate we need to divide by one million so that it makes comparison much easier (is this now a mix between metric and imperial?)

For example, 1 tonne of LNG gas is 51.7 mmBTU or 51,700,000 BTU.

I would like to use this library however my users will not be typing GBTU they will be using mmBTU, is there a way to extend the acronyms?

p.s. congratulations on a very interesting library

angularsen:
@d-barker 
I believe this change was made after adding kilo/mega/giga prefix support. I'm not familiar with this energy domain myself, so I'll have to lean on what the wiki says.

At least MBTU (1,000 BTUs) is problematic, because it is ambiguous with the metric Mega (1,000,000 BTUs).
I can only assume this change was meant for GBTU to be consistent with MBTU.

From wiki:
> The units MBtu and MMBtu are used in the natural gas and other industries to indicate 1,000 and 1,000,000 BTUs, respectively.[11][12] However, there is an ambiguity in that the metric system uses the prefix "M" to indicate one million (1,000,000), and consequently "MBtu" is often used to indicate one million BTUs.[13] Some authors have deprecated the use of MBtu, and avoided its use to reduce confusion. [12]


Back to your question, yes, we should be able to add back support for `mmBTU` _given_ that there are no existing `Energy` units with the same abbrevation. I did a quick check, and it seems available:

image


If you would be interested in doing a pull request, I see two ways of adding it:

### Option 1 - Add an extra abbreviation to the unit
Add `mmBTU` to the array of abbreviations for the `GigabritishThermalUnit`
https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/Energy.json#L43-L54

, by adding this override:
```
      "Localization": [
        {
          "Culture": "en-US",
          "Abbreviations": [ "BTU" ],
          "AbbreviationsForPrefixes": { "Giga": ["GBTU", "mmBTU"] }
        }
```

`ToString()` would then default to `GBTU` as before, no breaking change, but it would also support 
```
var myEnergy = Energy.Parse("1 mmBTU");
Console.WriteLine($"{myEnergy:v} {myEnergy:a1}"); // "mmBTU", a1 specifies index of a specific unit abbreviation
```

### Option 2 - Add an entirely new unit for mmBTU
We could add a new unit `MillionBritishThermalUnit`.
The advantage is that `Energy.Parse("1 mmBTU").ToString()` would return `"1 mmBTU"`, no need to specify the abbreviation index etc. But I generally want to avoid duplicating the `GigabritishThermalUnit` just to get a different name and different primary unit abbreviation. We can make exceptions if someone familiar with the domain can argue that it makes a lot more sense in the energy domain to have a different naming of the unit or the abbreviation than we already have.

d-barker:
Thanks for the detailed reply, I will have a look at your suggested changes. I am away on leave for a couple of weeks so will have a look when I am back.

PR #425, opened 2018-03-25
Create test project for UnitsNet.Serialization.JsonNet library in Nuget

gojanpaolo:
Fixes https://github.com/angularsen/UnitsNet/issues/417
This is another solution as discussed in https://github.com/angularsen/UnitsNet/pull/423#issuecomment-375937926

The new test project is UnitsNet.Serialization.JsonNet.CompatibilityTests.

UnitsNetJsonConverterTests.cs is added as a link.
`    <Compile Include="..\UnitsNet.Serialization.JsonNet.Tests\UnitsNetJsonConverterTests.cs" Link="UnitsNetJsonConverterTests.cs" />`

The project runs the serialization tests on the latest version of the UnitsNet.Serialization.JsonNet library in Nuget.
`<PackageReference Include="UnitsNet.Serialization.JsonNet" Version="*" />`

### DIFF
--- a/Build/build-functions.psm1
+++ b/Build/build-functions.psm1
@@ -58,7 +58,8 @@ function Start-Build([boolean] $skipUWP = $false) {
 function Start-Tests {
   $projectPaths = @(
     "UnitsNet.Tests\UnitsNet.Tests.NetCore.csproj",
-    "UnitsNet.Serialization.JsonNet.Tests\UnitsNet.Serialization.JsonNet.Tests.NetCore.csproj"
+    "UnitsNet.Serialization.JsonNet.Tests\UnitsNet.Serialization.JsonNet.Tests.NetCore.csproj",
+		"UnitsNet.Serialization.JsonNet.CompatibilityTests\UnitsNet.Serialization.JsonNet.CompatibilityTests.NetCore.csproj"
     )
 
   # Parent dir must exist before xunit tries to write files to it
--- a/Build/build-functions.psm1
+++ b/Build/build-functions.psm1
@@ -59,7 +59,7 @@ function Start-Tests {
   $projectPaths = @(
     "UnitsNet.Tests\UnitsNet.Tests.NetCore.csproj",
     "UnitsNet.Serialization.JsonNet.Tests\UnitsNet.Serialization.JsonNet.Tests.NetCore.csproj",
-		"UnitsNet.Serialization.JsonNet.CompatibilityTests\UnitsNet.Serialization.JsonNet.CompatibilityTests.NetCore.csproj"
+    "UnitsNet.Serialization.JsonNet.CompatibilityTests\UnitsNet.Serialization.JsonNet.CompatibilityTests.NetCore.csproj"
     )
 
   # Parent dir must exist before xunit tries to write files to it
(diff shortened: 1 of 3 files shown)


gojanpaolo:
I have a couple questions:
1. The other test projects has the following line:
`<DotNetCliToolReference Include="dotnet-xunit" Version="2.3.0-beta2-build3683" />`
I'm not sure what that does so I didn't add it to the new test project. Let me know if that needs to be added. A little explanation on what it does will be appreciated :). I think it's related to the `dotnet` commands, but I'm not familiar with that tool yet.

2. As stated in https://github.com/angularsen/UnitsNet/pull/423#issuecomment-375933028

> ... will it be better to force Assert.Inconclusive (yellow) if the JSON nuget fails? This way, we can easily distinguish it from the branch's code test fails (red).

The same idea can still be applied in this solution. :)

angularsen:
Absolutely.
1. `dotnet.exe` command line tooling supports extensions, so the xunit team has written plugin support via `dotnet-xunit` nuget, in order to run our XUnit based tests when typing `dotnet test` in the same folder as a solution file or a project file (see `/Build/build.ps1` and `/Build/build-functions.psm1` for how we run tests). The advantage here is a dead simple way of building, testing, packing nugets and pushing nugets to nuget.org with `dotnet` commands.

2. Inconclusive semantically means "something is up, but it's not necessarily wrong". We could use it, but I personally have seen only trouble using that with both NUnit and XUnit test runners in that subsequent tests are not run when they encounter inconclusive tests. Also, usually the project name is shown next to the test name so it should be fairly apparent what is going on. TL;DR I don't think we need to introduce inconclusive.

gojanpaolo:
Awesome! So if I understood correctly, it's a `runner` for `dotnet` commands similar to the need to include xunit.runner.visualstudio for the tests to run in VS?

angularsen:
That's how I understand it, plus `<DotNetCliToolReference>` also has a special meaning towards the `dotnet` CLI tooling and I assume it will search all nugets referenced that way for any plugin methods it recognizes I guess.

angularsen:
Testing out breaking change in #426

angularsen:
The new compat tests don't seem to run, I added them here:
https://github.com/angularsen/UnitsNet/pull/426/commits/5afd127f1c592e772520fa9cf3325c0c48d5e07c

My attempt to break the tests didn't work out though, will have to take another look later.

gojanpaolo:
Last two commits seemed to do the trick. 

Not sure though about the name of projects in the `Start-Tests` script function having a `.NetCore`.
See https://github.com/angularsen/UnitsNet/pull/425/commits/72ca6c5363eb37a60600d49b5b483a69b3cfa9df#diff-db6c5cef0c4daed00620abcef2c9d724R58

Build/build-functions.psm1
```diff
@@ -58,7 +58,8 @@ function Start-Build([boolean] $skipUWP = $false) {
 function Start-Tests {
   $projectPaths = @(
     "UnitsNet.Tests\UnitsNet.Tests.NetCore.csproj",
-    "UnitsNet.Serialization.JsonNet.Tests\UnitsNet.Serialization.JsonNet.Tests.NetCore.csproj"
+    "UnitsNet.Serialization.JsonNet.Tests\UnitsNet.Serialization.JsonNet.Tests.NetCore.csproj",
+    "UnitsNet.Serialization.JsonNet.CompatibilityTests\UnitsNet.Serialization.JsonNet.CompatibilityTests.NetCore.csproj"
```
angularsen: Oh god I love github and that I can fix minor things like whitespace issues directly in the browser instead of nagging about it :heart:
It was just slightly misaligned, possibly a tab char gone wrong. Now fixed.

angularsen:
Good stuff! Looks good to me and it seems you got the tests to fail correctly too, very nice!

gojanpaolo:
Great! Thank you, I learned a lot by solving this issue. :D

angularsen:
You put a lot of effort into this, much appreciated!

PR #428, opened 2018-04-11
Rename MyUnit to MyQuantity

### LINKED ISSUE #352: Rename occurences of 'MyUnit' to 'MyQuantity' in generated headers
This is a remnant of an earlier rename from "units" to "quantities".

Example:
https://github.com/angularsen/UnitsNet/blob/6e23232dc6bf9d87943b750db25fb31ce1722bb4/UnitsNet/GeneratedCode/Units/AmountOfSubstanceUnit.g.cs#L11

Fix this by both updating all .cs files (some are only generated once) and the .ps1 powershell code generator files, then running generate_code.bat.

bplubell:
This cleans up after an earlier rename from "units" to "quantities".

This resolves #352.

I know this is not my first issue, but it has been untouched for few months now.

### DIFF
--- a/UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
+++ b/UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
@@ -11,9 +11,9 @@
 //
 //     See https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit for how to add or edit units.
 //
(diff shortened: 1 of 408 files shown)


angularsen:
Thanks! No worries about "stealing" it, the only intention of the tag is to help newcomers find stuff they can help out on. We're not intentionally keeping them around :-)

PR #429, opened 2018-04-13
Optimize QuantityValue

jnyrup:
This PR avoid calls to [`Nullable<Decimal>.GetValueOrDefault()`](https://referencesource.microsoft.com/#mscorlib/system/nullable.cs,60) and [`Decimal.ToDouble`](https://referencesource.microsoft.com/#mscorlib/system/decimal.cs,892) when `number._value` is not null.


angularsen:
Ah, good catch. Thanks!

angularsen:
Nuget 3.95 on the way out.

PR #433, opened 2018-05-02
Adding heat transfer coefficient

tmilnthorp:


angularsen:
Nuget 3.98 on the way out.

PR #438, opened 2018-05-06
Fix #436 Flaky test: UnitsNet.Tests.UnitSystemTests.PositiveInfinityFormatting

### LINKED ISSUE #436: Flaky test: UnitsNet.Tests.UnitSystemTests.PositiveInfinityFormatting
This only happens very rarely on AppVeyor (first time I can think of), but @gojanpaolo reports it happens all the time on his setup. I assume this is due to manipulating static values in other tests, such as:

```
GetDefaultAbbreviationFallsBackToUsEnglishCulture()
{
    var zuluCulture = new CultureInfo("zu-ZA");
    CultureInfo.CurrentCulture = CultureInfo.CurrentUICulture = zuluCulture;
```

https://ci.appveyor.com/project/angularsen/unitsnet/build/724/tests

`UnitsNet.Tests.UnitSystemTests.PositiveInfinityFormatting`
```
Assert.Equal() Failure
          ↓ (pos 0)
Expected: ∞ m
Actual:   Infinity m
          ↑ (pos 0)
```

The following:
```
Console.WriteLine(double.PositiveInfinity.ToString(CultureInfo.InvariantCulture));
Console.WriteLine(double.PositiveInfinity.ToString(new CultureInfo("zu-ZA")));

/* Prints:
Infinity
∞
*/
```

So the fix is likely one or both of:
1) Ensure tests that rely on manipulating CultureInfo or other static values are NOT run in parallel (can xunit selectively configure this?)
2) Set CultureInfo properties back to their original values so it doesn't pollute other tests in same test class (I think all tests in one class are run in the same app domain and share static values)

gojanpaolo:
Fixes https://github.com/angularsen/UnitsNet/issues/436

Added `UnitSystemFixture` class which is a `CollectionDefinition` class used to group test classes:
1. that rely on manipulating CultureInfo. See https://github.com/angularsen/UnitsNet/issues/436
2. to avoid accessing static prop DefaultToString in parallel from multiple tests:
	a. UnitSystemTests.DefaultToStringFormatting()
	b. LengthTests.ToStringReturnsCorrectNumberAndUnitWithCentimeterAsDefualtUnit()

Applied `[Collection(nameof(UnitSystemFixture)]` attribute to test classes that:
  - Sets any of the following:
    - `UnitSystem.DefaultCulture`
    - `CultureInfo.CurrentCulture`
    - `CultureInfo.CurrentUICulture`
  - Originally uses `[Collection("DefaulToString")]` that also depends on `CultureInfo`
    - Only one `Collection` attribute per test class.

`UnitSystemTests.GetDefaultAbbreviationFallsBackToUsEnglishCulture`: Set CultureInfo properties back to their original values.

Updated to `xunit` and `xunit.runner.visualstudio` to v2.3.1 (was v2.3.0-beta4-build3742)
	- To have the [CollectionDefinition(DisableParallelization = true)] feature. See https://xunit.github.io/releases/2.3-beta5

### DIFF
--- a/UnitsNet.Serialization.JsonNet.CompatibilityTests/UnitsNet.Serialization.JsonNet.CompatibilityTests.csproj
+++ b/UnitsNet.Serialization.JsonNet.CompatibilityTests/UnitsNet.Serialization.JsonNet.CompatibilityTests.csproj
@@ -20,8 +20,8 @@
     <!--Get the latest released version of UnitsNet.Serialization.JsonNet in Nuget-->
     <PackageReference Include="UnitsNet.Serialization.JsonNet" Version="*" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
-    <PackageReference Include="xunit" Version="2.3.0-beta4-build3742" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.0-beta4-build3742" />
+    <PackageReference Include="xunit" Version="2.3.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
 		<DotNetCliToolReference Include="dotnet-xunit" Version="2.3.0-beta2-build3683" />
   </ItemGroup>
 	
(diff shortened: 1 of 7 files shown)


gojanpaolo:
Specifying `InvariantCulture` turns "∞" to "Infinity".
Should we still specify it then change the expected test result?

angularsen:
Sure, just change the expected result. --
*Andreas Gullberg Larsen* | +47 975 70 270 <+47%20975%2070%20270> | Leinsmoen
7, 7655 Verdal, Norway

gojanpaolo:
Done! :)

angularsen:
Awesome 🎉

PR #444, opened 2018-06-13
Base dimensions

tmilnthorp:
Hello all! I wanted to add base dimensions for all quantity classes. We can later add the automatic operator overload generation based on these. However this is useful information on its own, and does not need to wait for that work.


angularsen:
Cool, I like it. Will review later today.

UnitsNet/BaseDimensions.cs
```diff
@@ -0,0 +1,48 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using System;
+
+namespace UnitsNet
+{
+    public struct BaseDimensions
+    {
+        public readonly int
```
angularsen: You should add `/// <summary></summary>` to each of these.
I think it would also be better to use the unit name and not the measure name: Meter, Kilogram, Second, Ampere, Kelvin, Mole, Candela (https://en.wikipedia.org/wiki/SI_base_unit)

I would personally prefer to use "Meter" for consistency in the lib, but I see that I might be colored by coming from a germanic language and according to some googling "Metre" is the preferred spelling almost everywhere outside US and a couple of scandinavian countries. I still argue for consistency and rather changing all spelling in one go in a major version bump.

UnitsNet/UnitDefinitions/AmountOfSubstance.json
```diff
@@ -2,6 +2,9 @@
   "Name": "AmountOfSubstance",
   "BaseUnit": "Mole",
   "XmlDoc": "Mole is the amount of substance containing Avagadro's Number (6.02 x 10 ^ 23) of real particles such as molecules,atoms, ions or radicals.",
+  "BaseDimensions": {
+    "N": 1
```
angularsen: I think N can be removed altogether, since we store Value with the quantity?
So `2 Feet = 2 x 0.3048 Meter` where Value = 2 and M = 0.3048. Have I got it right?

Just a thought for a future PR, but it might be cool to add dimensions for all units too - not just the base units. This way I think we can possibly generate conversions between units instead of specifying from/to conversions functions in JSON. I have no idea how logarithmic and other non-linear units will work with such conversions, but it's probably doable.

tmilnthorp: I am a bit confused by your comment. N is the [SI base dimension symbol](https://en.wikipedia.org/wiki/SI_base_unit) for amount of substance. It is the base unit so only one base dimension is filled with a value of 1.

I have added the base dimensions for all units. For example Molarity is the amount of a substance in a volume: N / L³. It is represented in Molarity.json as such:

```json
  "BaseDimensions": {
    "L": -3,
    "N": 1
  },
```

And you can write this code in my branch to prove the division works correctly:

``` csharp
var result = AmountOfSubstance.BaseDimensions / Volume.BaseDimensions;
Assert.True( Molarity.BaseDimensions == result );
```
As I noted in my request: these are the dimensions that we can generate the conversions from. However knowing the base dimensions is very useful even without that work being done yet. However this also takes us a step closer to being able to do so!

tmilnthorp:
I just thought it would be nice to have the property on the instance types, as well as be able to cast instance to BaseDimensions implicitly. I also added some supporting tests that also show examples per our conversation above. Help that clears things up!

UnitsNet/GeneratedCode/Quantities/ApparentEnergy.g.cs
```diff
@@ -859,7 +869,7 @@ public string ToString(ApparentEnergyUnit unit)
         /// <returns>The value in the base unit representation.</returns>
         private double AsBaseUnitVoltampereHours()
         {
-			if (Unit == ApparentEnergyUnit.VoltampereHour) { return _value; }
```
tmilnthorp: I fixed some spacing issues I noticed in there also.

UnitsNet.Tests/BaseDimensionsTests.cs
```diff
@@ -0,0 +1,305 @@
...
+        public void CheckBaseDimensionMultiplicationWithForceEqualsMassTimesAccelerationOnInstanceProperty()
+        {
+            var mass = Mass.FromPounds(205);
+            var acceleration = Acceleration.FromMetersPerSecondSquared(9.8);
+
+            var calculatedDimensions = mass.Dimensions * acceleration.Dimensions;
+            Assert.True(calculatedDimensions == Force.BaseDimensions);
+        }
+
+        [Fact]
```
tmilnthorp: A couple examples here of where this can ultimately go.

tmilnthorp:
I would assume everything should be fixed with your build fixes?

tmilnthorp:
Oops didn't mean to close - sorry

tmilnthorp:
Ah. Can't do operator overloading with WCR. I suppose I will have to surround with #ifs like the rest of the code. Out on business but I can do that mid-week.

UnitsNet/UnitDefinitions/AmountOfSubstance.json
angularsen: You are absolutely right, I was confusing "Amount of substance" with "Amount" or "Value".

Regarding 
> add dimensions for all units

If I understand this right, `BaseDimensions` only describes the quantity's base unit - not all other units. So, for `Length`, base dimensions read `M=1` for `Length`'s base unit `Meter`. This allows us to generate operator overloads such as `Area (M=2) = Length (M=1) * Length (M=1)`. However, it does not allow us to generate the conversion from `Meter` to `Foot`, without also defining the dimensions of the `Foot` unit, right?

UnitsNet.Tests/BaseDimensionsTests.cs
```diff
@@ -254,5 +254,52 @@ public void LuminousIntensityDimensionsDivideCorrectly()
             Assert.True(result.Amount == 0);
             Assert.True(result.LuminousIntensity == 1);
         }
+
+        [Fact]
+        public void CheckBaseDimensionDivisionWithSpeedEqualsDistanceDividedByTimeOnStaticProperty()
+        {
+            var calculatedDimensions = Length.BaseDimensions / Duration.BaseDimensions;
```
angularsen: Nice, this looks like I imagined.

UnitsNet/Scripts/Include-GenerateQuantitySourceCode.ps1
```diff
@@ -199,16 +199,16 @@ namespace UnitsNet
...
+    {
 @"
-          /// <summary>
-          ///     The <see cref="BaseDimensions" /> of this quantity.
-          /// </summary>
-          public static BaseDimensions BaseDimensions => new BaseDimensions($baseDimensionLength, $baseDimensionMass, $baseDimensionTime, $baseDimensionElectricCurrent, $baseDimensionTemperature, $baseDimensionAmountOfSubstance, $baseDimensionLuminousIntensity);
+        /// <summary>
+        ///     The <see cref="BaseDimensions" /> of this quantity.
+        /// </summary>
+        public static BaseDimensions BaseDimensions => new BaseDimensions($baseDimensionLength, $baseDimensionMass, $baseDimensionTime, $baseDimensionElectricCurrent, $baseDimensionTemperature, $baseDimensionAmountOfSubstance, $baseDimensionLuminousIntensity);
```
angularsen: Use `BaseDimensions { get; } = new ...` instead to avoid allocating a new instance every time the property is called.

UnitsNet/Scripts/Include-GenerateQuantitySourceCode.ps1
```diff
@@ -235,7 +235,19 @@ namespace UnitsNet
 
         public static $quantityName Zero => new $quantityName(0, BaseUnit);
 
-"@; foreach ($unit in $units) {
+"@; 
+    if($quantity.BaseDimensions)
+    {
+@"
+        public static implicit operator BaseDimensions($quantityName quantity)
```
angularsen: Why is this implicit operator useful? I don't see it in your examples above.
I'm generally not too fond of implicit casts as they can trick the programmer up.

tmilnthorp: You can see the example in the following test:

```csharp
[Fact]
public void CheckBaseDimensionMultiplicationWithForceEqualsMassTimesAccelerationByImplicitOverload()
{
    var mass = Mass.FromPounds(205);
    var acceleration = Acceleration.FromMetersPerSecondSquared(9.8);

    BaseDimensions calculatedDimensions = mass * acceleration;
    Assert.True(calculatedDimensions == Force.BaseDimensions);
}
```

These kind of operations are quite useful. I completely understand about implicit casts. I would make them explicit, however it seems WCR doesn't support cast overloads, so I think I'll just remove them entirely and rely on the properties.

UnitsNet/UnitDefinitions/AmountOfSubstance.json
tmilnthorp: Correct it's only the base dimensions that the derived units boil down to.

Length for example is defined in base units as follows:

```json
  "BaseDimensions": {
    "L": 1
  },
```
I think you're confusing the base dimensions with the unit. M is the [Base Dimension Symbol](https://en.wikipedia.org/wiki/SI_base_unit) for Mass, not meters. 

Area is defined in base dimensions as Length * Length or L^2 (L: 2). So Length (L: 1) when multiplied adds the base dimensions giving you L: 2. You'd then look this up in a map of base dimensions -> quantity.  L:2 being area.

The conversion from Meter to Feet would still be in the Length quantity. Nothing else is necessary. Base dimensions apply to ALL units in length (feet, meters, fathoms etc.) They all have a base dimension of Length = 1 (L: 1).

angularsen: Thanks, I was indeed confused by this. As you say, I was of the understanding that the dimensions could be used to express *all units* - not just the base unit of each quantity. The confusion arise from that it is always the *base unit* of the quantity that expresses the dimensions, so for `Area` the base unit is `m²` and thus `L=2`.  In my mind, we could then somehow represent feet and inches the same way, since most units is a simple linear transformation of the base unit, but maybe that gets into the same territory why chose to have two conversion functions for `from` and `to`, since this function is not always trivial to invert.

Anyway, I think I'm onboard with the correct thinking now :-)

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -161,7 +161,7 @@ public Acceleration(double meterspersecondsquared)
         /// <summary>
         ///     The <see cref="BaseDimensions" /> of this quantity.
         /// </summary>
-        public static BaseDimensions BaseDimensions => new BaseDimensions(1, 0, -2, 0, 0, 0, 0);
+        public static BaseDimensions BaseDimensions = new BaseDimensions(1, 0, -2, 0, 0, 0, 0);
```
angularsen: You need to add either `readonly` to make that a field, or `{ get; } = ` to make it a statically initialized getter property. I prefer the latter, mostly because it can be included in an interface or base type later with less changes.

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -158,6 +158,11 @@ public Acceleration(double meterspersecondsquared)
         /// </summary>
         public static AccelerationUnit BaseUnit => AccelerationUnit.MeterPerSecondSquared;
 
+        /// <summary>
+        ///     The <see cref="BaseDimensions" /> of this quantity.
+        /// </summary>
+        public static BaseDimensions BaseDimensions = new BaseDimensions(1, 0, -2, 0, 0, 0, 0);
```
angularsen: You need to either make it a `readonly` field or add `{ get; } = new ...` to make it a statically initialized getter property. I prefer the latter, mostly because it can be included in an interface or base type later with less changes.

angularsen: Besides this, I think this is getting ready to merge.

tmilnthorp:
Sorry for all those commits for Windows Runtime Components. I don't have the Visual Studio support installed and I am not very familiar with them. I was relying on AppVeyor :)

Build succeeds now!

UnitsNet/UnitDefinitions/AmountOfSubstance.json
tmilnthorp: The word base in base dimensions is different than base as in base units. All derived units (like area) are derived from some combination of the 7 base dimensions. Length is a base dimension. Feet or meters are not. They are units that represent the length base dimension.

The base dimensions *are* expressed for **all** units inside a quantity, not just the base unit.

Area, whether m² or in² or ft², all have base dimensions of L=2.

UnitsNet/UnitDefinitions/ReactivePower.json
```diff
@@ -1,7 +1,12 @@
-{
+﻿{
   "Name": "ReactivePower",
   "BaseUnit": "VoltampereReactive",
   "XmlDoc": "Volt-ampere reactive (var) is a unit by which reactive power is expressed in an AC electric power system. Reactive power exists in an AC circuit when the current and voltage are not in phase.",
+  "BaseDimensions": {
+    "L": 2,
+    "M": 1,
+    "T": -2
```
angularsen: T: -3?
http://www.wolframalpha.com/input/?i=voltampere+reactive

UnitsNet/UnitDefinitions/RotationalSpeed.json
```diff
@@ -1,8 +1,11 @@
 ﻿{
-    "Name": "RotationalSpeed",
-    "BaseUnit": "RadianPerSecond",
-    "XmlDoc": "Rotational speed (sometimes called speed of revolution) is the number of complete rotations, revolutions, cycles, or turns per time unit. Rotational speed is a cyclic frequency, measured in radians per second or in hertz in the SI System by scientists, or in revolutions per minute (rpm or min-1) or revolutions per second in everyday life. The symbol for rotational speed is ω (the Greek lowercase letter \"omega\").",
-    "Units": [
+  "Name": "RotationalSpeed",
+  "BaseUnit": "RadianPerSecond",
+  "XmlDoc": "Rotational speed (sometimes called speed of revolution) is the number of complete rotations, revolutions, cycles, or turns per time unit. Rotational speed is a cyclic frequency, measured in radians per second or in hertz in the SI System by scientists, or in revolutions per minute (rpm or min-1) or revolutions per second in everyday life. The symbol for rotational speed is ω (the Greek lowercase letter \"omega\").",
+  "BaseDimensions": {
+    "T": -1
```
angularsen: I just realized something that might block automatically generating conversions between quantities.

Wolfram lists the dimensions like this: `[time]^(-2) [angle]`
http://www.wolframalpha.com/input/?i=radian+per+second+squared

Without angle, which is not a SI base unit, we don't represent that this is a rotational quantity so there is now ambiguity whether `Length / Time` should equal `Speed` or `RotationalSpeed`. Same for [illuminance](http://www.wolframalpha.com/input/?i=illuminance&assumption=%7B%22FS%22%7D+-%3E+%7B%7B%22IlluminanceFromLuminousIntensity%22,+%22I%22%7D%7D&assumption=%7B%22F%22,+%22IlluminanceFromLuminousIntensity%22,+%22r%22%7D+-%3E%221+m%22&assumption=%7B%22F%22,+%22IlluminanceFromLuminousIntensity%22,+%22Enu%22%7D+-%3E%221+lx%22&assumption=%7B%22C%22,+%22illuminance%22%7D+-%3E+%7B%22PhysicalQuantity%22%7D) having `[solid angle]` as a dimension.

The same goes for all the rotational vs linear quantities really.

Am I missing something?

UnitsNet/UnitDefinitions/ThermalResistance.json
```diff
@@ -2,6 +2,11 @@
   "Name": "ThermalResistance",
   "BaseUnit": "SquareMeterKelvinPerKilowatt",
```
angularsen: Not the scope of this PR, but it seems this is maybe not the correct base unit, but rather `KelvinPerWatt` `K/W`.
http://www.wolframalpha.com/input/?i=thermal+resistance

UnitsNet/BaseDimensions.cs
```diff
@@ -151,5 +151,19 @@ private static void AppendDimensionString( StringBuilder sb, string name, int va
                     sb.AppendFormat( "^{0}", value );
             }
         }
+
+        public int Length { get { return _length; } }
```
angularsen: In C#7 (or was it C#6?) you no longer need this field. You can simply have a property `public int Mass { get; }` that is then assigned its value in the constructor.

angularsen:
Awesome job on this, I'm going to bed now and will try to check back in tomorrow.

UnitsNet/BaseDimensions.cs
```diff
@@ -0,0 +1,169 @@
...
+            if( absoluteValue > 0 )
+            {
+                sb.AppendFormat( "[{0}]", name );
+
+                if( absoluteValue > 1 )
+                    sb.AppendFormat( "^{0}", value );
+            }
+        }
+
+        public int Length { get { return _length; } }
```
angularsen: In C#7 (or was it C#6?) you no longer need a backing field. You can simply have a property public int Mass { get; } that is then assigned its value in the constructor.

UnitsNet/UnitDefinitions/ThermalResistance.json
tmilnthorp: Indeed!

UnitsNet/UnitDefinitions/RotationalSpeed.json
tmilnthorp: Some can be the same. Take a look at PressureChangeRate and PowerDensity for example.

I am not entirely sure how to resolve this just yet. Pa/s is technically valid as both a power density as well as a pressure change rate, but only used in practice as a pressure change rate :)

tmilnthorp:
Just a bit of tidying up there. All should be ready to go!

UnitsNet/Scripts/Include-GenerateQuantityTypeSourceCode.ps1
```diff
@@ -43,12 +43,12 @@ function GenerateQuantityTypeSourceCode($quantityNames)
...
-	///     such as Length, Mass, Force etc.
-	///     This is useful for populating options in the UI, such as creating a generic conversion
-	///     tool with inputValue, quantityName, fromUnit and toUnit selectors.
-	/// </summary>
+    /// <summary>
+    ///     Lists all generated quantities with the same name as the quantity struct type,
+    ///     such as Length, Mass, Force etc.
+    ///     This is useful for populating options in the UI, such as creating a generic conversion
+    ///     tool with inputValue, quantityName, fromUnit and toUnit selectors.
+    /// </summary>
```
angularsen: Good thing to catch whitespace issues like this

tmilnthorp: I personally prefer tabs, but consistency rules the day :)

UnitsNet.Tests/BaseDimensionsTests.cs
```diff
@@ -0,0 +1,580 @@
...
+
+        [Fact]
+        public void LengthDimensionsMultiplyCorrectlyWithOperatorOverloads()
+        {
+            var baseDimensions1 = new BaseDimensions(1, 0, 0, 0, 0, 0, 0);
+            var baseDimensions2 = new BaseDimensions(2, 0, 0, 0, 0, 0, 0);
+
+            var result = baseDimensions1 * baseDimensions2;
+
+            Assert.True(result.Length == 3);
```
angularsen: We don't have to do this now, but you could DRY this up a bit by adding a custom `AssertDimensionsEqual(BaseDimension d, int length, int mass, int time, int current, int temperature, int amount, int luminousIntensity)` method.

UnitsNet.Tests/BaseDimensionsTests.cs
```diff
@@ -0,0 +1,580 @@
...
+            Assert.True(result.Mass == 0);
+            Assert.True(result.Time == 0);
+            Assert.True(result.Current == 0);
+            Assert.True(result.Temperature == 0);
+            Assert.True(result.Amount == 0);
+            Assert.True(result.LuminousIntensity == 1);
+        }
+
+        [Fact]
+        public void CheckBaseDimensionDivisionWithSpeedEqualsDistanceDividedByTimeOnStaticPropertyWithOperatorOverloads()
```
angularsen: I like descriptive method names, but maaybe a bit overboard here? :-D
How about `LengthDividedByDurationReturnsSpeed` ? We are already in a `BaseDimensionsTests` class, so the context is implicit.

Again, we can address this in a separate PR. I'm about merge this in now.

angularsen: Excellent work on these tests though! 👏

UnitsNet/Scripts/Include-GenerateQuantitySourceCode.ps1
```diff
@@ -189,12 +199,29 @@ namespace UnitsNet
...
+"@; 
+    if($baseDimensions)
+    {
+@"
+        private static readonly BaseDimensions _baseDimensions = new BaseDimensions($baseDimensionLength, $baseDimensionMass, $baseDimensionTime, $baseDimensionElectricCurrent, $baseDimensionTemperature, $baseDimensionAmountOfSubstance, $baseDimensionLuminousIntensity);
+
+        /// <summary>
+        ///     The <see cref="BaseDimensions" /> of this quantity.
+        /// </summary>
+        public static BaseDimensions BaseDimensions
```
angularsen: We could remove backing field here too, statically initialized getter property. Again, feel free to do in separate PR. No biggie.

UnitsNet.Tests/BaseDimensionsTests.cs
tmilnthorp: Indeed we could. I could also test Equals with a new BaseDimensions. I think I like that better - you?

UnitsNet.Tests/BaseDimensionsTests.cs
tmilnthorp: Perhaps a bit overly explicit :) Easy fix!

tmilnthorp:
Sounds good. Thanks!

angularsen:
@tmilnthorp Let me know when you want to do a nuget release, I'll hold off until you get any extra changes you want to include. We normally release after every PR.

Also, I have no better way of reaching out to you than here, but I would like to say I appreciate the quality work you have been putting in lately and I would very much like to invite you as a collaborator of Units.NET. We don't expect much other than helping out answering issues and reviewing PRs whenever you feel you have the time for it. You will also be in a good position to move the library in any desired direction as the collaborators are the decision makers. I'm not the only one with a say in this, but I thought I'd check in with you about it. Let me know what you think!

Cheers.

tmilnthorp:
Feel free to do a NuGet release. I am trying to squeeze in Units.NET fun between work, so not always sure how quickly I can code for this :)

Thank you very much for the compliment! I would be honored to be a collaborator . Units come up every day in my line of work (CAE), so I find them very interesting and love helping out.

angularsen:
Awesome! I'll ask for a vote of the collaborators, but I expect it to be positive.

Your PR made release number 100 for version 3.x 🎉 About time to consider that vNext I think ;-)
https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F3.100.0

PR #445, opened 2018-06-13
Add operator overloads

gojanpaolo:
Length = Pressure / SpecificWeight
Density = SpecificWeight / Acceleration


angularsen:
Appveyor build is broken for some (different) reason

Issue #391, opened 2018-01-26
Relative tolerance instead of absolute tolerance


ferittuncer:
As discussed in #350 we are planning to implement relative tolerance and use a standard relative tolerance value to test units instead of using a different absolute tolerance value for each unit.

Benefits are:
 - Standardized precision
 - Shorter source-code

We can implement it similar to [MathNet.Numerics.Precision.Equality](https://github.com/mathnet/mathnet-numerics/blob/master/src/Numerics/Precision.Equality.cs)

tmilnthorp:
I think this can be closed now unless anyone has any objections?

angularsen:
Yup!

Issue #409, opened 2018-02-13
Generate UnitsNet.WindowsRuntimeComponent as its own source code


angularsen:
The current UnitsNet source code is a bit of a mess due to supporting Windows Runtime Component (WinRTC) and adapting to all the constraints WinRTC imposes on the public API types and method signatures. This results in MANY `#if WINDOWS_UWP` directives that clutters the code.

Short story; WinRTC allows .NET code and nugets to be consumed by JavaScript and C++ Windows UWP apps, by sticking to some strict constraints on the public API. [Read more on Windows Runtime Component and its constraints.](https://docs.microsoft.com/en-us/windows/uwp/winrt-components/creating-windows-runtime-components-in-csharp-and-visual-basic)

I propose:
* Move WinRTC project to `/UnitsNet.WindowsRuntimeComponent` folder
* Generate WinRTC source code separately in new folder
* Remove `#if` directives from generated code of each set so it only has the code it needs
* Add linked files (not copies) of `/UnitsNet` source code that is fully compatible with both, with same folder structure
* Update build scripts with new paths
* Bonus: Copy/add some test cases specific to WinRTC, we currently have zero test coverage on it

I would rank this as medium difficulty, it definitely involves some mucking around with the source code, the code generator scripts and the build scripts. I'm happy to assist if anyone wants to take this on though!

angularsen:
In #520 we made an effort to simplify the generated code for WRC and NetFramework. It's still the same scripts, which makes sense since the output is so similar and it would require more effort to keep two sets of scripts in sync, but at least now there is generated separate sets of source code for WRC and NetFramework so we don't have all the `#if WINDOWS_UWP` sections in the code. That complexity is now contained in the generator scripts.

PR #451, opened 2018-06-21
Relative and absolute equality

tmilnthorp:
Initial go at #391


UnitsNet/UnitDefinitions/Energy.json
```diff
@@ -136,8 +136,8 @@
     {
       "SingularName": "ThermImperial",
       "PluralName": "ThermsImperial",
-      "FromUnitToBaseFunc": "x*1.05505585257348e+14",
-      "FromBaseToUnitFunc": "x/1.05505585257348e+14",
+      "FromUnitToBaseFunc": "x*1.05967e8",
```
tmilnthorp: [Was wrong](http://www.wolframalpha.com/input/?i=1+therm_39+to+J)

angularsen: There are actually 4 definitions: Therm EC/US/UK and BTU

https://en.wikipedia.org/wiki/Therm#Definitions

ThermImperial may be misleading in name, but refers to Therm UK.
There is already a unit for BritishThermalUnit (BTU) further up here.

So I think it is already correct, as is. Do you agree?

UnitsNet.Tests/AssertEx.cs
```diff
@@ -7,9 +8,22 @@ namespace UnitsNet.Tests
...
         {
-            Assert.True(actual >= expected - tolerance && actual <= expected + tolerance, $"Values are not equal within tolerance: {tolerance}\nExpected: {expected}\nActual: {actual}\nDiff: {actual - expected:e}");
+            if(comparisonType == ComparisonType.Relative)
+            {
+                bool areEqual = UnitsNet.Comparison.EqualsRelative(expected, actual, tolerance);
+
+                double difference = Math.Abs(expected - actual);
+                double percentDifference = (difference / actual) * 100.0d;
+
+                Assert.True( areEqual, $"Values are not equal within relative tolerance: {tolerance * 100.0d}%\nExpected: {expected}\nActual: {actual}\nDiff: {percentDifference}%" );
```
angularsen: You can use format P2 as in `{0.0354321:P2}` to get `"3.54%"` string including the `%` symbol. It also uses the right symbol for the current culture.

angularsen: I would perhaps argue that percentage should be relative to `expected` and not `actual`, since you don't really know what actual is, but expected is .. well.. what you expect :-) If expected is `10` and actual is `0.000001`, then a difference of `~10` becomes a huge percentage instead of having an error of `~100%`.

UnitsNet/Comparison.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using System;
+
+namespace UnitsNet
+{
+    public class Comparison
+    {
+        public static bool Equals( double value1, double value2, double tolerance = 0.00001, ComparisonType comparisonType = ComparisonType.Relative)
+        {
+            if(comparisonType == ComparisonType.Relative)
```
angularsen: A `switch` can be slightly more readable when comparing the same value multiple times and returning a value based on it.

UnitsNet/Comparison.cs
```diff
@@ -0,0 +1,49 @@
...
+                return EqualsRelative(value1, value2, tolerance);
+            else if(comparisonType == ComparisonType.Absolute)
+                return EqualsAbsolute(value1, value2, tolerance);
+            else
+                throw new InvalidOperationException("The ComparisonType is unsupported.");
+        }
+
+        public static bool EqualsRelative( double value1, double value2, double tolerance = 0.00001)
```
angularsen: The xmldoc should explain that the tolerance is relative to `value1`, which is kind of important I think.
Consider renaming `value1` to `referenceValue`, since we must choose a reference to compare relative to. The other could be named `value` maybe. Perhaps there are some naming conventions on this sort of thing.

UnitsNet/Comparison.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            double maxVariation = Math.Abs( value1 * tolerance );
```
angularsen: We should throw if tolerance is `< 0`, even if it would work the same.

UnitsNet/GeneratedCode/Extensions/Number/NumberToPowerDensityExtensions.g.cs
```diff
@@ -1132,6 +1166,40 @@ public static class NumberToPowerDensityExtensions
 
         #endregion
 
+        #region PetawattPerCubicMeter
+
+        /// <inheritdoc cref="PowerDensity.FromPetawattsPerCubicMeter(UnitsNet.QuantityValue)" />
+        public static PowerDensity PetawattsPerCubicMeter(this int value) => PowerDensity.FromPetawattsPerCubicMeter(value);
```
angularsen: Consider moving the PowerDensity units to a separate PR, even if that is probably why you started working on the abs vs relative comparison to begin with.

UnitsNet/GeneratedCode/Quantities/VolumeFlow.g.cs
```diff
@@ -1153,6 +1153,25 @@ public override bool Equals(object obj)
...
+        ///     Relative tolerance is when the difference between the two quantities is not greater than the scale of the values compared to the tolerance.
+        ///     Absolute tolerance is when the difference between the two quantities is not greater than exactly the tolerance.
+        ///     Note that it is advised against specifying zero difference, due to the nature
+        ///     of floating point operations and using System.Double internally.
+        /// </summary>
+        /// <param name="other">The other quantity to compare to.</param>
+        /// <param name="tolerance">The comparison tolerance.</param>
+        /// <param name="comparisonType">The comparison type: either relative or absolute.</param>
+        /// <returns>True if the difference between the two values is not greater than the specified relative or absolute tolerance.</returns>
+        public bool Equals(VolumeFlow other, double tolerance = 0.00001, ComparisonType comparisonType = ComparisonType.Relative)
```
angularsen: Tolerance probably needs to be of type `VolumeFlow`, since the consumer may be working with units of entirely different orders of magnitude than the base unit - so the default of `0.00001` may be way too small or way too large. 

A second option is to rely on the fact that we keep track of the `Unit` the quantity was constructed with, so we could pick one of the quantities' `Unit` and convert both quantities to a numeric value of that unit, then compare that with a default numeric tolerance. So when comparing a 1 nanometer with 2 nanometers, then the tolerance would 0.000001 nanometers by default rather than 0.000001 meters. The pitfall here is that it matters WHICH value you choose the unit for, as you get very different behavior when comparing 1 kilometer to 1e12 nanometers vs comparing 1e12 nanometers to 1 kilometer. In this particular example, I would expect `this` to be the reference - which I guess is intuitive.

UnitsNet/UnitDefinitions/Energy.json
tmilnthorp: I did assume it was therm(UK). I thought that was therm_39, but looks like I'm completely wrong there. It caught my eye immediately though because the scale is completely off (e14).

If you look at Wikipedia, it should be 1.05505585257348e8, not 1.05505585257348e14. Looks like the decimal on Wikipedia got overlooked :)

So just to make sure I'm not missing anything, 1.05505585257348e8 seems correct, do you agree?

angularsen: Ah I missed that part :-D Yup, should be `e8`!

tmilnthorp: I will change it to that then!

UnitsNet.Tests/AssertEx.cs
tmilnthorp: I didn't know you could do the P2 formatting outside of a ToString. Nice! I'll update that. And I agree that it should be against the expected.

UnitsNet/GeneratedCode/Extensions/Number/NumberToPowerDensityExtensions.g.cs
tmilnthorp: That is indeed why I started :) I just forgot to remove it. Good catch

UnitsNet/Comparison.cs
tmilnthorp: Done!

UnitsNet/Comparison.cs
tmilnthorp: I agree. I've updated this and added documentation that I hope is clear.

UnitsNet/Comparison.cs
tmilnthorp: Agreed. I was going to initially say <= 0 but I think it's perfectly valid to request being exact if desired.

UnitsNet/GeneratedCode/Quantities/VolumeFlow.g.cs
tmilnthorp: I think I like option #2 better. Absolute/relative can be a bit confusing and keeping it a double makes it easier to think of as a "percentage". Being typed would make sense for absolute differences, but since this is a generic method I think it's better as a double.

I made this in the same thinking of String.Equals where you can pass in the comparison type (ignore case, etc.).

I agree the reference being [this] is intuitive.

UnitsNet/Comparison.cs
```diff
@@ -31,21 +31,21 @@ public sealed class Comparison
         ///     Absolute tolerance is when the absolute difference between the two values is not greater than the tolerance value.
         /// </summary>
         /// <param name="referenceValue">The first value. If using relative tolerance, it is the value which the relative tolerance will be calculated against.</param>
-        /// <param name="value">The value to compare to.</param>
+        /// <param name="otherValue">The value to compare to.</param>
         /// <param name="tolerance">The absolute or relative tolerance value.</param>
         /// <param name="comparisonType">Whether to use tolerance as an absolute or relative tolerance.</param>
         /// <returns></returns>
-        public static bool Equals(double referenceValue, double value, double tolerance, ComparisonType comparisonType)
+        public static bool Equals(double referenceValue, double otherValue, double tolerance, ComparisonType comparisonType)
```
angularsen: WRC target is a big pain in the ass, I've contemplated dropping the support for it, but we have a Microsoft package that depends on it which is kind of cool.

I have created an issue to generate a separate code base for WRC, to avoid all the #if defs, which I think will be an improvement since most people don't care about WRC anyway. Then hopefully one will meet less friction when adding new stuff, since adding extra code like operator overloads would by default not be included in WRC.

https://github.com/angularsen/UnitsNet/issues/409

tmilnthorp: I did see that, and considered picking that up :)

UnitsNet.Tests/CustomCode/MassTests.cs
```diff
@@ -42,11 +42,11 @@ public class MassTests : MassTestsBase
 
         protected override double KilotonnesInOneKilogram => 1E-6;
 
-        protected override double LongTonsInOneKilogram => 0.000984207;
+        protected override double LongTonsInOneKilogram => 9.842065276110606e-4;
```
angularsen: This PR is getting a bit crowded and not so easy to review. Could you perhaps pull out the bits not directly related to absolute/relative functionality in a separate PR?

tmilnthorp: Unfortunately all of these changes are so the tests pass. These validation numbers were off quite a bit when I changed the tests to validate by % tolerance (not precise enough).

angularsen: Haha 😆 it's so good to finally fix this!
I understand about the precision changes, no problem!

angularsen: Is this getting ready now? I might find some time tomorrow to review.

tmilnthorp: Yup! I'm done making changes.

UnitsNet/Comparison.cs
```diff
@@ -0,0 +1,87 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+using System;
+
+namespace UnitsNet
+{
+    public sealed class Comparison
```
angularsen: Make this class `static`? Should it be `internal`?

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -790,6 +790,29 @@ public override bool Equals(object obj)
             return AsBaseUnitMetersPerSecondSquared().Equals(((Acceleration) obj).AsBaseUnitMetersPerSecondSquared());
         }
 
+        /// <summary>
+        ///     Compare equality to another Acceleration within the given absolute or relative tolerance.
+        ///     Relative tolerance is when the difference between the two quantities is not greater than the scale of the values compared to the tolerance.
```
angularsen: How about:
> Relative tolerance is defined by a fraction of this quantity, so a value of 0.01 means +/- 1% tolerance.
> Absolute tolerance is defined by the max difference between this quantity and <paramref name="other" /> in the <see cref="Unit" /> of this quantity. So if this quantity was constructed with 100 cm, then a tolerance of 0.01 means a tolerance of +/- 1 cm.

When writing this, I realize that I'm not sure this behavior for absolute tolerance is intuitive enough and whether we should have a separate method for absolute equality that instead take an `Acceleration` as tolerance? The relative tolerance implementation is fine I think.

UnitsNet/UnitDefinitions/Pressure.json
```diff
@@ -272,8 +272,8 @@
     {
       "SingularName": "TonneForcePerSquareCentimeter",
       "PluralName": "TonnesForcePerSquareCentimeter",
-      "FromUnitToBaseFunc": "x*98066501.9960652",
-      "FromBaseToUnitFunc": "x*1.01971619222242E-08",
+      "FromUnitToBaseFunc": "x*9.80665e7",
+      "FromBaseToUnitFunc": "x/9.80665e7",
```
angularsen: Nice that you are making this consistent with the guideline!

UnitsNet.Tests/CustomCode/MassTests.cs
```diff
@@ -42,11 +42,11 @@ public class MassTests : MassTestsBase
...
 
         protected override double MegapoundsInOneKilogram => 2.2046226218487757e-6;
 
-        protected override double MegatonnesInOneKilogram => 1E-6;
+        protected override double MegatonnesInOneKilogram => 1E-9;
```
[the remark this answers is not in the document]
tmilnthorp: I know!

UnitsNet/Comparison.cs
tmilnthorp: I thought I tried and Windows Runtime Component yelled at me. I'll double check because it should definitely be static. Not so sure about internal. Can be used anywhere if someone wishes!

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
tmilnthorp: I think having an example is good here. Makes it easier to understand. I'll add that. 

As for making the method typed I'm not so sure. When we want to implement an IQuantity interface, having the generic method that takes a double is the only way to make it reusable as an interface method.

UnitsNet/UnitDefinitions/Pressure.json
tmilnthorp: When the tests failed due to relative checks, I couldn't know if the conversion was wrong or the test, so I did both :)

UnitsNet.Tests/CustomCode/DurationTests.cs
```diff
@@ -36,19 +36,19 @@ public class DurationTests : DurationTestsBase
...
 
         protected override double NanosecondsInOneSecond => 1e+9;
 
         protected override double SecondsInOneSecond => 1;
 
-        protected override double WeeksInOneSecond => 1.6534e-6;
+        protected override double WeeksInOneSecond => 1.653439153439153e-6;
 
-        protected override double YearsInOneSecond => 3.1689e-8;
+        protected override double YearsInOneSecond => 3.170979198376458e-8;
```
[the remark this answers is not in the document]
tmilnthorp: Are you saying it is supposed to be different? Both are defines in the json files as x*365*24*3600, so I am not sure how they could be?

UnitsNet/Comparison.cs
tmilnthorp: It seems to be fine with static. I'll leave it as public though if that's ok with you?

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
tmilnthorp: That is true with the interface, but generally I'd expect a check, much like Equals(Object) does. If the types are different it would just always return false. But all debatable :)

Are you saying you'd like to see:

```csharp
var a = Length.FromMeters( 5.0 );
var isEqual = a.EqualsAbsolute( Length.FromMeters( 10.0 ), Length.FromNauticalMiles( 100000 ) )
```

tmilnthorp:
How on earth does public bool Equals($quantityName other, $quantityName maxError) in the source code generation not get emitted as a change in the generated files? :confused:

UnitsNet/Comparison.cs
```diff
@@ -26,14 +26,23 @@ namespace UnitsNet
...
-        ///     Absolute tolerance is when the absolute difference between the two values is not greater than the tolerance value.
+        ///     </para>
+        ///     <para>
+        ///     Relative tolerance is defined as the maximum allowable absolute difference between <paramref name="referenceValue"/> and
+        ///     <paramref name="otherValue"/> as a percentage of <paramref name="referenceValue"/>. For example, a relative tolerance
+        ///     of 0.01 means the difference must be within +/- 1% of <paramref name="referenceValue"/> to be considered equal.
+        ///     </para>
+        ///     <para>
+        ///     Absolute tolerance is defined as the maximum allowable absolute difference between <paramref name="referenceValue"/> and
+        ///     <paramref name="otherValue"/> as a fixed number.
```
angularsen: Fixed number, but in what unit?

UnitsNet/Comparison.cs
```diff
@@ -52,12 +61,14 @@ public static bool Equals(double referenceValue, double otherValue, double toler
...
-        ///     Relative tolerance is when the difference between the two values is not greater than referenceValue * tolerance.
+        ///     Checks if two values are equal with a given relative tolerance.
+        ///     Relative tolerance is defined as the maximum allowable difference between <paramref name="referenceValue"/> and
+        ///     <paramref name="otherValue"/> as a percentage of <paramref name="referenceValue"/>. For example, a relative tolerance
+        ///     of 0.01 means the difference must be within +/- 1% of <paramref name="referenceValue"/> to be considered equal.
         /// </summary>
         /// <param name="referenceValue">The reference value which the tolerance will be calculated against.</param>
         /// <param name="otherValue">The value to compare to.</param>
-        /// <param name="tolerance"></param>
+        /// <param name="tolerance">The relative tolerance. Must be greater than 0.</param>
```
angularsen: greater or equal

UnitsNet/Comparison.cs
```diff
@@ -26,14 +26,23 @@ namespace UnitsNet
...
+        ///     Absolute tolerance is defined as the maximum allowable absolute difference between <paramref name="referenceValue"/> and
+        ///     <paramref name="otherValue"/> as a fixed number.
+        ///     </para>
         /// </summary>
-        /// <param name="referenceValue">The first value. If using relative tolerance, it is the value which the relative tolerance will be calculated against.</param>
+        /// <param name="referenceValue">The reference value. If using relative tolerance, it is the value which the relative tolerance will be calculated against.</param>
         /// <param name="otherValue">The value to compare to.</param>
-        /// <param name="tolerance">The absolute or relative tolerance value.</param>
-        /// <param name="comparisonType">Whether to use tolerance as an absolute or relative tolerance.</param>
+        /// <param name="tolerance">The absolute or relative tolerance value. Must be greater than 0.</param>
```
angularsen: greater or equal

UnitsNet/Comparison.cs
```diff
@@ -70,7 +81,8 @@ public static bool EqualsRelative(double referenceValue, double otherValue, doub
 
         /// <summary>
         ///     Checks if two values are equal with a given absolute tolerance.
-        ///     Absolute tolerance is when the absolute difference between the two values is not greater than the tolerance value.
+        ///     Absolute tolerance is defined as the maximum allowable absolute difference between <paramref name="referenceValue"/> and
+        ///     <paramref name="otherValue"/> as a fixed number.
```
angularsen: Same, what unit?

UnitsNet/Scripts/Include-GenerateQuantitySourceCode.ps1
```diff
@@ -445,15 +445,25 @@ namespace UnitsNet
...
+        ///     <paramref name="other"/> as a fixed number.
+        ///     </para>
+        ///     <para>
         ///     Note that it is advised against specifying zero difference, due to the nature
         ///     of floating point operations and using System.Double internally.
+        ///     </para>
         /// </summary>
         /// <param name="other">The other quantity to compare to.</param>
-        /// <param name="tolerance">The comparison tolerance.</param>
+        /// <param name="tolerance">The absolute or relative tolerance value. Must be greater than 0.</param>
```
angularsen: greater or equal

UnitsNet/Scripts/Include-GenerateQuantitySourceCode.ps1
```diff
@@ -445,15 +445,25 @@ namespace UnitsNet
...
-        ///     Absolute tolerance is when the difference between the two quantities is not greater than exactly the tolerance value.
+        ///     </para>
+        ///     <para>
+        ///     Relative tolerance is defined as the maximum allowable absolute difference between this quantity's value and
+        ///     <paramref name="other"/> as a percentage of this quantity's value. For example, a relative tolerance
+        ///     of 0.01 means the difference must be within +/- 1% of this quantity's value to be considered equal.
+        ///     </para>
+        ///     <para>
+        ///     Absolute tolerance is defined as the maximum allowable absolute difference between this quantity's value and
+        ///     <paramref name="other"/> as a fixed number.
```
angularsen: Same, what unit?

UnitsNet/Comparison.cs
tmilnthorp: In this case it's just a double, so there is no unit. But when you're looking at the Equals on a quantity, the reference value is "this" so it's a fixed number of "this" unit.

```csharp
var a = Length.FromMeters( 100.0 );
var b = Length.FromInches( 5.0 );
a.Equals( b, 0.001 )
```

In this case it's absolute of a. So within 0.001 meters.

angularsen: Okay sure, that is one way to look at it. I will still argue that there is a unit in play here and it is the constructed unit of `a`, which is the most sensible choice by all means! However, I would have to read the current xmldoc more than once to understand what will actually happen, since there is conversions going on here and you are naturally not comparing the values 100 and 5 head on. If the xmldoc rather said the tolerance was in the `referenceValue`'s `Unit` then at least in my eyes it would be much more intuitive. I guess we just have different perspective on this and read it a bit differently.

So for me, a slight change in wording of xmldoc would make it a lot more intuitive.

tmilnthorp: I think changing the wording definitely makes sense regardless!

tmilnthorp:
Updating the documentation on the Equals method in quantity classes

UnitsNet/Scripts/Include-GenerateQuantitySourceCode.ps1
```diff
@@ -444,6 +444,57 @@ namespace UnitsNet
...
+        ///     </code>
+        ///     </example>
+        ///     </para>
+        ///     <para>
+        ///     Note that it is advised against specifying zero difference, due to the nature
+        ///     of floating point operations and using System.Double internally.
+        ///     </para>
+        /// </summary>
+        /// <param name="other">The other quantity to compare to.</param>
+        /// <param name="tolerance">The absolute or relative tolerance value. Must be greater than 0.</param>
```
angularsen: greater or equal to 0

angularsen: Also update `Comparison` with the same documentation updates.

tmilnthorp:
I will also obsolete Equals($quantityName other, $quantityName maxError) since we have a method that does relative and obsolete now. I'll also change the Equals(Object) implementation

angularsen:
Awesome! Looks good, nice touch on fixing the old equals methods too.
Nuget 3.101 on the way out.

PR #454, opened 2018-06-26
Changing Compare methods to compare current value to other in "this" …

tmilnthorp:
Changing Compare methods to compare current value to other in "this" units. This avoids scaling issues (and converting two quantities when we only need to potentially convert one) when both "this" and "other" are not defined in base units. "This" is the reference value.

### DIFF
--- a/UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
+++ b/UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
@@ -736,8 +736,8 @@ namespace UnitsNet
 
         public int CompareTo(object obj)
         {
-            if(obj is null) throw new ArgumentNullException("obj");
-            if(!(obj is Acceleration)) throw new ArgumentException("Expected type Acceleration.", "obj");
+            if(obj is null) throw new ArgumentNullException(nameof(obj));
+            if(!(obj is Acceleration)) throw new ArgumentException("Expected type Acceleration.", nameof(obj));
 
             return CompareTo((Acceleration)obj);
         }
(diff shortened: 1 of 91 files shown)


UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -736,9 +736,10 @@ public static string GetAbbreviation(AccelerationUnit unit)
 
         public int CompareTo(object obj)
         {
-            if (obj == null) throw new ArgumentNullException("obj");
-            if (!(obj is Acceleration)) throw new ArgumentException("Expected type Acceleration.", "obj");
-            return CompareTo((Acceleration) obj);
+            if(obj is null) throw new ArgumentNullException("obj");
+            if(!(obj is Acceleration)) throw new ArgumentException("Expected type Acceleration.", "obj");
```
angularsen: Let's add `nameof(obj)` too.

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -736,9 +736,10 @@ public static string GetAbbreviation(AccelerationUnit unit)
 
         public int CompareTo(object obj)
         {
-            if (obj == null) throw new ArgumentNullException("obj");
-            if (!(obj is Acceleration)) throw new ArgumentException("Expected type Acceleration.", "obj");
-            return CompareTo((Acceleration) obj);
+            if(obj is null) throw new ArgumentNullException("obj");
```
angularsen: Didn't know you could do `is null`, now I know!

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
tmilnthorp: Didn't know you could do this either 👍 

Done!

angularsen:
You have to sync again, just merged a PR.

tmilnthorp:
Done!

PR #455, opened 2018-06-26
Adding helper method to create quantity in different units

tmilnthorp:


tmilnthorp:
Oops. Need to do some more work here :)

PR #456, opened 2018-06-26
As conversions

tmilnthorp:
Doing As conversions in base type for accuracy sake. Only doing double conversion for As method.

Moving AsXXX methods to all be together in "Conversion" region.

I also renamed the method to convert to base units to make it consistent across all quantities. Potentially useful for future interface method. It also means the method signature won't change if a quantity's base unit is changed.


angularsen:
Nice! Good changes.

PR #458, opened 2018-06-27
Split wrc

tmilnthorp:
More work towards #409. Creating code generator script for .NET-only code. Using existing script to generate "common" code for both .NET and WRC. Removing #if WINDOWS_UWP from Value property in Common code and moving each to their separate .NET and WCR code generators.


tmilnthorp:
This is more baby steps FYI. There's a lot more code to pull into the separate implementations for .NET and WRC. It is already a lot of changed files so I didn't want to go too far immediately.

tmilnthorp:
Ultimately I'd like to move the scripts around to their appropriate folders/projects also. Possibly even move the common script out of the UnitsNet folder and into a "Common" folder (Common\Scripts).

I'd also like to move the generated code directories around a bit to split Common/.NET/WRC into their appropriate folders.

Common\GeneratedCode.
UnitsNet\GeneratedCode
WindowsRuntimeComponent\GeneratedCode

etc.

Trying to do this slowly since it will be large changes!

angularsen:
Consider naming it UnitsNet.Common to distinguish it from say json serialization code.

angularsen:
Awesome, looks good. The phone 📱 struggles with these huge diffs though 😁

PR #460, opened 2018-07-02
Moving code that is not valid in WRC from Common code to NetFramework…

tmilnthorp:
More work towards #409. I pulled out all code hidden behind the following preprocessor directive and moved it to the NetFramework code generator.

```csharp
#if !WINDOWS_UWP
```


UnitsNet/Scripts/Include-GenerateQuantitySourceCodeNetFramework.ps1
```diff
@@ -100,6 +100,110 @@ namespace UnitsNet
...
         public $baseType Value => _value;
+
+        #region Nullable From Methods
+
+"@; foreach ($unit in $units) {
+    $valueParamName = $unit.PluralName.ToLowerInvariant();@"
+        /// <summary>
+        ///     Get nullable $quantityName from nullable $($unit.PluralName).
+        /// </summary>
+        public static $($quantityName)? From$($unit.PluralName)($($quantityValueType)? $valueParamName)
```
angularsen: Outside this PR, but in light of binary size issues - nullable `From` methods add twice the method count for all our 800 units and I'm contemplating removing them from the lib in vNext. Thoughts?

It's only a convenience thing the way I see it, and this pattern is also a kind of weird pattern I've not seen anywhere else in .NET or third party libs. I would instead advocate checking null explicitly or using `Option<T>` or similar to box null value handling.

tmilnthorp: Generally I think that the consuming library would store the nullable type.

For example they would have an Acceleration? field. Then they either assign it somewhere, or not. I would generally not expect a null to come into a ?FromXXX() method. Rather the unit would just never be constructed.

I'm all for removing them.

PR #461, opened 2018-07-06
Split wrc

tmilnthorp:
More work towards #409. Moving some of the common code and unit definitions to top-level Common folder


angularsen:
Nice!

PR #463, opened 2018-07-12
Split WRC

tmilnthorp:
More work towards #409. Splitting ToString and Parse methods into the appropriate .NET and WRC code generators.


angularsen:
As mentioned on gitter, you need to push your version commit to `origin/master` first so that this PR doesn't introduce it.

PR #467, opened 2018-07-23
Use "kn" as default abbreviation for knots (ISO Standard)

gojanpaolo:
https://en.wikipedia.org/wiki/Knot_(unit)
> The ISO standard symbol for the knot is kn.

### DIFF
--- a/Common/UnitDefinitions/Speed.json
+++ b/Common/UnitDefinitions/Speed.json
@@ -198,7 +198,7 @@
       "Localization": [
         {
           "Culture": "en-US",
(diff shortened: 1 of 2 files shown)


angularsen:
I'll just merge it then @tmilnthorp, sounds good to me too. I don't consider this a breaking change, even if ToString() will change default abbreviation.

angularsen:
Nuget 104 on its way out.

tmilnthorp:
Unless you do a ToString followed by a Parse :)

angularsen:
Parse should still handle it, since we didn't remove any abbrevs - only changed the order of priority.

PR #480, opened 2018-09-21
No Undefined enum value for Units property

### LINKED ISSUE #437: Units property should not include Undefined value
https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/GeneratedCode/Quantities/Length.g.cs#L164

Similar for all quantities, Length only chosen as an example.

It doesn't make sense to include the `LengthUnit.Undefined = 0` value, since it is not meant to be used. UnitsNet will throw exception if you try to use this value for anything in methods accepting a unit value, as the intent is to help you catch bugs where you forgot to initialize the value such as when deserializing data.

bplubell:
See notes in the related issue (#437). Worth noting that this does not remove the Undefined value, but just does not expose it from each quantity classes Units property.

Resolves #437.

### DIFF
--- a/Common/GeneratedCode/Quantities/Acceleration.Common.g.cs
+++ b/Common/GeneratedCode/Quantities/Acceleration.Common.g.cs
@@ -101,7 +101,7 @@ namespace UnitsNet
 #if WINDOWS_UWP
         private
 #else
-        public 
+        public
 #endif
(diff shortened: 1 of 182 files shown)


angularsen:
Sweet! Great that you added a test for it too. 👍

Issue #343, opened 2017-12-14
"Length.TryParse" parses invalid values


nimishmistry:
``` Length.TryParse("2m,,,1", out length) ``` is parsed and the length is set to "2m". Is this intended?

I'm trying to use TryParse for validation.

angularsen:
Hi, thanks for reporting. This is not intentional, we try to be lenient about whitespace and combining multiple value/abbreviation combinations with different delimiters, but this is way beyond those usecases :-)
I'm sure it's a matter of tweaking the regex a little bit. Would you care to try to fix it? A pull request would be most welcome.

https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/CustomCode/QuantityParser.cs
https://github.com/angularsen/UnitsNet/blob/master/UnitsNet.Tests/CustomCode/ParseTests.cs

nimishmistry:
Thanks of the offer, will definitely attempt to fix this.

angularsen:
Awesome!

bplubell:
I took a look at this and found the problem is where the QuantityParser is trying to capture trailing, invalid data. It is only checking for alpha characters and, since comma is not an alpha character, it does not capture it.
https://github.com/angularsen/UnitsNet/blob/45c747db693d1f2e8a38651b7a37f695d7450d95/UnitsNet/CustomCode/QuantityParser.cs#L71

Trying to make this capture more characters (like comma) could be problematic with the current design where the string can have multiple sets of valid units (such as `"1m and 200mm"`). I think it would simplify things greatly and be more in keeping with framework parse methods if we did not allow multiple sets units. For example, `double.TryParse("1+2")` returns `false` throws an exception.

Proposal:
Fail when parsing with multiple sets of unit. The expectation would be that consumer code would pre-parse, since more complex maths than addition could be required and handled by the consumer.
- An additional library (Nuget package) could separate the parsing of complex string from the core handling of units (either part of this repo or one outside of UnitsNet)
- UnitNet should handle special cases like parsing FeetInches
- This would be a breaking change

Thoughts?

angularsen:
[...]

Issue #180, opened 2016-08-17
vNext: Wishlist for breaking changes


angularsen:
Although there are no immediate plans to do a vNext, this issue serves as a wishlist/reminder of what to consider changing in case we need to bump the major version in the future.

### To clarify
These are some things we should figure out before finalizing version 4.

A common denominator for design choices is binary size.
The library was nearing the 1MB mark and the majority of that is the large number of methods and properties for all our 700+ units. In particular number extension methods add 8 methods per unit (!). With some clever hacks we brought this back to 600kB without any breaking changes, but we want to keep binary size in mind for any new changes we introduce.

TODO Clean up and organize this brain dump:
- Numeric type for value representation: Keep `double`? Switch to `decimal`? Offer both? Separate nugets? What about `float`?
- Base types: To make it easier working with quantities generically, accessing value and unit, get units and their abbreviations, parsing and ToString() without know the quantity type in advance.
- `struct` vs `class`: Weigh all the pros and cons, performance (stack vs heap), inheritance vs interface
  - `class`: Can support all number types (float, double, decimal) without increasing binary size N times ([discussion](https://github.com/angularsen/UnitsNet/issues/285#issuecomment-351749507))
  - `struct`: More suitable for performance and memory constrained applications (what is the impact in practice?), avoids pressure on garbage collector, quantities match the semantics of being a value type

#### Removing things
- [x] https://github.com/anjdreas/UnitsNet/commit/7a455f0c12f5beed3790a7f5a1f6a65055f13c32: Remove default `Temperature` arithmetic, it is not correct (made this breaking change already, due to existing behavior not being correct).
- [ ] Remove `UnitClass` type, replaced by `QuantityType`
- [ ] Remove nullable `From` factory methods causing N additional methods for N units, see [comment](https://github.com/angularsen/UnitsNet/issues/372#issuecomment-402249801)
- [ ] Remove `==` `!=` and `Equals(object)` and `Equals<T>(T)` since they don't take a max error argument and is prone to floating point rounding issues, instead users should use `Equals()` methods that take a max error argument
- [ ] Remove code marked as `[Obsolete]`, such as `Volume.Teaspoon` unit that was too ambiguous
- [ ] Remove static methods on `UnitSystem`, should use `UnitSystem.Default` instead
- [ ] Remove unnecessary overloads on number types (8 overloads for each of the 700+ units) #372 

#### Changing behavior
- [ ] Throw exception on NaN values in unit class constructor methods (#176)
[...]

PR #481, opened 2018-09-22
Fix the use of '%' in ParseTests

McNeight:


angularsen:
Hi, thanks for the PR. Could you please refresh my memory on what this test actually tests and why you wanted to test for `%%` instead of `%`? I'm honestly puzzled by the purpose of this unit test, as it seems we want to allow parsing say `10 m!s` as 10 meters. I think it was part of supporting parsing things like `1' 2"` as feet/inches, but I don't think it makes much sense for `Length` and most other quantities.

This issue explains how the parsing is currently way too lenient and needs to be tightened a bit.
https://github.com/angularsen/UnitsNet/issues/343

This comment explains how we want to fix this, but it is a breaking change and can only be added in a major version bump.
https://github.com/angularsen/UnitsNet/issues/180#issuecomment-357100824

angularsen:
@McNeight Just a follow up, I'm waiting for your feedback on how to proceed here.

angularsen:
Just a heads up that https://github.com/angularsen/UnitsNet/pull/487 will re-implement the parse methods to be much more strict and only allow the format `"1 m"` and possibly `"1m"`.
I propose we close this issue as it will be no longer relevant after that change is made. Please reopen the issue if there is more to discuss here.

PR #483, opened 2018-09-24
Removing nullable methods from quantity classes as well as extension …

tmilnthorp:
Removing nullable methods from quantity classes as well as extension methods.

Work towards reducing library size #372


UnitsNet/GeneratedCode/Extensions/Number/NumberToAmountOfSubstanceExtensions.g.cs
```diff
@@ -49,475 +49,265 @@ public static class NumberToAmountOfSubstanceExtensions
...
-        public static AmountOfSubstance? Millimoles(this double? value) => AmountOfSubstance.FromMillimoles(value);
-
         /// <inheritdoc cref="AmountOfSubstance.FromMillimoles(UnitsNet.QuantityValue)" />
         public static AmountOfSubstance Millimoles(this float value) => AmountOfSubstance.FromMillimoles(value);
 
-        /// <inheritdoc cref="AmountOfSubstance.FromMillimoles(UnitsNet.QuantityValue)" />
-        public static AmountOfSubstance? Millimoles(this float? value) => AmountOfSubstance.FromMillimoles(value);
-
         /// <inheritdoc cref="AmountOfSubstance.FromMillimoles(UnitsNet.QuantityValue)" />
         public static AmountOfSubstance Millimoles(this decimal value) => AmountOfSubstance.FromMillimoles(Convert.ToDouble(value));
```
angularsen: I think we can remove decimal too, since the main scenarios I can think of are:
```
1.Meters();
1.5.Meters();
1.5m.Meters(); // For decimal, probably not so common
decimal myVal = 1.5m;
myVal.Meters(); // Also I'm guessing not common, I would not use it this way at least
```

UnitsNet/GeneratedCode/Extensions/Number/NumberToAmplitudeRatioExtensions.g.cs
```diff
@@ -49,135 +49,75 @@ public static class NumberToAmplitudeRatioExtensions
         /// <inheritdoc cref="AmplitudeRatio.FromDecibelMicrovolts(UnitsNet.QuantityValue)" />
         public static AmplitudeRatio DecibelMicrovolts(this int value) => AmplitudeRatio.FromDecibelMicrovolts(value);
 
-        /// <inheritdoc cref="AmplitudeRatio.FromDecibelMicrovolts(UnitsNet.QuantityValue)" />
-        public static AmplitudeRatio? DecibelMicrovolts(this int? value) => AmplitudeRatio.FromDecibelMicrovolts(value);
-
         /// <inheritdoc cref="AmplitudeRatio.FromDecibelMicrovolts(UnitsNet.QuantityValue)" />
         public static AmplitudeRatio DecibelMicrovolts(this long value) => AmplitudeRatio.FromDecibelMicrovolts(value);
```
angularsen: Can remove `long` too.

UnitsNet/GeneratedCode/Extensions/Number/NumberToAmplitudeRatioExtensions.g.cs
```diff
@@ -49,135 +49,75 @@ public static class NumberToAmplitudeRatioExtensions
...
-        public static AmplitudeRatio? DecibelMicrovolts(this long? value) => AmplitudeRatio.FromDecibelMicrovolts(value);
-
         /// <inheritdoc cref="AmplitudeRatio.FromDecibelMicrovolts(UnitsNet.QuantityValue)" />
         public static AmplitudeRatio DecibelMicrovolts(this double value) => AmplitudeRatio.FromDecibelMicrovolts(value);
 
-        /// <inheritdoc cref="AmplitudeRatio.FromDecibelMicrovolts(UnitsNet.QuantityValue)" />
-        public static AmplitudeRatio? DecibelMicrovolts(this double? value) => AmplitudeRatio.FromDecibelMicrovolts(value);
-
         /// <inheritdoc cref="AmplitudeRatio.FromDecibelMicrovolts(UnitsNet.QuantityValue)" />
         public static AmplitudeRatio DecibelMicrovolts(this float value) => AmplitudeRatio.FromDecibelMicrovolts(value);
```
angularsen: And `float`.

angularsen:
Good call about removing nullable From methods, twice the method count for a minor syntax convenience.

tmilnthorp:
Did one better... Rather than potentially breaking existing code, I changed the extension methods to be generic. This should allow _any_ type that can be called with Convert.ToDouble to be used. This actually expands the available use cases while also reducing the size of the dll by another 150 KB!

angularsen:
Awesome! I'm so puzzled, I was so sure I had already tried the generics route and it failed, but a quick prototype just now confirms it works. Very nice!

Soo, we have two ways to go about it then. Add back the nullable From methods and merge this into master now, or merge this into a v4 kind of branch and start getting a breaking change release going before merging any of it back to master.

PR #486, opened 2018-09-27
Obsolete nullable methods

tmilnthorp:
If we're going to remove nullable support per #483, we better mark them as obsolete now :)

### DIFF
--- a/UnitsNet/Comparison.cs
+++ b/UnitsNet/Comparison.cs
@@ -44,7 +44,7 @@ namespace UnitsNet
         ///     Absolute tolerance is defined as the maximum allowable absolute difference between <paramref name="referenceValue"/> and
         ///     <paramref name="otherValue"/> as a fixed number.
         ///     <example>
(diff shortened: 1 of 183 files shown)


UnitsNet/Comparison.cs
```diff
@@ -102,10 +102,10 @@ public static bool EqualsRelative(double referenceValue, double otherValue, doub
         /// <summary>
         ///     Checks if two values are equal with a given absolute tolerance.
         ///     <para>
-        ///     Absolute tolerance is defined as the maximum allowable absolute difference between <paramref name="referenceValue"/> and
-        ///     <paramref name="otherValue"/> as a fixed number.
+        ///     Absolute tolerance is defined as the maximum allowable absolute difference between <paramref name="value1"/> and
+        ///     <paramref name="value2"/> as a fixed number.
         ///     <example>
-        ///     In this example, the two values will be equal if abs(<paramref name="referenceValue"/> - <paramref name="otherValue"/>) <= 0.01
+        ///     In this example, the two values will be equal if abs(<paramref name="value1"/> - <paramref name="value2"/>) &lt;= 0.01
```
angularsen: Nice!

Issue #176, opened 2016-08-02
Support NaN


StevenLiekens:
Please add `NaN` properties for every unit.

For example
- `Energy.NaN`
- `Speed.NaN`

Also add `IsNaN(unit)`
- `Energy.IsNaN(Energy.NaN)`: true
- `Energy.IsNaN(Energy.MaxValue)`: false

Thanks!

anjdreas:
We could, but I question the value of adding this.

`double.IsNaN(Length.FromMeters(double.NaN).Centimeters)` returns `true`, as expected.
Is this a matter of syntactic convenience?

I also notice that `TimeSpan.FromSeconds(double.NaN)` throws `ArgumentException`
```
TimeSpan does not accept floating point Not-a-Number values
```

Not sure what is the best practice here. Personally I might prefer throwing exception, not a fan of NaN bugs making it harder to debug.

StevenLiekens:
> Is this a matter of syntactic convenience?

Yes. :)

> I also notice that TimeSpan.FromSeconds(double.NaN) throws ArgumentException

I guess because `TimeSpan` stores its state as Int64 (number of ticks), and there is no `Int64.NaN`.

anjdreas:
@eriove @bretthysuik @maherkassim @tongbong 

You are one of the top contributors, any input on NaN values?
a) Allow NaN values, can check `double.IsNaN(myLength.Meters)` (current behavior)
b) Allow NaN values, add `IsNaN` getter as syntactic convenience (suggested by @StevenLiekens)
c) Throw exception on trying to construct with NaN value, similar to how TimeSpan works (suggested by me, breaking change)

bretthysuik:
My vote would be c) and to not re-implement .NET BCL NaN detection functionality in this library. `double.IsNaN(myLength.Meters)` is simple and clean enough to me without having to expand the API of this library.

eriove:
Sorry for the late reply. Read the e-mail in my phone and forgot to reply.

c) is my favorite, since it catches errors early and NaN is usually an error when it comes to physical measures. Since it breaks backwards compatibility my second choice would be b. But I can't say I really care that much when choosing between a and b.

anjdreas:
It seems you both share my opinion on generally not facilitating the use of NaN. I'm adding a future reminder in #180 to consider option c) if we are going to bump the version later.

[...]

Issue #478, opened 2018-09-14
Suggestion: define conversions as rational values


R2D221:
I recently tried to do a simple conversion between US gallons and oil barrels. The conversion is simple, you just need to multiply or divide by 42. However, with this library, the result is the following:

    Volume.FromOilBarrels(1).UsGallons == 42.0000197938929

I have already read the [Precision](https://github.com/angularsen/UnitsNet/wiki/Precision) page, so I understand how the conversions are made and how it's not the first priority. However, I want to post this as a suggestion.

Instead of defining the units as a double, they could be defined as rational.

How would this work: Each conversion should have the numerator and denominator. This leaves things almost unchanged for e.g. meters to kilometers. But for conversions between metric and imperial systems, it would look as follows:

<a href="https://www.codecogs.com/eqnedit.php?latex=1&space;\textup{gal}&space;=&space;231&space;\textup{in}^3&space;=&space;231&space;*&space;\frac{254^3}{10000^3}&space;\textup{m}^3&space;=&space;\frac{473176473}{125000000000}\textup{m}^3" target="_blank"><img src="https://latex.codecogs.com/gif.latex?1&space;\textup{gal}&space;=&space;231&space;\textup{in}^3&space;=&space;231&space;*&space;\frac{254^3}{10000^3}&space;\textup{m}^3&space;=&space;\frac{473176473}{125000000000}\textup{m}^3" title="1 \textup{gal} = 231 \textup{in}^3 = 231 * \frac{254^3}{10000^3} \textup{m}^3 = \frac{473176473}{125000000000}\textup{m}^3" /></a>

<a href="https://www.codecogs.com/eqnedit.php?latex=1&space;\textup{bbl}&space;=&space;42&space;\textup{gal}&space;=&space;42*\frac{473176473}{125000000000}\textup{m}^3&space;=&space;\frac{9936705933}{62500000000}\textup{m}^3" target="_blank"><img src="https://latex.codecogs.com/gif.latex?1&space;\textup{bbl}&space;=&space;42&space;\textup{gal}&space;=&space;42*\frac{473176473}{125000000000}\textup{m}^3&space;=&space;\frac{9936705933}{62500000000}\textup{m}^3" title="1 \textup{bbl} = 42 \textup{gal} = 42*\frac{473176473}{125000000000}\textup{m}^3 = \frac{9936705933}{62500000000}\textup{m}^3" /></a>

[...]

PR #487, opened 2018-09-27
v4.0.0 (Work In Progress)

angularsen:
# 4.0.0 Release - DO NOT MERGE YET
This PR will serve as the list of items to complete and it will be updated to show the progress before finally merged into `master` when completed. We all have busy schedules, so **if you want to help move this work forward then that is much appreciated!** 

## The main theme is to reduce binary size
In two years it has grown from 280 kB to 1.4 MB and a lot of it is due to unnecessary syntactic sugar with many method overloads for various number types and nullable types - for every of our 800+ units! It simply adds up to a big total.

These items are chosen from #180, trying to include as many of the low hanging fruits as possible, while still keeping the list short and realistic to complete in a reasonably short time - we are all busy in our daily lives. We can always have more major version bumps later than trying to perfect it all now.

## Feature complete before November, merged before mid-December
I would like to aim for a "beta" pre-release nuget sometime in October with all the items completed, so we can have some time to test it before releasing the final, stable 4.0.0 version before Christmas holidays.
We should have the work items list more or less final before Monday, October 8th.

#### Removing things
- [ ] Remove code marked as `[Obsolete]`, such as `Volume.Teaspoon` unit that was too ambiguous
- [ ] Remove `UnitClass` type, replaced by `QuantityType`
- [ ] Remove nullable `From` factory methods
- [ ] Remove static methods on `UnitSystem`, should use `UnitSystem.Default` instead
- [ ] Remove extension methods on _nullable_ number types
- [ ] Move remaining extension methods on number types to a new `UnitsNet.NumberExtensions` nuget (OUT OF SCOPE?)
- [ ] Remove `==` `!=` and `Equals(object)` and `Equals<T>(T)` since they don't take a max error argument and is prone to floating point rounding issues, instead users should use `Equals()` methods that take a max error argument

#### Changing behavior
- [ ] Throw exception on NaN values in static constructor methods, like `FromMeters()` (https://github.com/angularsen/UnitsNet/issues/176#issuecomment-240434320)
- [ ] Throw if unit was not specified when constructing quantities, (https://github.com/angularsen/UnitsNet/pull/389#issuecomment-362811310)
- [ ] Stricter parsing (#343 and https://github.com/angularsen/UnitsNet/issues/180#issuecomment-357100824) (I THINK THIS ONE MAY BE OUT OF SCOPE?)

#### Renaming
- [ ] Correct SingularName for some Flow unit definitions (see #360)

#### Fixing
- [ ] Search for any TODO comments in the code to address, remove comment and either fix or create issue

## Milestones
- [ ] Finalize work items list (Monday, October 7)
- [ ] Release 4.0.0-beta1, feature complete (Wednesday, October 31)
- [ ] Release 4.0.0 (Monday, December 17)

### DIFF
--- a/UnitsNet.Serialization.JsonNet/UnitsNet.Serialization.JsonNet.Common.props
+++ b/UnitsNet.Serialization.JsonNet/UnitsNet.Serialization.JsonNet.Common.props
@@ -2,7 +2,7 @@
   <!-- NuGet properties -->
   <PropertyGroup>
     <PackageId>UnitsNet.Serialization.JsonNet</PackageId>
-    <Version>1.3.0</Version>
+    <Version>4.0.0-alpha1</Version>
     <Authors>Andreas Gullberg Larsen</Authors>
     <Title>Units.NET Serialization with Json.NET</Title>
     <Description>A helper library for serializing and deserializing types in Units.NET using Json.NET.</Description>
--- a/UnitsNet.WindowsRuntimeComponent/UnitsNet.WindowsRuntimeComponent.nuspec
+++ b/UnitsNet.WindowsRuntimeComponent/UnitsNet.WindowsRuntimeComponent.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
   <metadata>
     <id>UnitsNet.WindowsRuntimeComponent</id>
-    <version>3.105.0</version>
+    <version>4.0.0-alpha1</version>
     <title>Units.NET - Windows Runtime Component</title>
     <authors>Andreas Gullberg Larsen</authors>
     <owners>Andreas Gullberg Larsen</owners>
--- a/UnitsNet/Properties/AssemblyInfo.WindowsRuntimeComponent.cs
+++ b/UnitsNet/Properties/AssemblyInfo.WindowsRuntimeComponent.cs
@@ -38,6 +38,6 @@ using System.Runtime.InteropServices;
 [assembly: AssemblyTrademark("")]
 [assembly: AssemblyCulture("")]
 [assembly: NeutralResourcesLanguage("")]
-[assembly: AssemblyVersion("3.105.0")]
-[assembly: AssemblyFileVersion("3.105.0")]
+[assembly: AssemblyVersion("4.0.0")]
+[assembly: AssemblyFileVersion("4.0.0")]
 [assembly: InternalsVisibleTo("UnitsNet.WindowsRuntimeComponent.Tests")]
--- a/UnitsNet/UnitsNet.Common.props
+++ b/UnitsNet/UnitsNet.Common.props
@@ -2,7 +2,7 @@
   <!-- NuGet properties -->
   <PropertyGroup>
     <PackageId>UnitsNet</PackageId>
-    <Version>3.105.0</Version>
+    <Version>4.0.0-alpha1</Version>
     <Authors>Andreas Gullberg Larsen</Authors>
     <Title>Units.NET</Title>
     <Description>Get all the common units of measurement and the conversions between them. It is light-weight and thoroughly tested.</Description>


angularsen:
Down from 1.4MB to 583 KB as of now 🎉

angularsen:
@tmilnthorp In order to deploy nuget:
- Run the "bump suffix" .bat files in `/Build` folder for UnitsNet and Json lib
- Push to `v4`
- Merge `v4` into `release/v4` branch and push that

The reason for this second branch is that AppVeyor has config to avoid duplicate builds on PRs, by skipping branches that have open PRs and PR builds are special it seems - at least I couldn't get it to deploy on a few attempts.

angularsen:
Added work item
- Remove all number extension methods (#497)

angularsen:
Add work item: Update README with v4 changes #498

angularsen:
Since we've made such great progress, I'd really love for this one to in the work list:
- Stricter parsing (#343 and #180 (comment)) (OUT OF SCOPE?)

The number extensions nuget, however, I'm don't want to follow bring in for this release, but rather see if anyone actually complains about those methods being missing.

angularsen:
Added these:

Split/rename UnitSystem into UnitParser, GlobalConfiguration, UnitAbbreviationsCache
Do not try/catch in UnitConverter.Try-methods (#506)
Do not try/catch in Length.TryParse() and for other quantities

The split/rename idea is as follows:
- UnitParser (parses abbreviations)
- GlobalConfiguration (add/remove abbreviations dynamically, set default culture to use for ToString()
- UnitAbbreviationsCache (lookup lists per culture)

angularsen:
List of changes are now final and progress has been amazing! 
We should be feature complete within a week or two.

angularsen:
Added new item based on recent discussions.
- Change Temperature arithmetic back to use base unit Kelvin (see #518)

angularsen:
Added point
- [x] Remove unit parameter from `ToString()` methods (#546)

angularsen:
@eriove @gojanpaolo @ferittuncer @JKSnd Any input on these v4 changes before we lock it down would be greatly appreciated. I want to feature freeze and get a beta nuget out in the next couple of days. We can always do a v5 later, but more eyeballs may help avoid silly mistakes or forgetting something obvious.

gojanpaolo:
@angularsen i apologize for the late response. I tried updating our project to v4 and the first thing that popped up was `Length2d`. We do use it for coordinates. I can just write a custom class on our project if putting it back will be a hassle.

gojanpaolo:
Equality fails if `Quantity` in a containing `struct` with different `Unit` but same physical value.

Below code passes in v3. Might be a consequence of removing `!=` and `==` operators.

```csharp
public class Tests
{
    [Fact]
    public void LengthEquality()
    {
        Assert.Equal(
            new Parent { Length = Length.FromMeters(3) },
            new Parent { Length = Length.FromCentimeters(300) });
    }
}
public struct Parent
{
    public Length Length { get; set; }
}
```

tmilnthorp:
Without overriding the Equals methods, structs do a bitwise comparison, so it makes sense this fails now.

In thinking about it more, I think we should put the Equals methods and operators back in. Just because it may have some mathematical error doesn't mean it isn't a valid comparison. .NET certainly allows you to check two doubles for exact equality even if it's sometimes error prone. Checking should be up to the programmer if it applies or not. 

One always-valid example: 0 m is == 0 cm or 0 in or 0 any length!

Also if we ever move to a generic quantity class, you could have int or decimal quantities with conversion ratios that are exact (per #478): i.e. 12.0m inches is exactly == 1.0m feet.

angularsen:
Added point 
- Adding equality methods/operators back (#558)

angularsen:
@gojanpaolo Awesome that you tested this, thanks!

> @angularsen i apologize for the late response. I tried updating our project to v4.0.0-alpha5 and the first thing that popped up was `Length2d`. We do use it for coordinates. I can just write a custom class on our project if removing it is final or putting it back will be a hassle.

I understand, I think I am inclined to tell you and other consumers that you should create this type yourself.
It was largely replaced by `Area` in UnitsNet, and we don't have any geometric or geographic usecases in the library to warrant a Length2d right now. Maybe later in a separate package?

> Equality fails if Quantity in a containing struct with different Unit but same physical value.

Good catch, we merged a fix for it just now in #558 .

angularsen:
Added 3.x to 4.x upgrade guide.
https://github.com/angularsen/UnitsNet/wiki/Upgrading-from-3.x-to-4.x

gojanpaolo:
> Added 3.x to 4.x upgrade guide.
> https://github.com/angularsen/UnitsNet/wiki/Upgrading-from-3.x-to-4.x

I added [how to handle "Do not allow NaN or Infinity"](https://github.com/angularsen/UnitsNet/wiki/Upgrading-from-3.x-to-4.x#do-not-allow-nan-or-infinity). That's at least how I'm planning to do it on our project. :)

gojanpaolo:
Everything seems to be fine on our end after upgrading to v4. Got a few failing tests, but I believe these are all due to a different third party library not being able to handle `null` properly (was using `NaN` before).
:)

image

angularsen:
Added point
- Show release notes and upgrade guide when upgrading to 4.x nuget, if possible

angularsen:

Per discussion in:
https://github.com/angularsen/UnitsNet/pull/560#issuecomment-439413308

We can't justify this change just yet, we need to know if the majority of our userbase expects this new Temperature arithmetic or the old one.

angularsen:
4.0.0-beta2 released.

PR #488, opened 2018-09-27
Changing target frameworks to .NET standard 2.0 and .NET 4.0

### LINKED ISSUE #477: Add netstandard 2.0 target
To avoid all the dependencies that netstandard 1.0 brings with it.

We should also consider removing 1.0, since that no longer brings much value to the table I think, except for older versions of Xamarin and similar. 

People on NET 4.0-4.6.1 will simply use 4.0 dll and 4.6.2 and up will then use netstandard 2.0 dll.

Inspired by this tweet:
https://twitter.com/onovotny/status/1040053354465030144?s=19

tmilnthorp:
Fixes #477

### DIFF
--- a/UnitsNet/UnitsNet.Common.props
+++ b/UnitsNet/UnitsNet.Common.props
@@ -20,10 +20,8 @@
 
   <!-- Assembly and msbuild properties -->
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;net35;net40</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net40</TargetFrameworks>
     <RootNamespace>UnitsNet</RootNamespace>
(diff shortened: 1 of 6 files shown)


angularsen:
If you weren't aware, use `Fixes #477` syntax in PR description and it will automatically resolve the issue when merged :-)

UnitsNet/UnitsNet.Common.props
```diff
@@ -20,10 +20,8 @@
 
   <!-- Assembly and msbuild properties -->
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;net35;net40</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net40</TargetFrameworks>
     <RootNamespace>UnitsNet</RootNamespace>
-    <!-- Workaround for building with dotnet CLI and targeting .NET 3.5: https://github.com/Microsoft/msbuild/issues/1333#issuecomment-296346352 -->
```
angularsen: Ahh, what a relief seeing this go!

PR #490, opened 2018-09-27
Removing obsolete quantity units

tmilnthorp:


Common/GeneratedCode/Quantities/Duration.Common.g.cs
```diff
@@ -302,20 +290,6 @@ public static Duration FromMinutes(QuantityValue minutes)
...
         }
 
-        /// <summary>
-        ///     Get Duration from Months.
-        /// </summary>
-#if WINDOWS_UWP
-        [Windows.Foundation.Metadata.DefaultOverload]
-        public static Duration FromMonths(double months)
-#else
-        public static Duration FromMonths(QuantityValue months)
```
tmilnthorp: Interesting that this was marked obsolete in the .json file, but never made it here...

tmilnthorp: Surprised we've never noticed this before! Bug in the code generation. I've added #491 to be merged into master to address this.

PR #494, opened 2018-09-28
Fix SingularName for some units in VolumeFlow

tmilnthorp:
Fixing plural names placed in SingularName property (#360)


tmilnthorp:
I was a little unsure if MillionUsGallonsPerDay should remain as-is since Million already explicitly defines it as a plurality. "1 million us gallons per day" seems like it's correct.

angularsen:
I agree, original naming is good. I think both works, but the latter is on the form "2 millions **of** gallons" vs "2 million gallons".

PR #496, opened 2018-09-28
Remove UnitSystem static methods

angularsen:
Remove static where a similar instance method also exists.
Make `IsFallbackCulture` private, can't see how that is used externally.
Add some xmldoc while at it too.


UnitsNet/CustomCode/UnitSystem.cs
```diff
@@ -528,12 +528,11 @@ private void LoadDefaultAbbreviations([NotNull] IFormatProvider culture)
                 foreach (CulturesForEnumValue ev in localization.EnumValues)
                 {
                     int unitEnumValue = ev.Value;
-                    var usCulture = new CultureInfo("en-US");
```
angularsen: Nice

PR #497, opened 2018-09-28
Remove all number extension methods

angularsen:
Instead of waiting for us to get around to create a new nuget for this, let's remove it now that we have a breaking change release and rather add a new nuget if enough people actually ask for it.

Then we can have them do the work. \*evil laugh\*

But seriously, I suspect this is very little used and it adds significantly to binary size. Most people are well off creating a handful of extension methods themselves, or even copying source code from the repo history.


tmilnthorp:
Sounds good to me. I suspect they are used infrequently also. However even if anyone asks, I think scaling concerns with more units outweighs the additional API ability. It's not like we're removing functionality entirely, just removing one API in favor of another that supports the same.

PR #499, opened 2018-10-01
Adding ArgumentException to quantity if constructed with Undefned val…

tmilnthorp:
…ue. Add test for that behavior. A little script cleanup.


UnitsNet/Scripts/Include-GenerateQuantitySourceCodeCommon.ps1
```diff
@@ -19,25 +17,6 @@
     $baseDimensionAmountOfSubstance = if($baseDimensions.N){$baseDimensions.N} else{0};
     $baseDimensionLuminousIntensity = if($baseDimensions.J){$baseDimensions.J} else{0};
 
-    $convertToBaseType = switch ($baseType) {
```
angularsen: Good catch!

PR #505, opened 2018-10-01
Remove extension methods

tmilnthorp:


Common/GeneratedCode/Quantities/AmountOfSubstance.Common.g.cs
```diff
@@ -9,7 +9,6 @@
 //     See https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit for how to add or edit units.
 //
 //     Add CustomCode\Quantities\MyQuantity.extra.cs files to add code to generated quantities.
-//     Add Extensions\MyQuantityExtensions.cs to decorate quantities with new behavior.
```
angularsen: Good catch

UnitsNet/CustomCode/Quantities/AmplitudeRatio.extra.cs
```diff
@@ -61,6 +61,36 @@ public partial struct AmplitudeRatio
...
+        /// </summary>
+        /// <remarks>
+        ///     Provides a nicer syntax for converting an amplitude ratio back to a voltage.
+        ///     <example>
+        ///         <c>var voltage = voltageRatio.ToElectricPotential();</c>
+        ///     </example>
+        /// </remarks>
+        public ElectricPotential ToElectricPotential()
+        {
+            return AmplitudeRatio.ToElectricPotential(this);
```
angularsen: What about removing the static methods instead? I can't see how they add any value, since you need to provide it an instance anyway.

angularsen: Right, but removing those methods is a slightly different discussion I think, since DateTime/TimeStamp also has both a constructor and static factory methods. I think this pattern is fairly common in .NET, but it's extreme for us since we have 800 factory methods. 

A direct analog would be to have both `Length.AsCentimeters(myLength)` AND `myLength.Centimeters`, and clearly that is redundant in the same way having both static and instance methods for `ToElectricPotential()` is redundant.

So, let's remove the redundant static quantity conversion methods in this PR, then continue the conversation about removing static factory methods for quantities separately. My main gripe is losing a very familiar and dear API that has shorter syntax, but I absolutely see how it would slim the library a great deal more and perhaps offer a more consistent API on how to construct quantities. If it weren't for the binary size though, I would probably not want to remove them.

tmilnthorp: How should we handle methods that can't be handled via a constructor (same type for arguments)? For example:
```csharp
        public static Area FromCircleDiameter(Length diameter)
        {
            return System.Math.PI * diameter * diameter / 4;
        }

        public static Area FromCircleRadius(Length radius)
        {
            return System.Math.PI * radius * radius;
        }
```

Would you prefer ToAreaFromDiameter() and ToAreaFromRadius() instance methods on Length, or just leave these as is? They align well with the Area.FromSquareMeters methods that are generated on the quantity, so I'm tempted to leave these as is.

angularsen: I think these are fine as is too.

UnitsNet/CustomCode/Quantities/Area.extra.cs
```diff
@@ -47,16 +64,6 @@ public partial struct Area
         {
             return VolumeFlow.FromCubicMetersPerSecond(area.SquareMeters * speed.MetersPerSecond);
         }
-
-        public static Area FromCircleDiameter(Length diameter)
```
tmilnthorp: Noticed that this was in a #!if WINDOWS_UWP block. No reason to bar them from WRC clients. Had to change the logic to avoid operator overloads. It's a bit more explicit now too.

UnitsNet/CustomCode/Quantities/Area.extra.cs
```diff
@@ -31,6 +33,21 @@ public sealed partial class Area
     public partial struct Area
 #endif
     {
+        #region Static Methods
+
+        public static Area FromCircleDiameter(Length diameter)
+        {
+            var radiusInMeters = diameter.Meters / 2.0;
```
angularsen: Minor, but you can do `var radius = diameter / 2;`.

UnitsNet/CustomCode/Quantities/Area.extra.cs
angularsen: Nice find.

UnitsNet/CustomCode/Quantities/Area.extra.cs
angularsen: Fixed it.

PR #511, opened 2018-10-11
Splitting UnitSystem class into more focused classes

tmilnthorp:


tmilnthorp:
Ah. Found it. Some tests switch the DefaultCulture to CultureInfo.InvariantCulture and then do a UnitAbbreviationsCache.MapUnitToAbbreviation( LengthUnit.Meter, "m_s" ) for example.

This makes "m_s" the only abbreviation for CultureInfo.InvariantCulture so ToString and Parse expect that when tests use the InvariantCulture. The UnitAbbreviationsCache is now globally persistent so it doesn't go away. I think that actually makes sense but what do you think?

angularsen:
Yes, testing static types is always fun. I believe we can instruct xunit to run certain groups of tests in a separate appdomain than other tests - and if I recall correctly we already did such an effort back when either writing the tests or switching from nunit to xunit. 

Custom Test Collections
https://xunit.github.io/docs/running-tests-in-parallel.html#parallelism-in-test-frameworks

But, why is this a problem now and not before? Wasn't UnitSystem using static cache per cultures already?

tmilnthorp:
When in doubt, make them instance classes with a singleton pattern for maximum flexibility :smiley:

tmilnthorp:
Ok, I changed the tests back to match the previous behavior. I found the real issue -it was a multi-threading problem! Only saw it sometimes, especially in debug. I had to add [Collection(nameof(UnitSystemFixture))] to ParseTests, otherwise other tests were modifying GlobalConfiguration.DefaultCulutre at the same time the default parser was accessing it. Everything matches the old behavior and passes now 👍

UnitsNet.Tests/UnitSystemTests.cs
```diff
@@ -72,12 +72,14 @@ private enum CustomUnit
             IEnumerable<TUnitType> unitValues)
             where TUnitType : Enum
         {
+            var culture = GetCulture(cultureName);
```
angularsen: We should rename/split this test class too.

UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
```diff
@@ -45,7 +45,7 @@ internal string[] GetAllAbbreviations()
             return unitToAbbreviationMap.Values.SelectMany((abbreviations) =>
             {
                 return abbreviations;
-            } ).ToArray();
+            } ).Distinct().ToArray();
```
angularsen: Ah nice, we do indeed have duplicates in certain quantities.

angularsen:
```
[C:\projects\unitsnet\UnitsNet.WindowsRuntimeComponent\UnitsNet.WindowsRuntimeComponent.csproj]
GeneratedCode\Quantities\ElectricCharge.WindowsRuntimeComponent.g.cs(95,129): warning CS1574: XML comment has cref attribute 'UnitSystem' that could not be resolved 
```

A lot of spam on this one, maybe fix that first?
https://ci.appveyor.com/project/angularsen/unitsnet/builds/19466282

UnitsNet.Tests/UnitSystemTests.cs
tmilnthorp: 👍 done!

tmilnthorp:
Ok. Done. I do have a quick question though: any reason why GetDefaultAbbreviation was chosen to return  "(no abbreviation for ... when one doesn't exist?

I would expect an exception, or at least returning string.Empty to indicate failure. Otherwise you have to check if the string starts with "(no abbreviation for " for a failure. 

Perhaps throw an exception and add a bool TryGetDefaultAbbreviation method? Or would you prefer returning an empty string?

angularsen:
On the phone 📱. I don't recall, sounds reasonable to throw I think. It would only happen for json not defining abbreviations (do we have any like that?) I think. Even when adding units at runtime you do that by adding at least one abbrev.

tmilnthorp:
I think it would indeed be an exceptional case. I will add that.

UnitsNet.Tests/CustomCode/ParseTests.cs
```diff
@@ -93,16 +92,15 @@ public void ParseWithCultureUsingSpaceAsThousandSeparators(string s, double expe
...
-        public void ParseWithCultureUsingSpaceAsThousandSeparators_ThrowsExceptionOnInvalidString(string s, string expected)
+        public void ParseWithCultureUsingSpaceAsThousandSeparators_ThrowsExceptionOnInvalidString(string s, Type expectedExceptionType)
         {
             var numberFormat = (NumberFormatInfo) CultureInfo.InvariantCulture.NumberFormat.Clone();
             numberFormat.NumberGroupSeparator = " ";
             numberFormat.NumberDecimalSeparator = ".";
 
-            string actual = AssertExceptionAndGetFullTypeName(() => Length.Parse(s, numberFormat));
-            Assert.Equal(expected, actual);
+            Assert.Throws(expectedExceptionType, () => Length.Parse(s, numberFormat));
```
angularsen: Much better!

UnitsNet.Tests/GeneratedCode/AccelerationTestsBase.g.cs
```diff
@@ -327,5 +327,18 @@ public void UnitsDoesNotContainUndefined()
         {
             Assert.DoesNotContain(AccelerationUnit.Undefined, Acceleration.Units);
         }
+
+        [Fact]
+        public void AllUnitsHaveAtLeastOneAbbreviationSpecified()
```
angularsen: Good one

UnitsNet.Tests/GeneratedCode/AccelerationTestsBase.g.cs
```diff
@@ -327,5 +327,18 @@ public void UnitsDoesNotContainUndefined()
...
+        {
+            var units = Enum.GetValues(typeof(AccelerationUnit)).Cast<AccelerationUnit>();
+            foreach(var unit in units)
+            {
+                if(unit == AccelerationUnit.Undefined)
+                    continue;
+
+                var defaultAbbreviation = UnitAbbreviationsCache.Default.GetDefaultAbbreviation(unit);
```
angularsen: Just a thought, would `Default` be better named `SystemDefault` or `CurrentCulture` or something?
It's heavily based on `CultureInfo.CurrentCulture` and `CultureInfo.CurrentUICulture` behind the scenes.

angularsen: Also, it would be helpful to add a comment that we are testing that this does not throw exceptions. It might not be obvious to the reader. Removing the `var defaultAbbreviation` variable would also help highlight this.

angularsen:
I'm merging this now, so I can work on top of it in my strict parsing PR.
Feel free to address my comments in a separate PR.

PR #516, opened 2018-10-13
Stricter parsing

angularsen:
This PR grew a bit bigger than I wanted to, let me know if you want me to split it up a bit more to make it easier to review.

- Only parse the format `"{value} {unit abbreviation}"`, spacing is optional
- Add `Length.TryParseFeetInches()` and `ParseFeetInches()` to parse `1 ft 2 in` and `1' 2"`
- Change parsing exceptions to `FormatException`, matches .NET docs and `double.Parse()` better

Minor things at the same time:
- Remove Trim() from quantity code, instead trim in UnitParser
- Add new abbreviations to foot/inch based on wikipedia


Common/GeneratedCode/Quantities/Area.Common.g.cs
```diff
@@ -702,8 +701,7 @@ private static bool TryParseInternal([CanBeNull] string str, [CanBeNull] IFormat
 
             provider = provider ?? GlobalConfiguration.DefaultCulture;
 
-            return QuantityParser.Default.TryParse<Area, AreaUnit>(str, provider, TryParseUnitInternal, From,
-                (x, y) => From(x.SquareMeters + y.SquareMeters, BaseUnit), out result);
```
tmilnthorp: Real nice to get rid of the accumulator function!

UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -68,6 +69,67 @@ public static Length FromFeetInches(double feet, double inches)
...
+        /// https://en.wikipedia.org/wiki/Foot_(unit)
+        /// </summary>
+        /// <param name="str"></param>
+        /// <param name="formatProvider">Optionally specify the culture format numbers and localize unit abbreviations. Defaults to thread's culture.</param>
+        /// <returns>Parsed length.</returns>
+        public static Length ParseFeetInches(string str, IFormatProvider formatProvider = null)
+        {
+            if (!TryParseFeetInches(str, out Length result, formatProvider))
+            {
+                // A bit lazy, but I didn't want to duplicate this edge case implementation just to get more narrow exception descriptions.
```
tmilnthorp: I think this is actually the correct way to go 👍

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
```diff
@@ -34,10 +34,10 @@ namespace UnitsNet
 {
     public sealed partial class UnitAbbreviationsCache
     {
-        private Dictionary<IFormatProvider, UnitTypeToLookup> looksupsForCulture;
```
tmilnthorp: Oops :)

PR #520, opened 2018-10-14
Merge generator scripts for WRC and NetFramework

angularsen:
Merge #517 first.

I initially started out to see if I could easily create separate code generation scripts for WRC, but I quickly realized there is a lot in common and keeping both sets of scripts in sync on common code would be a big hassle. However, I think I found a nice middleground that improves on the current state.

Instead of using `partial` classes to share common code between WRC and NetFramework, we can instead handle WRC special cases in code generator scripts. This removes a bunch of #if sections and merges the "Common" partial code back into each of the WRC and NetFramework files so it is in one place and easy to read.

- Merge quantity code from Common partial files back into WRC/NetFramework specific code
- Handle WRC special cases in code generator script
- Arrange (huge) quantity code into #regions, easier to navigate
- Move code generation script into functions, easier to read
- Add Types.psm1 with types for intellisense on parsed JSON*
- Require PowerShell 5.1 (ships with Win10, needed for classes)

* VS and VSCode struggle with intellisense when importing types with 'using module', this is relatively new and will probably improve later.


angularsen:
Not quite ready to merge, still fixing WRC compile errors.

PR #521, opened 2018-10-15
Remove internal parse methods

angularsen:
Based on #520 so merge that first.
Parse/TryParseInternal and Parse/TryParseUnitInternal methods are redundant since the same logic is already in QuantityParser.
Also use UnitParser with the given UnitAbbreviationsCache instead of external unit parsing code via delegates.

This saves method declarations for all our 80 quantities, reduces binary size and helps keep things more DRY.


angularsen:
I'm fixing merge errors now.

angularsen:
Rebased.

tmilnthorp:
Just one tests fails expecting an exception I don't see. I think we may have removed it during refactoring. Intentional or accidental?

angularsen:
Test caught a flaw in the implementation actually, fixed now.

PR #523, opened 2018-10-15
Replace GetHashCode implementation to match library. Fixing exception…

tmilnthorp:
…s with null in methods/operators. Adding tests to validate.

### DIFF
--- a/UnitsNet/BaseDimensions.cs
+++ b/UnitsNet/BaseDimensions.cs
@@ -47,29 +47,21 @@ namespace UnitsNet
             if(obj is null || !(obj is BaseDimensions))
                 return false;
 
-            var baseDimensionsObj = (BaseDimensions)obj;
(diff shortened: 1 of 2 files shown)


UnitsNet/BaseDimensions.cs
```diff
@@ -47,29 +47,21 @@ public override bool Equals(object obj)
...
-            int hash = 17;
-            hash = hash * 23 + Length;
-            hash = hash * 23 + Mass;
-            hash = hash * 23 + Time;
-            hash = hash * 23 + Current;
-            hash = hash * 23 + Temperature;
-            hash = hash * 23 + Amount;
-            hash = hash * 23 + LuminousIntensity;
-            return hash;
+            return new {Length, Mass, Time, Current, Temperature, Amount, LuminousIntensity}.GetHashCode();
```
angularsen: Nice.

PR #524, opened 2018-10-15
Si unit system

tmilnthorp:


UnitsNet/BaseUnits.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        /// <summary>
+        /// Check if two dimensions are equal.
+        /// </summary>
+        /// <param name="left">Left.</param>
+        /// <param name="right">Right.</param>
+        /// <returns>True if equal.</returns>
+        public static bool operator ==(BaseUnits left, BaseUnits right)
+        {
+            return left?.Equals(right) == true;
```
angularsen: This fails to compare two `null` values I think.
https://stackoverflow.com/a/73732/134761

UnitsNet/UnitSystem.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public BaseUnits BaseUnits { get; }
+
+        private static readonly BaseUnits SIBaseUnits = new BaseUnits(LengthUnit.Meter, MassUnit.Kilogram, DurationUnit.Second,
+            ElectricCurrentUnit.Ampere, TemperatureUnit.Kelvin, AmountOfSubstanceUnit.Mole, LuminousIntensityUnit.Candela);
+
+        /// <summary>
+        /// Gets the SI unit system.
+        /// </summary>
+        public static UnitSystem SI { get; } = new UnitSystem(SIBaseUnits);
```
angularsen: Sweet.

UnitsNet/BaseUnits.cs
tmilnthorp: Left hand side null is handled by return left?. The right hand side check is handled in the Equals method:

```csharp
        public override bool Equals(object obj)
        {
            if(obj is null || !(obj is BaseUnits))
                return false;
```

tmilnthorp: I will add a test 😃

angularsen:
```
object left = null;
object right = null;
(left?.Equals(right) == true).Dump(); // False in linqpad
```

tmilnthorp: That would not call BaseUnits.Equals  ?

tmilnthorp: Fixed :)

UnitsNet/BaseDimensions.cs
```diff
@@ -41,6 +42,18 @@ public BaseDimensions(int length, int mass, int time, int current, int temperatu
             LuminousIntensity = luminousIntensity;
         }
 
+        public bool IsBase()
+        {
+            var dimensionsArray = new int[]{Length, Mass, Time, Current, Temperature, Amount, LuminousIntensity};
+            bool onlyOneEqualsOne = dimensionsArray.Where(dimension => dimension == 1).Take(2).Count() == 1;
```
angularsen: `dimensionsArray.Count(dimension => dimension == 1) == 1`

UnitsNet/BaseDimensions.cs
```diff
@@ -41,6 +42,18 @@ public BaseDimensions(int length, int mass, int time, int current, int temperatu
             LuminousIntensity = luminousIntensity;
         }
 
+        public bool IsBase()
```
angularsen: `IsBaseUnit` and `IsDerivedUnit` maybe? I had to think a bit when reading usages to get exactly what `IsBase` meant.

angularsen: Add xmldoc to public members, that also helps with above comment.

UnitsNet/BaseUnits.cs
```diff
@@ -0,0 +1,177 @@
...
+// THE SOFTWARE.
+
+using System;
+using System.Text;
+using UnitsNet.Units;
+
+namespace UnitsNet
+{
+#if !WINDOWS_UWP
+    public sealed partial class BaseUnits : IEquatable<BaseUnits> { }
```
angularsen: Don't you need to also add `IEquatable` to cover the non-generic version?

UnitsNet/BaseUnits.cs
```diff
@@ -0,0 +1,177 @@
...
+    {
+        /// <summary>
+        /// Creates an instance of if the base units class that represents the base units for a quantity.
+        /// All quantities, both base and derived, can be represented by a combination of these seven base units.
+        /// </summary>
+        /// <param name="length">The length unit (L).</param>
+        /// <param name="mass">The mass unit (M).</param>
+        /// <param name="time">The time unit (T).</param>
+        /// <param name="current">The electric current unit (I).</param>
+        /// <param name="temperature">The temperature unit (Θ).</param>
```
angularsen: Just a heads up, I had trouble with this character `Θ` in powershell. Had to use `P` instead there, or at least I couldn't figure it out at the time.

UnitsNet/BaseUnits.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+#endif
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            var sb = new StringBuilder();
+
+            sb.AppendFormat("[Length]: {0}, ", UnitAbbreviationsCache.Default.GetDefaultAbbreviation(Length));
```
angularsen: Is abbreviation or dimensional number more useful? I don't know myself.

UnitsNet/UnitSystem.cs
```diff
@@ -0,0 +1,125 @@
...
+// THE SOFTWARE.
+
+using System;
+using System.Text;
+using UnitsNet.Units;
+
+namespace UnitsNet
+{
+#if !WINDOWS_UWP
+    public sealed partial class UnitSystem : IEquatable<UnitSystem> { }
```
angularsen: Same here about `IEquatable`.

UnitsNet/BaseDimensions.cs
tmilnthorp: That would be less efficient as it must go through the entire array. This will stop once it finds 2.

It's small array though so it isn't a huge deal. Is this a readability concern?

UnitsNet/BaseUnits.cs
tmilnthorp: Where is IEquatable? I only see the generic version.

UnitsNet/BaseUnits.cs
tmilnthorp: These are the actual units, so the abbreviation is appropiate.

UnitsNet/BaseUnits.cs
tmilnthorp: Where is it used in PowerShell?

UnitsNet/BaseDimensions.cs
angularsen: Ah yes I found it confusing. I would favor readability since we're talking 7 items here, but either way works for me.

UnitsNet/BaseUnits.cs
angularsen: My bad, I was remembering `IComparable`.

UnitsNet/BaseUnits.cs
angularsen: Parsing base dimensions from JSON, then passing on its value to generated c# code.

UnitsNet/BaseDimensions.cs
angularsen: Never mind, merging this now.

UnitsNet.Tests/BaseDimensionsTests.cs
```diff
@@ -28,13 +28,13 @@ public void ConstructorImplementedCorrectly()
...
         [InlineData(0, 0, 0, 0, 0, 0, 1)]
-        public void IsBaseImplementedSuccessfully(int length, int mass, int time, int current, int temperature, int amount, int luminousIntensity)
+        public void IsBaseQuantityImplementedProperly(int length, int mass, int time, int current, int temperature, int amount, int luminousIntensity)
         {
             var baseDimensions = new BaseDimensions(length, mass, time, current, temperature, amount, luminousIntensity);
             var derivedDimensions = new BaseDimensions(length * 2, mass * 2, time * 2, current * 2, temperature * 2, amount * 2, luminousIntensity * 2);
 
-            Assert.True(baseDimensions.IsBase());
-            Assert.False(derivedDimensions.IsBase());
+            Assert.True(baseDimensions.IsBaseQuantity());
```
angularsen: Aw gad I don't know why I keep getting confused about what are units and what are quantities with these BaseUnits thing. I keep thinking about "SI base units" I guess.

PR #525, opened 2018-10-16
Prune generated tests

tmilnthorp:
Merge #524 first

Removing generated tests to match the reference files we have for generated code.


tmilnthorp:
That diff just looks ugly until #524 is pushed. Ill wait.

angularsen:
Merged #526  instead.

angularsen:
Btw, instead of creating a new PR you can use `interactive rebase` in git to remove unrelated commits and force push that to update the same PR.

PR #527, opened 2018-10-16
Implementing dimensionless base dimension

tmilnthorp:


tmilnthorp:
I'm losing track of what I based what off :)

SIUnits need to go first. Will move some code there.

PR #529, opened 2018-10-17
Cleaning up some unused code and adding ValueTuple support to make un…

tmilnthorp:
…it abbreviations a bit clearer


UnitsNet/CustomCode/UnitAbbreviationsCache.cs
```diff
@@ -65,8 +65,8 @@ private void LoadGeneratedAbbreviations()
         {
             foreach(var localization in GeneratedLocalizations)
             {
-                var culture = new CultureInfo(localization.Item1);
-                MapUnitToAbbreviation(localization.Item2, localization.Item3, culture, localization.Item4);
+                var culture = new CultureInfo(localization.CultureName);
+                MapUnitToAbbreviation(localization.UnitType, localization.UnitValue, culture, localization.UnitAbbreviations);
```
angularsen: Aw, much better!

UnitsNet.WindowsRuntimeComponent/UnitsNet.WindowsRuntimeComponent.csproj
```diff
@@ -210,6 +210,9 @@
     <PackageReference Include="JetBrains.Annotations">
       <Version>11.1.0</Version>
     </PackageReference>
+    <PackageReference Include="System.ValueTuple">
```
angularsen: Does this dependency add any extra constraints for consumers?

angularsen: No nuget dependencies listed on these at least. If I understand this nuget correctly, it is a Roslyn package that extends C# language capabilities and should look no different to consumers than `ValueTuple<T1,T2,...>` so if we used this on a public API consumers can still use Item1, Item2 etc. so I think it should be safe.

https://www.nuget.org/packages/System.ValueTuple/
https://blog.marcgravell.com/2017/04/exploring-tuples-as-library-author.html

tmilnthorp: Correct!

PR #530, opened 2018-10-17
Quantities namespace

tmilnthorp:
Organizing quantities into UnitsNet.Quantities namespace

### DIFF
--- a/UnitsNet.Serialization.JsonNet.CompatibilityTests/UnitsNet.Serialization.JsonNet.CompatibilityTests.csproj
+++ b/UnitsNet.Serialization.JsonNet.CompatibilityTests/UnitsNet.Serialization.JsonNet.CompatibilityTests.csproj
@@ -13,12 +13,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Compile Include="..\UnitsNet.Serialization.JsonNet.Tests\UnitsNetJsonConverterTests.cs" Link="UnitsNetJsonConverterTests.cs" />
-  </ItemGroup>
-
-  <ItemGroup>
     <!--Get the latest released version of UnitsNet.Serialization.JsonNet in Nuget-->
-    <PackageReference Include="UnitsNet.Serialization.JsonNet" Version="*" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
     <PackageReference Include="xunit" Version="2.3.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
@@ -26,6 +21,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\UnitsNet.Serialization.JsonNet\UnitsNet.Serialization.JsonNet.csproj" />
     <ProjectReference Include="..\UnitsNet\UnitsNet.csproj" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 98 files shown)


angularsen:
Failing on 
```
 UnitsNet.UnitsNetException : No private fields found in type.
      Stack Trace:
           at UnitsNet.Serialization.JsonNet.UnitsNetJsonConverter.GetPrivateInstanceField(Type quantityType, String fieldName)
           at UnitsNet.Serialization.JsonNet.UnitsNetJsonConverter.GetValueOfQuantity(Object value, Type quantityType)
           at UnitsNet.Serialization.JsonNet.UnitsNetJsonConverter.WriteJson(JsonWriter writer, Object obj, JsonSerializer 
```

tmilnthorp:
Yup. I fixed the one half of those but the failures now are in the compatibility tests. It's using an old version of the json classes against a new version of UnitsNet. The namespace changed and it was previously hard coded. Why do we use an old version of the json formatter against a new version of UnitsNet? Shouldn't they match in case of any changes?

angularsen:
That test was specifically written to ensure that quantities serialized to JSON 2 years ago can still be deserialized with latest version. I don't see the test code right now, but that was at least the intent.

tmilnthorp:
Those tests are invalid then. It uses the same test file as the current json tests. Rather than using an old json converter to serialize/deserialize current quantities, I'd expect the tests to have old style strings that get resumed into current objects. That would also mean the converter would need version info in the strings and be able to read old ones.

tmilnthorp:
Ok. I've added a new set of tests for compatibility that takes the old or expected json strings and converts them into current quantities.

UnitsNet.Serialization.JsonNet.CompatibilityTests/UnitsNet.Serialization.JsonNet.CompatibilityTests.csproj
```diff
@@ -12,20 +12,16 @@
...
     <!--Get the latest released version of UnitsNet.Serialization.JsonNet in Nuget-->
-    <PackageReference Include="UnitsNet.Serialization.JsonNet" Version="*" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.3.0" />
     <PackageReference Include="xunit" Version="2.3.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
 		<DotNetCliToolReference Include="dotnet-xunit" Version="2.3.0-beta2-build3683" />
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\UnitsNet.Serialization.JsonNet\UnitsNet.Serialization.JsonNet.csproj" />
```
angularsen: With this change, you are only verifying that the current JSON code can serialize/deserialize the current quantity code.
We also need to ensure that some arbitrary old JSON dll can serialize/deserialize new quantity code. The purpose of this CompatibilityTests project was to link in different versions of dependencies than the normal JSON tests project.

With old I mean some version deployed prior to v4 branch.
With current I mean whatever code is local starting from v4 branch and master branch when v4 is merged.

In gitter I suggested some tests like this:
1. A test that uses v3 dll of JSON serializer to serialize "current" quantities and can deserialize again
1. A test that uses "current" JSON code to serialize "current" quantities and can deserialize again
1. A test that deserializes a "v3" JSON string into "current" quantities using "current" JSON code

PR #531, opened 2018-10-22
Adding additional functionality to IQuantity<UnitType>

tmilnthorp:


angularsen:
Nice, I thought that would conflict with WRC but I see WRC only uses the non-generic interface.

PR #534, opened 2018-10-29
Adding dyn/cm² and lbm/(in·s²) units for pressure

tmilnthorp:


tmilnthorp:
1 microbar is the same unit as dyn/cm². How do we want to handle units with multiple definitons? Add a new unit (as I have in this PR), or just add dyn/cm² as an abbreviation for microbars?

angularsen:
Separate units as in the pr I think, since I assume you would want to refer to both Microbars and DynesPerSquareCentimeter even if they are equivalent.

Common/UnitDefinitions/Pressure.json
```diff
@@ -321,6 +321,30 @@
...
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "dyn/cm²" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "PoundPerInchSecondSquared",
+      "PluralName": "PoundsPerInchSecondSquared",
```
angularsen: Is this a pressure unit? `Second` seems out of place to me and I can't Google it in context of pressure.

tmilnthorp: Yes, see [here](https://www.wolframalpha.com/input/?i=lbm%2F(in+sec%5E2)).

You can also see it corresponds to the base dimensions: mass / (length * time²)

```json
  "BaseDimensions": {
    "L": -1,
    "M": 1,
    "T": -2
  }
```

angularsen: Thanks

tmilnthorp: 👍

angularsen:
Released in [3.107.1](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F3.107.1).

PR #536, opened 2018-10-29
Adding lbf/(ft·s) and lbm/s³ units for heat flux

tmilnthorp:


Common/UnitDefinitions/HeatFlux.json
```diff
@@ -116,6 +116,30 @@
...
+    },
+    {
+      "SingularName": "PoundForcePerFootSecond",
+      "PluralName": "PoundsForcePerFootSecond",
+      "FromUnitToBaseFunc": "x*1.459390293720636e1",
+      "FromBaseToUnitFunc": "x/1.459390293720636e1",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "lbf/(ft·s)", "lb/(ft·s)" ]
```
angularsen: Is `lb` (not `lbf`) a valid abbreviation for this? I've always thought it was to distinguish mass from force, similar to `kgf` vs `kg`.

Common/UnitDefinitions/HeatFlux.json
```diff
@@ -116,6 +116,30 @@
...
+    },
+    {
+      "SingularName": "PoundPerSecondCubed",
+      "PluralName": "PoundsPerSecondCubed",
+      "FromUnitToBaseFunc": "x*4.5359237e-1",
+      "FromBaseToUnitFunc": "x/4.5359237e-1",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "lbm/s³", "lb/s³" ]
```
angularsen: Is `lbm` the preferred notation over `lb`?

Common/UnitDefinitions/HeatFlux.json
tmilnthorp: I think you're right. I have seen it listed this way a few places, but I think I'll remove it. Otherwise it's too easy to confuse with lbm/(ft·s) which is for dynamic viscosity.

Common/UnitDefinitions/HeatFlux.json
tmilnthorp: I will switch these. I hate imperial units :)

angularsen: Haha, me too 😆 Come on world, let's standardize on something more sane than your own foot and a big, heavy stone you happened to have lying around.

tmilnthorp:
👍 abbreviations fixed!

Issue #532, opened 2018-10-26
Adapt to Open-Source Library Guidance from Microsoft


angularsen:
# Open-source library guidance
https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/

This guidance provides recommendations for developers to create high-quality .NET libraries. This documentation focuses on the *what* and the *why* when building a .NET library, not the *how*.

Aspects of high-quality open-source .NET libraries:

* **Inclusive** - Good .NET libraries strive to support many platforms and applications.
* **Stable** - Good .NET libraries coexist in the .NET ecosystem, running in applications built with many libraries.
* **Designed to evolve** - .NET libraries should improve and evolve over time, while supporting existing users.
* **Debuggable** - .NET libraries should use the latest tools to create a great debugging experience for users.
* **Trusted** - .NET libraries have developers' trust by publishing to NuGet using security best practices.

- First step: Create a checklist of work items for what UnitsNet needs to do be compliant
- Create issues with milestone `Open source library guidance compliance`

### Tentative work list (not complete)
- [x] Provide debugging sources with SourceLink #439
- [x] Use strong name signing and deprecate `UnitsNet.Signed` packages

angularsen:
I think the main points are solved now. Closing this.

Issue #439, opened 2018-05-08
Provide source with nuget using SourceLink


angularsen:
https://github.com/dotnet/sourcelink

I haven't looked much into it yet, but I assume this replaces the previous solution of shipping a symbols package to SymbolSource.

**Update 2018-09-30:** Good article on producing and consuming nugets with sourcelink
https://www.hanselman.com/blog/ExploringNETCoresSourceLinkSteppingIntoTheSourceCodeOfNuGetPackagesYouDontOwn.aspx

angularsen:
@tmilnthorp I reverted the sourcelink PR #538 in `master`. Let's revisit with your comment in mind:
> Either that or we can add them back and use the following to clean up the GitHub diffs: https://robots.thoughtbot.com/github-diff-supression
> 
> This would be nice too so you don't have to rerun generate every time you check out, switch branches, etc.

I think this sounds like an interesting idea, and I agree it's annoying you have to generate source locally when switching branches. We could do it as a pre-build step, but that would increase local build times significantly.

angularsen:
This was added in v4 in #548 after a bit back and forth.

PR #538, opened 2018-10-29
Adding SourceLink properties and NuGet package

tmilnthorp:
Adding SourceLink per #532


tmilnthorp:
I'm not 100% sure how this works so dunno how to test it :)

angularsen:
Looks good to me, let's try it!

angularsen:
@tmilnthorp I tried it and can't get it to work in Visual Studio 15.8.7, also running `sourcelink test` on it reveals some hashes don't match with the github URLs. Will need to look more into it.

Another thing I just realized; I don't think we can use SourceLink after we removed all the generated code - since that source code no longer exists on github. 😞 Maybe we should just stick to symbols nuget packages, which we already have.

tmilnthorp:
Either that or we can add them back and use the following to clean up the GitHub diffs: https://robots.thoughtbot.com/github-diff-supression

This would be nice too so you don't have to rerun generate every time you check out, switch branches, etc.

angularsen:
Let's continue discussion in #439.

angularsen:
This was first reverted, then added back again in #548.

PR #540, opened 2018-10-29
Fixed parsing of NaN.

### LINKED ISSUE #537: Can't parse NaN
I have some very big numbers coming in from another system. For some reason, these can be larger than double.MaxValue and will come in as double.NaN. They're converted to string then sent to UnitsNet's Parse method, which fails throwing an exception: "Invalid string detected: a".

I think what should happen is that you get a length (or whatever) with a value of NaN.

I've written a test and made a change to the number regex which seems to pass the test. I'll make a pull request soon.

ajf2:
Fixes #537. Values of double.NaN can now be parsed from strings.


angularsen:
Hi, we actually decided to not support Nan or Infinity since decimal and other numeric types don't support it.

See https://github.com/angularsen/UnitsNet/pull/502

PR #541, opened 2018-10-30
Fixing nasty bug in GetHashCode where two quantities with the same nu…

tmilnthorp:
Fixing nasty bug in GetHashCode where two quantities with the same numerical value/unit pair returned the same hash code. Type needed in GetHashCode implementation to guarantee uniqueness.

This is especially important now we have an IQuantity interface. If you had a HashSet<IQuantity>, Dictionary with IQuantity keys, etc., we need to avoid hash collisions.

Added test also.

### DIFF
--- a/UnitsNet.Tests/QuantityTests.cs
+++ b/UnitsNet.Tests/QuantityTests.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com).
+// https://github.com/angularsen/UnitsNet
+// 
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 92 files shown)


UnitsNet.Tests/QuantityTests.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public partial class QuantityTests
+    {
+        [Fact]
+        public void GetHashCodeForDifferentQuantitiesWithSameValuesAreNotEqual()
+        {
+            var length = new Length( 1.0, (LengthUnit)2 );
+            var area = new Area( 1.0, (AreaUnit)2 );
+
+            Assert.NotEqual( length.GetHashCode(), area.GetHashCode() );
```
tmilnthorp: Previously, this would fail! 💩

angularsen:
Released in [3.107.1](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F3.107.1).

PR #542, opened 2018-10-30
Fix for explicit AbbreviationsWithPrefixes not lining up with the pre…

tmilnthorp:
…fixes.

### DIFF
--- a/Common/UnitDefinitions/Force.json
+++ b/Common/UnitDefinitions/Force.json
@@ -70,7 +70,7 @@
         {
           "Culture": "ru-RU",
           "Abbreviations": [ "Н" ],
-          "AbbreviationsWithPrefixes": [ "мкН", "мН", "даН", "кН" ]
+          "AbbreviationsWithPrefixes": [ "мкН", "мН", "даН", "кН", "МН" ]
         }
       ]
     },
--- a/Common/UnitDefinitions/Force.json
+++ b/Common/UnitDefinitions/Force.json
@@ -70,7 +70,7 @@
         {
           "Culture": "ru-RU",
           "Abbreviations": [ "Н" ],
-          "AbbreviationsWithPrefixes": [ "мкН", "мН", "даН", "кН", "МН" ]
+          "AbbreviationsWithPrefixes": [ "мкН", "мН", "даН", "кН" ]
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


Common/UnitDefinitions/Force.json
```diff
@@ -70,7 +70,7 @@
         {
           "Culture": "ru-RU",
           "Abbreviations": [ "Н" ],
-          "AbbreviationsWithPrefixes": [ "даН", "кН" ]
+          "AbbreviationsWithPrefixes": [ "мкН", "мН", "даН", "кН", "МН" ]
```
tmilnthorp: I forgot to update these yesterday and it messed up the localization. I added a error in the script file for these situations and also found torque had the same issue.

UnitsNet/Scripts/GenerateUnits.ps1
```diff
@@ -169,6 +169,14 @@ function Add-PrefixUnits {
 
         foreach ($unit in $quantity.Units)
         {
+            foreach ($localization in $unit.Localization){
+                if($localization.AbbreviationsWithPrefixes.Count -gt 0){
+                    if($unit.Prefixes.Count -ne $localization.AbbreviationsWithPrefixes.Count){
+                        Write-Error "The prefix count ($($unit.Prefixes.Count)) does not match the abbreviations with prefixes count ($($unit.Localization.AbbreviationsWithPrefixes.Count)) for $($quantity.Name).$($unit.SingularName)"
```
tmilnthorp: Should we abort generation if this error is discovered?

angularsen: Good point, I guess a failing build would be nice. Downside is that we need to do all translations to add a new unit, but better than getting wrong translations.

tmilnthorp: Not all the translations really. The check is only if an explicit list of AbbreviationsWithPrefixes is specified.

Would it make sense to just have the prefixes localized somewhere? For example there are Russian AbbreviationsWithPrefixes where it looks like it's just a prefix (such as д for Deci) + the abbreviation. Similar to English. Maybe this falls apart for some languages?

tmilnthorp: Ok. The build now fails when they do not match (see failed build [here](https://ci.appveyor.com/project/angularsen/unitsnet/builds/19920611)).

angularsen:
Released in https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F3.107.1

PR #544, opened 2018-10-30
Making Obsolete attribute message consistent across scripts

tmilnthorp:


angularsen:
Released in https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F3.107.1

Issue #545, opened 2018-11-01
Remove conversion overloads from ToString implementations


tmilnthorp:
I would like to split up the code responsible for unit conversion and ToString.

Since we have the ToUnit method, we could simplify the ToString implementations and implement IFormattable.

This would leave the following ToString overloads:
```csharp
public override string ToString();
public string ToString(string format);
public string ToString(IFormatProvider provider);
public string ToString(string format, IFormatProvider provider);
```

For example to get a length string in inches, we currently do the following:

```csharp
var length = Length.FromMeters( 3.0 );
var lengthInInchesString = length.ToString( LengthUnit.Inch );
```

Today, this will achieve the same thing
```csharp
var length = Length.FromMeters( 3.0 );
var lengthInInchesString = length.ToUnit( LengthUnit.Inch ).ToString();
```

We could even change the conversion properties to return a converted quantity rather than a double (public Length Inches in this example) and do the following:
```csharp
var length = Length.FromMeters( 3.0 );
var lengthInInchesString = length.Inches.ToString();
```

angularsen:
My initial thought is, I like it. It feels more consistent that ToString() only uses whatever value and unit it was created with and if you want something else you convert it first. It also resembles how `DateTimeOffset.UtcNow.ToString()` is different than `DateTimeOffset.Now.ToString()`, if you consider different timezones as a kind of unit conversion, and you can't _convert_ to UTC using only format string - you need to call `.ToUniversalTime()` or `.ToLocalTime()` first.

However, it does might make it less discoverable (not obvious you need to call ToUnit() first, but we could mention it in ToString xmldoc). Also, I assume we will break a lot of consumer code since ToString()'ing into a specific unit is a pretty common usecase.

#### IFormattable format string
IFormattable takes a format string. I skimmed through [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.iformattable?view=netframework-4.7.2), it even has a Temperature example. Here they demo how you can pass "F" to get Fahrenheit or "C" for Celsius, but the number formatting is hard coded.

Maybe we could do something clever to allow you to control both unit and number formatting in the format string? Or maybe it's just a deathtrap waiting to cause bugs for consumers, who will never memorize the magic strings. I was thinking something like this:

```c#
var meter = Length.FromMeters(1);
meter.ToString("Foot_F2"); // "3.28 ft"
meter.ToString("Foot_F3"); // "3.281 ft"
[...]

PR #546, opened 2018-11-03
Removing unit conversion from ToString methods. ToUnit must be called…

tmilnthorp:
… first. ToString now has a single responsibility of converting to quantity into a string representation #545

### DIFF
--- a/UnitsNet.WindowsRuntimeComponent/GeneratedCode/Quantities/Information.WindowsRuntimeComponent.g.cs
+++ b/UnitsNet.WindowsRuntimeComponent/GeneratedCode/Quantities/Information.WindowsRuntimeComponent.g.cs
@@ -937,64 +937,51 @@ namespace UnitsNet
         /// <returns>String representation.</returns>
         public override string ToString()
         {
-            return ToString(Unit);
-        }
-
-        /// <summary>
-        ///     Get string representation of value and unit. Using current UI culture and two significant digits after radix.
-        /// </summary>
-        /// <param name="unit">Unit representation to use.</param>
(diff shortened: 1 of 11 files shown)


UnitsNet.WindowsRuntimeComponent/GeneratedCode/Quantities/Information.WindowsRuntimeComponent.g.cs
```diff
@@ -937,64 +937,51 @@ private decimal AsBaseNumericType(InformationUnit unit)
...
 
         /// <summary>
         ///     Get string representation of value and unit.
         /// </summary>
-        /// <param name="unit">Unit representation to use.</param>
         /// <param name="significantDigitsAfterRadix">The number of significant digits after the radix point.</param>
         /// <returns>String representation.</returns>
         /// <param name="cultureName">Name of culture (ex: "en-US") to use for localization and number formatting. Defaults to <see cref="GlobalConfiguration.DefaultCulture" /> if null.</param>
-        public string ToString(InformationUnit unit, [CanBeNull] string cultureName, int significantDigitsAfterRadix)
+        public string ToString(string cultureName, int significantDigitsAfterRadix)
```
angularsen: For future reference, `significantDigitsAfterRadix` is a param I think should be baked into the format string instead - similar to `double.ToString("F2")`. We did put in a whole lot of magic though on number formatting, to use scientific notation (`1.01e5`) for values smaller/larger than a certain threshold, as well as this "digits after radix" concept. Just some bits of info to bring into the whole format string discussion later.

PR #550, opened 2018-11-04
Normalize Temperature arithmetic

### LINKED ISSUE #518: Temperature times a double
Is there a specific reason that I cant do:

`Temperature * double`

even though I can do it with every other unit?

angularsen:
Fixes #518 

- Generate default arithmetic (no special case returning `TemperatureDelta`)
- Remove `Multiply()` and `Divide()` methods
- Remove custom `-` operator overload
- Add `Temperature.ToDelta(other)` instance method
- Add `TemperatureDelta.FromTemperatures(left, right)` static method


UnitsNet/CustomCode/Quantities/TemperatureDelta.extra.cs
```diff
@@ -47,6 +50,21 @@ public partial struct TemperatureDelta
...
+        ///     would operate on Kelvins and likely give an unexpected result.
+        ///     Example:
+        ///     double deltaCelsius = TemperatureDelta.FromTemperatures(celsius30, celsius20).DegreesCelsius; // 10 C
+        ///     double wrongDeltaCelsius = (celsius30 - celsius20).DegreesCelsius; // 303.15 K - 293.15 K = 10 K = -263.15 degrees Celsius
+        /// </summary>
+        /// <returns>A temperature delta.</returns>
+        public static TemperatureDelta FromTemperatures(Temperature left, Temperature right)
+        {
+            // TemperatureDeltaUnit enum has identical values to TemperatureUnit, so we can simply cast between them
+            return new TemperatureDelta(left.Value - right.As(left.Unit), (TemperatureDeltaUnit)left.Unit);
```
tmilnthorp: A straight enum cast relies on the ordering in the json file. Perhaps a switch would be better?

angularsen: I guess, I figured since they were generated it would be safe. How about instead enum value ToString() and enum parse? Then it will throw if they are not identical.

angularsen: @tmilnthorp Pushed new commit with a new ToDeltaUnit() method.

angularsen:
Self-merging this now just to get a beta nuget out. @tmilnthorp Please let me know if you have some reservations on my latest change.

angularsen:

PR #552, opened 2018-11-04
Remove InternalsVisibleTo in reflection helper (v3)

angularsen:
Instead copy dependency into JSON lib.
We have had enough problems with sharing this tiny bit of code, with strong name signing and what not.
It broke the compatibility tests in v4 branch for instance.


angularsen:
Can't delete internalsvisibleto in master (v3) branch, it would break backwards compatibility with older JSON libs.

Issue #555, opened 2018-11-05
Serialization in human readable form


Batboa:
I have a use case where we want  to serialize Length and other IQuantity, but instead of having them in the format that is currently provided of `"{\n  \"Unit\": \"MassUnit.Kilogram\",\n  \"Value\": 10.0\n}";` We want to keep the serialized version in a human readable format of `10 kg`. Since the `ToString ` and `Parse` functionality are already great it should be easy to leverage those for this. Example of https://github.com/Batboa/UnitsNet/commit/66553b5664894a020150e26879737357519321a6.

Batboa:
I've very willing to work with the maintainers of this project to get this commit into a format this fits what you want for this project.

angularsen:
Hi, this could definitely work for most quantities and units. However, there are several units where you would get exceptions due to either ambiguous unit abbreviations (f.ex. `Length.DtpPoint` and `.PrinterPoint`) or missing unit abbreviations (f.ex. `Ratio.DecimalFraction`).

```csharp
var cm = Length.FromCentimeters(1);
Console.WriteLine(cm.ToString(LengthUnit.DtpPoint)); // 28.35 pt
Console.WriteLine(cm.ToString(LengthUnit.PrinterPoint)); // 28.45 pt

var fiftyPercent = Ratio.FromPercent(50);
Console.WriteLine(fiftyPercent.ToString(RatioUnit.DecimalFraction)); // 0.5
Console.WriteLine(fiftyPercent.ToString(RatioUnit.PartPerMillion)); // 500,000 ppm
```

Batboa:
Is the missing unit abbreviation an actual issue? It seems that the serializer I proposed will maintain the existing behavior of a units `Parse` method. So if there exists a unit that does not have an abbreviations like in your example of a `DecimalFraction` for `Ratio` the value will serialize and deserialize properly. Then if an abbreviation is required and is missing an error will be thrown. I will add tests to display this behavior.

As for ambiguous unit abbreviations the user of the API will need to provided some type of object that contains the priority unit type in case of conflict. I'll push an example of a way this could be done to my project and open a PR.

tmilnthorp:
The purpose of serialization is to store or transport data. Being human-readable is not the goal. Perhaps we can help more if we understand what you are trying to achieve?

angularsen:
I agree, I don't fully understand the motivation for this kind of serialization. It would help to know more what you want to accomplish.

[...]

PR #557, opened 2018-11-06
Typed json converter

Batboa:
#555 Serializer that can read and write to in formats similar to `Parse`.


angularsen:
Please continue discussion in #555 first.

PR #560, opened 2018-11-08
Fix arithmetic

angularsen:
As part of creating the [4.x upgrade guide](https://github.com/angularsen/UnitsNet/wiki/Upgrading-from-3.x-to-4.x) and testing out v4 binary I realized that Temperature arithmetic was wrong after changing to the standard arithmetic implementation.

- Create tests for addition and subtraction of Temperature
- Update arithmetic for all quantities
- Regen

### DIFF
--- a/UnitsNet.Tests/CustomCode/TemperatureTests.cs
+++ b/UnitsNet.Tests/CustomCode/TemperatureTests.cs
@@ -46,6 +46,30 @@ namespace UnitsNet.Tests.CustomCode
         protected override double KelvinsInOneKelvin => 1;
 
         [Theory]
+        [InlineData(TemperatureUnit.DegreeCelsius, 10, 20, "303.15 °C")]
+        [InlineData(TemperatureUnit.DegreeFahrenheit, 100, 100, "659.67 °F")]
+        [InlineData(TemperatureUnit.Kelvin, 50, 100, "150 K")]
+        public void TemperaturePlusTemperatureEqualsTemperatureAddedInKelvins(TemperatureUnit unit, int left, int right, string expected)
+        {
+            var leftTemp = Temperature.From(left, unit);
+            var rightTemp = Temperature.From(right, unit);
+            Assert.Equal(expected, (leftTemp + rightTemp).ToString());
+        }
+
+        [Theory]
+        [InlineData(TemperatureUnit.DegreeCelsius, 30, 20, "-263.15 °C")]
+        [InlineData(TemperatureUnit.DegreeCelsius, 30, 30, "-273.15 °C")]
+        [InlineData(TemperatureUnit.DegreeFahrenheit, 100, 70, "-429.67 °F")]
+        [InlineData(TemperatureUnit.Kelvin, 300, 200, "100 K")]
+        [InlineData(TemperatureUnit.Kelvin, 200, 300, "-100 K")]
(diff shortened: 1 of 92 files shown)


angularsen:
@tmilnthorp  Waiting for a review on this, then we release a new beta nuget.

UnitsNet.Tests/CustomCode/TemperatureTests.cs
```diff
@@ -45,6 +45,30 @@ public class TemperatureTests : TemperatureTestsBase
...
+        {
+            var leftTemp = Temperature.From(left, unit);
+            var rightTemp = Temperature.From(right, unit);
+            Assert.Equal(expected, (leftTemp + rightTemp).ToString());
+        }
+
+        [Theory]
+        [InlineData(TemperatureUnit.DegreeCelsius, 30, 20, "-263.15 °C")]
+        [InlineData(TemperatureUnit.DegreeCelsius, 30, 30, "-273.15 °C")]
+        [InlineData(TemperatureUnit.DegreeFahrenheit, 100, 70, "-429.67 °F")]
```
tmilnthorp: http://www.wolframalpha.com/input/?i=100%C2%B0F+-+70%C2%B0F

tmilnthorp:
From an engineering standpoint, I agree 100%. But at what point do you switch from a general use library to an engineering/science specific library?

I would expect 30C - 20C = 10C for "general" use.

If I was using for engineering purpose I'd convert 30/20C to K before doing the subtraction anyways.

angularsen:
A lot of discussion and arguments back and forth were made in https://github.com/angularsen/UnitsNet/issues/518.

**Problem definition:**
- "Normal users" will expect 100F - 70F = 30F
- WolframAlpha agrees on this, a widely used scientific tool
- "Engineers" working with temperatures expect 100F - 70F = -429.67 °F = 16.67 K, since you are subtracting two temperatures (levels of energy) and in that context you naturally get a smaller temperature (lower energy level)
- MathCad agrees on this, a widely used scientific tool (I'm not familiar with it)
- The result of temperature addition/subtraction WILL be confusing to one of the two groups of people
- I don't know what is the larger portion of our userbase, can only speculate

**My conclusion:**
I think either way works and its decision can be justified, but I currently see this approach as slightly better because it is consistent across quantities. Either way will be counter-intuitive to _some_ ratio of our userbase, but I don't know what that ratio is and if this change is worse or better in that regard. I only know that I recognize the consistency as an improvement and I am in the not-doing-science camp of users.

I'm open to be convinced otherwise, but we need to settle on this. Worst case pull the change out of v4 and think more on it. What do you think @tmilnthorp ?

angularsen:
I think you are missing my point. I believe we are all biased, we all have our own expectations of what Temperature1 - Temperature2 should yield. A delta or a new temperature?
There is no right or wrong, and I would need a metric for how many % of our current users are expecting one over the other to truly decide with confidence.

However, I recognize that adding a special case for Temperature is more complex than treating all quantities alike. That is the reason I favor the new behavior.

I can agree with the argument that before _changing_ the behavior we must first justify it by measuring the expectations in a representative set of our userbase. 
But I don't agree with keeping it just because it's more appropriate for a "broad-use" library, we don't know the context our users are using this library.

I propose just that, let's revert the change back, close this fix and revisit when we know that a change will indeed improve things. It's pretty clear to mewe are not able to fully agree on this just now, and then we shouldn't change anything either.

angularsen:

PR #561, opened 2018-11-12
Add kg pm

tweet1992:
Adding KgPM and KgalPM


UnitsNet.Tests/CustomCode/MassFlowTests.cs
```diff
@@ -60,6 +60,8 @@ public class MassFlowTests : MassFlowTestsBase
 
         protected override double MegapoundsPerMinuteInOneGramPerSecond => 0.132277e-6;
 
+        protected override double KilogramsPerMinuteInOneGramPerSecond => 0.01666667e-3;
```
angularsen: [This source](http://www.kylesconverter.com/mass-flow/grams-per-second-to-kilograms-per-minute) says `0.060000000000000005`?

UnitsNet.Tests/CustomCode/VolumeFlowTests.cs
```diff
@@ -96,6 +96,8 @@ public class VolumeFlowTests : VolumeFlowTestsBase
 
         protected override double OilBarrelsPerMinuteInOneCubicMeterPerSecond => 3.77388865E2;
 
+        protected override double KilousGallonsPerMinuteInOneCubicMeterPerSecond => 15850.323141489e-3;
```
angularsen: Correct number, but please use scientific notation: `1.5850323141489e1`. Verified with [source](https://www.unitjuggler.com/convert-flowrate-from-m3s-to-us-gpm.html).

UnitsNet.Tests/CustomCode/MassFlowTests.cs
tweet1992: Good catch. I accidentally did the conversion the other way... Sorry!

UnitsNet.Tests/CustomCode/VolumeFlowTests.cs
tweet1992: Fixed with latest commit

UnitsNet.Tests/CustomCode/VolumeFlowTests.cs
```diff
@@ -96,7 +96,7 @@ public class VolumeFlowTests : VolumeFlowTestsBase
 
         protected override double OilBarrelsPerMinuteInOneCubicMeterPerSecond => 3.77388865E2;
 
-        protected override double KilousGallonsPerMinuteInOneCubicMeterPerSecond => 15850.323141489e-3;
+        protected override double KilousGallonsPerMinuteInOneCubicMeterPerSecond => 1.5850323141489e1;
```
angularsen: I think this value is correct, but the conversion function is perhaps wrong?

https://ci.appveyor.com/project/angularsen/unitsnet/builds/20278134
Values are not equal within relative tolerance: 0.0010 %
Expected: 15.850323141489
Actual: 15850323.141489
Diff: 99,999,900.0000 %
      Expected: True
      Actual:   False

angularsen: Also a side-question, is KiloUSGallonsPerMinute a widely used unit? I couldn't find any references to it when googling. Only found US gallons per minute, without kilo prefix.

tweet1992: I dont think it is widely used, but it is something that we are using in the oil and gas industry

tweet1992: FIxed a bug in the conversions :( Sorry for all the bugs! I'm glad you have a good continuous integration pipeline setup, lol.

angularsen: No problem! Yup, the builds sure help :-)

angularsen:
Nuget 3.109 on the way out
https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F3.109.0

PR #566, opened 2018-12-03
Added serialization support of array of units

dschuermans:
Support for serializing an array of units


dschuermans:
@angularsen I'm stumped... Tests are green in VS2017 -.-'

PR #567, opened 2018-12-03
Fix array serialisation & deserialisation

dschuermans:
This PR fixes the inability of the UnitsNetJsonConverter to work with arrays of units when serializing or deserializing

### DIFF
--- a/UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonConverterTests.cs
+++ b/UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonConverterTests.cs
@@ -139,6 +139,38 @@ namespace UnitsNet.Serialization.JsonNet.Tests
 
                 Assert.Equal(expectedJson, json);
             }
+
+            [Fact]
+            public void ArrayValue_ExpectJsonArray()
+            {
+                Frequency[] testObj = new Frequency[] { Frequency.FromHertz(10), Frequency.FromHertz(10) };
+
+                string expectedJson = "[\n" +
+                                      "  {\n" +
+                                      "    \"Unit\": \"FrequencyUnit.Hertz\",\n" +
+                                      "    \"Value\": 10.0\n" +
+                                      "  },\n" +
+                                      "  {\n" +
+                                      "    \"Unit\": \"FrequencyUnit.Hertz\",\n" +
+                                      "    \"Value\": 10.0\n" +
+                                      "  }\n" +
+                                      "]";
+
+                string json = SerializeObject(testObj);
+
+                Assert.Equal(expectedJson, json);
+            }
+
+            [Fact]
+            public void EmptyArrayValue_ExpectJsonArray()
+            {
+                Frequency[] testObj = new Frequency[0];
+
+                string expectedJson = "[]";
+
+                string json = SerializeObject(testObj);
+                Assert.Equal(expectedJson, json);
+            }
         }
 
         public class Deserialize : UnitsNetJsonConverterTests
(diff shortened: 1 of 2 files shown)


dschuermans:
@angularsen help please 😢

angularsen:
You do see the failing test error messages?
https://ci.appveyor.com/project/angularsen/unitsnet/builds/20739879/tests

I can't look into it today sorry, but if you are still stuck then I'll try to find time tomorrow or soon.

dschuermans:
> You do see the failing test error messages?
> https://ci.appveyor.com/project/angularsen/unitsnet/builds/20739879/tests
> 
> I can't look into it today sorry, but if you are still stuck then I'll try to find time tomorrow or soon.

Yes, it are tests that succeed in the original UnitsNetJsonConverter test but which fail in the compatibility project due to the fact that the version used in the compatibility project does not contain my changes to the json converter but it re-uses the same test as the original

Since i didn't setup this whole compatibility thing and there's no documentation regarding it I don't have a clue on how to fix it.

To be honest, i don't understand the purpose of the compatibility project. Conversion to and from JSON should work, irregardless of how UnitsNet works internally.

dschuermans:
Okay so to recap:

1.) This has nothing to do with UnitsNet itself, i didn't change anything internally to UnitsNet
2.) I only changed the UnitsNetJsonConverter and the corresponding unit test to check serialization and deserialization of arrays
3.) The build fails because the compatibility project uses the nuget package of an older version (1.3.1) of the serialization but still re-uses the same unit test for the UnitsNetJsonConverter. Seeing as how the UnitsNetJsonConverter in that nuget package does not contain my fixes, it's obvious that it fails.

So the nuget package used in the compatibility project should be updated to include my fixes, but that can't happen because the build fails so we can't release a new version >.>

dschuermans:
I'm still not entirely convinced...

The way i see it (mind you i haven't been keeping close tabs on where UnitsNet is going) you currently have 4 major versions:

- UnitsNet v3.*
- UnitsNet v4.* (in the making)
- UnitsNet v5.* (future)
- UnitsNet.Serialization.JsonNet v1.*

The UnitsNet.Serialization.JsonNet is a single package, which ideally should work with all those 3 major versions of UnitsNet. It requires you to have at least v3.107.1 i think.
This means that when you make changes in any of those major versions, you'll still want the JSON conversion to work.
This seems logical to test in a separate compatibility test, although I'd constrain that test to just 1 serialization and 1 deserialization test of an object  which contains all types currently test in the converter test (eg non-nullable and nullable unit, array of units, sub-object with units etc etc)
So if you change something inside one of those 3 major versions which causes the output to change, the test should fail and it should be addressed.
You'd might need to have a compatibility project per major version of UnitsNet though...
So I assume that if you're running the tests while developing, the normal conversion test will run against the latest major version of UnitsNet at that moment and the compatibility test(s) will run against older version(s)

Only when there's a bump in major version for the UnitsNet.Serialization.JsonNet (eg due to output changes), it makes sense to have a compatibility project which references an older version of the UnitsNet.Serialization.JsonNet package.

Anyway,

Duplicating the test won't have any affect in my opinion, it will just sit there unchanged and there will be scenario's in which it won't alert you to issues caused by change in functionality to the json conversion.
It will essentially be a snapshot of how UnitsNet is serialized and deserialized at that time

When testing against UnitsNet v4.*, everything should still work. The compatibility test should reference UnitsNet v3.* and everything should still work.
I mean, JSON is a standard, there aren't any edge cases that can differ. The only thing that can change is the way UnitsNet is serialized or deserialized.
If something changes in v4.* which would require UnitsNet.Serialization.JsonNet to have a different output or a complete overhaul of conversion logic (eg caused by renaming of internal fields etc), then compatibility test should add UnitsNet v1.* as nuget package and UnitsNet.Serialization.JsonNet should get a major bump, eg v2.*
That way when there are bugs in UnitsNet.Serialization.JsonNet v1.*, you can fix it in that version and release a new UnitsNet.Serialization.JsonNet v1.* package and the compatibility test will still work.

But for now, UnitsNet.Serialization.JsonNet v1.* is compatible with UnitsNet v3.* and UnitsNet v4.* so there shouldn't really be a need for the compatibility test...

Scenario:

UnitsNet v3.* + UnitsNet.Serialization.JsonNet v1.* works, UnitsNet v4.* + UnitsNet.Serialization.JsonNet v1.* works too. 
Changes made in UnitsNet.Serialization.JsonNet should be tested against UnitsNet v3.* and v4.*. 
If a change turns out to be incompatible with UnitsNet v3.* and there's no way to make it compatible (maybe due to internal changes in UnitsNet) then UnitsNet.Serialization.JsonNet should be bumped in major (v2.*)

Now we have the following situation:
- UnitsNet v3.* + UnitsNet.Serialization.JsonNet v1.*
- UnitsNet v4.* + UnitsNet.Serialization.JsonNet v2.*

When a bug is reported in UnitsNet.Serialization.JsonNet, it should be checked that
a.) it exists in both versions (v1.* and v2.*)
b.) should be fixed in both versions

dschuermans:
In the meanwhile i'll be grabbing a copy of the json converter and add it directly to our project, since it's we're working on json serialization and deserialization atm and I can't wait for this to be resolved. I'll be posting other bugs as I find them so that I don't forget they need fixing

dschuermans:
So you'll have a "master" branch per major version then?

What you proposed sounds good to me, i'm still a newb when it comes down to versioning lol

angularsen:
We're using github flow, which means `master` is always the latest version. Right now it is v3, in December it  will merge in `v4` branch and from that point on it will be v4. Later, v5 will merge in.

If you want to maintain older major versions then you branch out that. In this case, just before merging in v4 branch into master we will create a `v3` branch for the latest master commit and if we later need to publish a new v3 nuget with a bugfix we can do so on that branch.

dschuermans:
Yup sounds about right 😄

angularsen:
Just pinging this, do you still intend to follow through with this PR?
Please note that v4 branch was merged into master on December 16, so you have a merge conflict to resolve.

No rush, enjoy the holidays :-)

dschuermans:
Yea, i'll have a look at it the coming weeks if i have some time

angularsen:
Just a friendly ping :-) There is no rush, but if you don't have time to follow up on this anytime soon we can close it and reopen it later instead. Let me know what you want, cheers!

angularsen:
I'm sorry friend, but I am closing this pull request due to inactivity. Please reopen if you want to continue the work.

PR #575, opened 2018-12-16
Temperature standard arithmetic (WIP)

### LINKED ISSUE #518: Temperature times a double

angularsen:
**Do not merge.** 

This change is still up for debate. See discussions in #518, #550 and #560.

Created this PR just to have all the work of #550 and #560 in one place in case we want to standardize temperature arithmetic later. We decided to postpone this for v4 as we couldn't agree on whether this change was for the better or not.

- Generate default arithmetic (no special case returning `TemperatureDelta`)
- Remove `Multiply()` and `Divide()` methods
- Remove custom `-` operator overload
- Add `Temperature.ToDelta(other)` instance method
- Add `TemperatureDelta.FromTemperatures(left, right)` static method
- Create tests for addition and subtraction of Temperature
- Fix arithmetic code generation for all quantities, it produced wrong results for Temperature and would also be wrong if any other quantities had units that have different zero values (`ax + b` linear relationship instead of just a factor conversion `ax`).

Fixes #518


angularsen:
Closing this. We can keep the branch alive for some time and can always reopen the PR later.

PR #578, opened 2018-12-17
Added MapUnitToDefaultAbbreviation

### LINKED ISSUE #577: Custom abbreviation doesn't ToString() well
#### Problem

If I add a custom abbreviation:

```csharp
UnitAbbreviationsCache.Default.MapUnitToAbbreviation(VolumeUnit.UsCustomaryCup, "Cup");
```

it parses just fine.  However, when I call `.ToString()`, I still get the default abbreviation of `""`.

> **NOTE** I understand that there is ambiguity between "cup" units, and that's why it's defaulted to an empty string.  I'm using the code above to override this default and use US Customary Cup for my app.

#### Cause

The above method merely appends the abbreviation to the list of acceptable values.

https://github.com/angularsen/UnitsNet/blob/87dc3c79e6b7d9b5719c6b310c5c33bd98349a6c/UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs#L72-L82

However, the `.ToString()` methods on classes only use the first abbreviation in the list.

https://github.com/angularsen/UnitsNet/blob/87dc3c79e6b7d9b5719c6b310c5c33bd98349a6c/UnitsNet/CustomCode/UnitAbbreviationsCache.cs#L194

#### Proposal

Add a new method

```csharp
void MapUnitToDefaultAbbreviation<TUnitType>(TUnitType unit, string abbreviation) where TUnitType : Enum
{
     ...
}
```

that would insert the new abbreviation at the beginning of the list rather than appending it to the end.  Then when `.ToString()` is called, it can still take the first abbreviation.

gregsdennis:
Resolves #577

### DIFF
--- a/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
+++ b/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
@@ -309,12 +309,22 @@ namespace UnitsNet.Tests
         }
 
         [Fact]
-        public void MapUnitToDefaultAbbreviation_AddCustomAbbreviation_UsesNewAbbreviationWithToString()
+        public void MapUnitToDefaultAbbreviation_GivenUnitAndCulture_SetsDefaultAbbreviationForUnitAndCulture()
         {
             var cache = new UnitAbbreviationsCache();
             cache.MapUnitToDefaultAbbreviation(AreaUnit.SquareMeter, AmericanCulture, "m^2");
 
-            Assert.Equal("m^2", cache.GetDefaultAbbreviation(AreaUnit.SquareMeter));
+            Assert.Equal("m^2", cache.GetDefaultAbbreviation(AreaUnit.SquareMeter, AmericanCulture));
+        }
+
+        [Fact]
+        public void MapUnitToDefaultAbbreviation_GivenCustomAbbreviation_SetsAbbreviationUsedByQuantityToString()
+        {
+            // Use a distinct culture here so that we don't mess up other tests that may rely on the default cache.
+            var newZealandCulture = GetCulture("en-NZ");
+            UnitAbbreviationsCache.Default.MapUnitToDefaultAbbreviation(AreaUnit.SquareMeter, newZealandCulture, "m^2");
+
+            Assert.Equal("1 m^2", Area.FromSquareMeters(1).ToString(newZealandCulture));
         }
 
         /// <summary>
--- a/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
+++ b/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
@@ -308,6 +308,15 @@ namespace UnitsNet.Tests
             Assert.Equal("m²", cache.GetDefaultAbbreviation(AreaUnit.SquareMeter));
         }
 
+        [Fact]
+        public void MapUnitToDefaultAbbreviation_AddCustomAbbreviation_UsesNewAbbreviationWithToString()
+        {
+            var cache = new UnitAbbreviationsCache();
+            cache.MapUnitToDefaultAbbreviation(AreaUnit.SquareMeter, AmericanCulture, "m^2");
+
+            Assert.Equal("m^2", cache.GetDefaultAbbreviation(AreaUnit.SquareMeter));
+        }
+
         /// <summary>
         ///     Convenience method to the proper culture parameter type.
         /// </summary>
--- a/UnitsNet/CustomCode/UnitAbbreviationsCache.cs
+++ b/UnitsNet/CustomCode/UnitAbbreviationsCache.cs
@@ -91,6 +91,26 @@ namespace UnitsNet
         }
 
         /// <summary>
+        /// Adds a unit abbreviation for the given unit enum value and sets it as the default.
+        /// This is used to dynamically add abbreviations for existing unit enums such as <see cref="UnitsNet.Units.LengthUnit"/> or to extend with third-party unit enums
+        /// in order to <see cref="UnitParser.Parse{TUnitType}"/> or <see cref="GetDefaultAbbreviation{TUnitType}"/> on them later.
+        /// </summary>
+        /// <param name="unit">The unit enum value.</param>
+        /// <param name="abbreviation">Unit abbreviations to add as default.</param>
+        /// <typeparam name="TUnitType">The type of unit enum.</typeparam>
+        [PublicAPI]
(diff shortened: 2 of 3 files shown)


UnitsNet/CustomCode/UnitAbbreviationsCache.cs
```diff
@@ -90,6 +90,26 @@ private void LoadGeneratedAbbreviations()
...
+        /// <param name="abbreviation">Unit abbreviations to add as default.</param>
+        /// <typeparam name="TUnitType">The type of unit enum.</typeparam>
+        [PublicAPI]
+        // Windows Runtime Component does not allow public methods/ctors with same number of parameters: https://msdn.microsoft.com/en-us/library/br230301.aspx#Overloaded methods
+#if WINDOWS_UWP
+        internal
+#else
+        public
+#endif
+            void MapUnitToDefaultAbbreviation<TUnitType>(TUnitType unit, string abbreviation) where TUnitType : Enum
```
angularsen: Can we maybe add a `bool setAsDefault = false` parameter to the existing `MapUnitToAbbreviation` methods instead? Windows Runtime Component should be fine since it's already `internal` for that target.

UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
```diff
@@ -69,15 +69,21 @@ internal List<int> GetUnitsForAbbreviation(string abbreviation)
             return units.Distinct().ToList();
         }
 
-        internal void Add(int unit, string abbreviation)
+        internal void Add(int unit, string abbreviation, bool useAsDefault = false)
```
angularsen: If you add `setAsDefault` as I propose above, make sure these two names match. I slightly prefer `setAs` over `useAs`, but it's not a big deal.

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
gregsdennis: I thought of that, but adding default values to `params` methods is gross.  Also, I only want a single default.

UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
gregsdennis: I can change the parameter name.

gregsdennis: ✔️

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
angularsen: Ah I forgot it had a `params`, then I agree. 

But it seems we have a bit of duplicate code in the new and old methods, can we maybe:
1. Extract common code to a private method
2. Rename `MapUnitToDefaultAbbrevation` to `SetDefaultAbbreviationForUnit`, I'm not too fond of the old naming scheme
3. Add unit tests for the new methods, see how we test the old methods (I think we have some)

gregsdennis: Sure.  I'll put some of that together and update.

gregsdennis: ✔️ see https://github.com/angularsen/UnitsNet/pull/578/files#diff-96c7d0bea2108a11d25f686fd1687acaR209

gregsdennis: Um... regarding the tests... it seems that only a few of the generated test base classes are being produced.  I'm missing the vast majority of them.

gregsdennis: Nevermind.  Found the batch file.  It's really named quite well. 😄

gregsdennis: I still like the "map" word here since it's still technically mapping, if the abbreviation doesn't exist.  But it's your library. If you really want "set," I'll change it to "set."

angularsen: I see your point and I can go either way. It's not a big deal for me and we can keep it as-is. 
My biggest gripe with the naming is that it seems backwards to me. In my mind, we don't map a unit to an abbreviation, we map abbrevation(s) to a unit. But can't change the old method name now, so let's just go with it for now. It's fine.

UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
```diff
@@ -308,6 +308,15 @@ public void MapUnitToAbbreviation_AddCustomUnit_DoesNotOverrideDefaultAbbreviati
             Assert.Equal("m²", cache.GetDefaultAbbreviation(AreaUnit.SquareMeter));
         }
 
+        [Fact]
+        public void MapUnitToDefaultAbbreviation_AddCustomAbbreviation_UsesNewAbbreviationWithToString()
```
angularsen: `UsesNewAbbreviationWithToString` naming doesn't seem to match what you are testing here. I think you want two test cases here:
1. Rename this test case to `MapUnitToDefaultAbbreviation_GivenUnitAndCulture_SetsDefaultAbbreviationForUnitAndCulture`, and make sure to specify the same culture in `GetDefaultAbbreviation()` call
2. Add new test case `MapUnitToDefaultAbbreviation_GivenCustomAbbreviation_SetsAbbreviationUsedByQuantityToString` that tests how `Area.FromSquareMeters(1).ToString(IFormatProvider)` returns the new custom abbreviation

angularsen: Implementation looks good to me, so just some improvements to tests and this should be good to go 👍

gregsdennis: Interesting... that new test fails.  Somehow I think you knew it would...

gregsdennis: Ah... the `ToString()` uses the default cache, so creating a new cache in the test is useless.

angularsen: I didn't :-) The new code structure in 4.0 is still a bit new to me, but yes, these ToString() methods that rely on static types is a bit of a pain to test.

angularsen:
Nuget 4.1.0 on the way out 👏 

https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.1.0

PR #582, opened 2019-01-11
Added two units: Mass.Grain & SpecificEnergy.BtuPerPound

mp9007:
Add SpecificEnergy.BtuPerPound
Add Mass.Grains


Common/UnitDefinitions/Mass.json
```diff
@@ -165,6 +165,20 @@
           "Abbreviations": [ "cwt" ]
         }
       ]
+    },
+    {
+      "SingularName": "Grain",
+      "PluralName": "Grains",
+      "FromUnitToBaseFunc": "x/15432.35811478375",
```
angularsen: My calculator based on the xmldoc summary gives me: `1/(64.79891 / 1e6)=15432.358352941430650608166094152`

This source agrees: `1.5432358353e4`
https://www.convertworld.com/en/mass/grain-avoirdupois-us.html

Is there perhaps slightly different definitions of this unit?

mp9007: I think your value 15432.358352941431 is better, I had trouble to find online converter that show enough decimal. I revised other sources and they all fit your calculation. I will update the files.

PR #583, opened 2019-01-14
Adding pressure change rate units

ngeslin:
Adding MPa/min and KPa/min units


ngeslin:
Incrementation of the version number was unintentional (bad manipulation of scripts). I reverted it before adding the new units

build.bat
```diff
@@ -1,2 +1,3 @@
 @echo off
 powershell -ExecutionPolicy Bypass -NoProfile -File %~dp0\Build\build.ps1
+pause
```
angularsen: I get why you added this, it's convenient when double-clicking, however it also means it will pause when run from command line, which is my primary usecase for it.

May I suggest one of these options:
1. Create a shortcut (outside the repo folder) that runs `cmd /K .\UnitsNet\build.bat` (see [answer](https://stackoverflow.com/a/886858/134761))
2. Create a new bat file `build-with-pause.bat`

I personally prefer alternative 1.

angularsen: The rest of the PR looks good!

ngeslin: I encountered some errors when using the script and, indeed, I forgot to remove that line. Thanks for the tips.

angularsen:
Nuget on the way out [UnitsNet/4.4.0](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.4.0)

PR #584, opened 2019-01-16
Case insensitive unit parsing and conversion.

TrevorVonSeggern:
The purpose of this is to allow this library to allow for "year" and "Year" in the unit parser and UnitConverter without adding different casing to every abbreviation in the unit definitions.

### DIFF
--- a/UnitsNet.Tests/UnitConverterTest.cs
+++ b/UnitsNet.Tests/UnitConverterTest.cs
@@ -37,13 +37,13 @@ namespace UnitsNet.Tests
         }
 
         [Fact]
-        public void ConvertByName__QuantityCaseInsensitive()
+        public void ConvertByName_QuantityCaseInsensitive()
         {
             Assert.Equal(0, UnitConverter.ConvertByName(0, "length", "Meter", "Centimeter"));
         }
 
         [Fact]
-        public void ConvertByName__UnitTypeCaseInsensitive()
+        public void ConvertByName_UnitTypeCaseInsensitive()
         {
             Assert.Equal(0, UnitConverter.ConvertByName(0, "Length", "meter", "Centimeter"));
         }
--- a/UnitsNet.Tests/UnitConverterTest.cs
+++ b/UnitsNet.Tests/UnitConverterTest.cs
@@ -26,6 +26,7 @@ namespace UnitsNet.Tests
     public class UnitConverterTest
     {
         [Theory]
+        [InlineData(0, 0, "length", "meter", "centimeter")]
         [InlineData(0, 0, "Length", "Meter", "Centimeter")]
         [InlineData(100, 1, "Length", "Meter", "Centimeter")]
         [InlineData(1, 1000, "Mass", "Gram", "Kilogram")]
@@ -33,6 +34,18 @@ namespace UnitsNet.Tests
         public void ConvertByName_ConvertsTheValueToGivenUnit(double expectedValue, double inputValue, string quantityTypeName, string fromUnit, string toUnit)
         {
             Assert.Equal(expectedValue, UnitConverter.ConvertByName(inputValue, quantityTypeName, fromUnit, toUnit));
+        }
+
+        [Fact]
+        public void ConvertByName__QuantityCaseInsensitive()
+        {
+            Assert.Equal(0, UnitConverter.ConvertByName(0, "length", "Meter", "Centimeter"));
+        }
+
+        [Fact]
+        public void ConvertByName__UnitTypeCaseInsensitive()
+        {
+            Assert.Equal(0, UnitConverter.ConvertByName(0, "Length", "meter", "Centimeter"));
         }
 
         [Theory]
--- a/UnitsNet.Tests/UnitParserTests.cs
+++ b/UnitsNet.Tests/UnitParserTests.cs
@@ -44,30 +44,24 @@ namespace UnitsNet.Tests
         [Fact]
         public void Parse_AbbreviationCaseInsensitive_Lowercase_years()
         {
-            // Arrange
             var abbreviation = "years";
             var expected = DurationUnit.Year365;
             var parser = UnitParser.Default;
 
-            // Act
             var actual = parser.Parse<DurationUnit>(abbreviation);
 
-            // Assert
             Assert.Equal(expected, actual);
         }
 
         [Fact]
         public void Parse_AbbreviationCaseInsensitive_Uppercase_Years()
         {
-            // Arrange
             var abbreviation = "Years";
             var expected = DurationUnit.Year365;
             var parser = UnitParser.Default;
 
-            // Act
             var actual = parser.Parse<DurationUnit>(abbreviation);
 
-            // Assert
             Assert.Equal(expected, actual);
         }
 
--- a/UnitsNet.Tests/UnitParserTests.cs
+++ b/UnitsNet.Tests/UnitParserTests.cs
@@ -42,19 +42,32 @@ namespace UnitsNet.Tests
         }
 
         [Fact]
-        public void Parse_AbbreviationCaseInsensitive()
+        public void Parse_AbbreviationCaseInsensitive_Lowercase_years()
         {
+            // Arrange
             var abbreviation = "years";
             var expected = DurationUnit.Year365;
             var parser = UnitParser.Default;
 
+            // Act
             var actual = parser.Parse<DurationUnit>(abbreviation);
 
+            // Assert
             Assert.Equal(expected, actual);
+        }
 
-            abbreviation = "Years";
-            actual = parser.Parse<DurationUnit>(abbreviation);
+        [Fact]
+        public void Parse_AbbreviationCaseInsensitive_Uppercase_Years()
+        {
+            // Arrange
+            var abbreviation = "Years";
+            var expected = DurationUnit.Year365;
+            var parser = UnitParser.Default;
 
+            // Act
+            var actual = parser.Parse<DurationUnit>(abbreviation);
+
+            // Assert
             Assert.Equal(expected, actual);
         }
 
--- a/UnitsNet.Tests/UnitParserTests.cs
+++ b/UnitsNet.Tests/UnitParserTests.cs
@@ -42,6 +42,23 @@ namespace UnitsNet.Tests
         }
 
         [Fact]
+        public void Parse_AbbreviationCaseInsensitive()
+        {
+            var abbreviation = "years";
+            var expected = DurationUnit.Year365;
+            var parser = UnitParser.Default;
+
+            var actual = parser.Parse<DurationUnit>(abbreviation);
+
+            Assert.Equal(expected, actual);
+
+            abbreviation = "Years";
+            actual = parser.Parse<DurationUnit>(abbreviation);
+
+            Assert.Equal(expected, actual);
+        }
+
+        [Fact]
         public void Parse_UnknownAbbreviationThrowsUnitNotFoundException()
         {
             Assert.Throws<UnitNotFoundException>(() => UnitParser.Default.Parse<AreaUnit>("nonexistingunit"));
--- a/UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
+++ b/UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
@@ -63,22 +63,21 @@ namespace UnitsNet
 
         internal List<int> GetUnitsForAbbreviation(string abbreviation)
         {
-            abbreviation = abbreviation.ToLower();
-            if(!abbreviationToUnitMap.TryGetValue(abbreviation, out var units))
-                abbreviationToUnitMap[abbreviation] = units = new List<int>();
+            var lowerCaseAbbreviation = abbreviation.ToLower();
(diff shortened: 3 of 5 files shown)


UnitsNet.Tests/UnitParserTests.cs
```diff
@@ -41,6 +41,23 @@ public void Parse_ReturnsUnitMappedByCustomAbbreviation(string customAbbreviatio
...
+        [Fact]
+        public void Parse_AbbreviationCaseInsensitive()
+        {
+            var abbreviation = "years";
+            var expected = DurationUnit.Year365;
+            var parser = UnitParser.Default;
+
+            var actual = parser.Parse<DurationUnit>(abbreviation);
+
+            Assert.Equal(expected, actual);
```
angularsen: I would like to advocate [AAA-pattern](https://docs.telerik.com/devtools/justmock/basic-usage/arrange-act-assert).
Basically, one unit test should test one thing, and it should be split in arrange-act-assert parts.

Either split into two tests or consider using `[InlineData]` to pass in parameters with input values and expected value.

UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
```diff
@@ -63,10 +63,9 @@ internal List<string> GetAbbreviationsForUnit(int unit)
 
         internal List<int> GetUnitsForAbbreviation(string abbreviation)
         {
-            if(!abbreviationToUnitMap.TryGetValue(abbreviation, out var units))
-                abbreviationToUnitMap[abbreviation] = units = new List<int>();
-
-            return units.Distinct().ToList();
+            return abbreviationToUnitMap
+                .Where(x => x.Key.Equals(abbreviation, StringComparison.OrdinalIgnoreCase))
+                .SelectMany(x => x.Value).Distinct().ToList();
```
angularsen: Put Distinct and ToList on each a new line.
The exception is if the total chain is so short the entire thing can stay on a single line and still be readable.

UnitsNet/UnitConverter.cs
```diff
@@ -384,8 +385,9 @@ private static bool HasParameterTypes(MethodInfo methodInfo, params Type[] expec
         private static bool TryParseUnit(Type unitType, string unitName, out object unitValue)
         {
             unitValue = null;
-
-            if(!Enum.IsDefined(unitType, unitName))
+            var eNames = Enum.GetNames(unitType);
+            unitName = eNames.FirstOrDefault(x => x.Equals(unitName, StringComparison.OrdinalIgnoreCase));
+            if(unitName is null)
```
angularsen: Let's use `== null` for consistency.

UnitsNet/UnitConverter.cs
```diff
@@ -395,22 +397,30 @@ private static bool TryParseUnit(Type unitType, string unitName, out object unit
...
+                x.Name.Equals(quantityName, StringComparison.OrdinalIgnoreCase));
 
-            unitType = UnitsNetAssembly.GetType(unitTypeName); // ex: UnitsNet.Units.LengthUnit enum
             if(unitType == null)
                 return false;
 
             return true;
         }
 
+        private static List<Type> QuantityTypes = UnitsNetAssembly.GetTypes()
```
angularsen: Make it `readonly` and of type `IReadOnlyList<Type>`. Same for `UnitTypes` above. 
Move all fields to the top, above any methods or constructors.

UnitsNet/UnitConverter.cs
```diff
@@ -395,22 +397,30 @@ private static bool TryParseUnit(Type unitType, string unitName, out object unit
...
+            .Where(x => x.FullName.StartsWith(QuantityNamespace))
+            .ToList();
+
         private static bool TryGetQuantityType(string quantityName, out Type quantityType)
         {
-            string quantityTypeName = $"{QuantityNamespace}.{quantityName}";
-
-            quantityType = UnitsNetAssembly.GetType(quantityTypeName); // ex: UnitsNet.Length struct
```
angularsen: Nice improvement of caching this lookup of types.

UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
```diff
@@ -63,10 +63,9 @@ internal List<string> GetAbbreviationsForUnit(int unit)
 
         internal List<int> GetUnitsForAbbreviation(string abbreviation)
         {
-            if(!abbreviationToUnitMap.TryGetValue(abbreviation, out var units))
-                abbreviationToUnitMap[abbreviation] = units = new List<int>();
-
-            return units.Distinct().ToList();
+            return abbreviationToUnitMap
+                .Where(x => x.Key.Equals(abbreviation, StringComparison.OrdinalIgnoreCase))
```
angularsen: I'm a bit on the fence on whether to use OrdinalIgnoreCase or InvariantCultureIgnoreCase.
https://stackoverflow.com/a/72766/134761

I guess since we are not parsing random user input, but our own type names and unit abbreviations, it's perhaps fine have exact matching (ordinal) instead of culture-specific matching. It is faster too.

UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
tmilnthorp: This will be slower than since it's doing a linear search rather than a dictionary lookup. Perhaps we could achieve both speed and case insensitivity by just doing a ToLower when storing or looking up things instead?

angularsen: Ah, good point, I missed that. Yup, I agree with you.

UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
TrevorVonSeggern: I originally wanted to use InvariantCultureIgnoreCase, but it wouldn't build. Because we are defining our abbreviations statically, either string comparison methods will work.

UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
TrevorVonSeggern: If I change the abbreviationToUnitMap to store the lower case version, then it breaks several unit tests. Specifically where we are using this to get the abbreviation on ToString.
10 °C => 10 °c (on .ToString(CultureInfo.InvariantCulture, "{0:0} {1}"))
We could add another abbreviationToUnitMap, or possibly change one of the types of the dictionary to accommodate having a dictionary lookup. Do you think it's worth the added complexity/memory space?

UnitsNet.Tests/UnitParserTests.cs
angularsen: This is good, but I would suggest dropping the comments. At least for me, when code is divided in 3 paragraphs like this, it is already dead clear what each do. I also omit `// Arrange`, because it's already clear to me that everything before `// Act` is setup code.

UnitsNet/UnitConverter.cs
```diff
@@ -45,6 +45,14 @@ public static class UnitConverter
         private static readonly string UnitTypeNamespace = typeof(LengthUnit).Namespace;
         private static readonly Assembly UnitsNetAssembly = typeof(Length).GetAssembly();
 
+        private static readonly List<Type> QuantityTypes = UnitsNetAssembly.GetTypes()
```
angularsen: Since we are targeting .NET 4.0 we probably can't use `IReadOnlyList<T>` as I originally advised, I believe that came in .NET 4.5. I would recommend using an array and `.ToArray()` then, just to make it more clear that this is not supposed to be manipulated.

UnitsNet/UnitConverter.cs
```diff
@@ -384,8 +393,9 @@ private static bool HasParameterTypes(MethodInfo methodInfo, params Type[] expec
         private static bool TryParseUnit(Type unitType, string unitName, out object unitValue)
         {
             unitValue = null;
-
-            if(!Enum.IsDefined(unitType, unitName))
+            var eNames = Enum.GetNames(unitType);
+            unitName = eNames.FirstOrDefault(x => x.Equals(unitName, StringComparison.OrdinalIgnoreCase));
```
tmilnthorp: We should probably use Enum.TryParse which includes a bool parameter to specify ignoring case. See [here](https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=netframework-4.7.2#System_Enum_TryParse__1_System_String_System_Boolean___0__).

UnitsNet.Tests/UnitConverterTest.cs
```diff
@@ -35,6 +36,18 @@ public void ConvertByName_ConvertsTheValueToGivenUnit(double expectedValue, doub
...
         }
 
+        [Fact]
+        public void ConvertByName__QuantityCaseInsensitive()
+        {
+            Assert.Equal(0, UnitConverter.ConvertByName(0, "length", "Meter", "Centimeter"));
+        }
+
+        [Fact]
+        public void ConvertByName__UnitTypeCaseInsensitive()
```
angularsen: Is the double underscore intentional here and above?

UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
```diff
@@ -63,6 +63,7 @@ internal List<string> GetAbbreviationsForUnit(int unit)
 
         internal List<int> GetUnitsForAbbreviation(string abbreviation)
         {
+            abbreviation = abbreviation.ToLower();
```
angularsen: I would make this a separate variable `lowercaseAbbreviation` to make it more clear. You can choose the name, but make the variable `lower` in the Add method below the same name for consistency.

UnitsNet/CustomCode/UnitValueAbbreviationLookup.cs
```diff
@@ -71,11 +72,13 @@ internal List<int> GetUnitsForAbbreviation(string abbreviation)
 
         internal void Add(int unit, string abbreviation, bool setAsDefault = false)
         {
+            // abbreviation = abbreviation.ToLower();
```
angularsen: Remove commented code.

UnitsNet/UnitConverter.cs
```diff
@@ -397,9 +407,10 @@ private static bool TryParseUnit(Type unitType, string unitName, out object unit
 
         private static bool TryGetUnitType(string quantityName, out Type unitType)
         {
-            string unitTypeName = $"{UnitTypeNamespace}.{quantityName}Unit";
+            quantityName += "Unit";
```
angularsen: Make it a variable, f.ex. `unitTypeName`. Also I think we can keep the example comment, but modify to `// Example: "LengthUnit"`. I think this is helpful when reading the code.

UnitsNet/UnitConverter.cs
TrevorVonSeggern: This would also prefer this method, but I can't use that here without using reflection, which I was trying to avoid. The try parse method you linked takes a generic type argument, which means that the type has to be known at compile type. We have a Type parameter known only at run time. Is there another TryParse that you know of that takes the enum type "Type" as parameter?

UnitsNet.Tests/UnitConverterTest.cs
TrevorVonSeggern: Unintentional.

UnitsNet/UnitConverter.cs
```diff
@@ -407,9 +407,9 @@ private static bool TryParseUnit(Type unitType, string unitName, out object unit
 
         private static bool TryGetUnitType(string quantityName, out Type unitType)
         {
-            quantityName += "Unit";
+            var quantityTypeName = quantityName += "Unit"; // ex. LengthUnit
```
angularsen: It should be named `unitTypeName`, quantity would be `Length`.

Edit: I fixed it in the PR.

UnitsNet/UnitConverter.cs
angularsen: `Enum.TryParse<T>(string, bool, out T)` exists, but I suspect since we target .NET 4.0 is why we can't use it.

UnitsNet/UnitConverter.cs
```diff
@@ -44,6 +45,14 @@ public static class UnitConverter
         private static readonly string UnitTypeNamespace = typeof(LengthUnit).Namespace;
         private static readonly Assembly UnitsNetAssembly = typeof(Length).GetAssembly();
 
+        private static readonly Type[] QuantityTypes = UnitsNetAssembly.GetTypes()
```
angularsen: These include a lot more types than just the quantity structs and the unit enums.
I'm adding a fix.

angularsen:
Build is failing on WRC target with my latest changes, fixing.

angularsen:
Woops, found a bug. We lowercase the prefixes so we can no longer distinguish `milli` from `mega` :-)

image

As a result, `Pressure.Parse("3 Mbar")` fails, because there are two abbreviations it matches.
I'm looking into if there is a quick fix. Creating a new issue #590.

angularsen:
Nuget 4.5.1 on the way out.

https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.5.1

PR #588, opened 2019-01-25
Dynamic unit conversions

tmilnthorp:
- UnitConverter: Change from hard coded switch statement to lookup of conversion functions
- Allow consumers to add their own conversion functions between their own quantity types

Idea for #478. I'll add more commentary there.


tmilnthorp:
This isn't completely fleshed out yet. Just to show an idea, so don't merge it yet.

Comments in #478

UnitsNet/UnitConverter.cs
```diff
@@ -20,30 +20,124 @@
...
+        static UnitConverter()
+        {
+            Default = new UnitConverter();
+
+            RegisterDefaultConversions(Default);
+        }
+
+        public UnitConverter()
+        {
+            conversionFunctions = new Dictionary<ConversionFunctionLookup, ConversionFunction>();
```
angularsen: Can initialize this on field and also make field readonly. Also use `_` prefix on private instance fields.

UnitsNet/GeneratedCode/Quantities/ElectricInductance.NetFramework.g.cs
```diff
@@ -582,14 +582,18 @@ public ElectricInductance ToUnit(ElectricInductanceUnit unit)
         ///     This is typically the first step in converting from one unit to another.
         /// </summary>
         /// <returns>The value in the base unit representation.</returns>
-        private double AsBaseUnit()
+        internal ElectricInductance AsBaseUnit()
```
angularsen: `ToBaseUnit()` to be consistent with `ToUnit()`?

angularsen: This change seems like extra overhead to me, is it needed? If we later convert to `class`, doesn't this introduce an additional object allocation for every unit conversion? 

I think the order seems reverse to me.
- Instead of `double AsBaseNumericType()` calling the "heavier" `MyQuantity AsBaseUnit()` method, shouldn't rather `AsBaseUnit()` call `AsBaseNumericType()`? Or am I missing something?

While touching this code:
- Can we please rename `AsBaseNumericType()` to something more intuitive, maybe `GetValueInBaseUnit()`? I keep forgetting what it does and it's name is just confusing to me.
- Can we also use explicit types instead of `var` in places where it is not obvious what the type is, such as https://github.com/angularsen/UnitsNet/blob/b4a42777ba3f7d383c50a1c617c4c61fe875f99b/UnitsNet/GeneratedCode/Quantities/ElectricInductance.NetFramework.g.cs#L607-L608

tmilnthorp: I think those are all fair enough!

UnitsNet.WindowsRuntimeComponent/GeneratedCode/Quantities/Acceleration.WindowsRuntimeComponent.g.cs
```diff
@@ -671,48 +671,45 @@ public Acceleration ToUnit(AccelerationUnit unit)
...
             }
         }
 
-        private double AsBaseNumericType(AccelerationUnit unit)
+        /// <summary>
+        ///     Converts the current value + unit to the base unit.
+        ///     This is typically the first step in converting from one unit to another.
+        /// </summary>
+        /// <returns>The value in the base unit representation.</returns>
+        internal Acceleration ToBaseUnit()
```
angularsen: Ah, this read so much better!

UnitsNet/GeneratedCode/UnitConverter.g.cs
```diff
@@ -49,1689 +49,1689 @@ public static void RegisterDefaultConversions(UnitConverter unitConverter)
...
+            unitConverter.SetConversionFunction<Density>(DensityUnit.PoundPerCubicFoot, Density.BaseUnit, (q) => ((Density)q).ToBaseUnit());
             unitConverter.SetConversionFunction<Density>(Density.BaseUnit, DensityUnit.PoundPerCubicInch, (q) => ((Density)q).ToUnit(DensityUnit.PoundPerCubicInch));
-            unitConverter.SetConversionFunction<Density>(DensityUnit.PoundPerCubicInch, Density.BaseUnit, (q) => ((Density)q).AsBaseUnit());
+            unitConverter.SetConversionFunction<Density>(DensityUnit.PoundPerCubicInch, Density.BaseUnit, (q) => ((Density)q).ToBaseUnit());
             unitConverter.SetConversionFunction<Density>(Density.BaseUnit, DensityUnit.PoundPerImperialGallon, (q) => ((Density)q).ToUnit(DensityUnit.PoundPerImperialGallon));
-            unitConverter.SetConversionFunction<Density>(DensityUnit.PoundPerImperialGallon, Density.BaseUnit, (q) => ((Density)q).AsBaseUnit());
+            unitConverter.SetConversionFunction<Density>(DensityUnit.PoundPerImperialGallon, Density.BaseUnit, (q) => ((Density)q).ToBaseUnit());
             unitConverter.SetConversionF
```
angularsen: Eeek 😱 That is a big one!

UnitsNet/UnitConverter.cs
```diff
@@ -133,11 +131,20 @@ public bool TryGetConversionFunction(Type fromType, Enum from, Type toType, Enum
...
         }
 
 #endif
 
+        private static readonly Type[] QuantityTypes = UnitsNetAssembly.GetTypes()
+            .Where(typeof(IQuantity).IsAssignableFrom)
+            .Where(x => x.IsClass() || x.IsValueType()) // Future-proofing: we are discussing changing quantities from struct to class
+            .ToArray();
+
+        private static readonly Type[] UnitTypes = UnitsNetAssembly.GetTypes()
```
angularsen: There is some overlap here and in my PR, we should later merge so we have a single way to enumerate all unit types and quantity types.

angularsen:
Adding a summary from prior discussion here:

As for direct conversions, I think that sounds like a good idea regardless:

1. A single piece of code to handle all conversions, generically
1. Route existing unit-base and base-unit conversions through it
1. Ability to add higher-precision direct conversions
1. Ability to add cross-quantity conversions
1. Ability to plug in 3rd party conversions of 3rd party unit enums
1. Compatible with the new dynamic Quantity in #576 for converting dynamically

angularsen:
Re-reading this now.

UnitsNet/GeneratedCode/UnitConverter.g.cs
```diff
@@ -0,0 +1,1742 @@
...
+// ReSharper disable RedundantCommaInArrayInitializer
+// ReSharper disable once CheckNamespace
+
+namespace UnitsNet
+{
+    public sealed partial class UnitConverter
+    {
+        public static void RegisterDefaultConversions(UnitConverter unitConverter)
+        {
+            unitConverter.SetConversionFunction<Acceleration>(Acceleration.BaseUnit, AccelerationUnit.CentimeterPerSecondSquared, (q) => ((Acceleration)q).ToUnit(AccelerationUnit.CentimeterPerSecondSquared));
```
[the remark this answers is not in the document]
angularsen: To be clear, we don't have to wait for that benchmark to merge. It just made me think of it.

angularsen:
When this is merged, this wiki should be updated on the new capability to add unit conversions:
https://github.com/angularsen/UnitsNet/wiki/Extending-with-Custom-Units

tmilnthorp:
More work to do here (including doc obviously after warnings-as-errors). Just merging to keep up to date.

angularsen:
@tmilnthorp  I took this for spin locally and added some improvements. I think this is ready to merge now, but waiting for you to check out the latest changes.

I was specifically interested in seeing how we can improve support for third party quantities and units and added some test cases for that. It works pretty well by now, you can parse units, parse quantities and convert between quantities using only custom types. We still enforce the requirement of implementing `IQuantity`, which I did find awkward when adding an example `HowMuch` quantity since that interface is pretty bloated and most of it is not used for simple conversions and parsing.

I think there is definitely an opportunity to support simple POCO classes and structs without them having to implement `IQuantity`, but I didn't want to clutter this PR with that.

If you have time, I'd also love to merge #656 .

angularsen:
This needs to port the powershell changes to the new C# codegen scripts.

angularsen:
@tmilnthorp I think this one is good to go. Two months ago I made some changes to it and I'd like you to look over them whenever you find the time so we can get this puppy merged :)

tmilnthorp:
Looks good to me!

angularsen:
Sweet!

angularsen:
[Release UnitsNet/4.30.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.30.0)

PR #591, opened 2019-01-28
Fix parsing ambiguous abbreviations when lowercase

### LINKED ISSUE #590: Parsing mega og milli abbreviations fail with AmbiguousUnitParseException
After #584, we lowercase abbreviations, including the prefixes, which means we can no longer distinguish `milli` from `mega`.

As a result, `Pressure.Parse("3 Mbar")` fails, because there are two abbreviations it matches.

image

angularsen:
Fixes #590 

- Add separate abbreviation-to-unit mapping that preserves case sensitivity
- Fall back to case sensitive mapping if more than 1 unit is found
- Add test

I'm sure we can optimize this with a single map, but I don't think we're talking any significant memory footprint here.

### DIFF
--- a/UnitsNet.Tests/UnitParserTests.cs
+++ b/UnitsNet.Tests/UnitParserTests.cs
@@ -66,6 +66,13 @@ namespace UnitsNet.Tests
         }
 
         [Fact]
+        public void Parse_GivenAbbreviationsThatAreAmbiguousWhenLowerCase_ReturnsCorrectUnit()
+        {
+            Assert.Equal(PressureUnit.Megabar, Pressure.ParseUnit("Mbar"));
+            Assert.Equal(PressureUnit.Millibar, Pressure.ParseUnit("mbar"));
+        }
+
+        [Fact]
         public void Parse_UnknownAbbreviationThrowsUnitNotFoundException()
         {
             Assert.Throws<UnitNotFoundException>(() => UnitParser.Default.Parse<AreaUnit>("nonexistingunit"));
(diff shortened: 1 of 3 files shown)


angularsen:
Yeah, it feels yucky to have missed this and I also have second thoughts on the whole case insensitive support.

I think it's nice that you can parse both `5 PSI` and `5 psi` for free, without having to specify both abbreviations. On the other hand, adding some entries in JSON is a one-time effort and also not much work. Also, then we don't allow weird things like `5 PsI`, which could potentially mean something else entirely and then they would get the wrong unit instead of UnitNotFoundException.

I'm a bit torn, on one hand we are in a good position to make it super easy to parse user input without being too strict on perfect casing that I think many often get wrong. It's hard to do this outside the library. 

On the other hand, it feels a bit hacky, but in the implementation at least ensures there is no ambiguity by falling back to case sensitivity matching if more than one unit matches. 

The only problem I can think of is when trying to parse a unit we haven't added yet, that happens to be identical to an existing unit with different casing and instead of getting UnitNotFoundException you get the wrong unit. 

To counter this, I propose to make parsing methods case sensitive by default, with a new parameter `ignoreCase`. That should give us the best of both worlds, allowing users to explicitly parse with ignore case?

angularsen:
I'm merging this now, just pushed a new nuget out and I don't want this bug in it. We can revise this later.

angularsen:
https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.5.1

PR #594, opened 2019-01-29
Static properties should be read-only properties with inline initiali…

tmilnthorp:
Static properties should be read-only properties with inline initialization rather than lambdas.

For example, using a lambda for the Zero property would create a new Acceleration(0, BaseUnit) every time it's called, rather than just initializing it once.

A quick test of calling these properties 100,000,000 times showed the lambdas were roughly 10x slower. Not a huge deal but, thought I'd fix it 😃


angularsen:
Nice catch! I didn't notice before now.

PR #596, opened 2019-02-01
Implementing IConvertible

tmilnthorp:


UnitsNet.WindowsRuntimeComponent/GeneratedCode/Quantities/Acceleration.WindowsRuntimeComponent.g.cs
```diff
@@ -784,5 +784,107 @@ private static IFormatProvider GetFormatProviderFromCultureName([CanBeNull] stri
         {
             return cultureName != null ? new CultureInfo(cultureName) : (IFormatProvider)null;
         }
+
+        #region IConvertible Methods
+
+#if !WINDOWS_UWP
```
angularsen: Can't we skip/remove this entire section in WRC files? I think we have a `$wrc` variable in powershell you can use.

angularsen:
Ok, this is kinda neat! A bit unsure about converting to numbers like this, but hey, I guess it's the caller's funeral if they convert to number type and expected a different value :-) Is there any good way to communicate what to expect here? I don't see where to put that xmldoc though. Remarks on the class maybe?

UnitsNet.WindowsRuntimeComponent/GeneratedCode/Quantities/Acceleration.WindowsRuntimeComponent.g.cs
tmilnthorp: Good call. Done!

angularsen:
Ok nice, there's a merge conflict here, but other than that you can just merge it.

tmilnthorp:
Sounds good.

angularsen:
[Release UnitsNet/4.8.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.8.0)

Issue #563, opened 2018-11-22
v5 Roadmap


angularsen:
**TODO Summarize what will make the cut for v5**
#1158 merges v5 to master, prerelease nugets are already out.

What remains before we can call it a release and postpone the rest for v6?

### Temperature arithmetic
In #518 we discussed changing this. Long story short, it's ambiguous what addition and subtraction of `Temperature` should result in, because `0 Celsius != 0 Fahrenheit != 0 Kelvins`. Some people expect the temperature difference (ex: `30 C - 20 C = 10 delta C`), others expect a new temperature (ex: `30 C - 20 C = 303 K - 293 K = 10 K = -263 C`). We currently don't know what ratio of people favor one way over the other, so to make an informed decision I think we need to poll our userbase somehow or find a way to implement this that will cause the least amount of confusion.

PRs #550 + #560 implemented this change, but we realized we were not able to agree on whether to make this change or not in time for v4.

Mobz87:
An Idea:
A global-default unit.

An example:

We could set the global unit for pressure:
`GlobalUnit.Pressure = Pressure.Psi`

If I set a new pressure:
`_pressure.Value = 10;`
Then it is loaded in as 10 Psi

Using `_pressure.Value`
will return `10`

Using `_pressure.Unit`
return 'Psi'

using `_pressure.ToString();`
return `10 Psi`

My idea is not to change the unit for which it does it calculations behind the scene, just changing the unit it using on its in/outputs if nothing else is specified.

tmilnthorp:
Doesn't seem like anything the client application couldn't do where you stash a static "global" or default unit and pass it into the constructor or From method.

I think you're also potentially talking about the idea in #547

angularsen:
I think in general it is not a good idea to tamper with global static variables in a library. If you're using another library or project that also depend on UnitsNet, but expects different defaults, you can easily run into issues here. Also the default behavior shipped with UnitsNet could change in a major version bump, possibly causing similar issues if you have a lot of code just passing in values and assuming some unit will be used. It just sounds flaky with not much benefit in my eyes, but I can understand the convenience sounds appealing.

This kind of convenience is also very easy to build per-application, just create a factory type like this: 
```c#
static class Quantitator2000
{
    public PressureUnit DefaultPressureUnit { get; set; } = PressureUnit.Psi;
    public Pressure Pressure(double value) => new Pressure(value, DefaultPressureUnit);
}

[...]

PR #599, opened 2019-02-04
Initial implementation of IFormattable

### LINKED ISSUE #579: Enhanced IFormattable/ToString Support
Following some discussion on #545:

It would be nice to extend the support for IFormattable and ToString.

From the other thread:

So for IFormattable we should make a list:

- "G" = current default implementation
- "A1", "A2"... "An" = abbreviation 1 or 2 or n in the list
- "A" = first abbreviation (equal to "A1")
- "Q" = quantity name (e.g. "Length")
- "V" value using default ToString() for double/decimal/etc.

```csharp
var length = Length.FromMeters( 3.0 );
var str = length.ToString( "A2" ); 
```

For more complicated strings I think you'd want to specify on a more granular level. We don't want to parse all the number format types. For formatting the value as double, etc. you might want to do the following:

```csharp
var length = Length.FromMeters( 3.0 );
string.Format( "{0,10:E3} [{1:A1}]", length.Value, length );
```

tmilnthorp:
Fixes #579 

Initial go at #579. Didn't get much input there, so see if we like it!


tmilnthorp:
While working on this I realized... any reason we use GlobalConfiguration.DefaultCulture for everything? Why not just use CultureInfo.CurrentCulture? I think that is more what a user would expect. If you changed CurrentCulture right now, you'd also have to change the GlobalConfiguration.DefaultCulture for it to be picked up.

angularsen:
Hm, good point. I don't really see a good reason to keep it, I don't see what it gives us.

To add to it, there is the distinction of `CultureInfo.CurrentCulture` and `CultureInfo.CurrentUICulture` and I think we only have a configuration for the latter. If I recall correctly, the guidelines say something along the lines of
- `CurrentUICulture` for translations/localizations 
- `CurrentCulture` for number formatting

It's probably more nuanced than that, but that's how I remember it. I think you have to manipulate CultureInfo.CurrentCulture to change the default number formatting used by ToString(). If that is correct, then what you propose would be more consistent too. It's a breaking change though, so something to add to #563 .

angularsen:
I'll review the PR later.

tmilnthorp:
I  re-merged with after the script split. I'm curious your thoughts!

UnitsNet/GeneratedCode/Quantities/AmountOfSubstance.NetFramework.g.cs
```diff
@@ -859,6 +859,60 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+        /// <param name="formatProvider">Format to use for localization and number formatting. Defaults to <see cref="GlobalConfiguration.DefaultCulture" /> if null.</param>
+        /// <returns>The string representation.</returns>
+        public string ToString(string format, IFormatProvider formatProvider)
+        {
+            formatProvider = formatProvider ?? GlobalConfiguration.DefaultCulture;
+
+            int digits = 0;
+            string formatString = format;
+
+            if(string.IsNullOrWhiteSpace(formatString))
```
angularsen: NullOrEmpty. If format does not match any specifiers then it should return the format string as-is, if I'm reading the docs right. See earlier comment.

UnitsNet.Tests/QuantityIFormattableTests.cs
```diff
@@ -0,0 +1,57 @@
...
+namespace UnitsNet.Tests
+{
+    public class QuantityIFormattableTests
+    {
+        private static Length length = Length.FromFeet(3.0);
+
+        [Fact]
+        public void GFormatStringEqualsToString()
+        {
+            Assert.Equal(length.ToString("G"), length.ToString());
```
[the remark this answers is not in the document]
tmilnthorp: Indeed it is. I do a ToUpper in the code so it's insensitive. I will add a test though!

UnitsNet.Tests/QuantityIFormattableTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        [Fact]
+        public void SFormatEqualsSignificantDigits()
+        {
+            Assert.Equal(length.ToString(null, 1), length.ToString("S"));
+            Assert.Equal(length.ToString(null, 1), length.ToString("S1"));
+            Assert.Equal(length.ToString(null, 2), length.ToString("S2"));
+            Assert.Equal(length.ToString(null, 3), length.ToString("S3"));
+            Assert.Equal(length.ToString(null, 4), length.ToString("S4"));
```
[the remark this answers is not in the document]
tmilnthorp: 👍

UnitsNet.Tests/QuantityIFormattableTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            Assert.Equal(length.ToString("G"), length.ToString());
+        }
+
+        [Fact]
+        public void EmptyOrNullFormatStringEqualsGFormat()
+        {
+            Assert.Equal(length.ToString("G"), length.ToString(string.Empty));
+            Assert.Equal(length.ToString("G"), length.ToString((string)null));
+            Assert.Equal(length.ToString("G"), length.ToString(" "));
```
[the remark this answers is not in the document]
tmilnthorp: I think it's better to follow more closely and keep it at null or empty with the FormatException for anything else. See example of DateTime [here](https://dotnetfiddle.net/QZomyx).

UnitsNet/GeneratedCode/Quantities/AmountOfSubstance.NetFramework.g.cs
tmilnthorp: I think FormatException is more widely used. See my earlier comment.

UnitsNet.Tests/QuantityIFormattableTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        [Fact]
+        public void VFormatEqualsValueToString()
+        {
+            Assert.Equal(length.Value.ToString(), length.ToString("V"));
+        }
+
+        [Fact]
+        public void QFormatEqualsQuantityName()
+        {
+            Assert.Equal(Length.Info.Name, length.ToString("Q"));
```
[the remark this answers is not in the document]
tmilnthorp: Good idea! Added.

We could certainly do that. Do you want to discuss more and get a broader list before we merge, or merge this and then get more input as we go?

UnitsNet.Tests/QuantityIFormattableTests.cs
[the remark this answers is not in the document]
angularsen: I tried this:
`Console.WriteLine(DateTime.Now.ToString("        y    "));` and it outputs the whitespace with a `19` in there. Interestingly, with no whitespace, it gives me `February 2019` as the output - not just the number 19 anymore. 🤷‍♂️

UnitsNet.Tests/QuantityIFormattableTests.cs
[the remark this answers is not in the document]
angularsen: Good point, I guess there is no standard here with .NET itself using a mix of case sensitive/insensitive.

UnitsNet.Tests/QuantityIFormattableTests.cs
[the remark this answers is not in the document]
angularsen: Human readable can be added later, let's take some baby steps here. We don't even have the human readable forms ready to use.

tmilnthorp:
Looking at the **standard** format strings for:
- [Numeric types](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings)
- [DateTime](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings)

It seems that they use case insensitivity for the default format strings. 

Custom format strings tends to be case sensitive though. Should we follow a similar paradigm? What denotes a custom format vs standard though is what we should probably figure out! 😃

UnitsNet.Tests/QuantityIFormattableTests.cs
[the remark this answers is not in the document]
tmilnthorp: Oops. Replied below.

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -827,6 +827,62 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+
+        /// <summary>
+        /// Gets the string representation of this instance in the specified format string using the specified format provider, or <see cref="GlobalConfiguration.DefaultCulture" /> if null.
+        /// </summary>
+        /// <param name="format">The format string.</param>
+        /// <param name="formatProvider">Format to use for localization and number formatting. Defaults to <see cref="GlobalConfiguration.DefaultCulture" /> if null.</param>
+        /// <returns>The string representation.</returns>
+        public string ToString(string format, IFormatProvider formatProvider)
+        {
+            formatProvider = formatProvider ?? GlobalConfiguration.DefaultCulture;
```
angularsen: Can we perhaps extract this code to a helper class? Seems a lot to duplicate N times.

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -827,6 +827,62 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+                formatString = formatString.Substring(0, 1);
+            }
+
+            switch(formatString)
+            {
+                case "G":
+                    return ToString(formatProvider, 2);
+                case "A":
+                    var abbreviations = UnitAbbreviationsCache.Default.GetUnitAbbreviations<AccelerationUnit>(Unit, formatProvider);
+                    return abbreviations[digits];
```
angularsen: Rename `digits` to `number` maybe? Since we are using it as an index in the case of `"A"` here, and possibly for number of digits for the case of `"V"` below.

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -827,6 +827,62 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+                case "V":
+                    return Value.ToString(formatProvider);
```
angularsen: Maybe support significant digits after radix point here? `Mass.FromKilograms(15.3).ToString("V4")` could output `15.3000`.

angularsen: In the same spirit `"V####"` could maybe output "15.3", since `#` is used to omit trailing zeros after radix point (and leading zeros before radix point).

tmilnthorp: That's what S does (must be given a number). You want both to do the same thing?

angularsen:
> It seems that they use case insensitivity for the default format strings.

Actually, it seems to me they are in fact different.

["g" short time](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#the-general-date-short-time-g-format-specifier) vs ["G" long time](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#the-general-date-long-time-g-format-specifier).

Same for "d"/"D", "f"/"F", "t"/"T", "u"/"U" differing on short/long or short/full formats.
These are all standard formats and that to me adds weight to the argument to keep it all case sensitive - to avoid any confusion. For instance "o" and "O" are identical in Date, which to me just adds confusion when most other standard formats have different outputs for lower and upper case.

> What denotes a custom format vs standard though is what we should probably figure out! 😃

From [Date docs](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings):
> Any date and time format string **that contains more than one character, including white space, is interpreted as a custom date and time format string**; for more information, see Custom Date and Time Format Strings.

The numeric docs say pretty much the same thing, except the also accept an optional number `xx` behind the letter. 

The way I read it, "standard" here only means their vetted, recommended, typical formats. The formats they want the .NET ecosystem to use. While custom is anything else, with a combination of date and time parts as well as free text fashioned together in any way you like.

In our case, I'm thinking standard format specifiers could maybe be like this:
- g: 15.1 kg
- s4: 15.1000 kg
- S4: 15.1 kg (using `#` to omit trailing zeros)

Custom format specifiers:
- `a`: `kg`
- `a2`: (2nd abbreviation for kilogram, or empty string(?) if not exists)
- `q`: `Mass`
- `u`: `Kilogram`
- `v`: `15.1`
- `v4`: `15.1000`
- `v####`: `15.1`

Then later, if we add human readable strings, we could add
- `Q`: `Bit Rate`
- `U`: `Bit per Second`
- `P` or `Up`: `Bits per Second`

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
tmilnthorp: 👍 done!

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -845,7 +844,7 @@ public string ToString(string format)
         /// <returns>The string representation.</returns>
         public string ToString(string format, IFormatProvider formatProvider)
         {
-            return QuantityFormatter.Format<Acceleration, AccelerationUnit>(this, format, formatProvider);
+            return QuantityFormatter.Format<AccelerationUnit>(this, format, formatProvider);
```
angularsen: Nice!

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
tmilnthorp: I always wonder at what point do we stop with the strings and let the user just do a string.format("0.####", quantity.Value) ❓

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
tmilnthorp: 👍 done!

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
tmilnthorp: Made it a bit simpler now 🤡

tmilnthorp:
You're right... I think I'm going crazy 🤡 I've had a sick baby at home and not much sleep 😆 

I'll make them lower-case for now.

angularsen:
I can relate :-D  a solid case of the chickenpox the past two days here reaching its peak just about now :-)

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
angularsen: Fair points, I've had the same thought.
Doesn't `S` also include the abbreviation? Maybe I read it wrong.

angularsen: `Length.FromCentimeters(15.34000).ToString(null, 5)` => `15.34 cm`, so yes, it does the `#` thing.

So my proposal would be:
`Length.FromCentimeters(15.34000).ToString("V####")` to get `15.34` without the unit abbreviation.

It's valid on its own, but I agree with you, at some point people should just do standard/custom numeric formatting of the `Value` property. I can agree to don't do the `V####` usecase on that ground.

angularsen:
I think with your latest changes, this is getting ready to merge. We don't need to get this perfect on the first try, let's just try something and see how it feels. We can leave `v4` and `v####` out for now.

tmilnthorp:
Oh man. Hope everything's better!

I've added doc about the valid format strings.

Also, I think we should obsolete the following methods as part of this PR:
- ```public string ToString([CanBeNull] IFormatProvider provider)```
  - Use ```ToString("g", provider)```
- ```public string ToString([CanBeNull] IFormatProvider provider, int significantDigitsAfterRadix)```
  - Use ```ToString("s2", provider)``` (or however many digits in format string)
- ```public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string format, [NotNull] params object[] args)```
  - Use string.format using value, unit, etc.

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -836,6 +836,16 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
         /// </summary>
         /// <param name="format">The format string.</param>
         /// <returns>The string representation.</returns>
+        /// <remarks>
```
angularsen: To avoid repeating xmldoc for method overloads in one type we can use `/// <inheritdoc cref="SomeMethod" />`.  Then you only need to specify the extra parameters or override any elements that need a different text.

We can also do the same to avoid repeating for N quantities, but maybe that's overdoing it. It's also more readable on github to add it to all quantities.

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -836,6 +836,16 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+        /// The valid format strings are as follows:
+        /// "g": The quantity formatted as the value with 2 significant digits after the radix, and abbreviation. For example: 1.23 m
```
angularsen: Can remove "The quantity formatted as", it's implicit in the context.

Maybe something like this: 
`The value with 2 significant digits after the radix followed by the unit abbreviation, such as "1.23 m".`

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -836,6 +836,16 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+        /// "a": The default abbreviation for the quantity's unit.
```
angularsen: `The default unit abbreviation for <see cref="Unit" />, such as "m".`

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -836,6 +836,16 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+        /// "a0", "a1", "a2", etc.: The abbreviation with index where multiple abbreviations exist. "a0" is the same as "a" or the default abbreviation.
```
angularsen: `"a0", "a1", ..., "aN": The Nth unit abbreviation for <see cref="Unit" />. "a0" is the same as "a".`

What happens if you specify a value out of range? That should be documented too.

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -836,6 +836,16 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+        /// "v": The quantity's value outputted as a string using the default representation.
```
angularsen: `"v": String representation of <see cref="Value" />.`

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -836,6 +836,16 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+        /// "u": The quantity's unit
```
angularsen: `"u": The enum name of <see cref="Unit" />, such as "Meter".`

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -836,6 +836,16 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+        /// "q": The quantity's name.
```
angularsen: `"q": The quantity name, such as "Length".`

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -836,6 +836,16 @@ public string ToString([CanBeNull] IFormatProvider provider, [NotNull] string fo
...
+        /// "s1", "s4", etc.: The quantity formatted as the value with n significant digits after the radix, and abbreviation. For example, "s4" would give: 1.2345 m
```
angularsen: `"s1", "s2", ..., "sN": The value with N significant digits after the radix followed by the unit abbreviation. For example, "s4" would return "1.2345 m" if <see cref="Value" /> is 1.2345678. Trailing zeros are omitted.`

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
tmilnthorp: Done. I wonder if we should be using         /// <see cref="QuantityFormatter.Format{TUnitType}(IQuantity{TUnitType}, string, IFormatProvider)"/> rather than inheritdoc though for cross-referencing though?

UnitsNet.Tests/CustomCode/TemperatureTests.cs
```diff
@@ -63,7 +64,8 @@ public void MultiplyByTemperatureDeltaEqualsTemperature(TemperatureUnit unit, in
             // Act
             Temperature resultTemp = temperature.Multiply(factor, unit);
 
-            string actual = resultTemp.ToUnit(unit).ToString(CultureInfo.InvariantCulture, "{0:0} {1}");
+            var resultTempInUnit = resultTemp.ToUnit(unit);
```
angularsen: Minor, but I think it read slightly better as a single line. Same below.

UnitsNet.Tests/QuantityIFormattableTests.cs
```diff
@@ -46,12 +46,12 @@ public void QFormatEqualsQuantityName()
         [Fact]
         public void SFormatEqualsSignificantDigits()
         {
-            Assert.Equal(length.ToString(null, 0), length.ToString("s"));
-            Assert.Equal(length.ToString(null, 1), length.ToString("s1"));
-            Assert.Equal(length.ToString(null, 2), length.ToString("s2"));
-            Assert.Equal(length.ToString(null, 3), length.ToString("s3"));
-            Assert.Equal(length.ToString(null, 4), length.ToString("s4"));
-            Assert.Equal(length.ToString(null, 5), length.ToString("s5"));
+            Assert.Equal("1 ft", length.ToString("s"));
```
angularsen: Much better to explicitly state the values 👍 
But isn't this test a duplicate of `CustomNumberOfSignificantDigitsAfterRadixFormatting` ? I like this test method name `SFormatEqualsSignificantDigits` better though, if you merge the two.

UnitsNet.Tests/QuantityTests.ToString.cs
```diff
@@ -124,9 +124,9 @@ public void FormatsNumberUsingGivenDigitsAfterRadix()
                 try
                 {
                     GlobalConfiguration.DefaultCulture = CultureInfo.InvariantCulture;
-                    Assert.Equal("0.05 m", Length.FromCentimeters(5).ToUnit(LengthUnit.Meter).ToString(null, 4));
-                    Assert.Equal("1.97 in", Length.FromCentimeters(5).ToUnit(LengthUnit.Inch).ToString(null, 2));
-                    Assert.Equal("1.9685 in", Length.FromCentimeters(5).ToUnit(LengthUnit.Inch).ToString(null, 4));
+                    Assert.Equal("0.05 m", Length.FromCentimeters(5).ToUnit(LengthUnit.Meter).ToString("s4"));
+                    Assert.Equal("1.97 in", Length.FromCentimeters(5).ToUnit(LengthUnit.Inch).ToString("s2"));
```
angularsen: Maybe construct the meter and inch quantities directly in that unit, so it's easier to compare the input and the output numeric values - instead of using `ToUnit()`.

UnitsNet.Tests/CustomCode/TemperatureTests.cs
tmilnthorp: No sense in using strings to check values. I've converted these tests to MemberData driven tests to avoid the issue entirely.

UnitsNet.Tests/QuantityTests.ToString.cs
tmilnthorp: Not sure I follow.

I just left the code as-is and replaced the format string due to the obsolete warning causing a build failure.

UnitsNet.Tests/QuantityIFormattableTests.cs
tmilnthorp: Replaced redundant test and made data-driven.

UnitsNet.Tests/QuantityTests.ToString.cs
angularsen: Ah yes, sorry, what I commented on was probably not a change introduced in your PR but rather something I just noticed while reviewing and maybe something we could fix while at it.

What I was trying to convey was that, instead of this
```c#
Assert.Equal("1.97 in", Length.FromCentimeters(5).ToUnit(LengthUnit.Inch).ToString("s2"));
```
Maybe do this:
```c#
Assert.Equal("1.2345 in", Length.FromInches(1.2345).ToString("s2"));
```

UnitsNet.Tests/CustomCode/TemperatureTests.cs
```diff
@@ -27,106 +25,103 @@ public class TemperatureTests : TemperatureTestsBase
...
-            Temperature temperature = Temperature.From(temperatureVal, unit);
-
-            // Act
-            Temperature resultTemp = temperature.Divide(divisor, unit);
-
-            var resultTempInUnit = resultTemp.ToUnit(unit);
-            var actual = string.Format(CultureInfo.InvariantCulture, "{0:0} {1:a}", resultTempInUnit.Value, resultTempInUnit);
-            Assert.Equal(expected, actual);
+            Temperature resultTemp = temperature.Divide(divisor, temperature.Unit);
+            Assert.True(expected.Equals(resultTemp, 1e-5, ComparisonType.Absolute));
```
angularsen: Shouldn't use our implementation of `Equals` here since that would be testing one implementation with another implementation. Better to just write the comparison check manually as I don't think Xunit has good support for it (I may be wrong). Something like this, can be added as a util method for test code.
`Assert.True(Math.Abs(a-b) <= 1e-5);`

Same thing in other tests below.

UnitsNet.Tests/CustomCode/TemperatureTests.cs
```diff
@@ -27,106 +25,103 @@ public class TemperatureTests : TemperatureTestsBase
...
             Justification = "R# incorrectly identifies method as impure, due to internal method calls.")]
         [Theory]
-        [InlineData(TemperatureUnit.DegreeCelsius, 10, 1, "10 °C")]
-        [InlineData(TemperatureUnit.DegreeCelsius, 10, 5, "2 °C")]
-        [InlineData(TemperatureUnit.DegreeCelsius, 10, -10, "-1 °C")]
-        [InlineData(TemperatureUnit.DegreeFahrenheit, 10, 1, "10 °F")]
-        [InlineData(TemperatureUnit.DegreeFahrenheit, 10, 5, "2 °F")]
-        [InlineData(TemperatureUnit.DegreeFahrenheit, 10, -10, "-1 °F")]
-        public void DividedByTemperatureDeltaEqualsTemperature(TemperatureUnit unit, int temperatureVal, int divisor, string expected)
+        [MemberData(nameof(DividedByTemperatureDeltaEqualsTemperatureData))]
```
angularsen: Nice, I agree, asserting on string representation is not good. Testing two things instead of one thing.

UnitsNet/QuantityFormatter.cs
```diff
@@ -35,16 +36,19 @@ public class QuantityFormatter
         /// <typeparam name="TUnitType">The quantity's unit type, for example <see cref="LengthUnit"/>.</typeparam>
         /// <param name="quantity">The quantity to format.</param>
         /// <param name="format">The format string.</param>
-        /// <param name="formatProvider">The format provider to use for localization and number formatting. Defaults to <see cref="GlobalConfiguration.DefaultCulture" /> if null.</param>
+        /// <param name="formatProvider">The format provider to use for localization and number formatting. Defaults to
+        /// <see cref="GlobalConfiguration.DefaultCulture" /> if null.</param>
```
angularsen: This needs to be updated to `CultureInfo.CurrentUICulture` after recent PR merge.

angularsen:
You also have a bit of merge conflicts to deal with ;-)

UnitsNet.Tests/QuantityTests.ToString.cs
tmilnthorp: I agree! However I would have to do all the conversions to make sure the double has enough decimal places on construction to make sure it's testing the same thing. I'd rather not do that as part of this PR!

UnitsNet.Tests/CustomCode/TemperatureTests.cs
tmilnthorp: I don't think it's completely avoidable. Other tests would fail too (Equals tests) if it was wrong. All tests rely on the code to some degree - even ```Temperature.Fromxxx```

UnitsNet.Tests/CustomCode/TemperatureTests.cs
tmilnthorp: 👍

UnitsNet/QuantityFormatter.cs
tmilnthorp: Done!

tmilnthorp:
Merged and updated. I think it's done! 🎉

UnitsNet.Tests/CustomCode/TemperatureTests.cs
angularsen: I suppose you are right to some extent it is unavoidable. Equals could be avoidable though, but I agree with your reasoning that Equals is already tested and should be fine.

UnitsNet.Tests/QuantityTests.ToString.cs
angularsen: Fair enough

angularsen:
Wohoo :) Awesome job on this!

MrFoged:
Could you give a little insight into how to use the new string.Format?

Right now i have this code:
`.ToString(new CultureInfo("en-US"), "{0:N1} {1}")` which it tells me to change but to what? :-)

tmilnthorp:
The equivalent of that should be:
'''csharp
string.Format(new CultureInfo("en-US"), "{0:N1} {1:a}", quantity.Value, quantity);
''

angularsen:
Updated docs in [String Formatting · angularsen/UnitsNet Wiki](https://github.com/angularsen/UnitsNet/wiki/String-Formatting)

Issue #547, opened 2018-11-03
Add quantity constructor overload for UnitSystem


angularsen:
Per discussion in #519, we want to construct quantities using a unit system such as SI or a custom unit system. The advantage is consistent base units and not having to specify the unit when constructing quantities.

So you can do
```
new Length(1); // 1 meter, SI unit system
new Length(1, UnitSystem.SI); // same
new Length(1, myUnitSystem); // whatever length unit you configured in your system
```

### Implementation
- Add constructor `MyQuantity(double numericValue, UnitSystem unitSystem) {}`
- Add constructor `MyQuantity(double numericValue) : MyQuantity(numericValue, UnitSystem.SI) {}`

#### Quantities with dimensions matching base units is easy
```c#
public Length(double numericValue, UnitSystem unitSystem) 
    : this(numericValue, unitSystem.BaseUnits.Length) { }
```

#### Dimensionless quantities?
[`Information`](http://www.wolframalpha.com/input/?i=information&assumption=%7B%22C%22,+%22information%22%7D+-%3E+%7B%22PhysicalQuantity%22%7D), `Ratio`, `Angle`, etc...

```c#
public Information(decimal numericValue, UnitSystem unitSystem) 
    : this(numericValue, unitSystem.BaseUnits./*???*/) { }
```

These are not coupled to any unit system, so we can keep the existing choice of base unit?

#### Logarithmic quantities?
[`Level`](http://www.wolframalpha.com/input/?i=power+quantity+level&lk=1&assumption=%22ClashPrefs%22+-%3E+%7B%22PhysicalQuantity%22%7D) (dB), `PoweRatio`, `AmplitudeRatio`

```c#
public PowerRatio(decimal numericValue, UnitSystem unitSystem) 
    : this(numericValue, unitSystem.BaseUnits./*???*/) { }
```

[AmplitudeRatio dBV](http://www.wolframalpha.com/input/?i=dBV&assumption=%7B%22C%22,+%22dBV%22%7D+-%3E+%7B%22Unit%22%7D) 
> Unit system: (used in any unit system)

[PoweRatio dBW](http://www.wolframalpha.com/input/?i=dBW)
> Unit system: (not in a named unit system)

They seem uncoupled from unit systems, so we can keep the existing choices of base unit?

#### Non-trivial quantities?
Most quantities fall in this category. How do we construct them given a non-SI instance of `UnitSystem`?

[...]

PR #601, opened 2019-02-04
Adding base units to some quantities (first steps)

tmilnthorp:
UnitSystem constructors (#547)


tmilnthorp:
And no, it doesn't work on WRC 😄 I did it quick. Have to pull code and do locally without WRC to see it.

angularsen:
Probably a good idea to merge in #607 first.

tmilnthorp:
More of a proof of concept. Not ready to merge :)

UnitsNet/GeneratedCode/Quantities/Mass.NetFramework.g.cs
```diff
@@ -64,7 +64,32 @@ public partial struct Mass : IQuantity<MassUnit>, IEquatable<Mass>, IComparable,
         static Mass()
         {
             BaseDimensions = new BaseDimensions(0, 1, 0, 0, 0, 0, 0);
-            Info = new QuantityInfo<MassUnit>(QuantityType.Mass, Units, BaseUnit, Zero, BaseDimensions);
+
+            Info = new QuantityInfo<MassUnit>(QuantityType.Mass, new UnitInfo<MassUnit>[] {
+                new UnitInfo<MassUnit>(MassUnit.Centigram, new BaseUnits(LengthUnit.Undefined, MassUnit.Undefined, DurationUnit.Undefined, ElectricCurrentUnit.Undefined, TemperatureUnit.Undefined, AmountOfSubstanceUnit.Undefined, LuminousIntensityUnit.Undefined)),
```
angularsen: It's quite verbose right now, since the majority is assigning a single value we can later use default values to omit most of them.

UnitsNet/GeneratedCode/Quantities/Mass.NetFramework.g.cs
```diff
@@ -64,7 +64,32 @@ public partial struct Mass : IQuantity<MassUnit>, IEquatable<Mass>, IComparable,
...
+                new UnitInfo<MassUnit>(MassUnit.Decagram, new BaseUnits(LengthUnit.Undefined, MassUnit.Undefined, DurationUnit.Undefined, ElectricCurrentUnit.Undefined, TemperatureUnit.Undefined, AmountOfSubstanceUnit.Undefined, LuminousIntensityUnit.Undefined)),
+                new UnitInfo<MassUnit>(MassUnit.Decigram, new BaseUnits(LengthUnit.Undefined, MassUnit.Undefined, DurationUnit.Undefined, ElectricCurrentUnit.Undefined, TemperatureUnit.Undefined, AmountOfSubstanceUnit.Undefined, LuminousIntensityUnit.Undefined)),
```
angularsen: Prefix units all undefined, but you already mentioned that to be WIP.

UnitsNet.Tests/CustomCode/LengthTests.cs
```diff
@@ -169,5 +169,12 @@ public void NegativeLengthToStonePoundsReturnsCorrectValues()
             Assert.Equal(-2.0, feetInches.Feet);
             Assert.Equal(-1.0, feetInches.Inches);
         }
+
+        [Fact]
+        public void LengthCreatedInSISystem()
```
angularsen: Can we have tests for the imperial units? Those seem to be added in JSON already.

tmilnthorp:
I can't reply directly to your comments so here we go:

- DefaultUnit would probably be a better name!
- The only derived unit is area right now. The others just need adding!
- Yes it is quite verbose. I should
  - Skip adding any BaseUnits if they're all undefined (make it null)
  - Use the default values or object initializers to skip the undefined base units.
- We can add something for imperial units, but what I'm not sure. I'm open to suggestions. There's so real imperial system in the sense of default units. Even restricting it to the US customary units has inches, feet, yards, and miles for length! No standard like SI.

tmilnthorp:
I'm actually pretty happy with this now if you'd like to take a look!

UnitsNet.Tests/QuantityInfoTest.cs
```diff
@@ -15,18 +15,17 @@ public class QuantityInfoTest
         public void Constructor_AssignsProperties()
         {
             var expectedZero = Length.FromCentimeters(10);
-            var expectedUnits = new Enum[] {LengthUnit.Centimeter, LengthUnit.Kilometer};
-            var expectedBaseUnit = LengthUnit.Meter;
+            var expectedUnitInfos = new UnitInfo[]{ new UnitInfo(LengthUnit.Centimeter, new BaseUnits(LengthUnit.Centimeter)), new UnitInfo(LengthUnit.Kilometer, new BaseUnits(LengthUnit.Kilometer))};
```
angularsen: Chop long line here and other places.

UnitsNet.Tests/QuantityInfoTest.cs
```diff
@@ -15,18 +15,17 @@ public class QuantityInfoTest
...
+            var expectedBaseUnit = LengthUnit.Centimeter;
             var expectedQuantityType = QuantityType.Length;
             var expectedBaseDimensions = Length.BaseDimensions;
 
-            var info = new QuantityInfo(expectedQuantityType, expectedUnits, expectedBaseUnit, expectedZero, expectedBaseDimensions);
+            var info = new QuantityInfo(expectedQuantityType, expectedUnitInfos, expectedBaseUnit, expectedZero, expectedBaseDimensions);
 
+#pragma warning disable 618
```
angularsen: Pragma comment was lost

UnitsNet.Tests/QuantityInfoTest.cs
angularsen: Would it make sense to add a more convenient `UnitInfo` ctor that take a single enum value and assigns `BaseUnits` with that value unless otherwise specified? It seems a very common call pattern in this PR.

UnitsNet.Tests/QuantityInfoTest.cs
```diff
@@ -38,18 +37,17 @@ public void Constructor_AssignsProperties()
...
-            var expectedBaseUnit = LengthUnit.Meter;
+            var expectedUnitInfos = new UnitInfo<LengthUnit>[]{ new UnitInfo<LengthUnit>(LengthUnit.Centimeter, new BaseUnits(LengthUnit.Centimeter)), new UnitInfo<LengthUnit>(LengthUnit.Kilometer, new BaseUnits(LengthUnit.Kilometer))};
+            var expectedBaseUnit = LengthUnit.Centimeter;
             var expectedQuantityType = QuantityType.Length;
             var expectedBaseDimensions = Length.BaseDimensions;
 
-            var info = new QuantityInfo<LengthUnit>(expectedQuantityType, expectedUnits, expectedBaseUnit, expectedZero, expectedBaseDimensions);
+            var info = new QuantityInfo<LengthUnit>(expectedQuantityType, expectedUnitInfos, expectedBaseUnit, expectedZero, expectedBaseDimensions);
 
+#pragma warning disable 618
```
angularsen: Pragma comment lost

UnitsNet.Tests/QuantityTest.cs
```diff
@@ -171,5 +171,19 @@ public void Types_ReturnsKnownQuantityTypes()
             Assert.Superset(knownQuantities.ToHashSet(), types.ToHashSet());
             Assert.Equal(QuantityCount, types.Length);
         }
+
+        [Fact]
+        public void GetUnitForUnitSystemTest()
```
angularsen: Separate test cases as mentioned above

UnitsNet.Tests/QuantityTest.cs
```diff
@@ -171,5 +171,19 @@ public void Types_ReturnsKnownQuantityTypes()
...
+        {
+            Assert.Throws<ArgumentNullException>(() => Length.GetUnitFor((UnitSystem)null));
+            Assert.Equal(LengthUnit.Meter, Length.GetUnitFor(UnitSystem.SI));
+        }
+
+        [Fact]
+        public void GetUnitForBaseUnitsTest()
+        {
+            Assert.Throws<ArgumentNullException>(() => Length.GetUnitFor((BaseUnits)null));
```
angularsen: Here too, I don't like the duality of `BaseUnits` and `UnitSystem`. It smells like the two should merge.

UnitsNet.Tests/QuantityTests.cs
```diff
@@ -51,5 +51,28 @@ private static void AssertQuantityInfoRepresentsLength(QuantityInfo<LengthUnit>
...
+        public void TestGenericMultiplication()
+        {
+            var length1 = Length.FromInches(2.0);
+            var length2 = Length.FromMeters(2.0);
+
+            var calculated = Multiply(length1, length2);
+            Assert.Equal(length1 * length2, calculated);
+        }
+
+        private IQuantity Multiply(IQuantity left, IQuantity right)
```
angularsen: Is this Multiply method supposed to be moved into an implementation somewhere? Also it seems work in progress still.

UnitsNet.Tests/QuantityTests.cs
```diff
@@ -51,5 +51,28 @@ private static void AssertQuantityInfoRepresentsLength(QuantityInfo<LengthUnit>
             Assert.Equal(lengthUnitNames, quantityInfo.UnitNames);
 #pragma warning restore 618
         }
+
+        [Fact]
+        public void TestGenericMultiplication()
+        {
+            var length1 = Length.FromInches(2.0);
+            var length2 = Length.FromMeters(2.0);
```
angularsen: Can we add an example of multiplying different quantities, such as `Force * Length = Torque`.

UnitsNet/BaseUnits.cs
```diff
@@ -25,8 +30,10 @@ public sealed class BaseUnits: IEquatable<BaseUnits>
         /// <param name="temperature">The temperature unit (Θ).</param>
         /// <param name="amount">The amount of substance unit (N).</param>
         /// <param name="luminousIntensity">The luminous intensity unit (J).</param>
-        public BaseUnits(LengthUnit length, MassUnit mass, DurationUnit time, ElectricCurrentUnit current,
-            TemperatureUnit temperature, AmountOfSubstanceUnit amount, LuminousIntensityUnit luminousIntensity)
+        public BaseUnits(LengthUnit length = LengthUnit.Undefined, MassUnit mass = MassUnit.Undefined,
```
angularsen: Chop and align with one param per line

UnitsNet/GeneratedCode/Quantities/Capacitance.NetFramework.g.cs
```diff
@@ -68,6 +79,32 @@ public Capacitance(double numericValue, CapacitanceUnit unit)
             _unit = unit;
         }
 
+        /// <summary>
+        /// Creates an instance of the quantity with the given numeric value in units compatible with the given <see cref="UnitSystem"/>.
+        /// </summary>
+        /// <param name="numericValue">The numeric value  to contruct this quantity with.</param>
+        /// <param name="unitSystem">The unit system to create the quantity with.</param>
+        public Capacitance(double numericValue, UnitSystem unitSystem)
```
angularsen: Already mentioned, but I think we should have only one ctor for either UnitSystem or BaseUnits. Not both.

UnitsNet/GeneratedCode/Quantities/Capacitance.NetFramework.g.cs
```diff
@@ -684,6 +721,36 @@ private double AsBaseNumericType(CapacitanceUnit unit)
             }
         }
 
+        /// <summary>
+        /// Gets the unit for this quantity compatible with the given <see cref="UnitSystem"/>.
+        /// </summary>
+        /// <param name="unitSystem">The <see cref="UnitSystem"/> to get the compatible units for.</param>
+        /// <returns>The unit for this quantity compatible with the given  <see cref="UnitSystem"/></returns>
+        public static CapacitanceUnit GetUnitFor(UnitSystem unitSystem)
```
angularsen: Same here, one method not two.

UnitsNet/UnitInfo.cs
```diff
@@ -17,13 +17,16 @@ namespace UnitsNet
...
-        /// <summary>Creates an instance from a unit enum value.</summary>
-        /// <param name="value">The unit enum value.</param>
-        /// <example>new UnitInfo(LengthUnit.Meter)</example>
-        public UnitInfo(Enum value)
+        /// <summary>
+        /// Creates an instance of the UnitInfo class.
+        /// </summary>
+        /// <param name="value">The enum value for this class, for example <see cref="LengthUnit.Meter"/>.</param>
+        /// <param name="baseUnits">The <see cref="BaseUnits"/> for this unit.</param>
+        public UnitInfo(Enum value, BaseUnits baseUnits)
```
angularsen: Add `[NotNull]` to both params.

UnitsNet.Tests/QuantityTest.cs
angularsen: Also, can we have a test case for non-SI units? Maybe something like 
```c#
var unitSystem = new UnitSystem(MassUnit.Pound, LengthUnit.Foot);
Assert.Equal(TorqueUnit.PoundForceFeet, Torque.GetUnitFor(unitSystem));
```

UnitsNet.Tests/BaseUnitsTests.cs
```diff
@@ -91,6 +91,21 @@ public void InequalityOperatorIsImplementedCorrectly()
             Assert.False(nullBaseUnits1 != nullBaseUnits2);
         }
 
+        [Fact]
+        public void ExistsInWorksCorrectly()
```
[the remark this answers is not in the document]
tmilnthorp: Done.

UnitsNet.Tests/CustomCode/LengthTests.cs
```diff
@@ -151,5 +152,18 @@ public void NegativeLengthToStonePoundsReturnsCorrectValues()
...
+        {
+            Assert.Throws<ArgumentNullException>(() => new Length(1.0, (UnitSystem)null));
```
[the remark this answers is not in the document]
tmilnthorp: Done.

UnitsNet/UnitInfo.cs
tmilnthorp: Done.

UnitsNet/BaseUnits.cs
tmilnthorp: Done.

UnitsNet.Tests/QuantityTests.cs
tmilnthorp: I removed this entirely. Was a test method for me more than anything. I'll do this in another PR once this is in

UnitsNet.Tests/CustomCode/EnergyTests.cs
```diff
@@ -55,7 +55,7 @@ public class EnergyTests : EnergyTestsBase
         protected override double WattHoursInOneJoule => 0.000277777778;
 
         [Fact]
-        public void EnergyCreatedInSISystem()
+        public void Constructor_UnitSystem_SIUsesSIUnits()
```
angularsen: I know  there is no consistent pattern in our test suite, but if you are dividing up the name like this I would expect it to read as {what is tested}\_{what are the inputs}\_{what is the result}. At least that is a common naming scheme I've seen a lot.

So the middle part should then say this is a SI unit system and the final part say something about SI units being the result - or maybe just spell it out directly that Joule is the unit.

Something along the lines of `Constructor_UnitSystemSI_AssignsSIUnit` or `Constructor_GivenUnitSystemSI_AssignsJouleUnit`. Same for test methods below.

tmilnthorp: Done!

UnitsNet/GeneratedCode/Quantities/Torque.NetFramework.g.cs
```diff
@@ -64,7 +64,30 @@ public partial struct Torque : IQuantity<TorqueUnit>, IEquatable<Torque>, ICompa
         static Torque()
         {
             BaseDimensions = new BaseDimensions(2, 1, -2, 0, 0, 0, 0);
-            Info = new QuantityInfo<TorqueUnit>(QuantityType.Torque, Units, BaseUnit, Zero, BaseDimensions);
+
+            Info = new QuantityInfo<TorqueUnit>(QuantityType.Torque, new UnitInfo<TorqueUnit>[] {
+                new UnitInfo<TorqueUnit>(TorqueUnit.KilogramForceCentimeter, new BaseUnits(LengthUnit.Undefined, MassUnit.Undefined, DurationUnit.Undefined, ElectricCurrentUnit.Undefined, TemperatureUnit.Undefined, AmountOfSubstanceUnit.Undefined, LuminousIntensityUnit.Undefined)),
```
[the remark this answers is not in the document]
tmilnthorp: Area and Energy have some for example. Just need to add as we go.

UnitsNet/GeneratedCode/Quantities/Capacitance.NetFramework.g.cs
tmilnthorp: I think both are useful. I wanted to keep them as two classes for separation of concerns and single responsibility. UnitSystems are defined with BaseUnits, but BaseUnits do not make up a UnitSystem. You could do something like this for example when you want the SI prefixes:
```csharp
    public class SIUnitSystem : UnitSystem
    {
        private static readonly BaseUnits SIBaseUnits = new BaseUnits( LengthUnit.Meter, MassUnit.Kilogram, DurationUnit.Second,
            ElectricCurrentUnit.Ampere, TemperatureUnit.Kelvin, AmountOfSubstanceUnit.Mole, LuminousIntensityUnit.Candela );

        /// <summary>
        /// Gets the SI unit system.
        /// </summary>
        public static SIUnitSystem Instance { get; } = new SIUnitSystem();

        private SIUnitSystem() : base( SIBaseUnits )
        {
        }

        public const double KiloFactor = 1e3;
    }
```

UnitsNet/GeneratedCode/Quantities/Capacitance.NetFramework.g.cs
tmilnthorp: Continue above.

UnitsNet/BaseUnits.cs
```diff
@@ -73,25 +73,26 @@ public bool Equals(BaseUnits other)
...
+        /// <param name="other">The other <see cref="BaseUnits"/> to check for existance in.</param>
+        /// <returns>True if the base units exist in other, otherwise false.</returns>
+        public bool ExistsIn(BaseUnits other)
         {
             if(other is null)
                 return false;
 
-            if(Equals(Undefined) || other.Equals(Undefined))
+            // If everything is undefined, there's nothing to exist in other.
+            if(Equals(Undefined))
```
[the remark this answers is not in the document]
tmilnthorp: 👍

UnitsNet.Tests/BaseUnitsTests.cs
```diff
@@ -91,6 +91,21 @@ public void InequalityOperatorIsImplementedCorrectly()
...
+        {
+            Assert.False(siBaseUnits.ExistsIn(null));
+
+            Assert.False(BaseUnits.Undefined.ExistsIn(siBaseUnits));
+            Assert.False(siBaseUnits.ExistsIn(BaseUnits.Undefined));
+
+            var meterBaseUnits = new BaseUnits(LengthUnit.Meter);
+            Assert.True(meterBaseUnits.ExistsIn(siBaseUnits));
```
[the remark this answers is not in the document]
tmilnthorp: I like IsSubsetOf! Matches HashSet

UnitsNet.Tests/BaseUnitsTests.cs
```diff
@@ -91,6 +91,21 @@ public void InequalityOperatorIsImplementedCorrectly()
             Assert.False(nullBaseUnits1 != nullBaseUnits2);
         }
 
+        [Fact]
+        public void ExistsInWorksCorrectly()
+        {
+            Assert.False(siBaseUnits.ExistsIn(null));
+
+            Assert.False(BaseUnits.Undefined.ExistsIn(siBaseUnits));
```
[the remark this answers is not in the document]
tmilnthorp: True if you're thinking of set theory, which this essentially is!

UnitsNet/GeneratedCode/Quantities/AmountOfSubstance.NetFramework.g.cs
```diff
@@ -877,7 +877,7 @@ public static AmountOfSubstanceUnit GetUnitFor(BaseUnits baseUnits)
             if(baseUnits == null)
                 throw new ArgumentNullException(nameof(baseUnits));
 
-            var unit = Info.UnitInfos.Where((unitInfo) => unitInfo.BaseUnits.ExistsIn(baseUnits)).FirstOrDefault();
+            var unit = Info.UnitInfos.Where((unitInfo) => unitInfo.BaseUnits.IsSubsetOf(baseUnits)).FirstOrDefault();
             if(unit == null)
                 throw new NotImplementedException($"No LengthUnit was found for the given BaseUnits.");
```
angularsen: Substitute `LengthUnit`

UnitsNet/GeneratedCode/Quantities/VolumeFlow.NetFramework.g.cs
```diff
@@ -65,6 +117,32 @@ public VolumeFlow(double numericValue, VolumeFlowUnit unit)
...
+            _value = Guard.EnsureValidNumber(numericValue, nameof(numericValue));
+            _unit = GetUnitFor(unitSystem.BaseUnits);
+        }
+
+        /// <summary>
+        /// Creates an instance of the quantity with the given numeric value in units compatible with the given <see cref="BaseUnits"/>.
+        /// </summary>
+        /// <param name="numericValue">The numeric value  to contruct this quantity with.</param>
+        /// <param name="baseUnits">The base units to create the quantity with.</param>
+        public VolumeFlow(double numericValue, BaseUnits baseUnits)
```
angularsen: I think maybe you missed my last comment, so I'll add it back below. I think we should choose only one of the two ctors. I prefer `UnitSystem` over `BaseUnits`, because I think it's easier to relate to. Base units is in my mind a bit abstract/unfamiliar.

> It seems a bit odd to me that we can pass in both UnitSystem and BaseUnits to the ctor here.
Also, somewhat out of scope of PR, but what value does UnitSystem bring beyond BaseUnits? Do we expect it to have anything more than a list of base units at some point? Could we merge the two?

UnitsNet/GeneratedCode/Quantities/VolumeFlow.NetFramework.g.cs
```diff
@@ -1337,6 +1415,36 @@ private double AsBaseNumericType(VolumeFlowUnit unit)
             }
         }
 
+        /// <summary>
+        /// Gets the unit for this quantity compatible with the given <see cref="UnitSystem"/>.
+        /// </summary>
+        /// <param name="unitSystem">The <see cref="UnitSystem"/> to get the compatible units for.</param>
+        /// <returns>The unit for this quantity compatible with the given  <see cref="UnitSystem"/></returns>
+        public static VolumeFlowUnit GetUnitFor(UnitSystem unitSystem)
```
angularsen: Same here, I think we should have keep only one of the two `GetUnitFor` methods. I prefer `UnitSystem` over `BaseUnits` here as well, but at least it should be consistent.

UnitsNet/GeneratedCode/Quantities/VolumeFlow.NetFramework.g.cs
tmilnthorp: See my above statement about UnitSystem vs BaseUnits. However I'll remove the base units constructor and just keep UnitSystem.

UnitsNet/GeneratedCode/Quantities/Capacitance.NetFramework.g.cs
tmilnthorp: @angularsen

UnitsNet/GeneratedCode/Quantities/AmountOfSubstance.NetFramework.g.cs
tmilnthorp: I moved this to QuantityInfo anyways. Easier when dealing with interfaces or statics (for later)

tmilnthorp:
I will add more tests for QuantityInfo.GetUnitInfoForBaseUnitsSubset before merging

UnitsNet/QuantityInfo.cs
```diff
@@ -123,6 +123,25 @@ public QuantityInfo(QuantityType quantityType, [NotNull] UnitInfo[] unitInfos, [
...
+        /// <returns>The UnitInfo that has BaseUnits that are a subset.</returns>
+        public UnitInfo GetUnitInfoForBaseUnitsSubset(BaseUnits baseUnits)
+        {
+            if(baseUnits == null)
+                throw new ArgumentNullException(nameof(baseUnits));
+
+            var matchingUnitInfos = UnitInfos.Where((unitInfo) => unitInfo.BaseUnits.IsSubsetOf(baseUnits));
+            if(!matchingUnitInfos.Any())
+                throw new NotImplementedException("No units were found that are a subset of the given BaseUnits.");
+            else if(matchingUnitInfos.Skip(1).Any())
```
angularsen: `Skip(1).Any()` tickles me a bit :-) 

This is really minor, but I couldn't help myself trying out an alternative. How about something like this? It should be similarly optimized to enumerate as little as necessary.
```c#
           var matchingUnitInfos = UnitInfos
                .Where((unitInfo) => unitInfo.BaseUnits.IsSubsetOf(baseUnits))
                .Take(2)
                .ToArray();

            var firstUnitInfo = matchingUnitInfos.FirstOrDefault();
            if (firstUnitInfo == null)
                throw new NotImplementedException("No units were found that are a subset of the given BaseUnits.");
            
            if (matchingUnitInfos.Length > 1)
                throw new InvalidOperationException("More than one unit was found that is a subset of the given BaseUnits.");

            return firstUnitInfo;
```

UnitsNet.Tests/QuantityInfoTest.cs
```diff
@@ -119,5 +119,17 @@ public void GenericsConstructor_GivenNullAsBaseDimensions_ThrowsArgumentNullExce
         {
             Assert.Throws<ArgumentNullException>(() => new QuantityInfo<LengthUnit>(QuantityType.Length, Length.Info.UnitInfos, Length.BaseUnit, Length.Zero, null));
         }
+
+        //[Fact]
```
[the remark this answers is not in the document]
tmilnthorp: Still working!

UnitsNet.Tests/QuantityInfoTest.cs
tmilnthorp: No - that would only be applicable for base quantities (length, mass, etc.), and would require more code than necessary for an info holder (check for base dimensions being a base quantity, throwing exceptions if unable to map, case statement for applicable assignment, etc.)

Derived quantities have no applicable mapping.

tmilnthorp: Done!

UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -92,20 +92,7 @@ public Acceleration(double numericValue, UnitSystem unitSystem)
...
-        /// </summary>
-        /// <param name="numericValue">The numeric value  to contruct this quantity with.</param>
-        /// <param name="baseUnits">The base units to create the quantity with.</param>
-        public Acceleration(double numericValue, BaseUnits baseUnits)
-        {
-            if(baseUnits == null) throw new ArgumentNullException(nameof(baseUnits));
-
-            _value = Guard.EnsureValidNumber(numericValue, nameof(numericValue));
-            _unit = GetUnitFor(baseUnits);
+            _unit = Info.GetUnitInfoForBaseUnitsSubset(unitSystem.BaseUnits).Value;
```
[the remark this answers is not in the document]
tmilnthorp: Done!

UnitsNet/QuantityInfo.cs
tmilnthorp: Sounds good to me. Fixed!

tmilnthorp:
Ok, I think I got through it all! 🎉

UnitsNet.Tests/QuantityInfoTest.cs
```diff
@@ -137,16 +137,30 @@ public void GenericsConstructor_GivenNullAsBaseDimensions_ThrowsArgumentNullExce
...
+
+        [Fact]
+        public void GetUnitInfoFor_GivenBaseUnitsWithNoMatch_ThrowsInvalidOperationException()
+        {
+            var baseUnitsWithNoMatch = new BaseUnits(mass: MassUnit.Kilogram);
+            Assert.Throws<InvalidOperationException>(() => Length.Info.GetUnitInfoFor(baseUnitsWithNoMatch));
+        }
+
+        [Fact]
+        public void GetUnitInfoFor_GivenBaseUnitsWithMultipleMatches_ThrowsInvalidOperationException()
```
angularsen: Good test!

angularsen:
I think so too :-) Looks good to me, great job on this one too!

PR #607, opened 2019-02-12
Separate WindowsRuntimeComponent into its own source code and code generators

### LINKED ISSUE #597: Separate WindowsRuntimeComponent into its own source code and code generators
See https://github.com/angularsen/UnitsNet/pull/576#issuecomment-459633641.

Basically WRC has been a major friction for maintainers and contributors of the library for a long time.
We don't seem to have a lot of consumers of the WRC target so the pain outweighs the gain and the WRC target as a means of sharing .NET libraries with JavaScript and C++ UWP apps don't seem to have a lot of interest in general.

> Adding new units and quantities is still a breeze in WRC, so what we need to do is split out the code generator scripts for WRC and get its full copy of all the shared source code, then we can still add new units to both UnitsNet and WRC and produce new versions as before - it just won't get things like Quantity and QuantityInfo in the future.

angularsen:
Fixes #597

- Copy scripts and source code into `/WindowsRuntimeComponent` folder
- Simplify scripts for each target
- Simplify code and remove `#if WINDOWS_UWP` sections
- Simplify codegen scripts and remove WRC logic
- Change to opt-in to build WRC (most people don't have it)
- Update AppVeyor to run new `build-with-wrc.bat`
- Unit definition `.json` files are still shared

This should be a refactor and not introduce any functional changes, except for maybe some minor things I found in WRC regarding the recently added `Quantity` types that are not very well supported in WRC anyway. I wouldn't consider it a breaking change.

I think this should remove the majority of pain associated with WRC:
- People can't build without UWP workload installed in Visual Studio (most don't have it and it's big)
- The AppVeyor build very often breaks when doing any changes beyond adding new units, due to complicated rules in WRC and previously sharing source code and codegen scripts with WRC

We can always branch out and remove WRC entirely if we find it still adds pain, but I don't think that will be necessary anytime soon unless we do a major overhaul of the `.json` file approach.


angularsen:
Oh God it feels good to remove this stuff ❤️ 
Need a pair of fresh eyes to look over the changes though, but everything compiles and nugets are packed for all targets.

angularsen:
Sweeet!

angularsen:
[Release UnitsNet/4.8.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.8.0)

PR #611, opened 2019-02-22
Added BeatPerMinute (unit used for measuring heart rate) as a frequency.

russelljahn:
I'm using BPM and heart rate in my own projects, so I figured it would be a good addition.

### DIFF
--- a/Common/UnitDefinitions/Frequency.json
+++ b/Common/UnitDefinitions/Frequency.json
@@ -55,6 +55,18 @@
           "Abbreviations": [ "cph" ]
         }
       ]
+    },
+    {
+      "SingularName": "BeatPerMinute",
+      "PluralName": "BeatsPerMinute",
+      "FromUnitToBaseFunc": "x/60",
+      "FromBaseToUnitFunc": "x*60",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "bpm" ]
+        }
+      ]
     }
   ]
 }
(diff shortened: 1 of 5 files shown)


angularsen:
Hi and thanks for contributing!

Would it be possible for you to use CyclePerMinute unit instead?

I can understand BPM would be useful for things like `Frequency.Parse("120 bpm")`, and `heartbeat.ToString()` to output `120 BPM`, but I think this is a slippery slope because I can think of many variations of "per second" or "per minute" counts - such as RPM (revolutions per minute) [a RotationalSpeed unit](https://github.com/angularsen/UnitsNet/blob/e80240bbaf2e5f68c148b63e77d9b0d8fbcf982b/Common/UnitDefinitions/RotationalSpeed.json#L79), and more silly examples like popcorn pops per second or cars per hour as a measure of traffic. All of these can be considered frequency measures. Where should we draw the line for what should be added to the library and not? 

I'm just not sure. On one hand BPM is widely used and it would let you parse and ToString() the "BPM" unit abbrevation. On the other hand, I don't think it adds much value in terms of unit conversion (what other heat rate units are used?) and we already have the CPM unit. Maybe it's better for you to create your own application specific type `HeartRate` to represent this quantity and possibly convert to other frequencies?

What do you think @tmilnthorp ?

tmilnthorp:
While I agree about things like popcorn pops per second and being careful about what we allow, I think BPM is so common for heartrates and music tempos that it is valid to be included.

angularsen:
Good point about music, didn't think of that usecase. That helps justify this as its own unit of frequency. I can stand behind that 👍

angularsen:
Nuget on the way out.
[Release UnitsNet/4.11.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.11.0)

PR #612, opened 2019-02-23
Add IQuantity.Value

angularsen:
Using value type double since that is the standard conversion property getters, until we add support for different value types. Those handful few quantities that use decimal internally for precision reasons explicitly implement the interface with a cast to double.


angularsen:
Nuget on the way out.
[Release UnitsNet/4.10.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.10.0)

PR #614, opened 2019-02-23
Add TeraWh

mivano:
In our current project we use the Energy class, but need to output to TeraWh. This was not included, so I added it here. Also did the ru_RU thing which I also have in the other PR. So only one of the PRs need merging.


angularsen:
Perfect, looks good to me!

angularsen:
Nuget on the way out.
[Release UnitsNet/4.10.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.10.0)

PR #616, opened 2019-02-25
Obsolete GlobalConfiguration

### LINKED ISSUE #600: Replace GlobalConfiguration.DefaultCulture with CultureInfo.CurrentCulture
Replace GlobalConfiguration.DefaultCulture with CultureInfo.CurrentCulture. Breaking change so belongs in #563. 

- CultureInfo.CurrentCulture for number formatting
- Use CultureInfo.CurrentCulture or CultureInfo.CurrentUICulture for abbreviations?

angularsen:
Fixes #600 .

So it can be more easily removed in v5.

- Mark GlobalConfiguration as obsolete
- Use `Thread.CurrentThread.CurrentUICulture` as backing value ⚠️ 
- Update usages to use `Thread.CurrentThread.CurrentUICulture` directly

This is kind of a breaking change. Before it would initialize to `CurrentUICulture` upon first static access to `GlobalConfiguration` and that value would no longer match per-thread - but rather be globally static across threads. With this PR, it would instead match .NET's per-thread setup - which one could argue is a bug or a feature that we didn't - but since GlobalConfiguration was introduced in v4 in December and I don't think many/anyone relies on this initialize once behavior, I vote to take this in now.

### DIFF
--- a/UnitsNet.Tests/CustomCode/StonePoundsTests.cs
+++ b/UnitsNet.Tests/CustomCode/StonePoundsTests.cs
@@ -35,7 +35,7 @@ namespace UnitsNet.Tests.CustomCode
         {
             Mass m = Mass.FromStonePounds(3500, 1);
             StonePounds stonePounds = m.StonePounds;
-            string numberInCurrentCulture =  3500.ToString("n0", GlobalConfiguration.DefaultCulture); // Varies between machines, can't hard code it
+            string numberInCurrentCulture =  3500.ToString("n0", CultureInfo.CurrentUICulture); // Varies between machines, can't hard code it
 
             Assert.Equal($"{numberInCurrentCulture} st 1 lb", stonePounds.ToString());
         }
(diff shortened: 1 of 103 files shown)


tmilnthorp:
I think we should probably be using CurrentCulture, not CurrentUICulture. At least for the number formatting. Maybe CurrentUICulture is more appropriate for the abbreviations? 

I found an interesting quote [here](https://stackoverflow.com/questions/329033/what-is-the-difference-between-currentculture-and-currentuiculture-properties-of):

> Often times they are both the same. But on my system they would be different: I prefer my numbers and dates in the German format, so the CurrentCulture would be German, but I also prefer all my applications in English, so the CurrentUICulture would be English.

angularsen:
Yes, that matches how I remember it when reading up on that a long time ago. CurrentCulture for numbers, CurrentUICulture for abbreviations and other texts tied to language/culture.
However, for this PR, we can't do that change I think. Changing backing value to current thread's culture is fine I think, should not affect many - only those actively manipulating this value from multiple threads.
Changing it from CurrentUICulture to CurrentCulture sounds like would affect a lot more people, such as that german guy you mentioned. 

I'm happy to address this in v5, but for this PR I think it's out of scope?

tmilnthorp:
Agreed!

PR #617, opened 2019-02-26
Fix msbuild warnings, re-enable and treat all as errors

### LINKED ISSUE #495: Re-enable and fix msbuild warnings
**Wait for v4 branch to merge before starting on this.**

Due to thousands of msbuild warnings, several msbuild warnings are muted.
The most prominent ones are due to missing XML documentation on public types and methods.

We should address this warning and the others that are currently disabled, but right now working on v4 branch it just gets in the way. 

At least these, but check the other .csproj files too:
https://github.com/angularsen/UnitsNet/blob/0211ecce19e966b913b30ab5a8d915e9c9c01e84/UnitsNet/UnitsNet.Common.props#L27

https://github.com/angularsen/UnitsNet/blob/0211ecce19e966b913b30ab5a8d915e9c9c01e84/UnitsNet.WindowsRuntimeComponent/UnitsNet.WindowsRuntimeComponent.csproj#L21

angularsen:
Fixes #495 .

This PR fixes several thousands of warnings. Left WRC projects alone as-is.
Moving forward, all public types/members should have xmldoc and this PR helps enforce that.

- Add tons of xmldoc for public types/members
- Remove `<NoWarn>` from projects to unmute
- Mute a few warnings
- Add pragma around internal uses of obsolete members

### DIFF
--- a/UnitsNet.Serialization.JsonNet.CompatibilityTests/UnitsNet.Serialization.JsonNet.CompatibilityTests.csproj
+++ b/UnitsNet.Serialization.JsonNet.CompatibilityTests/UnitsNet.Serialization.JsonNet.CompatibilityTests.csproj
@@ -4,6 +4,7 @@
     <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
     <RootNamespace>UnitsNet.Serialization.JsonNet.CompatibilityTests</RootNamespace>
     <LangVersion>7.3</LangVersion>
+    <MSBuildTreatWarningsAsErrors>true</MSBuildTreatWarningsAsErrors>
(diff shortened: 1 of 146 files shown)


tmilnthorp:
That's a lot of doc! Looks good! Good to keep errors as warnings on IMHO.

angularsen:
I agree and for a small library it's absolutely feasible to do.

tmilnthorp:
Just realized it's not turned on for WRC

angularsen:
Yeah I don't want to touch those anymore really unless I have to

tmilnthorp:
Understandable 😆

PR #618, opened 2019-02-27
Remove reflection

tmilnthorp:
IQuantity and QuantityInfo negates the need for a lot of reflection 👍

### DIFF
--- a/UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
+++ b/UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
@@ -117,8 +117,8 @@ namespace UnitsNet.Serialization.JsonNet
             {
                 // See ValueUnit about precision loss for quantities using decimal type.
                 Value = quantity.Value,
-                Unit = $"{quantity.QuantityInfo.UnitType.Name}.{quantity.Unit}"
-            });
+                Unit = $"{quantity.QuantityInfo.UnitType.Name}.{quantity.Unit}" // Example: "MassUnit.Kilogram"
+            } );
         }
 
         /// <summary>
--- a/UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
+++ b/UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
@@ -2,13 +2,9 @@
 // Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
 
 using System;
-using System.Linq;
-using System.Reflection;
 using JetBrains.Annotations;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
-using UnitsNet.Serialization.JsonNet.Internal;
-using UnitsNet.Units;
 
 namespace UnitsNet.Serialization.JsonNet
 {
@@ -27,11 +23,6 @@ namespace UnitsNet.Serialization.JsonNet
     public class UnitsNetJsonConverter : JsonConverter
     {
         /// <summary>
-        /// Numeric value field of a quantity, typically of type double or decimal.
-        /// </summary>
-        private const string ValueFieldName = "_value";
-
-        /// <summary>
         ///     Reads the JSON representation of the object.
         /// </summary>
         /// <param name="reader">The <see cref="T:Newtonsoft.Json.JsonReader" /> to read from.</param>
@@ -46,31 +37,22 @@ namespace UnitsNet.Serialization.JsonNet
             JsonSerializer serializer)
         {
             if (reader.ValueType != null)
-            {
                 return reader.Value;
-            }
-            object obj = TryDeserializeIComparable(reader, serializer);
+
+            var obj = TryDeserializeIComparable(reader, serializer);
             // A null System.Nullable value or a comparable type was deserialized so return this
             if (!(obj is ValueUnit vu))
-            {
                 return obj;
-            }
 
             // "MassUnit.Kilogram" => "MassUnit" and "Kilogram"
-            string unitEnumTypeName = vu.Unit.Split('.')[0];
-            string unitEnumValue = vu.Unit.Split('.')[1];
-
-            // "MassUnit" => "Mass"
-            string quantityTypeName = unitEnumTypeName.Substring(0, unitEnumTypeName.Length - "Unit".Length);
+            var unitEnumTypeName = vu.Unit.Split('.')[0];
+            var unitEnumValue = vu.Unit.Split('.')[1];
 
             // "UnitsNet.Units.MassUnit,UnitsNet"
-            string unitEnumTypeAssemblyQualifiedName = "UnitsNet.Units." + unitEnumTypeName + ",UnitsNet";
-
-            // "UnitsNet.Mass,UnitsNet"
-            string quantityTypeAssemblyQualifiedName = "UnitsNet." + quantityTypeName + ",UnitsNet";
+            var unitEnumTypeAssemblyQualifiedName = "UnitsNet.Units." + unitEnumTypeName + ",UnitsNet";
 
             // -- see http://stackoverflow.com/a/6465096/1256096 for details
-            Type unitEnumType = Type.GetType(unitEnumTypeAssemblyQualifiedName);
+            var unitEnumType = Type.GetType(unitEnumTypeAssemblyQualifiedName);
             if (unitEnumType == null)
             {
                 var ex = new UnitsNetException("Unable to find enum type.");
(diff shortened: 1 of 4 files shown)


UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
```diff
@@ -167,91 +96,31 @@ private static object TryDeserializeIComparable(JsonReader reader, JsonSerialize
...
-            string quantityUnitName = GetUnitFullNameOfQuantity(obj, quantityType); // Example: "MassUnit.Kilogram"
+            var quantity = obj as IQuantity;
 
             serializer.Serialize(writer, new ValueUnit
             {
                 // See ValueUnit about precision loss for quantities using decimal type.
-                Value = Convert.ToDouble(quantityValue),
-                Unit = quantityUnitName
+                Value = quantity.Value,
+                Unit = $"{quantity.QuantityInfo.UnitType.Name}.{quantity.Unit}"
```
angularsen: I would add the comment ` // Example: "MassUnit.Kilogram"` here, it's not entirely obvious what the result is.

UnitsNet.Tests/UnitConverterTest.cs
```diff
@@ -77,13 +77,6 @@ public void ConvertByAbbreviation_ConvertsTheValueToGivenUnit(double expectedVal
             Assert.Equal(expectedValue, UnitConverter.ConvertByAbbreviation(inputValue, quantityTypeName, fromUnit, toUnit));
         }
 
-        [Theory]
-        [InlineData(1, "UnknownQuantity", "m", "cm")]
-        public void ConvertByAbbreviation_ThrowsQuantityNotFoundExceptionOnUnknownQuantity(double inputValue, string quantityTypeName, string fromUnit, string toUnit)
```
angularsen: Hm, why was this test case removed? Is it no longer relevant/valid?

UnitsNet/UnitConverter.cs
```diff
@@ -54,7 +40,7 @@ public static double Convert(QuantityValue fromValue, Enum fromUnitValue, Enum t
         public static bool TryConvert(QuantityValue fromValue, Enum fromUnitValue, Enum toUnitValue, out double convertedValue)
         {
             convertedValue = 0;
-            if (!Quantity.TryFrom(fromValue, fromUnitValue, out IQuantity from)) return false;
+            if (!Quantity.TryFrom(fromValue, fromUnitValue, out var from)) return false;
```
angularsen: I'd add back `IQuantity` here, it's not obvious to me what type `from` is.

UnitsNet/UnitConverter.cs
```diff
@@ -99,17 +85,17 @@ public static bool TryConvert(QuantityValue fromValue, Enum fromUnitValue, Enum
         /// <exception cref="AmbiguousUnitParseException">More than one unit matches the abbreviation.</exception>
         public static double ConvertByName(QuantityValue fromValue, string quantityName, string fromUnit, string toUnit)
         {
-            if (!TryGetUnitType(quantityName, out Type unitType))
+            if (!TryGetUnitType(quantityName, out var unitType))
                 throw new UnitNotFoundException($"The unit type for the given quantity was not found: {quantityName}");
 
-            if (!TryParseUnit(unitType, fromUnit, out Enum fromUnitValue)) // ex: LengthUnit.Meter
+            if (!TryParseUnit(unitType, fromUnit, out var fromUnitValue)) // ex: LengthUnit.Meter
```
angularsen: Here too, in my mind it's not super obvious that the type is `Enum` although the comment helps hint at it and you can reason about it. I just prefer it to not have to think so much when reading.

UnitsNet/UnitConverter.cs
```diff
@@ -99,17 +85,17 @@ public static bool TryConvert(QuantityValue fromValue, Enum fromUnitValue, Enum
...
             {
                 var e = new UnitNotFoundException($"Unit not found [{fromUnit}].");
                 e.Data["unitName"] = fromUnit;
                 throw e;
             }
 
-            if (!TryParseUnit(unitType, toUnit, out Enum toUnitValue)) // ex: LengthUnit.Centimeter
+            if (!TryParseUnit(unitType, toUnit, out var toUnitValue)) // ex: LengthUnit.Centimeter
```
angularsen: Same here.

UnitsNet.Tests/UnitConverterTest.cs
angularsen: I see further down why, but I think we can keep the test case and test for UnitNotFoundException instead?

UnitsNet/UnitConverter.cs
```diff
@@ -229,24 +215,16 @@ public static double ConvertByAbbreviation(QuantityValue fromValue, string quant
         /// <exception cref="AmbiguousUnitParseException">More than one unit matches the abbreviation.</exception>
         public static double ConvertByAbbreviation(QuantityValue fromValue, string quantityName, string fromUnitAbbrev, string toUnitAbbrev, string culture)
```
angularsen: Should `culture` here be `IFormatProvider`? Is this a slip up from the WRC-splitting PR? If it was, I should fix that. If it wasn't, maybe we can add to v5 roadmap.

UnitsNet/UnitConverter.cs
```diff
@@ -229,24 +215,16 @@ public static double ConvertByAbbreviation(QuantityValue fromValue, string quant
...
                 throw new UnitNotFoundException($"The unit type for the given quantity was not found: {quantityName}");
 
             var cultureInfo = string.IsNullOrWhiteSpace(culture) ? GlobalConfiguration.DefaultCulture : new CultureInfo(culture);
 
             var fromUnitValue = UnitParser.Default.Parse(fromUnitAbbrev, unitType, cultureInfo); // ex: ("m", LengthUnit) => LengthUnit.Meter
-            var toUnitValue = UnitParser.Default.Parse(toUnitAbbrev, unitType, cultureInfo); // ex:("cm", LengthUnit) => LengthUnit.Centimeter
-
-            var fromMethod = GetStaticFromMethod(quantityType, unitType); // ex: UnitsNet.Length.From(double inputValue, LengthUnit inputUnit)
-            var fromResult = fromMethod.Invoke(null, new object[] {fromValue, fromUnitValue}); // ex: Length quantity = UnitsNet.Length.From(5, LengthUnit.Meter)
+            var fromResult = Quantity.From(fromValue, fromUnitValue);
```
angularsen: Nitpicky, but `from/toResult` could maybe be renamed to `from/to` or `from/toQuantity`? Same below.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
```diff
@@ -46,103 +37,41 @@ public class UnitsNetJsonConverter : JsonConverter
...
-        /// <returns></returns>
-        private static object GetFromMethodValueArgument(MethodInfo fromMethod, double value)
-        {
-            Type valueParameterType = fromMethod.GetParameters()[0].ParameterType;
-            if (valueParameterType == typeof(QuantityValue))
-            {
-                // We use this type that takes implicit cast from all number types to avoid explosion of method overloads that take a numeric value.
-                return (QuantityValue) value;
-            }
+            var value = vu.Value;
```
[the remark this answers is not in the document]
tmilnthorp: I did it based on the suggestion of Visual Studio to simplify. We can allow the IDE to help here as IntelliSense lets you know the type if that works?

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
```diff
@@ -46,103 +37,41 @@ public class UnitsNetJsonConverter : JsonConverter
...
-        /// <param name="quantityType">Type of quantity, such as <see cref="Mass"/>.</param>
-        /// <param name="value">Numeric value.</param>
-        /// <param name="unitValue">The unit, such as <see cref="MassUnit.Kilogram"/>.</param>
-        /// <returns>The constructed quantity, such as <see cref="Mass"/>.</returns>
-        private static object CreateQuantity(Type quantityType, double value, object unitValue)
-        {
-            // We want the non-nullable return type, example candidates if quantity type is Mass:
-            // double Mass.From(double, MassUnit)
-            // double? Mass.From(double?, MassUnit)
-            MethodInfo notNullableFromMethod = quantityType
```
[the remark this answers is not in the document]
tmilnthorp: 👍

UnitsNet.Tests/UnitConverterTest.cs
tmilnthorp: It fails because it doesn't throw it any more 😃 Code is unnecessary.

UnitsNet/UnitConverter.cs
tmilnthorp: See above. If you agree we can leave, if not I'll add it back!

UnitsNet/UnitConverter.cs
tmilnthorp: See above.

UnitsNet/UnitConverter.cs
tmilnthorp: See above.

UnitsNet/UnitConverter.cs
tmilnthorp: It probably should be. Let's add that.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
```diff
@@ -167,91 +96,31 @@ private static object TryDeserializeIComparable(JsonReader reader, JsonSerialize
...
         public override void WriteJson(JsonWriter writer, object obj, JsonSerializer serializer)
         {
-            Type quantityType = obj.GetType();
+            var quantityType = obj.GetType();
 
             // ValueUnit should be written as usual (but read in a custom way)
             if(quantityType == typeof(ValueUnit))
             {
-                JsonSerializer localSerializer = new JsonSerializer()
+                var localSerializer = new JsonSerializer()
```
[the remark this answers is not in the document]
tmilnthorp: The TypeNameHandling property of localSerializer depends on the method's serializer parameter, so I think it should be left as a local variable.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
tmilnthorp: Done!

UnitsNet/UnitConverter.cs
tmilnthorp: Done!

UnitsNet/UnitConverter.cs
tmilnthorp: I did rename this ```fromQuantity``` to match the other rename you suggested. That should make it more obvious too.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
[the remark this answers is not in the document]
angularsen: It helps, but I want to advocate code that reads well without the help of an IDE - such as when browsing code on GitHub or popping up a file in a text editor - or even just not having to hover the mouse over the variable or triggering intellisense to learn what type it is.

My personal preference is to only use `var` when it is evident what the type is such as assigning the int or double on the same line or a `var probablySomeType = SomeType.From("foo")`. 

ReSharper has a config for "use var if evident", but Visual Studio doesn't seem to have this notion.
In this particular place, it's not clear to me whether this is double, decimal or maybe even `object` without reading other lines to know what `vu` and its property `Value` really is.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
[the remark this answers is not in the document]
angularsen: Ah I missed that.

UnitsNet.Tests/UnitConverterTest.cs
angularsen: But.. I honestly don't know what result to expect when calling this method with these arguments? What results are we getting if not exception?

UnitsNet/UnitConverter.cs
angularsen: See my counter-argument above 😆 Here too, I wouldn't know if it was `object`, `Enum` or perhaps even a specific enum type if I don't know the code well.

UnitsNet/UnitConverter.cs
angularsen: Added to v5 issue, confirmed it was from before the WRC-splitting.

UnitsNet/UnitConverter.cs
angularsen: I would still prefer to add it back here.

UnitsNet.Tests/UnitConverterTest.cs
tmilnthorp: It will also give UnitNotFoundException. I'll add it back and catch that.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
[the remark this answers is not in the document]
tmilnthorp: Sounds good. I've added the types back!

UnitsNet/UnitConverter.cs
tmilnthorp: Done!

UnitsNet/UnitConverter.cs
tmilnthorp: Done!

UnitsNet/UnitConverter.cs
tmilnthorp: Done!

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
```diff
@@ -39,39 +39,39 @@ public class UnitsNetJsonConverter : JsonConverter
...
+            double value = vu.Value;
+            Enum unitValue = (Enum)Enum.Parse(unitEnumType, unitEnumValue); // Ex: MassUnit.Kilogram
 
             return Quantity.From(value, unitValue);
         }
 
         private static object TryDeserializeIComparable(JsonReader reader, JsonSerializer serializer)
         {
-            var token = JToken.Load(reader);
+            JToken token = JToken.Load(reader);
```
angularsen: I see you reverted all the types back. That's fine, although I only meant for the few places where the exact type wasn't obvious to have explicit types.

angularsen:
Very nice 🎉

PR #619, opened 2019-03-01
Renaming confusing methods

tmilnthorp:
Based on feedback in #588, adding this as its own PR. That one has some ways to go so figured I'd just get this in.


angularsen:
Ahh, so much better!

PR #620, opened 2019-03-01
Moving implemented interfaces to IQuantity

tmilnthorp:
Allows more to be done with IQuantity interface, for example:
- Use Linq on collections of IQuantity (sort, min/max, etc.)
- Pass an IQuantity instance to methods that take an IComparable or IConvertible


tmilnthorp:
I've actually decided I don't like this. It means IQuantity MUST implement those interfaces and I don't want to burden implementors with them if not needed

angularsen:
Hm yeah, interesting arguments for and against. I'm not sure which way I prefer yet. Another argument against is that you can't really sort a list of `IQuantity` if it contains different quantity types, such as Temperature and Length. On the other hand you can easily filter on quantity type etc. I don't know, can keep it on the backburner to see if a real need for it pops up sometime later.

PR #621, opened 2019-03-01
IQuantity As/ToUnit method exceptions

tmilnthorp:
IQuantity As/ToUnit methods would blindly cast an Enum to the unit type, even if it was the wrong unit type. Throwing an exception for wrong type.

### DIFF
--- a/UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
+++ b/UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
@@ -723,7 +723,14 @@ namespace UnitsNet
 
         #region Conversion Methods
 
-        double IQuantity.As(Enum unit) => As((AccelerationUnit)unit);
+        /// <inheritdoc />
+        double IQuantity.As(Enum unit)
+        {
+            if(!(unit is AccelerationUnit))
+                throw new ArgumentException("The given unit is not of type AccelerationUnit.", nameof(unit));
+
+            return As((AccelerationUnit)unit);
+        }
 
         /// <summary>
         ///     Convert to the unit representation <paramref name="unit" />.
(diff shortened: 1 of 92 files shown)


UnitsNet/GeneratedCode/Quantities/Acceleration.NetFramework.g.cs
```diff
@@ -738,9 +745,6 @@ public double As(AccelerationUnit unit)
             return Convert.ToDouble(converted);
         }
 
-        /// <inheritdoc />
-        public double As(Enum unit) => As((AccelerationUnit) unit);
```
tmilnthorp: Removed as there's no point in exposing the As(Enum) method on the class itself. People would just use As(AccelerationUnit) on an Acceleration instance. It's really only for IQuantity, so just keeping the explicit interface implementation.

angularsen: Good catch

UnitsNet/GeneratedCode/Quantities/AmountOfSubstance.NetFramework.g.cs
```diff
@@ -753,7 +753,14 @@ public override int GetHashCode()
 
         #region Conversion Methods
 
-        double IQuantity.As(Enum unit) => As((AmountOfSubstanceUnit)unit);
+        /// <inheritdoc />
+        double IQuantity.As(Enum unit)
+        {
+            if(!(unit is AmountOfSubstanceUnit))
```
angularsen: Optional, but you can rewrite this to latest c# for extra vanity points:
```c#
if (unit is AmountOfSubstanceUnit validUnit)
{
    return As(validUnit);
}

throw new ArgumentException("The given unit is not of type AmountOfSubstanceUnit.", nameof(unit));
```

angularsen: Also, we could add `unit.GetType()` to the exception message to give extra help figuring out what went wrong.

tmilnthorp: I only thought you could do that with ```as```. Sweet! And I agree, I updated the message.

PR #622, opened 2019-03-01
Adding As/ToUnit conversions to UnitSystem

tmilnthorp:


tmilnthorp:
Fat fingers + copy paste. A recipe for disaster in test naming 😆

UnitsNet/Scripts/Include-GenerateQuantitySourceCode.ps1
```diff
@@ -1010,9 +1024,29 @@ function GenerateConversionMethods([GeneratorArgs]$genArgs)
             return ToUnit(unitAs$unitEnumName);
         }
 
+        /// <summary>
+        ///     Converts the quantity into a quantity which has units compatible with the given <see cref="UnitSystem"/>.
```
angularsen: Grammar nazi 🎈 , but `which` should always be preceded with a comma.
I also changed the descriptions a bit in IQuantity, so after you review those then match this xmldoc up with those and regen the files.

UnitsNet.Tests/CustomCode/AreaTests.cs
```diff
@@ -90,5 +90,19 @@ public void Constructor_UnitSystemSI_AssignsSIUnit()
...
+
+        [Fact]
+        public void As_ImperialGivenSIUnitSystem_ReturnsSIValue()
+        {
+            var imperialArea = new Area(2.0, AreaUnit.SquareInch);
+            Assert.Equal(0.00129032, imperialArea.As(UnitSystem.SI));
+        }
+
+        [Fact]
+        public void ToUnit_ImperialGivenSIUnitSystem_ReturnsSIValue()
```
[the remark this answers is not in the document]
tmilnthorp: It will pass only for a unit system that has a base unit in meters. You could define one in feet and it would have a different quantity value. That conversion is handled by other tests, but I added it here regardless. I can change it for sure.

UnitsNet/BaseUnits.cs
```diff
@@ -101,6 +101,21 @@ public bool IsSubsetOf(BaseUnits other)
                 (LuminousIntensity == LuminousIntensityUnit.Undefined || LuminousIntensity == other.LuminousIntensity);
         }
 
+        /// <summary>
+        /// Checks if all of the base units are defined.
+        /// </summary>
+        /// <returns>True if all of the base units are defined, otherwise false.</returns>
+        public bool AreAllBaseUnitsDefined()
```
angularsen: Since BaseUnits is readonly, this could be a getter property initialized once or lazily. Prop name idea: `IsFullyDefined`

tmilnthorp: 👍 I like it better.

No need for lazy. Seven && checks is not expensive 😄 I'll just make it a property set in the constructor.

UnitsNet/Scripts/Include-GenerateQuantitySourceCode.ps1
angularsen: Pinging this one, please review my xmldoc revisions in IQuantity, and make this consistent with that, then regen.

tmilnthorp:
Yup! Sorry, I got busy with my daughter and didn't get to finish. Let me take a look now.

UnitsNet/Scripts/Include-GenerateQuantitySourceCode.ps1
tmilnthorp: Looks good to me. I just did an inheritdoc to keep things cleaner.

angularsen:
Ok nice!

UnitsNet.Tests/CustomCode/AreaTests.cs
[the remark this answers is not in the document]
angularsen: Wait, I forgot this one before merging.

Are you saying that this test will only pass because `ToUnit(UnitSystem.SI)` converts to the Meter unit?
I think that is false, because my the Area equality converts to the default/base unit before comparing values.

As an example: `Length.FromCentimeters(100).Equals(Length.FromMeters(1))` is `true`.

This is why I think this test does not actually test the  `ToUnit` behavior at all - because it could have incorrectly converted to nautical mile unit and it would still equal the left hand side.

tmilnthorp:
Oh yea. #629 to fix.

PR #624, opened 2019-03-04
Use auto properties where possible

tmilnthorp:
Just to match the rest of the library where we use them heavily.

I wish I could do the quantity classes (_value), but the old json still uses reflection to get it. Now that reflection is gone we could make the change but have to update the json and break forward compatibility.

### DIFF
--- a/UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonConverterTests.cs
+++ b/UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonConverterTests.cs
@@ -9,22 +9,22 @@ namespace UnitsNet.Serialization.JsonNet.Tests
 {
     public class UnitsNetJsonConverterTests
     {
-        private readonly JsonSerializerSettings _jsonSerializerSettings;
+        private JsonSerializerSettings JsonSerializerSettings { get; }
 
         protected UnitsNetJsonConverterTests()
         {
-            _jsonSerializerSettings = new JsonSerializerSettings {Formatting = Formatting.Indented};
-            _jsonSerializerSettings.Converters.Add(new UnitsNetJsonConverter());
+            JsonSerializerSettings = new JsonSerializerSettings {Formatting = Formatting.Indented};
+            JsonSerializerSettings.Converters.Add(new UnitsNetJsonConverter());
         }
 
         private string SerializeObject(object obj)
         {
-            return JsonConvert.SerializeObject(obj, _jsonSerializerSettings).Replace("\r\n", "\n");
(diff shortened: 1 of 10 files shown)


UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonConverterTests.cs
```diff
@@ -9,22 +9,22 @@ namespace UnitsNet.Serialization.JsonNet.Tests
 {
     public class UnitsNetJsonConverterTests
     {
-        private readonly JsonSerializerSettings _jsonSerializerSettings;
+        private JsonSerializerSettings JsonSerializerSettings { get; }
```
angularsen: This is probably only a preference thing, but I find private properties peculiar. Usually a property is used to be able to expose it in an interface or to add some implementation to its setters/getters. I don't see how a private getter property adds anything beyond a readonly field. I can appreciate the consistency argument, but in my experience this is a bit against the C# norm.

UnitsNet/CustomCode/Quantity.cs
```diff
@@ -7,7 +7,7 @@ namespace UnitsNet
 {
     public partial class Quantity
     {
-        private static readonly Lazy<QuantityInfo[]> InfosLazy;
+        private static Lazy<QuantityInfo[]> InfosLazy { get; }
```
angularsen: Same here, a static prop will never be part of an interface either, so I don't think this change adds value.

UnitsNet/CustomCode/QuantityParser.cs
```diff
@@ -124,7 +124,7 @@ static QuantityParser()
             bool matchEntireString = true)
             where TUnitType : Enum
         {
-            var unitAbbreviations = _unitAbbreviationsCache.GetUnitAbbreviations(unit, formatProvider);
+            var unitAbbreviations = UnitAbbreviationsCache.GetUnitAbbreviations(unit, formatProvider);
```
angularsen: This is one of the reasons I don't favor properties over fields, it's harder to tell from reading the code that we are using a private member vs something public.

tmilnthorp:
I don't mind either way. I like them, and hate them 😆 

Lines like ```UnitAbbreviationsCache.GetUnitAbbreviations``` make it hard to tell if it's static, although if going to something else like ```instance.UnitAbbreviationsCache.GetUnitAbbreviations```.

I'm happy to abandon it. I got bored last night and needed something mindless to mess with 😃

angularsen:
Great, I'm glad you didn't pour your soul into it :-)

PR #625, opened 2019-03-04
Adding BaseUnits for a handful of quantities

tmilnthorp:


Common/UnitDefinitions/Force.json
```diff
@@ -11,6 +11,11 @@
     {
       "SingularName": "Dyn",
       "PluralName": "Dyne",
+      "BaseUnits": {
+        "L": "Centimeter",
```
angularsen: Where do you look up these facts? I have no prior reference. Do you google each you are unsure about or do you have a goto-source? I found this one on wiki.

Common/UnitDefinitions/Force.json
```diff
@@ -93,6 +103,11 @@
     {
       "SingularName": "Poundal",
```
angularsen: I see there is a notion of foot-pound-second system and centimeter-gram-second system (CGS / Gauss).
I'll create a PR to discuss adding those.

angularsen:
Looks good to me

Common/UnitDefinitions/Force.json
tmilnthorp: Here's [one ](https://en.wikipedia.org/wiki/Dyne) for dyne:
> The dyne is defined as "the force required to accelerate a mass of one gram at a rate of one centimetre per second per second"

Or try [this](https://www.wolframalpha.com/input/?i=dyne):

> Unit system:
centimeter-gram-second (cgs)

Common/UnitDefinitions/Force.json
tmilnthorp: Indeed there is. CGS, FPS, MKS

Common/UnitDefinitions/Force.json
angularsen: Thanks

PR #626, opened 2019-03-04
New unit: Linear feet added as a new unit

Sonan1994:
Linear feet added as a new unit

### DIFF
--- a/Common/UnitDefinitions/Length.json
+++ b/Common/UnitDefinitions/Length.json
@@ -279,6 +279,18 @@
           "Abbreviations": [ "twip" ]
         }
       ]
+    },
+    {
+      "SingularName": "LinearFeet",
+      "PluralName": "LinearFeets",
+      "FromUnitToBaseFunc": "x*0.3048",
+      "FromBaseToUnitFunc": "x/0.3048",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "LF" ]
+        }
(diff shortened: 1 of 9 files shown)


tmilnthorp:
A linear foot is no different than LengthUnit.Foot.

I don't want to dismiss this prematurely as being application specific, as the abbreviation (LF) is indeed different. Is this perhaps just an abbreviation we should add to allow ToString/Parse support?

angularsen:
@Sonan1994 Any feedback on @tmilnthorp 's questions above?

Sonan1994:
> @Sonan1994 Any feedback on @tmilnthorp 's questions above?

I am so sorry for the late response. Like @tmilnthorp said, it's just abbreviation :) Thank you, for the quick response. I am going to create new pull request where I add just new abbreviation.

tmilnthorp:
No problem! I think that sounds good.

FYI parsing is easy, and when you do a ToString, I think length.ToString("a3") will work for you.

@angularsen I just thought - do we alphabetically sort abbreviations? We should not, and always add new ones to the end of the json file, and maintain that order. Otherwise ToString("a3") would be a behavior break.

angularsen:
@tmilnthorp yes I was thinking the same thing, we should always append and I think that has been the practice except for in very rare cases removing abbreviations, which is a breaking change anyway.

angularsen:
Closing this as the author would create a separate PR.

PR #627, opened 2019-03-04
Add KilogramPerLiter DensityUnit

jbrandin:


Common/UnitDefinitions/Density.json
```diff
@@ -189,6 +189,18 @@
           "Abbreviations": [ "ppg (imp.)" ]
         }
       ]
+    },
+    {
+      "SingularName": "KilogramPerLiter",
+      "PluralName": "KilogramsPerLiter",
+      "FromUnitToBaseFunc": "x*1e3",
```
tmilnthorp: @angularsen just a note when we add new units, we should make sure BaseUnits are defined.

jbrandin: @tmilnthorp Can you provide an example? It was my understanding that BaseUnits are defined at the top of the file (KilogramPerCubicMeter in this case).

tmilnthorp: Those are BaseDimensions (for example dimensions of area is Length: 2).

In my recent PR #625, you can see an example for Density [here](https://github.com/angularsen/UnitsNet/pull/625/files#diff-c93aba4c1b33543ddb186a74146b8324).

tmilnthorp: In this case, it's Kilogram and Decimeter

angularsen: @jbrandin This is a very recent addition to the library so it's quite understandable you didn't catch this :-) It will be easier once we have populated all the units with these base units. I will update the wiki with the extra step.

tmilnthorp: Completely forgot about the Wiki. Ah doc. My weakness 😞

jbrandin: Ah, very recent indeed!

```suggestion
      "BaseUnits": {
        "L": "Decimeter",
        "M": "Kilogram"
      },
      "FromUnitToBaseFunc": "x*1e3",
```

tmilnthorp: Regen and you'll be good to go!

UnitsNet/GeneratedCode/Quantities/Density.NetFramework.g.cs
```diff
@@ -68,6 +68,7 @@ static Density()
                     new UnitInfo<DensityUnit>(DensityUnit.KilogramPerCubicCentimeter, BaseUnits.Undefined),
                     new UnitInfo<DensityUnit>(DensityUnit.KilogramPerCubicMeter, BaseUnits.Undefined),
                     new UnitInfo<DensityUnit>(DensityUnit.KilogramPerCubicMillimeter, BaseUnits.Undefined),
+                    new UnitInfo<DensityUnit>(DensityUnit.KilogramPerLiter, BaseUnits.Undefined),
```
tmilnthorp: Was not regenerated after BaseUnits added

jbrandin: Fixed.

Common/UnitDefinitions/Density.json
```diff
@@ -193,6 +193,10 @@
     {
       "SingularName": "KilogramPerLiter",
       "PluralName": "KilogramsPerLiter",
+      "BaseUnits": {
+        "L": "Decimeter",
```
angularsen: @tmilnthorp Is decimeter correct? At first I found it strange, but then I realized one liter is decimeter^3 - the same way cubicmeter is meter^3. This base unit stuff is still maturing in my head :-)

tmilnthorp: Yes it is 😃

tmilnthorp:
Looks good to me

Common/UnitDefinitions/Density.json
angularsen: Awesome, please merge and release the nuget if you have the time @tmilnthorp . Or I can do it, just let me know.

PR #630, opened 2019-03-04
Experimental: Add CGS and FPS unit systems

angularsen:
Inspired by the previous PR, I figured adding these CGS and FPS unit systems would be a breeze.
It did not take long to stop me in my tracks. I truly am lost on what units to designate the 7 base units for CGS and FPS unit systems. Do you have some ideas on how to proceed here?

I couldn't find much information about the units and it almost seems like there is no 1:1 mapping between SI base units concept and CGS/FPS.  For instance, in CGS there is a table for converting from SI electric units to CGS units in ESU/EMU/Gauss systems - so maybe CSG is just a group of systems sharing those three base units and ESU/EMU/Gauss are really the systems we should consider adding instead?

https://no.wikipedia.org/wiki/CGS-systemet (Norwegian wiki)

I don't know. Maybe these systems are so old no one will care about them being available either. I'm just eager to add something more than just SI to help showcase the value of various unit systems, which is yet a fairly abstract concept to me.

Update:
After maturing this thought a bit, I think our current `UnitSystem` design needs a little bit of rethink. It is currently an SI unit system, with 7 SI base units. Other unit systems don't define all those units or define entirely different units. See #651 for discussion.

### DIFF
--- a/UnitsNet/UnitSystem.cs
+++ b/UnitsNet/UnitSystem.cs
@@ -81,12 +81,46 @@ namespace UnitsNet
         /// </summary>
         public BaseUnits BaseUnits{ get; }
 
-        private static readonly BaseUnits SIBaseUnits = new BaseUnits(LengthUnit.Meter, MassUnit.Kilogram, DurationUnit.Second,
-            ElectricCurrentUnit.Ampere, TemperatureUnit.Kelvin, AmountOfSubstanceUnit.Mole, LuminousIntensityUnit.Candela);
-
         /// <summary>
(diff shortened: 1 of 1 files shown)


tmilnthorp:
Seems they aren't specified! I could relax the fully defined for UnitSystems. But not all conversions will succeed.

angularsen:
Should also look into adding British Engineering units and English Engineering units, as a proof of concept.

angularsen:
Closing this for now.

PR #631, opened 2019-03-05
Avoid reflection in InfosLazy

tmilnthorp:


UnitsNet/GeneratedCode/Quantity.g.cs
```diff
@@ -31,18 +30,199 @@ namespace UnitsNet
...
+                case QuantityType.Torque:
+                    return Torque.From(value, Torque.BaseUnit);
+                case QuantityType.VitaminA:
+                    return VitaminA.From(value, VitaminA.BaseUnit);
+                case QuantityType.Volume:
+                    return Volume.From(value, Volume.BaseUnit);
+                case QuantityType.VolumeFlow:
+                    return VolumeFlow.From(value, VolumeFlow.BaseUnit);
+                default:
+                    return null;
```
angularsen: I think we should throw instead of return null here. I believe `ArgumentOutOfRangeException` is often used for invalid enum values, or at least R# generates that for auto-generated switch statements.

UnitsNet/GeneratedCode/Quantity.g.cs
```diff
@@ -376,288 +531,192 @@ public static bool TryParse([CanBeNull] IFormatProvider formatProvider, Type qua
...
+                case Type _ when quantityType == typeof(Torque):
+                    return parser.TryParse<Torque, TorqueUnit>(quantityString, formatProvider, Torque.From, out quantity);
+                case Type _ when quantityType == typeof(VitaminA):
+                    return parser.TryParse<VitaminA, VitaminAUnit>(quantityString, formatProvider, VitaminA.From, out quantity);
+                case Type _ when quantityType == typeof(Volume):
+                    return parser.TryParse<Volume, VolumeUnit>(quantityString, formatProvider, Volume.From, out quantity);
+                case Type _ when quantityType == typeof(VolumeFlow):
+                    return parser.TryParse<VolumeFlow, VolumeFlowUnit>(quantityString, formatProvider, VolumeFlow.From, out quantity);
+                default:
+                    throw new ArgumentException(
```
angularsen: If we go with ArgOutOfRangeException above, maybe we should change this to the same exception type.

tmilnthorp: Actually... this is a TryParse. It should return false!

UnitsNet/GeneratedCode/Quantity.g.cs
angularsen: And maybe some test cases for the new public method `FromQuantityType`, such as also passing in an invalid enum value.

UnitsNet/GeneratedCode/Quantity.g.cs
angularsen: Ouch :-D I didn't see, the PR review collapses too much code sometimes 😇

tmilnthorp: Added a test that made that fail. Also updated to return false 😃

UnitsNet/GeneratedCode/Quantity.g.cs
tmilnthorp: Done!

UnitsNet.Tests/QuantityTest.cs
```diff
@@ -135,6 +136,13 @@ public void TryFrom_GivenValueAndUnit_ReturnsQuantity()
             Assert.Equal(Pressure.FromMegabars(3), parsedPressure);
         }
 
+        [Fact]
+        public void TryParse_GivenInvalidQuantityType_ReturnsFalseAndNullQuantity()
+        {
+            Assert.False(Quantity.TryParse(typeof(DummyIQuantity), "3.0 cm", out IQuantity parsedLength));
```
angularsen: Nice

UnitsNet.Tests/QuantityTest.cs
```diff
@@ -171,5 +179,27 @@ public void Types_ReturnsKnownQuantityTypes()
...
         }
+
+        [Fact]
+        public void FromQuantityType_GivenUndefinedQuantityType_ThrowsArgumentException()
+        {
+            Assert.Throws<ArgumentException>(() => Quantity.FromQuantityType(QuantityType.Undefined, 0.0));
+        }
+
+        [Fact]
+        public void FromQuantityType_GivenInvalidQuantityType_ThrowsArgumentException()
```
angularsen: I added this test case

PR #636, opened 2019-03-06
Adding base units for volume

tmilnthorp:


Common/UnitDefinitions/Volume.json
```diff
@@ -189,6 +222,9 @@
     {
       "SingularName": "ImperialGallon",
       "PluralName": "ImperialGallons",
+      "BaseUnits": {
+        "L": "Decimeter"
```
angularsen: Will this throw an exception due to multiple units having the same base unit? 
`new Volume(1, new UnitSystem(LengthUnit.Decimeter);`

I wonder if it would be acceptable to just choose one in that case? I'm too far removed from the usecase to know what is expected behavior.

Common/UnitDefinitions/Volume.json
```diff
@@ -206,6 +242,9 @@
     {
       "SingularName": "ImperialOunce",
       "PluralName": "ImperialOunces",
+      "BaseUnits": {
+        "L": "Decimeter"
```
angularsen: How do you determine this is decimeter? Also, UsOunce below is inches. Should they be the same unit?

Common/UnitDefinitions/Volume.json
```diff
@@ -275,6 +323,9 @@
     {
       "SingularName": "AuTablespoon",
       "PluralName": "AuTablespoons",
+      "BaseUnits": {
+        "L": "Decimeter"
```
angularsen: Is this not inches because Australia uses metric units? And, could it just as well be centimeters? Why/why not?
 I struggle to see what these base units should be for non-trivial units.

Common/UnitDefinitions/Volume.json
tmilnthorp: It would. We should add a method to return all unit infos that are a subset of baseunits. We could pick the first one.

Common/UnitDefinitions/Volume.json
tmilnthorp: An imperial fluid ounce is ​1/20 of an imperial pint. An imperial pint is 1/8 of an imperial gallon. An imperial gallon is defined as exactly 4.54609 liters. A liter has base units of Decimeters.

Common/UnitDefinitions/Volume.json
tmilnthorp: An Australian tablespoon is 20 ml. A liter is 1 cubic decimeter

angularsen: But, `20 ml` is 20 cubic centimeters too? Could we choose either one, or is liter/cubicdecimeter the correct choice?

tmilnthorp:
Oops. Force pushed overwriting my changes.

Here are some base units that are only for SI right now.

PR #641, opened 2019-03-14
Adding A/in² and A/ft² for ElectricCurrentDensity. Also adding BaseUn…

tmilnthorp:
…its while I'm in there


angularsen:
Verified

PR #642, opened 2019-03-14
Get unit infos for

tmilnthorp:
@angularsen to stop the exceptions (default to first match) when constructing a quantity for a unit system.

Should we maybe give a bool flag to allow an exception? Or somehow indicate there may be more than one match.


tmilnthorp:
Maybe a From/TryFrom that returns an IEnumerable would be better?

angularsen:
> Maybe a From/TryFrom that returns an IEnumerable would be better?

I think that sounds better, yes. It clearly communicates that multiple matches should be expected.

tmilnthorp:
On each of the quantity classes, in addition to the UnitSystem constructor. I think this is still better than the exception though. I'll change the documentation before I merge to note the fact it returns the first match.

tmilnthorp:
Doc updated. I'll do the From/TryFrom in another PR.

angularsen:
[Release UnitsNet/4.17.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.17.0)

Issue #344, opened 2017-12-16
Support multiple abbreviations for kilo, mega and other prefixes of units


angularsen:
#### Problem
Currently, it is not possible to support parsing both `Area.Parse("1 km²")` and `Area.Parse("1 km^2")` due to a limitation in the JSON structure for prefix units (kilo, mega etc.) and how we parse it in powershell script.

#### Background
We support defining multiple abbreviations for a unit in JSON on the `Abbreviations` property, but many units add prefixes such as kilo, mega etc to generate those extra units by simply adding `k` and `M` prefix to the unit abbreviation. However, this design currently has two limitations:

1) Only the first abbreviation in `Abbreviations` is used when creating abbreviations for prefix units. Ex: Area may have `"Abbreviations": [ "m²", "m^2" ]` and `"Prefixes": [ "Kilo" ]`, but it only generates `km²` for the square kilometer unit.

2) If `AbbreviationsWithPrefixes` is defined, then this should be used instead of adding `k` and `M` prefixes. This is a list of abbreviations fully typed out and each position in the array matches the array `Prefixes`. This is useful for languages such as Russian where a different character should be used or where it's not as simple as prefixing a string. However, currently this is an array of strings and thus only supports 1 abbreviation per prefix. I propose to extend this to support an array of a mix of strings and arrays of strings. 

#### Proposal
Support defining multiple strings per prefix unit in JSON, but keep it backwards compatible so we can either define an array of strings (as we already do) or an array of arrays of strings.

The following (silly) example would create the following abbreviations: `m², m^2, km², km^2, Mm²`
I think in most cases it would be either an array of string OR an array of arrays of strings - and not a mix as in this example, but implementing it this way makes it flexible and should be a matter of checking each item in `AbbreviationsWithPrefixes` whether it is a string or an array.
```
"Localization": [
          {
            "Culture": "en-US",
            "Abbreviations": [ "m²", "m^2" ],
            "Prefixes": [ "Kilo", "Mega" ],
            "AbbreviationsWithPrefixes": [ ["km²", "km^2"], "Mm²" ] // contrived example
          }
  ]
```

angularsen:
Closing this due to inactivity.

angularsen:
Reopening this due to blocking #646 .

lipchev:
Here is what I tried with (first time PS): 
 ```
Localization=$unit.Localization | % { 
						foreach ($abbrSyno in $_.Abbreviations) {
[...]

PR #646, opened 2019-04-03
Concentration Units (mixtures/solutions)

lipchev:
Added the following [Concentration Units](https://en.wikipedia.org/wiki/Concentration):
- [Mass Concentration](https://en.wikipedia.org/wiki/Concentration#Mass_concentration): SI = kg/m3, typically mg/l
- [Volume Concentration ](https://en.wikipedia.org/wiki/Concentration#Volume_concentration): dimensionless, typically %
- [Mass Fraction](https://en.wikipedia.org/wiki/Concentration#Mass_fraction): SI = kg/kg, typically mg/kg (although not part of of the main group- still very useful in practice)

Modified the existing Molarity unit (AKA [Molar Concentration](https://en.wikipedia.org/wiki/Concentration#Molar_concentration)):
- Some operations that were originally based on the Density units now use the MassConcentration units instead (Note: despite the fact that they share the same measurement units- the Density is a distinct Quantity Type from the Mass Concentration)
- Marked all operators involving Molarity from Density as Obsolete

Defined some basic operations that were missing from the AmountOfSubstance/MolarMass/Mass units

Defined the triangular operations involving Mass/Molar/Volume concentrations (using the solute's Density and/or Molar Mass)

All unit tests included, most were defined by actual chemists (which I AM NOT).
Note: one of the tests (QuantityIFormattableTests.VFormatEqualsValueToString)- was failing on my machine- it passes if I add CultureInfo.CurrentUICulture to the value.ToString() - as I presume was the intended behavior


angularsen:
Thanks for contributing, I'm a bit busy these daysm but me or @tmilnthorp will get around to review it soon.

Common/UnitDefinitions/MassConcentration.json
```diff
@@ -0,0 +1,246 @@
...
+    {
+      "SingularName": "GramPerLiter",
+      "PluralName": "GramsPerLiter",
+      "FromUnitToBaseFunc": "x/1",
+      "FromBaseToUnitFunc": "x*1",
+      "Prefixes": [ "Pico", "Nano", "Micro", "Milli", "Centi", "Deci" ],
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "g/l"]
```
angularsen: I would suggest using capital `L` for liter as I believe that is the most common form? I do see we are not very consistent in our existing units and found several of both lower and uppercase.

Common/UnitDefinitions/MassConcentration.json
```diff
@@ -0,0 +1,246 @@
...
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "slug/ft³" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "GramPerLiter",
+      "PluralName": "GramsPerLiter",
+      "FromUnitToBaseFunc": "x/1",
```
angularsen: Missing `"BaseUnits"` here and several below. 

For your information, we recently introduced this concept and haven't added it to all our existing units yet, so I assume you copied some existing definitions that didn't have them.

Note that not all units make sense to define BaseUnits for, such as dimensionless units or units with no obvious mapping to SI base units.

angularsen: Can use simply `x` here, no need to multiply or divide by 1.

Common/UnitDefinitions/MassConcentration.json
```diff
@@ -0,0 +1,246 @@
...
+    {
+      "SingularName": "GramPerDeciliter",
+      "PluralName": "GramsPerDeciLiter",
+      "FromUnitToBaseFunc": "x/1e-1",
+      "FromBaseToUnitFunc": "x*1e-1",
+      "Prefixes": [ "Pico", "Nano", "Micro", "Milli", "Centi", "Deci" ],
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "g/dl" ]
```
angularsen: Same here and below, I propose using `dL` with uppercase `L`. We should standardize on a convention in the library.

Common/UnitDefinitions/MassConcentration.json
```diff
@@ -0,0 +1,246 @@
...
+    },
+    {
+      "SingularName": "PoundPerUSGallon",
+      "PluralName": "PoundsPerUSGallon",
+      "FromUnitToBaseFunc": "x*1.19826427e2",
+      "FromBaseToUnitFunc": "x/1.19826427e2",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "ppg (U.S.)" ]
```
angularsen: Good job following our conventions!

Common/UnitDefinitions/MassFraction.json
```diff
@@ -0,0 +1,106 @@
+{
+  "Name": "MassFraction",
```
angularsen: I'm a little unsure about this one. Generally, you should be able to use and extend [Ratio](https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/Ratio.json) to represent these units.

The reason I am hesitant is that there are probably infinite variations of units that are dimensionless and that would have their own unique unit abbreviations. Is it possible that you can do with using Ratio? Or is there a very compelling usecase for adding this that you can't get with Ratio? I'd like to learn more.

Common/UnitDefinitions/Molarity.json
```diff
@@ -30,7 +30,20 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "mol/L" ]
+          "Abbreviations": [ "mol/L"]
+        }
+      ]
+    },
+    {
+      "SingularName": "Molar",
```
angularsen: Does `Molar` replace `MolesPerLiter`? If so, we should maybe add `"ObsoleteText": "Deprecated, use Molar instead.",` to line 27 and maybe merge the abbreviations - or should they both exist?

Common/UnitDefinitions/VolumeConcentration.json
```diff
@@ -0,0 +1,106 @@
+{
+  "Name": "VolumeConcentration",
```
angularsen: Same as for MassFraction, I need to learn why `Ratio` is not sufficient. Keep the discussion in MassFraction thread.

Common/UnitDefinitions/MassConcentration.json
lipchev: fixed for g/l, g/dl, g/ml & kg/l

Common/UnitDefinitions/MassConcentration.json
lipchev: I was indeed copying over from the Density.. I've updated MassConcentration & Molarity types to use Kilogram/Meter & Mole/Meter respectively (incl ppg?). However I am still a little unclear as to what I should prefer when selecting between equivalent base units? I presume there is a difference in precision? So should I aim in the higher or lower range?  Could you give me an example?

Common/UnitDefinitions/MassConcentration.json
lipchev: Should I update the Density units as well?

Common/UnitDefinitions/Molarity.json
lipchev: No, they are [all equivalent](https://en.wikipedia.org/wiki/Molar_concentration#Units). I tried simply adding the abbreviations to the original MolesPerLiter unit, but failed to parse back anything except for "123 M" (like "123 mM").

Common/UnitDefinitions/MassFraction.json
lipchev: The first reason for including both entities & their respective units (like g/kg) is for parsing: I have "documents" that contain "concentrations" using any of the measure units in the list. Having the possibility to easily disambiguate the user input in a given context (i.e. chem. solutions) is key for both IR and HCI.
The second reason (and the actual reason I did not do separate PRs for each type) is the intricate way that the transformation operations bind Conc. Units < - Mass | Density | AmountOfSubstance MolarMass | Volume -> Conc. Units. For example MassConcentration = VolumeConcentration  * Density. Such operations wouldn't be possible with simple Ratios in the middle of it all.  Also a Ratio, IMAHO, is no more 'type-safe' than a Double. I've thought a little about how generics could help with the Ratios/Fractions but ended up with nothing useful- except maybe that you _have_ to know what you're referring to and not only in terms of units. A VolumeConcentration is the "Ratio< Volume >" between "constituent" / "mixture",  and that should not be "assignable from" any Ratio< Volume >, let alone Ratio. 
As a side note- I haven't implemented the previously mentioned transformation-operations for the MassFraction unit as they would have required working with lists of constituents. However, the transformations do become trivial if you're working in the getter/setter/function of a "Solute" (constituent) that is part of a MultiComponentSolution (mixture).

Common/UnitDefinitions/MassConcentration.json
```diff
@@ -12,8 +12,8 @@
       "SingularName": "GramPerCubicMillimeter",
       "PluralName": "GramsPerCubicMillimeter",
       "BaseUnits": {
-        "L": "Millimeter",
-        "M": "Gram"
+        "L": "Meter",
+        "M": "Kilogram"
```
angularsen: As mentioned in discussion, `Gram` and `Millimeter` is correct here. The point is that when doing `new MassConcentration(1, UnitSystem.SI)` it will look for any units with base units `Meter` and `Kilogram` (`KilogramPerCubicmeter`) and if found construct the quantity with that unit.

If you provide a custom `UnitSystem` with say `Length=Millimeter` and `Mass=Gram` then the above example would choose `GramPerCubicMillimeter` unit.

Common/UnitDefinitions/MassConcentration.json
```diff
@@ -165,46 +165,62 @@
...
+          "Abbreviations": [ "g/L"]
         }
       ]
     },
     {
       "SingularName": "GramPerDeciliter",
-      "PluralName": "GramsPerDeciLiter",
+      "PluralName": "GramsPerDeciLiter",    
+      "BaseUnits": {
+        "L": "Meter",
```
angularsen: Deciliter is `1cm*1cm*1cm` so `L=Centimeter`.

Common/UnitDefinitions/MassConcentration.json
angularsen: Good questions, we are kind of treading the path as we go on this new `UnitSystem` and `BaseUnits` concept so I don't have all the answers yet.

1. The choice of `BaseUnits` does not affect precision, you are maybe confused with `BaseUnit` on the quantity as opposed to `BaseUnits` on the unit? 

In https://github.com/angularsen/UnitsNet/issues/563#issuecomment-467029946 I propose renaming `Length.BaseUnit` to `Length.DefaultUnit` to avoid this ambiguity.

I updated the [wiki's step 1](https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#1-add-or-modify-json-file-for-unit-class) to be a bit more clear about `BaseUnit` vs `BaseUnits`, so please give that another read and let me know if that helps clear things up?

2. As for two equivalent units, they should specify the same `BaseUnits`. One of them (first match) will be selected when constructing quantities using `UnitSystem` argument. It has no effect if not using `UnitSystem` to construct quantities, as I have now tried to clarify in the wiki.

3. I see your latest commit updated `BaseUnits` from say `Gram` to `Kilogram`, but it was originally correct. Please see the examples in the linked wiki and let me know if that helps or should be further improved.

Common/UnitDefinitions/Molarity.json
angularsen: Ah, yes, this is currently not supported: https://github.com/angularsen/UnitsNet/issues/344

I really want to avoid adding a separate unit to work around this problem and rather fix the support for multiple abbreviations with prefixes to begin with. Would you be interested in taking a look at that? 

I don't have too much time to spend on this project these days, so it may take some time for me to get around to it.

UnitsNet.Tests/CustomCode/MassFractionTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        protected override double PartsPerBillionInOneDecimalFraction => 1e9;
+
+        protected override double PartsPerMillionInOneDecimalFraction => 1e6;
+
+        protected override double PartsPerThousandInOneDecimalFraction => 1e3;
+
+        protected override double PartsPerTrillionInOneDecimalFraction => 1e12;
+
+        [Fact]
+        public void MassFractionFromMassesConsturctedCorrectly()
```
angularsen: Typo `Constructed`

UnitsNet.Tests/CustomCode/MassFractionTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        [Fact]
+        public void TotalMassFromMassFraction()
+        {
+            var componentMass = Mass.FromKilograms(1);
+            var massFraction = MassFraction.FromPercent(50);
+
+            var totalMass = massFraction.GetTotalMass(componentMass);
+            AssertEx.EqualTolerance(2, totalMass.Kilograms, KilogramsPerKiloGramTolerance);
+        }
+        [Fact]
```
angularsen: Empty line between methods

UnitsNet.Tests/CustomCode/MassFractionTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            var one_kg = Mass.FromKilograms(1);
+            var two_kg = Mass.FromKilograms(2);
+
+            var massFraction = MassFraction.FromMasses(one_kg, two_kg);
+            AssertEx.EqualTolerance(50, massFraction.Percent, PercentTolerance);
+        }
+
+        [Fact]
+        public void TotalMassFromMassFraction()
```
angularsen: Nice that you added tests, helps understand the usage.

UnitsNet.Tests/CustomCode/MolarMassTests.cs
```diff
@@ -40,5 +41,17 @@ public class MolarMassTests : MolarMassTestsBase
         protected override double NanogramsPerMoleTolerance => 1e-3;
         protected override double NanogramsPerMoleInOneKilogramPerMole => 1e12;
         protected override double PoundsPerMoleInOneKilogramPerMole => 2.2046226218487757;
+
+        private static double MolarMassOfOxygen = 15.999;
```
angularsen: `private const`

UnitsNet.Tests/CustomCode/MolarMassTests.cs
```diff
@@ -40,5 +41,17 @@ public class MolarMassTests : MolarMassTestsBase
...
+        private static double MolesInTenGramsOfOxygen = 0.6250390649415588;
+
+        [Fact]
+        public void TenGramsOfOxygenContainExpectedNumberOfMoles()
```
angularsen: Good test

UnitsNet.Tests/CustomCode/MolarityTests.cs
```diff
@@ -30,29 +30,104 @@ public class MolarityTests : MolarityTestsBase
...
+        protected override double MillimolarInOneMolesPerCubicMeter => 1;
+        protected override double MicromolarInOneMolesPerCubicMeter => 1E3;
+        protected override double NanomolarInOneMolesPerCubicMeter => 1E6;
+        protected override double PicomolarInOneMolesPerCubicMeter => 1E9;
+
+        private static double MolarMassHClInGramsPerMole = 36.46;
+        private static double MassOfSubstanceInGrams = 5;
+        private static double VolumeOfSolutionInLiters = 1.2;
+
+        private static double ExpectedMolarityMolesPerLiter = 0.1142805; // molarity = 5 / (1.2 * 36.46) = 0.114 mol/l = 0.114 M
```
angularsen: `private const` for all these

Common/UnitDefinitions/MassFraction.json
angularsen: Thanks for the detailed clarification. I don't yet fully grasp all the domain concepts here, but I think I get the gist of it. If I am reading it right, you are looking for two benefits:

1. Parsing strings like "1 g/kg"
2. Type safety and result type inference for calculations

I think these are valid reasons to have its own type.

UnitsNet.Tests/CustomCode/MolarityTests.cs
```diff
@@ -30,29 +30,104 @@ public class MolarityTests : MolarityTestsBase
...
+        private static double ExpectedConcentrationInKgPerCubicMeter = 4.16667;
+        
+        private static double MolarMassOfEthanolInGramsPerMole = 46.06844;
+        private static double DensityOfEthanolInKgPerCubicMeter = 789;
+        private static double VolumeConcentration_0_5M_Ethanol = 29.19419518377693;
 
         [Fact]
-        public void ExpectDensityConvertedToMolarityCorrectly()
+        public void ExpectMassConcentrationConvertedToMolarityCorrectly()
```
angularsen: Hm, should we remove the old tests? Isn't it good to test both Density and MassConcentration? If I read your PR description right, you say that MassConcentration does not replace Density and they should both continue to exist?

UnitsNet.Tests/CustomCode/MolarityTests.cs
```diff
@@ -30,29 +30,104 @@ public class MolarityTests : MolarityTestsBase
...
+            var substanceMass = Mass.FromGrams(MassOfSubstanceInGrams);
+            var volumeSolution = Volume.FromLiters(VolumeOfSolutionInLiters);
+            AmountOfSubstance amountOfSubstance = substanceMass / molarMass;
+
+            Molarity molarity = amountOfSubstance / volumeSolution;
+            AssertEx.EqualTolerance(ExpectedMolarityMolesPerLiter, molarity.MolesPerLiter, MolesPerLiterTolerance);
+        }
+
+        [Fact]
+        public void HClSolutionConcentrationIsEqualToExpected()
```
angularsen: These tests are great!

UnitsNet.Tests/CustomCode/MolarityTests.cs
```diff
@@ -30,29 +30,104 @@ public class MolarityTests : MolarityTestsBase
...
         protected override double PicomolesPerLiterInOneMolesPerCubicMeter => 1e9;
-        protected override double MolesPerCubicMeterInOneMolesPerCubicMeter => 1;
+
+        protected override double MolarInOneMolesPerCubicMeter => 1E-3;
+        protected override double MillimolarInOneMolesPerCubicMeter => 1;
+        protected override double MicromolarInOneMolesPerCubicMeter => 1E3;
+        protected override double NanomolarInOneMolesPerCubicMeter => 1E6;
+        protected override double PicomolarInOneMolesPerCubicMeter => 1E9;
+
+        private static double MolarMassHClInGramsPerMole = 36.46;
```
angularsen: Verified.

UnitsNet.Tests/CustomCode/MolarityTests.cs
```diff
@@ -30,29 +30,104 @@ public class MolarityTests : MolarityTestsBase
...
+        private static double MassOfSubstanceInGrams = 5;
```
angularsen: If this is an arbitrarily chosen value for a particular test case/scenario then I recommend inlining the value directly into the test methods so it's easier to read each test method on its own.

If the value is well defined or well known, then it should be renamed to something descriptive.

If I am reading it right, only a handful of these fields are well defined constants:
- MolarMassHClInGramsPerMole 
- MolarMassOfEthanolInGramsPerMole 
- DensityOfEthanolInKgPerCubicMeter 
- VolumeConcentration_0_5M_Ethanol 

The rest of them I suggest to be inlined into the test methods, so you can read the actual test case numbers for each test method. That makes it easier to read, in my opinion.

UnitsNet.Tests/CustomCode/MolarMassTests.cs
```diff
@@ -40,5 +41,17 @@ public class MolarMassTests : MolarMassTestsBase
         protected override double NanogramsPerMoleTolerance => 1e-3;
         protected override double NanogramsPerMoleInOneKilogramPerMole => 1e12;
         protected override double PoundsPerMoleInOneKilogramPerMole => 2.2046226218487757;
+
+        private static double MolarMassOfOxygen = 15.999;
+        private static double MolesInTenGramsOfOxygen = 0.6250390649415588;
```
angularsen: Verified.

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
```diff
@@ -42,6 +42,17 @@ public class AmountOfSubstanceTests : AmountOfSubstanceTestsBase
         protected override double NanopoundMolesInOneMole => 0.002204622621848776 * 1e9;
         protected override double PoundMolesInOneMole => 0.002204622621848776;
         protected override double MegamolesInOneMole => 1e-6;
+        
+        private static double MolarMassOfOxygen = 15.999;
```
angularsen: 15.9994?
https://www.convertunits.com/from/moles+Oxygen/to/grams

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
```diff
@@ -42,6 +42,17 @@ public class AmountOfSubstanceTests : AmountOfSubstanceTestsBase
...
+        private static double MolesInTenGramsOfOxygen = 0.6250390649415588;
```
angularsen: Verified.

angularsen: I see you re-define the same constants in different classes. I recommend moving these to a `Constants` class in its own file, where we can keep all the constants in one place to avoid having to duplicate them for test classes.

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
```diff
@@ -42,6 +42,17 @@ public class AmountOfSubstanceTests : AmountOfSubstanceTestsBase
...
+
+
+        private static double MolarMassHClInGramsPerMole = 36.46;
+        private static double MassOfSubstanceInGrams = 5;
+        private static double VolumeOfSolutionInLiters = 1.2;
```
angularsen: Same as in earlier comment, inline "arbitrary" values so only well-defined constants remain as fields.

UnitsNet.Tests/CustomCode/DensityTests.cs
```diff
@@ -107,25 +107,7 @@ public static void DensityTimesKinematicViscosityEqualsDynamicViscosity()
             DynamicViscosity dynamicViscosity = Density.FromKilogramsPerCubicMeter(2) * KinematicViscosity.FromSquareMetersPerSecond(10);
             Assert.Equal(dynamicViscosity, DynamicViscosity.FromNewtonSecondsPerMeterSquared(20));
         }
-
-        [Fact]
```
angularsen: Same here, shouldn't we keep the Density tests?

UnitsNet.Tests/CustomCode/MassFractionTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+        protected override double PercentInOneDecimalFraction => 100;
+
+        protected override double PartsPerBillionInOneDecimalFraction => 1e9;
+
+        protected override double PartsPerMillionInOneDecimalFraction => 1e6;
+
+        protected override double PartsPerThousandInOneDecimalFraction => 1e3;
+
+        protected override double PartsPerTrillionInOneDecimalFraction => 1e12;
```
angularsen: All above constants verified.

UnitsNet.Tests/CustomCode/MolarityTests.cs
```diff
@@ -30,29 +30,104 @@ public class MolarityTests : MolarityTestsBase
...
+        protected override double MicromolesPerLiterInOneMolesPerCubicMeter => 1e3;
         protected override double NanomolesPerLiterInOneMolesPerCubicMeter => 1e6;
         protected override double PicomolesPerLiterInOneMolesPerCubicMeter => 1e9;
-        protected override double MolesPerCubicMeterInOneMolesPerCubicMeter => 1;
+
+        protected override double MolarInOneMolesPerCubicMeter => 1E-3;
+        protected override double MillimolarInOneMolesPerCubicMeter => 1;
+        protected override double MicromolarInOneMolesPerCubicMeter => 1E3;
+        protected override double NanomolarInOneMolesPerCubicMeter => 1E6;
+        protected override double PicomolarInOneMolesPerCubicMeter => 1E9;
```
angularsen: All above verified.

UnitsNet.Tests/CustomCode/VolumeConcentrationTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        protected override double PartsPerBillionInOneDecimalFraction => 1e9;
+
+        protected override double PartsPerMillionInOneDecimalFraction => 1e6;
+
+        protected override double PartsPerThousandInOneDecimalFraction => 1e3;
+
+        protected override double PartsPerTrillionInOneDecimalFraction => 1e12;
+
+        private static double MolarMassOfEthanolInGramsPerMole = 46.06844;
+        private static double DensityOfEthanolInKgPerCubicMeter = 789;
```
angularsen: 789.3?
https://www.aqua-calc.com/calculate/weight-to-volume/substance/ethanol

UnitsNet.Tests/CustomCode/VolumeConcentrationTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        private static double VolumeConcentration_0_5M_Ethanol = 29.19419518377693;
```
angularsen: Short on time here, couldn't find a source to verify this. Do you have a link?

UnitsNet.Tests/QuantityIFormattableTests.cs
```diff
@@ -43,7 +43,7 @@ public void AFormatWithInvalidIndexThrowsFormatException()
         [Fact]
         public void VFormatEqualsValueToString()
         {
-            Assert.Equal(length.Value.ToString(), length.ToString("v"));
+            Assert.Equal(length.Value.ToString(CultureInfo.CurrentUICulture), length.ToString("v"));
```
angularsen: Good catch

UnitsNet/CustomCode/Quantities/AmountOfSubstance.extra.cs
```diff
@@ -33,5 +33,37 @@ public double NumberOfParticles()
...
+        {
+            return Molarity.FromMolesPerCubicMeter(amountOfComponent.Moles / mixtureVolume.CubicMeters);
+        }
+
+        /// <summary>Get <see cref="Volume" /> from <see cref="AmountOfSubstance" /> divided by <see cref="Molarity" />.</summary>
+        public static Volume operator /(AmountOfSubstance amountOfSubstance, Molarity molarity)
+        {
+            return Volume.FromCubicMeters(amountOfSubstance.Moles / molarity.MolesPerCubicMeter);
+        }
+
```
angularsen: All these are verified.

UnitsNet/CustomCode/Quantities/Density.extra.cs
```diff
@@ -7,26 +7,6 @@ namespace UnitsNet
 {
     public partial struct Density
     {
-        /// <summary>
-        ///     Gets <see cref="Molarity" /> from this <see cref="Density" />.
-        /// </summary>
-        /// <param name="molecularWeight"></param>
-        public Molarity ToMolarity(Mass molecularWeight)
```
angularsen: 1. We can't remove these without breaking backwards compatibility. At best, we can mark with `[Obsolete("some text")]` and remove in #563 .
2. Since Density should continue to exist, shouldn't these methods keep existing too or do we want people to use `MassConcentration` instead?

UnitsNet/CustomCode/Quantities/Molarity.extra.cs
```diff
@@ -5,40 +5,70 @@ namespace UnitsNet
 {
     public partial struct Molarity
     {
+
         /// <summary>
-        ///     Construct from <see cref="Density"/> divided by <see cref="Mass"/>.
+        ///     Get a <see cref="MassConcentration"/> from this <see cref="Molarity"/>.
         /// </summary>
-        /// <seealso cref="Density.op_Division(UnitsNet.Density,UnitsNet.Mass)"/>
```
angularsen: Can't remove things, as mentioned earlier.

Common/UnitDefinitions/MassConcentration.json
lipchev: Ok, this makes sense now... I hadn't considered custom UnitSystems before- but one could imagine having one per App, User, Screen or even one for persistence.. The only problem that comes to mind is the possible overlap of units, and the accidental change of behavior when extending existing units. It would be nice to be able to register the default units for (some) types inside a unit system- maybe even have it serve as cache for the As(UnitSystem unitSystem) method.

Common/UnitDefinitions/MassConcentration.json
lipchev: Are you sure about [the Deciliter](https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#examples-on-baseunits-values)- here is how I did the mapping: 

1. 1 l = 1 dm3 -> Decimeter
2. 1 dl = 1 ? ([0.1 dm3](https://www.google.com/search?client=firefox-b-d&channel=trow&q=1+dl+to+dm3) | [100 cm3](https://www.google.com/search?client=firefox-b-d&channel=trow&q=1+dl+to+cm3))  - Decimeter is indeed closer (and if I understand this correctly would be hidden- as Decimeter is already defined) 
3. 1 ml = 1 cm3 -> Centimeter (I personally would always prefer this to the 'alternative' Centimeter option in 2.)

lipchev: And on second thought- does it even make sense to have base(default) types defined for the Liter derivatives? Given an exhaustive list of cubic-length units- they would always come 'second' in terms of matching BaseUnits.

Common/UnitDefinitions/MassConcentration.json
angularsen: `UnitSystem` is not my brainchild, @tmilnthorp can probably answer a lot better than me, but I believe the idea is to make it easier to construct quantities and perform arithmetic with consistent units of measurement - so an application can switch from say SI unit system to British Engineering unit system and all the GUI now renders with feet and pounds and calculations are performed with said system. In particular in applications where you work with pure numbers and not wrapped in quantity types like we have in Units.NET, I understand that choosing and sticking to a particular system is crucial to get consistent results.

The overlap of units is definitely something we need to figure out. I don't have the answer to how we should best solve that, but your idea of making it configurable sounds reasonable. As previously mentioned, this whole unit system thing is still kind of experimental so any feedback is very useful.

Common/UnitDefinitions/MassConcentration.json
angularsen: In a separate PR please :-) This one is way too big already 😅

Common/UnitDefinitions/MassConcentration.json
angularsen: My bad, you are of course right. I thought 1cm³ was 1 dL, but I thought wrong.
In these cases I must defer to @tmilnthorp what is reasonable. I too struggle to determine what we should specify as base units for "not exactly matching" base units. Is closest unit ok, or should it be exact?

The way I like to think about it, is what do I expect when do this?
```c#
// pesudo code
var unitSystem = new UnitSystem { Length = Centimeter, Mass = Gram };
new MassConcentration(1, unitSystem); // 1 g/dL or 1 g/cL or exception ?
```

Common/UnitDefinitions/MassConcentration.json
lipchev: It should probably be noted that if no base (default) unit is defined for a given type then the constructor for the quantity that takes a UnitSystem will fail with "No units were found" exception:

            Assert.Equal(new Length(1, UnitSystem.SI).Unit, Length.BaseUnit);  // BaseUnit = Meter 
            Assert.Equal(new Volume(1, UnitSystem.SI).Unit, Volume.BaseUnit);  // fails because no unit is defined
Result of the second test is System.ArgumentException : No units were found for the given UnitSystem.
Parameter name: unitSystem

PS We do expect that BaseUnit is always SI

Common/UnitDefinitions/MassConcentration.json
lipchev: Maybe something like:
```
        [Fact]
        public void DefaultUnitTypeRespectedForCustomUnitSystem()
        {
            UnitSystem customSystem = new UnitSystem(new BaseUnits(LengthUnit.Millimeter, MassUnit.Gram, DurationUnit.Millisecond,
                ElectricCurrentUnit.Ampere, TemperatureUnit.DegreeCelsius, AmountOfSubstanceUnit.Mole, LuminousIntensityUnit.Candela));

            var massConcentration = new MassConcentration(1, customSystem);
            Assert.Equal(MassConcentrationUnit.GramPerCubicMillimeter, massConcentration.Unit);
        }
```

Common/UnitDefinitions/MassConcentration.json
angularsen: > Result of the second test is System.ArgumentException : No units were found for the given 

I believe that is simply because we haven't gotten around to adding `BaseUnits` to all our units yet. In https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/Volume.json I see that only two units have it defined.

UnitsNet.Tests/CustomCode/MolarityTests.cs
lipchev: No according to my colleagues those operations don't make sense- we assumed that @JonathanDavies626 had meant to express concentrations instead, but they weren't able to tell me what substances were used. 
Another indicator was the name of the variable used- (... **Mass** _molecularWeight_  ..)

UnitsNet.Tests/CustomCode/VolumeConcentrationTests.cs
lipchev: This one was given to me by my colleges. I'm sorry, I also wanted to include one from the web but couldn't find a straightforward calculator.

UnitsNet/CustomCode/Quantities/Density.extra.cs
lipchev: My bad- I've re-introduced the methods/operators (marking them as Obsolete)

UnitsNet.Tests/CustomCode/MolarityTests.cs
angularsen: Aha, that is helpful to learn. I am not proficient in this domain of units. I guess I can agree to remove/change these methods if they truly are incorrect, without having to wait for v5 release first. 

Any chance you can comment on this @JonathanDavies626 whether you agree to this change or not?

JonathanDavies626: It's been some time since I wrote this. I'll take a look and get back to you this evening, if that's okay.

angularsen: Great! Absolutely, no problem.

> I guess I can agree to remove/change these methods if they truly are incorrect, without having to wait for v5 release first.

I'm going to change my mind on this one. We shouldn't take any shortcuts here even if it is convenient. If we decide to remove any methods, we should mark them as obsolete and remove them in v5.

JonathanDavies626: @lipchev is correct. I had intended to express concentrations in the molarity conversions. Example use case is converting blood glucose measurements from mg/dL to mmol/L.

Common/UnitDefinitions/Molarity.json
lipchev: Ok, my latest commit removes the redundant Molar unit, adding it as abbreviation instead. I've annotated the failing test as Skip= Awaiting fix for #344 effectively reversing the dependency. I suggest we close this first, and then I'll have a ready test for the fix in #344 (which I think I have).

UnitsNet.Tests/CustomCode/MolarityTests.cs
angularsen: Thanks @JonathanDavies626 , much appreciated!

UnitsNet.Tests/CustomCode/DensityTests.cs
lipchev: I think not- I've left the operators as obsolete, tests were "converted"

Common/UnitDefinitions/Molarity.json
angularsen: Ok, that is fine. Let's try to close this big one first then.

Common/UnitDefinitions/MassConcentration.json
angularsen: Yep, that looks right to me.

Common/UnitDefinitions/Molarity.json
lipchev: I just saw your last commit the proper usage of Theory + InlineData so started converting where appropriate.. I agree that it's much better this way- however, when done _properly_, tests become quite a bit more verbose- I'll commit a sample of two tests for you to review- then we can figure out what's the appropriate approach...

Common/UnitDefinitions/MassFraction.json
```diff
@@ -0,0 +1,106 @@
...
+      "Prefixes": [ "Nano", "Micro", "Milli", "Centi", "Deci", "Deca", "Hecto", "Kilo" ],
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "g/g" ]
+        }
+      ]
+    },
+    {
+      "SingularName": "GramPerKiloGram",
```
angularsen: `PerKilogram`

Common/UnitDefinitions/VolumeConcentration.json
```diff
@@ -0,0 +1,106 @@
...
+	{
+      "SingularName": "LitersPerLiter",
+      "PluralName": "LitersPerLiter",
+      "FromUnitToBaseFunc": "x",
+      "FromBaseToUnitFunc": "x",
+      "Prefixes": [ "Pico", "Nano", "Micro", "Milli", "Centi", "Deci" ],
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "l/l" ]
```
angularsen: `L/L`

Common/UnitDefinitions/VolumeConcentration.json
```diff
@@ -0,0 +1,106 @@
...
+	{
+      "SingularName": "LitersPerMililiter",
+      "PluralName": "LitersPerMililiter",
+      "FromUnitToBaseFunc": "x/1e-3",
+      "FromBaseToUnitFunc": "x*1e-3",
+      "Prefixes": [ "Pico", "Nano", "Micro", "Milli", "Centi", "Deci" ],
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "l/ml" ]
```
angularsen: `L/mL`

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
angularsen: This question is still unresolved

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
```diff
@@ -42,6 +42,17 @@ public class AmountOfSubstanceTests : AmountOfSubstanceTestsBase
         protected override double NanopoundMolesInOneMole => 0.002204622621848776 * 1e9;
         protected override double PoundMolesInOneMole => 0.002204622621848776;
         protected override double MegamolesInOneMole => 1e-6;
+        
+        private static double MolarMassOfOxygen = 15.999;
+        private static double MolesInTenGramsOfOxygen = 0.6250390649415588;
+
+
+        private static double MolarMassHClInGramsPerMole = 36.46;
```
angularsen: Reuse the new `KnownQuantities` class?
And still need to inline the test case values here.

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
```diff
@@ -58,5 +69,46 @@ public void NumberOfParticlesInTwoMolesIsDoubleAvogadroConstant()
             var numberOfParticles = twoMoles.NumberOfParticles();
             Assert.Equal(AmountOfSubstance.AvogadroConstant * 2, numberOfParticles);
         }
+
+        [Fact]
+        public void TenMolesOfOxygenToMassEqualToExpected()
+        {
+            AmountOfSubstance tenMoles = AmountOfSubstance.FromMoles(10);
+            MolarMass molarMass = MolarMass.FromGramsPerMole(MolarMassOfOxygen);
+            AssertEx.EqualTolerance(10 * MolarMassOfOxygen, (tenMoles * molarMass).Grams, MolesTolerance);
```
angularsen: These tests are good, but I struggle a bit to see what is actually tested here vs what is setup code.
Please restructure the tests on the format
```c#
// Arrange
/* setup code here */

// Act
/* ideally a single line calling a single method/operator returning a single result */

// Assert
/* Assert statements here */
```


I usually omit the `// Arrange` comment, because I find that obvious. I also sometimes omit `// Act` and `// Assert` comments if there are three clearly defined paragraphs of Arrange-Act-Assert code.

Reference: http://wiki.c2.com/?ArrangeActAssert

Once restructured, see if the naming of the method can be improved. We have no hard guideline here, but we try to follow the naming convention of `WhatIsTested_WhatAreTheInputs_WhatIsTheResult()`, such as `Multiply_Given10MolesAndOxygenMolarMass_Returns10MolarMassOfOxygen`. You can probably see from my example name that I didn't fully understand what was tested here, so please choose your own wording.

Another naming style I find just as functional is `VolumeTimesMassConcentrationEqualsMass` as you have used below. It works.

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        [Fact]
+        public static void VolumeTimesMassConcentrationEqualsMass()
+        {
+            Mass mass = Volume.FromCubicMeters(3) * MassConcentration.FromKilogramsPerCubicMeter(2);
+            Assert.Equal(mass, Mass.FromKilograms(6));
+        }
+
+        [Fact]
+        public void ExpectMolarityConvertedToMassConcentrationCorrectly()
```
angularsen: If this is testing `Molarity.ToMassConcentration()` then this test should be moved to `MolarityTests` and renamed slightly. "Correctly" is a bit vague, if feasible explain exactly what is the expected outcome.

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
```diff
@@ -0,0 +1,159 @@
...
+            var molarity = Molarity.FromMolesPerLiter(1.02698355);
+            var molarMass = MolarMass.FromGramsPerMole(58.443);
+
+            MassConcentration concentration = molarity.ToMassConcentration(molarMass);     // molarity * molarMass
+            AssertEx.EqualTolerance(60.02, concentration.KilogramsPerCubicMeter, KilogramsPerCubicMeterTolerance);
+        }
+
+        [Theory]
+        [InlineData(60.02, MassConcentrationUnit.KilogramPerCubicMeter, 58.443, MolarMassUnit.GramPerMole, 1026.98355, MolarityUnit.MolesPerCubicMeter)]
+        public void ExpectMassConcentrationConvertedToMolarityCorrectly(
```
angularsen: Empty line between Act-code and Assert-code would make it easier to read, per above discussion on AAA-pattern. 
Also consider renaming method to describe what method is being tested, what are the inputs and what is the expected outcome (if feasible). Same for other tests.

UnitsNet.Tests/CustomCode/MolarMassTests.cs
```diff
@@ -40,5 +41,17 @@ public class MolarMassTests : MolarMassTestsBase
         protected override double NanogramsPerMoleTolerance => 1e-3;
         protected override double NanogramsPerMoleInOneKilogramPerMole => 1e12;
         protected override double PoundsPerMoleInOneKilogramPerMole => 2.2046226218487757;
+
+        private const double MolarMassOfOxygen = 15.999;
```
angularsen: Suggestion of 15.9994 and moving/merging it to `KnownQuantities` still applies here.

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
```diff
@@ -69,16 +70,20 @@ public class MassConcentrationTests : MassConcentrationTestsBase
...
+            double volumeValue, VolumeUnit volumeUnit,
+            double expectedMassValue, MassUnit expectedMassUnit, double tolerance= 1e-5)
         {
-            Mass mass = MassConcentration.FromKilogramsPerCubicMeter(2) * Volume.FromCubicMeters(3);
-            Assert.Equal(mass, Mass.FromKilograms(6));
+            var massConcentration = new MassConcentration(massConcValue, massConcUnit);
+            var volume = new Volume(volumeValue, volumeUnit);
+
+            Mass mass = massConcentration * volume;
+            AssertEx.EqualTolerance(mass.As(expectedMassUnit), expectedMassValue, tolerance);
```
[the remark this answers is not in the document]
lipchev: Sorry, I didn't get your second point- do you mean I assert mass.Unit == Kilogram (the result of the operator)? I still would have to convert to the expected result which could be expressed in say grams instead, right? And do we want to enforce the unit type that is the result of the operator? 
Also side-note, in the operator I did FromKilograms but I guess that was subjective (I remember questioning my choice at the time, wondering about the loss of precision)..

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
```diff
@@ -117,15 +126,34 @@ public void MassConcentrationFromVolumeConcentrationAndDensity()
...
-            var density = Density.FromKilogramsPerCubicMeter(DensityOfEthanolInKgPerCubicMeter);
+            const double VolumeConcentration_0_5M_Ethanol = 29.19419518377693;
+            var density = Density.FromKilogramsPerCubicMeter(KnownQuantities.DensityOfEthanolInKgPerCubicMeter);
             var volumeConcentration = VolumeConcentration.FromMillilitersPerLiter(VolumeConcentration_0_5M_Ethanol);
 
             MassConcentration massConcentration = volumeConcentration.ToMassConcentration(density);
             AssertEx.EqualTolerance(23.03422, massConcentration.GramsPerLiter, GramsPerCubicMeterTolerance);
         }
+
+        [Fact(Skip = "No BaseUnit defined: see https://github.com/angularsen/UnitsNet/issues/651")]
```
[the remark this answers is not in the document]
lipchev: No, this time it is "no matching unit for SI"- SI uses Kg for BaseUnit and we have Gram + prefix [..,Kilo,..] but no joy. I think it's important that new types (at least) support SI fully, but didn't want to include a redundant KilogramPer** unit.

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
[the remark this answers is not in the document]
angularsen: Yes, I suggested to assert both unit and value. 
I suppose I see your point about not necessarily being interested in what unit it converts to, as long as the quantity is correct. I don't have a strong opinion either way, you can keep it this way. Still rearrange the expected/actual arguments though.

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
lipchev: Hmm, this one (along with another from VolumeConcentration) were actually duplicated in both source and target tests. I'm going to leave XX  = input.ToXX(..) in the TestsXX. Although I'm not exactly sure where to define the actual operator- in XX or in input.class :)

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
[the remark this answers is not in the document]
angularsen: Ah, makes sense.

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
lipchev: Now I even understand why they got duplicated- my first intuition was that I'd place them where the correct TolerenceXX value lies. Then I decided to test "the operator" class and there I think there must have been a brain crash :)

angularsen: If you are testing a method, then I think the test belongs to that class.
If you are testing arithmetic operators, then if I recall correctly I think we have tried to gather those on the first/left-hand argument type. 

Example: 
```c#
// put this implementation in Force.extra.cs and the test in ForceTests
Torque t = Force.Zero * Length.Zero;
```

Common/UnitDefinitions/MassConcentration.json
```diff
@@ -64,47 +64,78 @@
...
     },
     {
-      "SingularName": "PoundPerCubicInch",
-      "PluralName": "PoundsPerCubicInch",
+      "SingularName": "GramPerMilliliter",
+      "PluralName": "GramsPerMilliliter",
       "BaseUnits": {
-        "L": "Inch",
-        "M": "Pound"
+        "M": "Gram",
```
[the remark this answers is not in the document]
lipchev: I've removed BaseUnits from GramPerDeciliter(not exact + overlap), kept it in GramPerLiter (as exact & non-overlapping), also kept it for GramPerMilliliter(exact + overlapping) because I thought it would be useful to have at least one such case for future testing.

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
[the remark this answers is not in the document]
lipchev: Almost all of the tests now conform to this format (having flipped the expected values comparison- that was of course unintentional). Let's just hope that, with all those doubles everywhere, they don't cause headaches later on when converting to decimal /< T >

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
lipchev: Most tests renamed to follow something like the second convention: SomeResultFromSomethingAndSomethingElse

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
lipchev: Following the Method/FirstOperator : TestClass mapping, I've shuffled some of the tests around. 
Also moved the Mass/MolarMass operator to the Mass.extra class (removing the MolarMass.extra)

angularsen:
Very nice, I'm at the cabin this weekend and may not have the opportunity to look at this before Sunday.

UnitsNet.Tests/CustomCode/MassConcentrationTests.cs
angularsen: Nice

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
```diff
@@ -70,45 +60,63 @@ public void NumberOfParticlesInTwoMolesIsDoubleAvogadroConstant()
...
-            Mass tenGrams = Mass.FromGrams(10);
-            MolarMass molarMass = MolarMass.FromGramsPerMole(MolarMassOfOxygen);
-            AssertEx.EqualTolerance(MolesInTenGramsOfOxygen, (tenGrams / molarMass).Moles, MolesTolerance);
+            AssertEx.EqualTolerance(expectedMass, mass.As(expectedMassUnit), tolerence);
         }
 
-        [Fact]
-        public void HClSolutionVolumeIsEqualToExpected()
+        [Theory]
+        [InlineData(5, MassUnit.Gram,
```
angularsen: To be clear, this is fine, but I didn't mean that `[InlineData]` was my preferred method of testing. Only that the values would be inlined/inserted into the test code itself, so that you can read all the numbers that go into the calculation for each test method rather than reusing fields. `[Fact]` is probably more "correct" than `[Theory]` + `[InlineData]` if there is only one test case. If you have multiple test cases/inputs, then `[InlineData]` is a good fit.

At any rate, this works and I'm not going to bother you to change it again.

angularsen:
[Release UnitsNet/4.20.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.20.0)

PR #655, opened 2019-04-17
Fix #652 - Parse fails with custom formatprovider

### LINKED ISSUE #652: Parsing units fails with a custom FormatProvder
The following exception is thrown
> Unit not found with abbreviation [кг] for unit type [UnitsNet.Units.MassUnit].

when executing the following code
```
           try
            {
                var oneKg = Mass.Parse("1 кг", new CultureInfo("ru-RU"));
            }
            catch (Exception ex)
            {
                Console.WriteLine(ex);
            }
```
given that the abbreviation for the unit is present [here](https://github.com/angularsen/UnitsNet/blob/54420846fcdbf82441495b220f3f2756ef8895d5/Common/UnitDefinitions/Mass.json#L26)

I think the issue stems from [this line of code](https://github.com/angularsen/UnitsNet/blob/54420846fcdbf82441495b220f3f2756ef8895d5/UnitsNet/CustomCode/UnitParser.cs#L53) which doesn't pass on the culture provider 

using the latest version (4.18)

angularsen:
Fixes #652 

Format provider was accidentally not forwarded in one of the `UnitParser.Parse()` methods.
Add failing tests and then fix them.
Add tests for the most common quantities.

### DIFF
--- a/UnitsNet.Tests/CustomCode/ParseTests.cs
+++ b/UnitsNet.Tests/CustomCode/ParseTests.cs
@@ -143,5 +143,71 @@ namespace UnitsNet.Tests.CustomCode
             bool actual = Length.TryParse(s, usEnglish, out Length _);
             Assert.Equal(expected, actual);
         }
+
+        [Theory]
+        [InlineData("1 ng", "en-US", 1, MassUnit.Nanogram)]
+        [InlineData("1 нг", "ru-RU", 1, MassUnit.Nanogram)]
+        [InlineData("1 g", "en-US", 1, MassUnit.Gram)]
+        [InlineData("1 г", "ru-RU", 1, MassUnit.Gram)]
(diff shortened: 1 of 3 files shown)


angularsen:
@lipchev This fixes the bug you reported.
@tmilnthorp Do you have time to review?

angularsen:
The failing test is flaky, can't repro locally. I'll investigate, we probably modify some static values in some other tests that affect these.

angularsen:
Can repro with `build.bat`, but not with Rider test runner.

angularsen:
Meh, I think my test runner fooled me. It was consistently repro'ed now and fixed a bad test case.

tmilnthorp:
Surprised we missed it too! Looks good.

angularsen:
[Release UnitsNet/4.19.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.19.0)

PR #657, opened 2019-04-21
Create code coverage reports

angularsen:
Testing out codecov.io for test coverage reports on pull requests, with diffs against master.

- Use ~[OpenCoverage](https://github.com/opencover/opencover/)~ [dotCover](https://www.jetbrains.com/dotcover/download/#section=commandline) to produce statistics
- Use [ReportGenerator](https://github.com/danielpalme/ReportGenerator) to merge reports and produce an HTML report (not dependent on CodeCov.io)
- Upload coverage reports to https://codecov.io/gh/angularsen/UnitsNet for online reports

I tried OpenCoverage first, but I ran into heaps of trouble and it seems the project is not very actively maintained. dotCover CLI tools does not require a license as far as I can tell and is much faster to run.


angularsen:
Currently getting error messages about not finding .pdb files on AppVeyor, but the same commands work locally when building with `build.bat`. Not sure why exactly.

angularsen:
Well whoop, it finally works. The numbers are horrible of course, but we'll improve that by either filtering out stuff or converting tests to test for N units.

tmilnthorp:
Looks good - but yikes. Do you mean that coverage isn't calculated properly until we switch to NUnit from XUnit?

angularsen:
I didn't mean NUnit :-D

I meant going from "hey here is a test case that tests Length, Mass and some other widely used quantities" to maybe generating unique test cases for every quantity and unit - in order to get better test coverage.

But yeah, I'll have to look closer at the numbers. I was mostly happy that I got it working at all as a start :-P

angularsen:
I'll take a second look at why the numbers are so bad before merging this.

angularsen:
@tmilnthorp Wohoo! After switching from OpenCover to dotCover I got more sane results and way better performance. 56% coverage now, which sounds reasonable given that many tests only target specific quantities.

I believe the licensing for the CLI tooling is OK, but I have also applied for an open source license just to be sure. Merging this now so we can get a baseline going for future PRs.

tmilnthorp:
Sweet. Still surprised though given all the As/To methods

angularsen:
Added issue to follow up improving the coverage numbers:
https://github.com/angularsen/UnitsNet/issues/682

PR #659, opened 2019-04-24
Quantity parser parse powerd numbers

chromhelm:
Adds the ability to the unit parsed that unit that use the notation like m^2 still can be parsed just like m².


angularsen:
Let us know when this is ready for review.

chromhelm:
Hej
I would like to get some feed back on this.
This patch is allows the unit parse to recognize unit like m² even if it is written as m^2, as most people will do it when typing in a textbox.
The strategy right now is that the unitParser tries to parse the unit.
If it can recognize one or more abbreviations it continues the as it did before.
When it does not recognize the unit it will call the NormalizeUnitString function and try again.

I especially like get feedback to the NormalizeUnitString function.
I don't like that there are so many string.replace functions. But i could not come up with a more simple readable code.

UnitsNet/CustomCode/QuantityParser.cs
```diff
@@ -183,6 +183,13 @@ private static string GetRegexPatternForUnitAbbreviations(IEnumerable<string> ab
         private static bool ExtractValueAndUnit(Regex regex, string str, out string valueString, out string unitString)
         {
             var match = regex.Match(str);
+
+            if(match.Groups[0].Value == match.Groups[1].Value || match.Groups.Count < 3)
```
angularsen: A short comment could be useful here, on its own it's hard to read what this actually checks since I don't know the regex looks like at this point and what the groups represent.

UnitsNet/CustomCode/UnitParser.cs
```diff
@@ -77,6 +77,12 @@ public Enum Parse([NotNull] string unitAbbreviation, Type unitType, [CanBeNull]
 
             var unitIntValues = abbreviations.GetUnitsForAbbreviation(unitAbbreviation, ignoreCase: true);
 
+            if (unitIntValues.Count < 1)
```
angularsen: I'd rather do `== 0`, we never expect negative counts.

UnitsNet/CustomCode/UnitParser.cs
```diff
@@ -94,6 +100,34 @@ public Enum Parse([NotNull] string unitAbbreviation, Type unitType, [CanBeNull]
             }
         }
 
+        internal static string NormalizeUnitString(string unitAbbreviation)
+        {
+            // remove all whitespace in the string
+            unitAbbreviation = new string(unitAbbreviation.Where(c => !char.IsWhiteSpace(c)).ToArray());
```
angularsen: I think you asked about feedback on this method in particular.

As far as I know, the way you strip whitespace here is perfectly fine. You could use regex, but I suspect that will just be slower and not really bring any new value.

The below `.Replace()` calls can be optimized in terms of string copying, but we are talking extremely small strings here - typically less than 10 chars. I'm sure there are clever ways to reduce string allocations instead of many calls to `string.Replace()`, but I don't know of any and not sure it's worth much effort in this case.

If you just want the code to be shorter, you can use "method chaining" like this. This does not bring any optimizations though, only shorter syntax.
```c#
unitAbbreviation = unitAbbreviation
    .Replace("^-9", "⁻⁹")
    .Replace("^-8", "⁻⁸")
    .Replace("^-7", "⁻⁷");
```

UnitsNet/CustomCode/UnitParser.cs
```diff
@@ -165,6 +199,12 @@ public bool TryParse(string unitAbbreviation, Type unitType, [CanBeNull] IFormat
 
             var unitIntValues = abbreviations.GetUnitsForAbbreviation(unitAbbreviation, ignoreCase: true);
 
+            if (unitIntValues.Count < 1)
```
angularsen: Same here, `== 0`.

lipchev:
Btw, I remember also having to replace the "micro" sign (this might not be relevant anymore):
```
        private const char MicroUTF16 = 'μ';
        private const char MicroASCII = 'µ';
```

angularsen:
> Btw, I remember also having to replace the "micro" sign (this might not be relevant anymore):
> 
> ```
>         private const char MicroUTF16 = 'μ';
>         private const char MicroASCII = 'µ';
> ```

I was not aware, that would be good to include in the normalization as well and add test cases for.

chromhelm:
This will make the test case ParseUnit_ShouldUseCorrectMicroSign redundant
Should I remove it?

angularsen:
Huh, good catch. I'm honestly not sure if we are doing a good thing or bad thing by normalizing the lower case greek letter Mu to micro. For me personally this seems like an improvement, I would never distinguish the two anyway. I don't know of any cases where greek letter Mu is used as an abbreviation, but who knows if someone wants to add Greek localizations. I don't expect that anytime ever, so...

I think I'm leaning towards removing the test as you propose and normalize - and fix it if someone later complains on what the heck we were thinking? :-)

UnitsNet/CustomCode/QuantityParser.cs
```diff
@@ -184,7 +184,11 @@ private static bool ExtractValueAndUnit(Regex regex, string str, out string valu
         {
             var match = regex.Match(str);
 
-            if(match.Groups[0].Value == match.Groups[1].Value || match.Groups.Count < 3)
+            // the regex coming in contains all allowed units as strings.
+            // That means if the unit in str is not formatted right
+            // the regex.Match will ether put str or string.empty into Groups[0] and Groups[1]
+            // Therefor a mismatch can be detected by comparing the values of this two groups.
```
angularsen: Nice

angularsen:
@chromhelm I think this PR is good to go. Did you want to touch up the `.Replace()` code first, or should I just merge?

chromhelm:
No, I think it is good as it is right now.

angularsen:
[Release UnitsNet/4.24.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.24.0)

angularsen:
Excellent! Good job on this, 4.24 nuget is on the way out.

PR #661, opened 2019-05-01
codegen: Make abbreviations for prefixes explicit

angularsen:
To avoid relying on exact same count and ordering of "Prefixes" in unit and "AbbreviationsWithPrefixes", which is bugprone when adding new prefixes.

- Rename AbbreviationsWithPrefixes to AbbreviationsForPrefixes
- Make AbbreviationsForPrefixes an object with prefixes as keys and string or string array as value
- Update JSON files


angularsen:
[Release UnitsNet/4.22.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.22.0)

PR #663, opened 2019-05-01
Support localized SI prefixes, fix Russian

### LINKED ISSUE #315: Auto-Generated Abbreviations with Prefixes for Russian
As discussed [here](https://github.com/angularsen/UnitsNet/pull/305) we should extend the script to support Russian.

angularsen:
Fixes #315.

The majority of AbbreviationsForPrefixes was adding Russian abbreviations.
This is no longer necessary for unit abbreviations that simply prepend the Russian SI prefix.
This change fixes many unit abbreviations in Russian that incorrectly used international SI prefixes instead of the Russian equivalents.

- Add localization support to SI prefixes
- Map Russian SI prefixes
- Remove redundant JSON properties of AbbreviationsForPrefixes, not just Russian


UnitsNet.sln.DotSettings
```diff
@@ -1,7 +1,13 @@
...
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=EnforceForeachStatementBraces/@EntryIndexedValue">SUGGESTION</s:String>
```
angularsen: JetBrains Rider code cleanup profiles started aggressively removing curly braces in `if` and `foreach`, so these settings seems to calm it down a bit.

angularsen:
[Release UnitsNet/4.22.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.22.0)

PR #665, opened 2019-05-04
Add revolutions per second squared unit to rotational acceleration

G18SSY:
Add a unit of _revolutions per second squared_ to the [RotationalAcceleration](https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/RotationalAcceleration.json) quantity. Used same conventions as r/s unit in [RotationalSpeed](https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/RotationalSpeed.json)

### DIFF
--- a/Common/UnitDefinitions/RotationalAcceleration.json
+++ b/Common/UnitDefinitions/RotationalAcceleration.json
@@ -41,6 +41,18 @@
           "Abbreviations": [ "rpm/s" ]
         }
       ]
+    },
+    {
+      "SingularName": "RevolutionPerSecondPerSecond",
+      "PluralName": "RevolutionsPerSecondPerSecond",
+      "FromUnitToBaseFunc": "(2*Math.PI)*x",
+      "FromBaseToUnitFunc": "(1/(2*Math.PI))*x",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "r/s²" ]
+        }
+      ]
     }
   ]
 }
--- a/Common/UnitDefinitions/RotationalAcceleration.json
+++ b/Common/UnitDefinitions/RotationalAcceleration.json
@@ -43,8 +43,8 @@
       ]
     },
     {
-      "SingularName": "RevolutionPerSecondPerSecond",
-      "PluralName": "RevolutionsPerSecondPerSecond",
+      "SingularName": "RevolutionPerSecondSquared",
+      "PluralName": "RevolutionsPerSecondSquared",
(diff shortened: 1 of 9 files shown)


Common/UnitDefinitions/RotationalAcceleration.json
```diff
@@ -41,6 +41,18 @@
           "Abbreviations": [ "rpm/s" ]
         }
       ]
+    },
+    {
+      "SingularName": "RevolutionPerSecondPerSecond",
```
angularsen: You named it "revolutions per second squared" in the description. Should maybe the unit also be named that? What is more natural in your opinion?
I see the other units are named `PerSecondSquared` too, so it would be more consistent I think.

UnitsNet.Tests/CustomCode/RotationalAccelerationTests.cs
```diff
@@ -30,5 +30,6 @@ public class RotationalAccelerationTests : RotationalAccelerationTestsBase
         protected override double DegreesPerSecondSquaredInOneRadianPerSecondSquared => 180 / Math.PI;
         protected override double RadiansPerSecondSquaredInOneRadianPerSecondSquared => 1;
         protected override double RevolutionsPerMinutePerSecondInOneRadianPerSecondSquared => 9.549296586;
+        protected override double RevolutionsPerSecondPerSecondInOneRadianPerSecondSquared => 1 / (2 * Math.PI);
```
angularsen: Use a constant here instead of calculation, preferably a constant found in some source. The idea is to double check we got the calculation right and easy to verify, as well as the conversion functions are in the right "direction" (not mixing up base to unit and unit to base, which is very easy to do).

Common/UnitDefinitions/RotationalAcceleration.json
G18SSY: Agreed, I'll change it to `RevolutionPerSecondSquared` in the next commit

UnitsNet.Tests/CustomCode/RotationalAccelerationTests.cs
G18SSY: Used a constant value from [here](https://www.smartconversion.com/unit_conversion/angular_acceleration_converter.aspx) in the next commit

angularsen:
Looks good, that was quick! Will publish nuget when I get home.

angularsen:
[Release UnitsNet/4.23.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.23.0)

PR #678, opened 2019-07-17
Adding RatioChangeRate unit

ngeslin:


Common/UnitDefinitions/RatioChangeRate.json
```diff
@@ -0,0 +1,32 @@
+﻿{
+  "Name": "RatioChangeRate",
+  "BaseUnit": "PercentPerSecond",
```
angularsen: I would choose `DecimalFractionPerSecond` as base unit, to be consistent with [Ratio](https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/Ratio.json).

You would then have to swap some of the conversion functions and test case values too.

Common/UnitDefinitions/RatioChangeRate.json
```diff
@@ -0,0 +1,32 @@
...
+  "XmlDoc": "The quantity of percent flowing per unit of time.",
```
angularsen: How about `The change in ratio per unit of time.` ?

Common/UnitDefinitions/RatioChangeRate.json
```diff
@@ -0,0 +1,32 @@
...
+    },
+    {
+      "SingularName": "DecimalFractionPerSecond",
+      "PluralName": "DecimalFractionsPerSecond",
+      "FromUnitToBaseFunc": "x*1e2",
+      "FromBaseToUnitFunc": "x/1e2",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "" ]
```
angularsen: Hmm.. should this not have an abbreviation? Not sure what, but would perhaps `1/s` work? Just something to signify the time dimensions we clearly have.

angularsen:
Nuget on the way out
[Release UnitsNet/4.30.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.30.0)

PR #687, opened 2019-08-09
Removing default parameter as an overload for ParseUnit already exist…

tmilnthorp:
…s. Adding CanBeNull attribute.

### DIFF
--- a/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
@@ -1,4 +1,4 @@
-// Licensed under MIT No Attribution, see LICENSE file at the root.
+﻿// Licensed under MIT No Attribution, see LICENSE file at the root.
 // Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
 
 using System;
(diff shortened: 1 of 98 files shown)


CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -499,13 +499,13 @@ public static bool TryParse([CanBeNull] string str, [CanBeNull] IFormatProvider
         ///     Parse a unit string.
         /// </summary>
         /// <param name=""str"">String to parse. Typically in the form: {{number}} {{unit}}</param>
+        /// <param name=""provider"">Format to use when parsing number and unit. Defaults to <see cref=""CultureInfo.CurrentUICulture"" /> if null.</param>
```
tmilnthorp: Just keeping params grouped together.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -499,13 +499,13 @@ public static bool TryParse([CanBeNull] string str, [CanBeNull] IFormatProvider
...
         /// <example>
         ///     Length.ParseUnit(""m"", new CultureInfo(""en-US""));
         /// </example>
         /// <exception cref=""ArgumentNullException"">The value of 'str' cannot be null. </exception>
         /// <exception cref=""UnitsNetException"">Error parsing string.</exception>
-        /// <param name=""provider"">Format to use when parsing number and unit. Defaults to <see cref=""CultureInfo.CurrentUICulture"" /> if null.</param>
-        public static {_unitEnumName} ParseUnit(string str, IFormatProvider provider = null)
+        public static {_unitEnumName} ParseUnit(string str, [CanBeNull] IFormatProvider provider)
```
tmilnthorp: The default null parameter was useless (and surprisingly not ambiguous) given there's already a ParseUnit overload that only takes a string.

angularsen: Good catch

PR #691, opened 2019-08-14
Update abbreviations

lingyunyumo:


tmilnthorp:
While this is a good change, it is a breaking one. 

I think that's easily remedied by adding the new abbreviations to the list rather than replacing them (like you did for VolumeFlow).

lingyunyumo:
@tmilnthorp You got me. I completely neglected the PR is breaking.

Some of my thoughts for my PR:
Aadding new abbreviation to VolumeFlow is on purpose while the LPS, LPM, LPH are all OK as I see.
But other three commits are really supposed to be some corrections:
- "Abbreviations": [ "PaS" ] --> "Abbreviations": [ "Pa·s" ]: a lowercase 's' usually for second.  '·' for multiplication
- "Abbreviations": [ "g/S" ] --> "Abbreviations": [ "g/s" ]: a lowercase 's' usually for second
- "Abbreviations": [ "Btu/hr" ] --> "Abbreviations": [ "Btu/h" ]: here "hr" is a exception while other units all use "h" in UnitsNet. and "h" is a usual abbreviation for hour

Anyhow, I aggree with you that adding abbrevations to the list is reasonable for avoiding breaking changes.

tmilnthorp:
If you add the abbreviation to the beginning, it should make it the new default. That would be a behavior change, but at least is would still allow parsing of the old units. @angularsen what are you thoughts on this kind of behavior change?

I believe it should be ok as these changes are more standard. It would only break applications that do something with the default abbreviation (say parsing it themselves for whatever reason). But that we can't help. Any code doing round-trip ToString and Parse within UnitsNet will still work so long as the old abbreviations are kept at the end of the list.

angularsen:
@tmilnthorp @lingyunyumo Yup, I'm fine with changing default abbreviation. I don't consider that a breaking change. Consumers should not depend on ToString() never changing in my opinion, we have unit enums they can rely on instead.

lingyunyumo:
@angularsen I re-pushed some commits to this PR. Could you have a look if this PR can be merged?

tmilnthorp:
Looks good to me, although it looks like you did not run the code generator. The build does not automatically do that and commit the generated files, so it must still be done manually

angularsen:
I just merged this anyway, will fix on master.

angularsen:
[Release UnitsNet/4.35.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.35.0)

tmilnthorp:
Sounds good. I wonder if we could set up the builds to automatically commit generated code to the source branch if any were found.

angularsen:
It should be possible by doing a local git diff on the build agent. Would be a nice addition.

PR #697, opened 2019-08-21
Adding deformation and electric ratio units

salido-m:
Pull request for adding 2 (ratio) units:

- Deformation: a dimensionless quantity used to measure deformations in continuum mechanics.
- ElectricRatio: Used to measure an electrical signal which is proportional to the applied excitation voltage (also called ratio-metric system).


tmilnthorp:
Does the existing ratio unit not suffice for these? Just curious as it seems to just be a named ratio, but I'm not familiar with these use cases.

salido-m:
> 
> 
> Does the existing ratio unit not suffice for these? Just curious as it seems to just be a named ratio, but I'm not familiar with these use cases.

In the facts, yes, there are ratio units, but where the current Ratio unit is not linked to physical quantities, theses new units are linked to Length (for Deformation), or ElectricPotential (for ElectricRatio).

The uses cases is to condition a measurement signal from [strain gauges](https://en.wikipedia.org/wiki/Strain_gauge).

angularsen:
@salido-m The only reason we are a bit hesitant to add these quantities is that we are simply trying to avoid the library growing too fat with little-used quantities or if we don't immediately see that it adds enough value, so please bear with us trying to understand the usecase better.

This is not the first time we are debating whether to add quantities and ratio-quantities is a recurring theme since there are so many variations of them. We don't yet have a good guideline for whether a quantity should be added or not, so it's not always easy to decide.

The main reasons I can think of for adding quantities to the library in general are:
- [x] You get `Parse()` and `ToString()` out of the box
- [x] Consistent API design and behavior as all the other quantities in Units.NET
- [x] Dynamic conversion and parsing via `Quantity` and `IQuantity`
- [x] Arithmetic with the same quantity (add, subtract, multiply and divide)
- [ ] Arithmetic with other quantities, f.ex: `Torque = Length x Mass`
- [ ] Convert between units of the same quantitiy
- [ ] It is widely used and others will benefit from it or expect it to be in the library


I've checked the boxes I feel you obtain with `ElectricRatio`, which does have multiple units even though it only has the simpler prefix units like kilo and milli. It also has unit abbreviations. `Deformation` only has one unit and it is dimensionless so I feel perhaps it adds less value over simply using `Ratio`?

Could you perhaps give a short pseudo code example that illustrates why adding these quantities would be useful to you and others, and why using `Ratio` or providing your own app-specific types would not be as good or not very useful at all? It would help us decide, thanks!

angularsen:
Following up on this, do you have any thoughts on my previous comments and questions? It would be helpful to better understand your use cases. Thanks!

Issue #699, opened 2019-08-30
Syntax Sugar (Implicit ValueTuple Conversion)


ZacharyPatten:
There is a feature scheduled for C# 9 that allows type inference on construction (Example: "Acceleration a = new(...);"). That will be some nice syntax sugar for defining measurements.

However... It is possible to simplify things even further with implicit casting operators from ValueTuple types. Example:
```csharp
Acceleration a = (7, AccelerationUnit.MeterPerSecondSquared);
```
You could even simplify things further to something like this, but it will require additional types:
```csharp
Acceleration a = (7, Meters / Seconds / Seconds);
// or
Acceleration a = (7, MetersPerSecondSquared);
```
_(In my opinion)_ These are improvements over using the static factory methods or constructors. Current required code:
```csharp
Acceleration a = new Acceleration(7, AccellerationUnit.MeterPerSecondSquared);
// or
Acceleration a = Acceleration.FromMetersPerSecondSquared(7);
```

tmilnthorp:
I like the ValueTuple conversions, really nice! Boost units does something similar with multiplication e.g. 7 * AccellerationUnit.MeterPerSecondSquared

ZacharyPatten:
The implementation of this issue should be pretty obvious (just add implicit casting operators from ValueTuple types that funnel into the constructors of the measurement types)... but just in case, I have examples in my Towel project if needed for reference.
```csharp
Force<double> force1 = (1d, Kilograms * Meters / Seconds / Seconds);
Force<double> force2 = (2d, Newtons);
```
I quite like the operators in the unit definitions personally, but **Units.Net currently doesn't allow for mix-matching like unit types in derived measurements (Example: Meters / Seconds / Hours)**, so Units.Net probably shouldn't do that unless it is re-structured to store all base types.

With its current structure... Units.Net should just go with the other example I mentioned:
```csharp
Acceleration a = (7, MetersPerSecondSquared);
```
because that is compatible with the current code base.

Either way, it will require additional types (and the help of a using statement to a static class) to get the desired syntax.

PR #701, opened 2019-09-05
Allow units to be constructed from ValueTuples

tmilnthorp:
#699

### DIFF
--- a/UnitsNet.Tests/GeneratedCode/LengthTestsBase.g.cs
+++ b/UnitsNet.Tests/GeneratedCode/LengthTestsBase.g.cs
@@ -549,5 +549,14 @@ namespace UnitsNet.Tests
         {
             Assert.False(Length.BaseDimensions is null);
         }
+
+        [Fact]
+        public void CanCreateInstanceFromTuple()
+        {
+            Length instance = (3.0, LengthUnit.Meter);
+
+            Assert.Equal<QuantityValue>(3.0, instance.Value);
+            Assert.Equal(LengthUnit.Meter, instance.Unit);
+        }
     }
 }
--- a/UnitsNet.Tests/QuantityValueTests.cs
+++ b/UnitsNet.Tests/QuantityValueTests.cs
@@ -1,0 +1,131 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+
+namespace UnitsNet.Tests
+{
+    public class QuantityValueTests
+    {
+        [Fact]
+        public void ImplicitCreationFromByte()
+        {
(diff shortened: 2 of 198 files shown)


ZacharyPatten:
In my towel project I added a "MeasurementUnitSyntax" static class that had static fields for all the supported measurement units of all supported measurement types. Then I could add a "using static Towel.Measurements.MeasurementUnitSyntax" at the top of any file. That allows you to remove the enum type in the value tuple.

ZacharyPatten:
Without the support of a "using static" it really isn't worth adding this feature because it doesn't change the syntax much.

UnitsNet.Tests/QuantityValueTests.cs
```diff
@@ -0,0 +1,131 @@
...
+            QuantityValue decimalValue1 = 3.0m;
+            QuantityValue decimalValue2 = 3.0m;
+            QuantityValue decimalValue3 = 5.0m;
+
+            Assert.True(decimalValue1.Equals(decimalValue2));
+            Assert.False(doubleValue1.Equals(decimalValue3));
+        }
+
+        [Fact]
+        public void GetHashCodeEqualsValueGetHashCode()
```
angularsen: These are nice tests!

UnitsNet.Tests/GeneratedCode/LengthTestsBase.g.cs
```diff
@@ -549,5 +549,14 @@ public void BaseDimensionsShouldNeverBeNull()
         {
             Assert.False(Length.BaseDimensions is null);
         }
+
+        [Fact]
+        public void CanCreateInstanceFromTuple()
+        {
+            Length instance = (3.0, LengthUnit.Meter);
```
angularsen: Can you add this code example to the README for how to use ValueTuple syntax.

ZacharyPatten:
@angularsen I personally would wait on pulling this in until you have a static class you can use in an alias such as "using static SYNTAX." Example:

```csharp
using static Measurements.MeasurementSyntax; // <--- alias

namespace Example
{
  public static class MeasurementSyntax // <--- static class for alias
  {
    public static LengthUnit Meter = LengthUnit.Meter;
  }

  class Program
  {
    static void Main(string[] args)
    {
      Length length = (3.0, Meter); // <--- gets "Meter" from alias
    }
  }  
}
```

Without the static class to allow for an alias, the syntax isn't any cleaner IMO (and is probably worse than the current). However, with the static using, I think that is pretty clean as "Length" does not appear multiple times. It is just a bit of syntax sugar. Is it worth it? IDK... up to you...

PR #703, opened 2019-09-12
Added fuel efficiency unit

aidbal:


angularsen:
As for coverage, I think what you are seeing is that generated code for a new quantity is lower than the code coverage of the entire codebase. So when adding a new quantity, it automatically reduces the overall coverage percentage. I've started some small steps in improving coverage by moving test cases to the code generator so that we can properly test N quantities and M units instead of a handful picked ones. Things should improve then.

angularsen:
[Release UnitsNet/4.36.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.36.0)

PR #705, opened 2019-09-19
Rename constructor argument 'numericalValue' to 'value'

rohahn:
In a project we would like to serialize quantity objects (e.g. Pressure) with [MessagePack](https://github.com/neuecc/MessagePack-CSharp) for inter-process communication. We don't want to persist the data and all processes will use the same version of Units.NET.

Aligning the name of the constructor argument with the property 'Value' allows MessagePack deserialization of types generated by _QuantityGenerator_ (with some minor changes to MessagePack). MessagePack can deserialize immutable types by passing the deserialized values of properties without a setter to an appropriate constructor.

I did not regenerate the classes in _GeneratedCode_ to keep the changes as small as possible.


angularsen:
@tmilnthorp I don't see any problems with this, do you? If it helps some reflection libraries auto-map ctor params to properties then I guess that is a benefit.

@rohahn Just a word of caution that you might be better off mapping to your own types for serialization. We are very likely to introduce breaking changes without increasing the major version, and this PR reflects that situation perfectly (renaming ctor params). See `UnitsNet.Serialization.JsonNet` project for how we added support for JSON.net serialization.

rohahn:
@angularsen Thanks for the warning. I read some of the discussions about this topic and I am aware of the potential problems. Our current workaround is quite similar and based on a custom resolver and formatter for MessagePack. But since MessagePack is optimized for performance (code generation instead of reflection), we would prefer a solution without custom code in the serialization pipeline. In our case the problem of breaking changes is mitigated, since we control all parts of the system and the serialization is not part of a the public API.

angularsen:
Very well. This PR needs to merge in latest `origin/master` to resolve the conflicts.
Also, you should run `generate-code.bat` to also include the changes in generated code in the PR.

angularsen:
Looks good to me! Nuget on the way out.

[Release UnitsNet/4.37.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.37.0)

Issue #651, opened 2019-04-15
Extending the functionality of BaseUnits & UnitSystem


lipchev:
Following the many lengthy exchanges concerning the BaseUnits (default)  and the UnitSystem in #646 here is a summary of limitations that I think should be addressed _at some point_.
First let me start by stating the general use cases (the way that I see them- feel free to add/correct me here): 

1. Domain  separation
- Stock Screen works with quantities in kg | liters (UnitSystem)
- Recipe Screen works with quantities in mg | ml
- Persistence- use a well defined UnitSystem for storing values- forget about Quantity.To/FromXX- As(..), new (..) instead
- Reporting- I want my report to use this and that measure unit (ideally UnitSystem would be serializable)
2. Presentation customization
- default unit, default abbreviation- ideally it is also possible to associate those with a range- [1e, 1e3] -> [Gram], [1e3,1e6] -> [Kilogram]
- list of common units- what do you put in the combo-box

Currently, the default (abbreviation for) BaseType for a UnitSystem is determined by taking the first matching unit, that satisfies the equality comparison with the UnitSystem.BaseUnits. The following two operations depend on the success of the previous lookup:

- Constructing quantities using the (double, UnitSystem) constructor
- Converting using the As(UnitSystem) operator 

This approach has the following limitations:

1. Possible run-time exceptions due to some combination of quantity types in a UnitSystem not having a matching base type definition- it makes sense to throw an exception for custom UnitSystems- but at least constructing/converting using SI should be safe
2. BaseUnits cannot be specified for prefixed units: [Kilo]Gram is not matched for SI as the BaseUnits apply to the Gram only
3. Ambiguous BaseUnits definitions: using the "dimensions ignored" conversion may sometimes lead to ambiguous conversions such as with the  Liter -> CubicMeter derivatives. Consider the following mapping: 

- 1 l = 1 dm3 -> Decimeter
- 1 dl = 1 ? (0.1 dm3 | 100 cm3) 
- 1 ml = 1 cm3 -> Centimeter 
4. Overlapping BaseUnits- having the find-first approach makes it easy to both override/have your default unit definition overridden by mistake
5. Mapping BaseUnits definitions for UnitSystems in run-time: there is currently no way to define missing or override existing BaseUnits for a UnitSystem

I think the last point is the key to solving most of the limitations on the list. My proposition would basically consist of storing a dictionary representing the default unit for a given quantity type inside the UnitSystem:

[...]

PR #708, opened 2019-09-26
Add benchmarks for UnitSystem methods

lipchev:
Added benchmarks for UnitSystem methods (UnitSystem.SI) in order to use as base line for comparison (I see there is a [tool](https://github.com/dotnet/performance/tree/master/src/tools/ResultsComparer) for that) with the modifications envisioned in #651

Issue #700, opened 2019-09-02
 No units were found for the given UnitSystem: Convert non-base units to SI


SGStino:
**Is your feature request related to a problem? Please describe.**
Converting L/h to SI results in "No units were found for the given UnitSystem"

**Describe the solution you'd like**
I'd love to see the values as m³/s in the SI case.
 
The use-case is a .NET tool that collects data from various sources in various units (including some weird ones like hL/min, which causes another problem for the moment) and outputs them on a single channel in SI unit by convention.
This channel is used by multiple tools written in multiple languages, hence the SI convention.

Currently we have to remove the Unit from the value and apply a custom scale factor that we have to set it manually (=little more error prone)

this means:
`output = input * 4.41631375E-05`
 instead of 
 `output = VolumeFlow.FromOilBarrelsPerHour(input).As(UnitSystem.SI)`

angularsen:
Thanks for describing the usecase. The UnitSystem concept is still quite new and we have not yet mapped all the units that should be the result of converting a quantity to SI base units. We are also debating a redesign of the system, since it can't easily be used to represent other units than SI - such as CGS, FPS, British Engineering units and so on.

There are some discussions in #651 and #630 on a possible redesign.

However, I believe your scenario can be implemented by adding `BaseUnits` to the cubic meters per second unit in `VolumeFlow.json`, such that `.As(UnitSystem.SI)` will convert to that unit. Remember to run `generate-code.bat` after changing .json files.

If you are willing to do a pull request, I'm happy to assist.

Issue #666, opened 2019-05-08
Generic Parameters For Measurement Quantities


ZacharyPatten:
### Generic Types

Right now, UnitsNet can only support "double" types. If you use generic types on the quantities you can allow for support of other types than just double.

I have working prototypes for Length, Mass, and Angle measurement types using generic paramters here:
https://github.com/ZacharyPatten/Towel

I wanted to share my patterns with UnitsNet, because they could help improve UnitsNet should they be adopted. However, there are pros/cons to my pattern, and it would be a massive overhaul of UnitsNet to incorporate generics, so it might not be feasible for the project to adopt the patterns.

### Performance Optimization

The conversions in UnitsNet require a double conversion to/from a base unit. In that Towel project, I am using multiplication tables (in the form of jagged arrays) to cache the conversions between units. The index of the multiplication tables are provided by the values on the unit enums. This allows Towel to perform unit conversions with a single operation rather than double as UnitsNet is currently doing.

angularsen:
Thanks for reporting @ZacharyPatten! As you described in the email, you are not interested in pursuing these improvements in UnitsNet yourself, just pointing out some source of reference we can look at when we address these two points.

There are a number of related issues on this already:
#623 #588 #500 #478 #371 #285 

Below is my reply in the email for future reference.


> To address your points:
> 
> 1. Yup, this is a pain point for some users. We are currently restricted by using `struct` instead of `class`, which is by design in order to get value vs reference semantics (no null checking, pass by value, more similar to .NET types like DateTime and TimeSpan). We do however have some open discussions on this topic already and although I have probably been the biggest opponent to changing the semantics without carefully evaluating the pros/cons first, I do see that there are considerable wins by moving to `class`; generics, inheritance, a lot less duplicated code and smaller binary size.
> 
> 2. Performance has not been a priority in the past, but I am always open to ideas. We don't do any caching so that sounds fairly straight forward to add. We also recently refactored the conversions from hard coded switches (in generated code) to dynamic lookup tables. It adds another method invocation in the conversion, not sure about that perf impact, but it gives us a constant lookup time and the possibility of adding direct conversion functions from say meter to centimeter - and maybe even automating conversions between prefixes (centi- to kilo-). It also opens up the option for plugging in third party conversions by the consumers. I guess we'll just have to profile it and see how well it performs, but again, not a big priority for now.
> 
[...]

PR #709, opened 2019-09-28
Redesign UnitSystem to support non-SI systems and configurable default units

lipchev:
(WIP) for #651

- [ ] 1. Redesign/Remove BaseUnits type and remove related from JSON and UnitSystem **(breaking change)**

- [x] 2. UnitSystem: Add mapping of default unit, such as QuantityType.Length => LengthUnit.Centimeter

- [ ] 3. UnitSystem: Add mapping of default abbreviation per unit enum value, such as LengthUnit.Feet => "ft" (**_optional- also a breaking change if UnitAbbreviationCache is to become part of the UnitSystem_**)

- [x] 4. Length.ctor(double, UnitSystem) constructs with unit system's default unit for Length

- [x] 5. double val = myLength.As(UnitSystem) converts to unit system's default unit for Length

- [x] 6. UnitSystem should remain immutable, so to change its configuration at runtime you create a new instance either manually or by cloning an existing system with extension methods for adding/modifying mappings like in use case 5. Then you inject this instance into your view or whatever.

- [ ] 7. Add UnitSystem.SI and UnitSystem.EnglishEngineering with mappings

- [x] 8. Add tests for the the new public methods in UnitSystem (got one in LenghtTests- but none that test for the "exceptional" cases)

### DIFF
--- a/CodeGen/Generators/NanoFrameworkGen/NuspecGenerator.cs
+++ b/CodeGen/Generators/NanoFrameworkGen/NuspecGenerator.cs
@@ -25,7 +25,7 @@ namespace CodeGen.Generators.NanoFrameworkGen
 <package xmlns=""http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd"">
   <metadata>
     <id>UnitsNet.nanoFramework.{_quantity.Name}</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET {_quantity.Name} - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/AbsorbedDoseOfIonizingRadiation/UnitsNet.NanoFramework.AbsorbedDoseOfIonizingRadiation.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/AbsorbedDoseOfIonizingRadiation/UnitsNet.NanoFramework.AbsorbedDoseOfIonizingRadiation.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.AbsorbedDoseOfIonizingRadiation</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET AbsorbedDoseOfIonizingRadiation - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/Acceleration/UnitsNet.NanoFramework.Acceleration.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/Acceleration/UnitsNet.NanoFramework.Acceleration.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.Acceleration</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET Acceleration - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/AmountOfSubstance/UnitsNet.NanoFramework.AmountOfSubstance.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/AmountOfSubstance/UnitsNet.NanoFramework.AmountOfSubstance.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.AmountOfSubstance</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET AmountOfSubstance - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/AmplitudeRatio/UnitsNet.NanoFramework.AmplitudeRatio.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/AmplitudeRatio/UnitsNet.NanoFramework.AmplitudeRatio.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.AmplitudeRatio</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET AmplitudeRatio - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/Angle/UnitsNet.NanoFramework.Angle.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/Angle/UnitsNet.NanoFramework.Angle.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.Angle</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET Angle - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/ApparentEnergy/UnitsNet.NanoFramework.ApparentEnergy.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/ApparentEnergy/UnitsNet.NanoFramework.ApparentEnergy.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.ApparentEnergy</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET ApparentEnergy - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/ApparentPower/UnitsNet.NanoFramework.ApparentPower.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/ApparentPower/UnitsNet.NanoFramework.ApparentPower.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.ApparentPower</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET ApparentPower - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/Area/UnitsNet.NanoFramework.Area.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/Area/UnitsNet.NanoFramework.Area.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.Area</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET Area - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/AreaDensity/UnitsNet.NanoFramework.AreaDensity.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/AreaDensity/UnitsNet.NanoFramework.AreaDensity.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.AreaDensity</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET AreaDensity - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/AreaMomentOfInertia/UnitsNet.NanoFramework.AreaMomentOfInertia.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/AreaMomentOfInertia/UnitsNet.NanoFramework.AreaMomentOfInertia.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.AreaMomentOfInertia</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET AreaMomentOfInertia - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/BitRate/UnitsNet.NanoFramework.BitRate.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/BitRate/UnitsNet.NanoFramework.BitRate.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.BitRate</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET BitRate - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/BrakeSpecificFuelConsumption/UnitsNet.NanoFramework.BrakeSpecificFuelConsumption.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/BrakeSpecificFuelConsumption/UnitsNet.NanoFramework.BrakeSpecificFuelConsumption.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.BrakeSpecificFuelConsumption</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET BrakeSpecificFuelConsumption - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/Capacitance/UnitsNet.NanoFramework.Capacitance.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/Capacitance/UnitsNet.NanoFramework.Capacitance.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.Capacitance</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET Capacitance - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/CoefficientOfThermalExpansion/UnitsNet.NanoFramework.CoefficientOfThermalExpansion.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/CoefficientOfThermalExpansion/UnitsNet.NanoFramework.CoefficientOfThermalExpansion.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.CoefficientOfThermalExpansion</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET CoefficientOfThermalExpansion - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/Compressibility/UnitsNet.NanoFramework.Compressibility.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/Compressibility/UnitsNet.NanoFramework.Compressibility.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.Compressibility</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET Compressibility - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/Density/UnitsNet.NanoFramework.Density.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/Density/UnitsNet.NanoFramework.Density.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.Density</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET Density - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/Duration/UnitsNet.NanoFramework.Duration.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/Duration/UnitsNet.NanoFramework.Duration.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.Duration</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET Duration - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
--- a/UnitsNet.NanoFramework/GeneratedCode/DynamicViscosity/UnitsNet.NanoFramework.DynamicViscosity.nuspec
+++ b/UnitsNet.NanoFramework/GeneratedCode/DynamicViscosity/UnitsNet.NanoFramework.DynamicViscosity.nuspec
@@ -2,7 +2,7 @@
 <package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
   <metadata>
     <id>UnitsNet.nanoFramework.DynamicViscosity</id>
-    <version>5.60.0</version>
+    <version>5.61.0</version>
     <title>Units.NET DynamicViscosity - nanoFramework</title>
     <authors>Andreas Gullberg Larsen,nanoframework</authors>
     <owners>UnitsNet</owners>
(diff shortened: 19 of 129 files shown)


lipchev:
Initially I went with a dictionary (as proposed in #651) but being stunned by the lack of improvement in performance compared with the test in #708 I dropped it in favor of an array (of QuantityInfo), indexed by the (int)QuantityType - 1, which gave me some improvement- but not nearly as much as I expected initially..

lipchev:
Method | Mean | Error | StdDev | Median | Gen 0 | Gen 1 | Gen 2 | Allocated
-- | -- | -- | -- | -- | -- | -- | -- | --
Constructor | 10.261 ns | 0.2568 ns | 0.3683 ns | 10.102 ns | - | - | - | -
Constructor_SI | 341.489 ns | 20.1771 ns | 57.5664 ns | 300.000 ns | - | - | - | -
FromMethod | 23.169 ns | 0.1087 ns | 0.0907 ns | 23.156 ns | - | - | - | -
ToProperty | 7.105 ns | 0.0894 ns | 0.0836 ns | 7.111 ns | - | - | - | -
As | 7.171 ns | 0.1547 ns | 0.1447 ns | 7.095 ns | - | - | - | -
As_SI | 336.364 ns | 19.1489 ns | 56.1604 ns | 300.000 ns | - | - | - | -
ToUnit | 16.206 ns | 0.1705 ns | 0.1424 ns | 16.205 ns | - | - | - | -
ToUnit_SI | 456.667 ns | 23.4713 ns | 65.4286 ns | 400.000 ns | - | - | - | -
ToStringTest | 2,022.562 ns | 14.2156 ns | 13.2973 ns | 2,020.125 ns | 0.1945 | - | - | 832 B
Parse | 51,561.651 ns | 990.9471 ns | 1,179.6520 ns | 51,123.584 ns | 10.1929 | - | - | 43000 B
TryParseValid | 49,676.343 ns | 808.6956 ns | 756.4544 ns | 49,425.769 ns | 10.1929 | - | - | 42976 B
TryParseInvalid | 52,532.164 ns | 612.5532 ns | 543.0125 ns | 52,483.185 ns | 10.1318 | - | - | 42576 B
QuantityFrom | 68.093 ns | 1.4228 ns | 1.5814 ns | 67.222 ns | 0.0132 | - | - | 56 B
IQuantity_As | 15.401 ns | 0.2076 ns | 0.1942 ns | 15.320 ns | 0.0057 | - | - | 24 B
IQuantity_As_SI | 380.723 ns | 14.8686 ns | 39.6873 ns | 400.000 ns | - | - | - | -
IQuantity_ToUnit | 23.538 ns | 0.2679 ns | 0.2506 ns | 23.453 ns | 0.0133 | - | - | 56 B
IQuantity_ToStringTest | 2,037.505 ns | 39.4539 ns | 61.4250 ns | 2,013.519 ns | 0.1945 | - | - | 832 B

Note that the performance of my machine is slightly lower today (I think it's either the room temperature or some process eating up my resources).

lipchev:
Still some things I'm not sure about:

- what would be the expected equality comparison for two UnitSystems (currently depends on the BaseType)
- previous tests state that for example "AmplitudeRatio is unitless. Can't have any matches" (when constructing AmplituedRatio using SI)- yet it seems there are 4 "dimensionless units" to chose from when selecting a default unit
- I struggled to validate within WithDefaultUnit(QuantityType, UnitInfo) that UnitInfo corresponds to the provided QuantityType (any ideas? lookup the list of Info's again? change the signiture?)
- I struggled to create the generic overload corresponding to WithDefaultUnit(QuantityType, UnitInfo):
`WithDefaultUnit<TUnit>(UnitInfo<TUnit>) // how do I get the QuantityType value from this`

angularsen:
We can't merge this into `master` branch until all breaking changes are ready, so I just created `release/v5` branch. Please point this PR to that branch instead, or create a new PR if it can't be changed.

Other breaking changes: #666

angularsen:
> which gave me some improvement- but not nearly as much as I expected initially..

I think this basically comes down to the fact that we have ~100 quantities today, and enumerating 100 items with a simple condition check is really really fast to begin with. At any rate, a lookup of sorts seems reasonable anyway.

angularsen:
> Still some things I'm not sure about:
> 
> * what would be the expected equality comparison for two UnitSystems (currently depends on the BaseType)

I propose using the default implementation that `class` brings, which checks if they are the exact same object reference. We could compare all the mappings and any values, but I really question the value of this.

> * previous tests state that for example "AmplitudeRatio is unitless. Can't have any matches" (when constructing AmplituedRatio using SI)- yet it seems there are 4 "dimensionless units" to chose from when selecting a default unit

I believe you are referencing this test:
https://github.com/angularsen/UnitsNet/blob/master/UnitsNet.Tests/CustomCode/LengthTests.cs#L189-L194

First, I don't understand why we are testing `AmplitudeRatio` in `LengthTests`.
Second, it's likely a limitation of the old unit system design where there are no mappings from SI base units to dimensionless units. I don't see why we shouldn't preserve this behavior in the new design, so this test can be removed I think.

> * I struggled to validate within WithDefaultUnit(QuantityType, UnitInfo) that UnitInfo corresponds to the provided QuantityType (any ideas? lookup the list of Info's again? change the signiture?)

I'll have to look closer on this one.

> * I struggled to create the generic overload corresponding to WithDefaultUnit(QuantityType, UnitInfo):
>   `WithDefaultUnit<TUnit>(UnitInfo<TUnit>) // how do I get the QuantityType value from this`

I'll have to look closer on this one.

UnitsNet.Tests/UnitSystemTests.cs
```diff
@@ -149,5 +150,55 @@ public void SIUnitSystemHasCorrectBaseUnits()
...
+        {
+            // TODO do we expect to preserve this behavior?
+            // AmplitudeRatio might be unitless- but there are (more than one) ways to express ratios. 
+            Assert.Null(UnitSystem.SI.GetDefaultUnitInfo(AmplitudeRatio.QuantityType));
+        }
+
+        [Fact]
+        public void WithDefaultUnitThrowsIfQuantityTypeIsUndefined()
+        {
+            Assert.Throws<ArgumentException>(() => UnitSystem.SI.WithDefaultUnit(QuantityType.Undefined, null));
```
angularsen: Pass in a proper `UnitInfo` here, so that the test doesn't accidentally throw ArgumentException due to a `null` value passed in. Test one thing at a time.

UnitsNet.Tests/UnitSystemTests.cs
```diff
@@ -149,5 +150,55 @@ public void SIUnitSystemHasCorrectBaseUnits()
...
+        }
+
+        [Fact]
+        public void WithDefaultUnitUsesOldBaseUnitsIfNotSpecified()
```
angularsen: Please use naming convention of `WhatIsTested_WhatAreTheInputsOrConditions_ExpectedOutcome`.
In this case, `WithDefaultUnit_GivenNullForBaseUnits_ReturnsUnitSystemWithOldBaseUnits` or something like that. 

I don't fully understand the intent of this test case, however. Aren't we removing BaseUnits concept altogether?

UnitsNet.Tests/UnitSystemTests.cs
```diff
@@ -149,5 +150,55 @@ public void SIUnitSystemHasCorrectBaseUnits()
...
+        [InlineData(LengthUnit.Meter, MassUnit.Kilogram, DurationUnit.Undefined, ElectricCurrentUnit.Ampere, TemperatureUnit.Kelvin, AmountOfSubstanceUnit.Mole, LuminousIntensityUnit.Candela)]
+        [InlineData(LengthUnit.Meter, MassUnit.Kilogram, DurationUnit.Second, ElectricCurrentUnit.Undefined, TemperatureUnit.Kelvin, AmountOfSubstanceUnit.Mole, LuminousIntensityUnit.Candela)]
+        [InlineData(LengthUnit.Meter, MassUnit.Kilogram, DurationUnit.Second, ElectricCurrentUnit.Ampere, TemperatureUnit.Undefined, AmountOfSubstanceUnit.Mole, LuminousIntensityUnit.Candela)]
+        [InlineData(LengthUnit.Meter, MassUnit.Kilogram, DurationUnit.Second, ElectricCurrentUnit.Ampere, TemperatureUnit.Kelvin, AmountOfSubstanceUnit.Undefined, LuminousIntensityUnit.Candela)]
+        [InlineData(LengthUnit.Meter, MassUnit.Kilogram, DurationUnit.Second, ElectricCurrentUnit.Ampere, TemperatureUnit.Kelvin, AmountOfSubstanceUnit.Mole, LuminousIntensityUnit.Undefined)]
+        public void WithDefaultUnitThrowsIfSpecifiedBaseUnitsNotFullyDefined(LengthUnit length, MassUnit mass, DurationUnit time, ElectricCurrentUnit current,
```
angularsen: Same here, aren't we removing `BaseUnits`?

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -89,16 +89,16 @@ public Acceleration(double value, AccelerationUnit unit)
...
-            var unitInfos = Info.GetUnitInfosFor(unitSystem.BaseUnits);
-            var firstUnitInfo = unitInfos.FirstOrDefault();
+            if(unitSystem == null) throw new ArgumentNullException(nameof(unitSystem)); 
 
             _value = Guard.EnsureValidNumber(value, nameof(value));
-            _unit = firstUnitInfo?.Value ?? throw new ArgumentException("No units were found for the given UnitSystem.", nameof(unitSystem));
+
+            var defaultUnitInfo = unitSystem.GetDefaultUnitInfo(QuantityType) as UnitInfo<AccelerationUnit>;
+
+            _unit = defaultUnitInfo?.Value ?? throw new ArgumentException("No default unit was defined for the given UnitSystem.", nameof(unitSystem));
```
angularsen: This looks good

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -745,13 +745,12 @@ public double As(UnitSystem unitSystem)
...
+            var defaultUnitInfo = unitSystem.GetDefaultUnitInfo(QuantityType) as UnitInfo<AccelerationUnit>;
 
-            var firstUnitInfo = unitInfos.FirstOrDefault();
-            if(firstUnitInfo == null)
-                throw new ArgumentException("No units were found for the given UnitSystem.", nameof(unitSystem));
+            if(defaultUnitInfo == null)
+                throw new ArgumentException("No default unit was found for the given UnitSystem.", nameof(unitSystem));
 
-            return As(firstUnitInfo.Value);
+            return As(defaultUnitInfo.Value);
```
angularsen: This looks good

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -788,13 +787,12 @@ public Acceleration ToUnit(UnitSystem unitSystem)
...
+            var defaultUnitInfo = unitSystem.GetDefaultUnitInfo(QuantityType) as UnitInfo<AccelerationUnit>;
 
-            var firstUnitInfo = unitInfos.FirstOrDefault();
-            if(firstUnitInfo == null)
-                throw new ArgumentException("No units were found for the given UnitSystem.", nameof(unitSystem));
+            if(defaultUnitInfo == null)
+                throw new ArgumentException("No default unit was found for the given UnitSystem.", nameof(unitSystem));
 
-            return ToUnit(firstUnitInfo.Value);
+            return ToUnit(defaultUnitInfo.Value);
```
angularsen: This looks good

UnitsNet/UnitSystem.cs
```diff
@@ -26,6 +31,27 @@ public UnitSystem(BaseUnits baseUnits)
                 throw new ArgumentException("A unit system must have all base units defined.", nameof(baseUnits));
 
             BaseUnits = baseUnits;
+
+            // default implementation (does not fix the prefixed entities matching issue)
+            _defaultUnits = new Lazy<UnitInfo[]>(() => Quantity.Infos.Select(i => i.GetUnitInfosFor(baseUnits).FirstOrDefault()).ToArray());
```
angularsen: Same here, we are later removing BaseUnits, right?

UnitsNet/UnitSystem.cs
```diff
@@ -26,6 +31,27 @@ public UnitSystem(BaseUnits baseUnits)
...
+        }
+
+        /// <summary>
+        /// Creates an instance of a unit system with the specified base units and default unit associations.
+        /// </summary>
+        /// <param name="baseUnits">The base units for the unit system.</param>
+        /// <param name="defaultUnits">The default units associated with each quantity type</param>
+        private UnitSystem(BaseUnits baseUnits, UnitInfo[] defaultUnits)
```
angularsen: Later, we remove `baseUnits` and only pass in `defaultUnits`, right?

UnitsNet/UnitSystem.cs
```diff
@@ -88,5 +114,55 @@ public override int GetHashCode()
...
+        /// <summary>
+        ///     Gets the default unit information for the given quantity type associated with the current unit system.
+        ///     For example: the default unit of length in SI is the 'meter' abbreviated with the letter 'm'.
+        ///     It is possible to customize the associations by creating derived UnitSystems (immutable) using the method
+        ///     <see cref="WithDefaultUnit" />.
+        /// </summary>
+        /// <param name="quantityType">The quantity type of interest.</param>
+        /// <returns>
+        ///     The default UnitInfo for the given quantity type, if such an association exists,
+        ///     and <see langword="null" /> otherwise.
```
angularsen: .NET docs often write this as `The default [...] -or- null if no association exists.

UnitsNet/UnitSystem.cs
```diff
@@ -88,5 +114,55 @@ public override int GetHashCode()
...
+        /// </returns>
+        /// <exception cref="ArgumentException">
+        ///     Quantity type can not be undefined.
+        /// </exception>
+        public UnitInfo GetDefaultUnitInfo(QuantityType quantityType)
+        {
+            if (quantityType == QuantityType.Undefined)
+                throw new ArgumentException("Quantity type can not be undefined.", nameof(quantityType));
+            return _defaultUnits.Value[(int)quantityType - 1];
```
angularsen: A comment should explain the `- 1` here, that it's due to not including the `QuantityType.Undefined` value.

UnitsNet/UnitSystem.cs
```diff
@@ -88,5 +114,55 @@ public override int GetHashCode()
...
+        }
+
+        /// <summary>
+        ///     Create a derived unit system by specifying a default unit for a given quantity type.
+        ///     It is possible to configure multiple associations by chaining calls to this method:
+        ///     SI.WithDefaultUnit(..).WithDefaultUnit(..)...
+        /// </summary>
+        /// <param name="quantityType">The quantity type of interest.</param>
```
angularsen: You write excellent xmldocs!

UnitsNet/UnitSystem.cs
```diff
@@ -88,5 +114,55 @@ public override int GetHashCode()
...
+        /// <param name="defaultUnitInfo">The default UnitInfo to associate with the given quantity type.</param>
+        /// <param name="baseUnits">Optionally provide a new definition for the base units of the new unit system.</param>
+        /// <returns>
+        ///     A new UnitSystem that defines <paramref name="defaultUnitInfo" /> as the default unit for
+        ///     <paramref name="quantityType" />
+        /// </returns>
+        /// <exception cref="ArgumentException">
+        ///     Quantity type can not be undefined.
+        /// </exception>
+        public UnitSystem WithDefaultUnit(QuantityType quantityType, UnitInfo defaultUnitInfo, BaseUnits baseUnits = null)
```
angularsen: BaseUnits is going away here too, right?

UnitsNet/UnitSystem.cs
```diff
@@ -88,5 +114,55 @@ public override int GetHashCode()
...
+        {
+            if (quantityType == QuantityType.Undefined)
+                throw new ArgumentException("Quantity type can not be undefined.", nameof(quantityType));
+
+            // TODO any way to check if UnitInfo is of QuantityType?
+            var newBaseUnits = baseUnits ?? BaseUnits;
+
+            var newDefaultUnits = _defaultUnits.Value.ToArray();
+            newDefaultUnits[(int)quantityType - 1] = defaultUnitInfo;
```
angularsen: Add comment explaining `- 1`.

UnitsNet/UnitSystem.cs
```diff
@@ -88,5 +114,55 @@ public override int GetHashCode()
...
+        /// </returns>
+        /// <exception cref="ArgumentException">
+        ///     Quantity type can not be undefined.
+        /// </exception>
+        public UnitSystem WithDefaultUnit(QuantityType quantityType, UnitInfo defaultUnitInfo, BaseUnits baseUnits = null)
+        {
+            if (quantityType == QuantityType.Undefined)
+                throw new ArgumentException("Quantity type can not be undefined.", nameof(quantityType));
+
+            // TODO any way to check if UnitInfo is of QuantityType?
```
angularsen: The two ways I can think of:
1. Add `QuantityType` field to `UnitInfo` so we can read it, but it feels unnecessary
2. Check that unit info exists in `Quantity.GetInfo(QuantityType.Length).UnitInfos`.

lipchev: used option 2

UnitsNet.Tests/UnitSystemTests.cs
```diff
@@ -149,5 +150,55 @@ public void SIUnitSystemHasCorrectBaseUnits()
...
+        {
+            Assert.Throws<ArgumentException>(() => UnitSystem.SI.GetDefaultUnitInfo(QuantityType.Undefined));
+        }
+
+        [Fact]
+        public void GetDefaultUnitInfoReturnsNullForQuantitiesWithNoDefaultUnits()
+        {
+            // TODO do we expect to preserve this behavior?
+            // AmplitudeRatio might be unitless- but there are (more than one) ways to express ratios. 
+            Assert.Null(UnitSystem.SI.GetDefaultUnitInfo(AmplitudeRatio.QuantityType));
```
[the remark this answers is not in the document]
lipchev: Right- so, in theory, the only way to get a null here would be if one creates a custom unit system, explicitly setting the association to null (as presumable the original unit system would have all of its associations defined).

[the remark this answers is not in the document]
angularsen: Sounds about right. We should also create some unit tests that breaks as soon as a new quantity is added and if one forgets to add a mapping of default unit for included unit systems such as `SI` and `EnglishEngineering`. I don't think we need to add a mapping for default abbreviation, since I think it's natural to default to the first abbreviation as we already do and simply use that as a means to allow overriding the default.

lipchev:
Regarding all comments about BaseUnits going away- you are right- they should be going away in v5.
 
I just thought I might make this a two step PR (as to not make it too difficult to review)- where the first part would be the 'non-breaking' change (that allows for issue like the one described in #700 to be patched in the client code). 

However, now that I think of it- I'm not sure how good a fix is to pollute one's code with something like SI_Ex (having correctly mapped all necessary defaults by hand) just to have it fixed in the next major release (and going about changing back to SI once more)...

angularsen:
> Regarding all comments about BaseUnits going away- you are right- they should be going away in v5.
> 
> I just thought I might make this a two step PR (as to not make it too difficult to review)- where the first part would be the 'non-breaking' change (that allows for issue like the one described in #700 to be patched in the client code).
> 
> However, now that I think of it- I'm not sure how good a fix is to pollute one's code with something like SI_Ex (having correctly mapped all necessary defaults by hand) just to have it fixed in the next major release (and going about changing back to SI once more)...

Got it, good thinking, but I suspect the PR might be easier to review if we do a clean break to begin with instead of having both old and new designs in our head at the same time. It's up to you, just do whatever you feel makes the most sense.

lipchev:
Here is an example of the modified unit definition schema (haven't removed the BaseUnits from the CubicHectometer & ImperialPint as to not trigger the code-regeneration)- tell me what you think.

The two associations would be used to construct the corresponding list of UnitInfo's and default UnitInfo for a given quantity and unit system. 

Modifying the whole list of UnitDefinitions in such a way would be quite the task- but as long as there is some review process- I think we can manage. 

As we do that- what I think would be really cool is to associate units & quantities with their actual ontological references: like for instance our [MassFraction](https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/MassFraction.json) is defined in UO as [MassPercentage](http://purl.obolibrary.org/obo/UO_0000163) and again as [MassFraction in OM](http://www.ontology-of-units-of-measure.org/resource/om-2/MassFraction).

We could thus (maybe in the future) run some ontological alignment validations on our unit definitions, as described in this awesome paper: [Comparison and Evaluation of Ontologies for Units of Measurement](http://www.semantic-web-journal.net/system/files/swj1825.pdf). 

Not to mention the nice extension methods that one could think of..

angularsen:
I like it. Explicit and much simpler to reason about than before. I originally thought about defining unit system as its own file, but keeping it per quantity is actually a much better idea since it's much more visible and just a single JSON file to edit when adding new units.

It's still something new that contributors will have to think about, but with some updated steps in the wiki I think we should manage.

Also, unit systems are really opt-in the way I see it. Meaning, if we decide to not map all units they should still continue to work for everything else but methods that take a unit system as argument.

lipchev:
Ok, so I refactored UnitSystem- I couldn't bear to remove the BaseUnits so I moved it the BaseUnitSystem extension of the UnitSystem (along with the IEquatable interface). It is thus still present for UnitSystem.SI (which is a BaseUnitSystem)- I don't yet know how we could is use- but I don't think it hurts to keep it for now. I've also left out the original BaseUnits constructor (and the corresponding mapping behavior)- marked as obsolete. Thus it seems we are mostly backward-compatible. (still I presume we will move this to v5?)

I've replaced the previous a UnitInfo[] mapping (in the UnitSystem) with a UnitSystemInfo- that includes the default unit (optional) and list of "common" (also "derived" or "named") units for a quantity in a unit system. This class itself is currently not exposed (maybe make internal)- only the UnitInfos are exposed via the GetDefaultUnitInfo(QuantityType quantityType) & GetCommonUnitsInfo(QuantityType quantityType) methods in the UnitSystem.

Next I updated the JSON types & added UnitSystem generator- that completes the partial part of the concrete UnitSystem implementations (SI, GGS, EE etc..) and finally added several more UnitSystem mappings to the UnitDefinitions of Acceleration, AmountOfSubstance, Area, Energy & Length (pretty much the same thing as with Volume- note however that CGS often maps default unit to prefixed-quantities).

Still some tests need to be added (note for instance how it's not really possible to argument check the Lazy in the constructor). Also I've commented(temporary) parts of the Ctor_WithValueAndSIUnitSystem_ReturnsQuantityWithSIUnitOrThrowsArgum, and two of my own tests for the WithDefaultUnit registration- that used to test for the BaseUnit- haven't created an overload for it in BaseUnitSystem yet)

lipchev:
Also- the benchmark showed some unexpected results- I didn't expect to see any change over the previous tests (we are still doing the same array lookup) - yet the results turned up even faster than the original As/ToUnit methods- what, am I not doing enough argument checking somewhere? Or maybe it is something worse :)

lipchev:
Method | Mean | Error | StdDev | Median | Gen 0 | Gen 1 | Gen 2 | Allocated
-- | -- | -- | -- | -- | -- | -- | -- | --
Constructor | 9.703 ns | 0.1571 ns | 0.1469 ns | 9.747 ns | - | - | - | -
Constructor_SI | 14.399 ns | 0.2823 ns | 0.3138 ns | 14.327 ns | - | - | - | -
FromMethod | 22.979 ns | 0.3424 ns | 0.3036 ns | 22.930 ns | - | - | - | -
ToProperty | 7.132 ns | 0.1730 ns | 0.2188 ns | 7.109 ns | - | - | - | -
As | 7.117 ns | 0.1795 ns | 0.2631 ns | 7.158 ns | - | - | - | -
As_SI | 8.965 ns | 0.1758 ns | 0.1727 ns | 8.984 ns | - | - | - | -
ToUnit | 16.113 ns | 0.3489 ns | 0.5634 ns | 16.020 ns | - | - | - | -
ToUnit_SI | 19.220 ns | 0.5484 ns | 1.5909 ns | 18.631 ns | - | - | - | -
ToStringTest | 2,143.655 ns | 42.0048 ns | 72.4562 ns | 2,124.925 ns | 0.1945 | - | - | 832 B
Parse | 50,824.466 ns | 1,167.4845 ns | 1,674.3713 ns | 50,253.683 ns | 10.1929 | - | - | 43000 B
TryParseValid | 50,085.794 ns | 1,109.7561 ns | 1,519.0467 ns | 49,460.004 ns | 10.1929 | - | - | 42976 B
TryParseInvalid | 51,682.055 ns | 1,388.3917 ns | 1,946.3312 ns | 51,121.497 ns | 10.1318 | - | - | 42576 B
QuantityFrom | 64.901 ns | 1.1769 ns | 1.0433 ns | 64.964 ns | 0.0132 | - | - | 56 B
IQuantity_As | 14.626 ns | 0.2431 ns | 0.2274 ns | 14.591 ns | 0.0057 | - | - | 24 B
IQuantity_As_SI | 10.176 ns | 0.1887 ns | 0.1576 ns | 10.154 ns | - | - | - | -
IQuantity_ToUnit | 23.791 ns | 0.4649 ns | 0.4348 ns | 23.867 ns | 0.0133 | - | - | 56 B
IQuantity_ToStringTest | 1,985.983 ns | 34.6024 ns | 27.0153 ns | 1,994.777 ns | 0.1945 | - | - | 832 B

angularsen:
I've kind of forgotten to follow up on this. Will try to get to it sometime soon.

lipchev:
Sorry for the long delay - I'm just so late on my main project that I've had to imposed myself a restriction on working on other solutions for the time being.. :) 
Regardless - I've had a little time to consider a couple of things with respect to this feature, the BaseUnits and the exact unit representations (#666 ). Here's what I think:

1. Given that every SI derived quantity has a default unit (in SI), it seems to me like it should be possible to define/generate the Multiplication/Division operators between any 3 quantities (TResult, TLeft, TRight) where:

  - {TResult, TLeft, TRight} are all part of SI (lets assume this for- more on this later)
  - TResult.BaseDimension == TLeft.BaseDimensions * TRight.BaseDimensions (for multiplication)
  - TResult.BaseDimension == TLeft.BaseDimensions / TRight.BaseDimensions (for division)
2. Such operators would naturally result in TResult having the DefaultUnit as per the UnitSystem (SI in the case under consideration). I guess this is _fine_ for the most part (like as a default/fail-back behavior)- but it's not ideal for the following two use cases:

- 1 mm * 1 mm -> (0.01 m * 0.01 m) - > 1E-6 m^2 (SI) _**// ideally you would get 1 mm^2)**_
- 1 ft * 1 ft -> (??? m * ??? m) -> ??? m^2 (SI) _**// ideally you would get 1 ft^2**_
- most integer multiplications (as per #666) would fail  _**// ideally the result is Area< int >**_
3. As it currently stands- users adding/updating units write operators by hand- putting whatever unit the like in the result- thus imposing a convention for writing operators that are compliant with use cases stated above- seems like a tall order. Let alone having to do so for more than one value type (int,double,decimal).

...let me grab another coffee and will try to outline the proposal..

lipchev:
Lets assume that the operation TResult = TLeft * TRight makes sense (more on that later): it should be possible to obtain the triplet <TLeftUnit, TRightUnit, TResultUnit> such that we can:
`return new TResult(Multiply(TLeft.As(TLeftUnit), TRight.As(TRightUnit), TResultUnit)) // as per #666  `
You probably have guessed by now that I'm looking at the BaseUnits (_not_ the Dimensions) for the rescue. My suggestion is that we construct those _correctly_ for each quantity/unit/prefix. Lets start unrolling this _slowly_:

1. Let's consider a basic unit SI unit with no prefixes and no specific information with regards to the BasicUnits: say Density (ignoring the prefixes for now)- we can safely construct the BaseUnits for it using the DefaultUnit for SI (which is required) giving us something like {Mass = Killogram, Length=Meter}
2. Prefixed units (SI only) with no BaseUnits information - here we have to consider the problem outlined a few times already (see  #651 point 3.) about  selecting appropriate BaseUnits for something like Deciliter : 1 dl = 1 ? (0.1 dm3 | 100 cm3). In this case, I don't see any problem in simply selecting the bigger one- e.g. 1 dl = 100 cm3. The correction factor (100 instead of the implicit value of 1) should either be included in the BaseUnits definition (or somehow accounting for it by the means of the Quantity.Dimensions)
3. Another option would be to restructure the BaseUnits- such that it can represent a map of {Quantity, Unit} with the entries representing any of the Quantity/Units in the system (not only the ones in SI). Given that the list contains the prefixed units as well- it shouldn't be a problem to create a non-SI base unit for Density - e.g. {Mass = Gram, Volume= Liter}
4. As it stands (in the current version of this feature)- we have the a list of associated UnitInfos for each quantity inside each UnitSystem - it shouldn't be a problem to allow for different UnitInfo instances per different UnitSystem (containing the customized BaseUnits like for the Density given before).
5. However, in order to benefit from (4) we have to preserve the UnitSystem that was used for the construction of a given quantity...

lipchev:
Stale - but not abandoned..

angularsen:
We can add `pinned` label to prevent bot from stale-closing it.

lipchev:
This is almost done- still need to complete mapping in the JSON files, but as it stands I can only see one client call that is _breaking_ : `new UnitSystem(baseUnits).BaseUnits` - the constructor works as before(marked as obsolete)- it is only the BaseUnits property that was moved to `BaseUnitSystem`- so `UnitSystem.SI.BaseUnits` is still valid. 

I've kept the equality contract for BaseUnitSystem as before- only comparing the default SI base units, where as for all other unit systems (not derived from BaseUnitSystem) we compare _all unit associations_. 

I've added tests akin to the ones you've suggested in #844. 
There are a few more tests that can be added to UnitSystemTests (I don't think that those that I've marked as Obsolete contribute to the coverage- while the constructor they relied on is indeed obsolete- the tested behavior is still valid).

angularsen:
Awesome, I'll try to take a closer look this weekend.

angularsen:
@lipchev I have totally neglected this one, but stumbled over it today. Would you like me to review this still?

angularsen:
Also, related #864.

Here we deprecate QuantityType in favor of strings, which is more extensible. It seems like something this PR should take into account as well.

inflectionpoint:
I would suggest the naming the alternative Unit System of "EnglishEngineering" units be defined as "USCustomary" as I believe that is the proper name. I believe the "English" or "Imperial" is historical naming and not used anymore.

angularsen:
Thank you @inflectionpoint , I didn't know that!

lipchev:
@angularsen I've merged everything from upstream, removing the SupportsSIUnitSystem tests in favor of the auto-generated equivalents. 
There is one small thing that's been bugging me for a while- that is the Undefined unit. As far as I can tell this is only used for the SI BaseUnit definitions (where I think it could easily be replaced it with a nullable property). Do you know of any other reason for having it? Wouldn't it be better if we made BaseUnit the zero-index (default) unit for all quantities? At the very least- this would allow us to remove the nullable _unit field (well maybe not _remove_ it, but at least make it non-nullable :)).

angularsen:
@lipchev I think the reasoning for Undefined was to catch bugs where someone didn't specify a value, such as when deserializing JSON.
I'm fine with making it nullable instead since we are already making breaking changes, and I don't expect anyone to use Undefined for anything meaningful that can't be replaced with null.

lipchev:
I don't know where to put this (maybe a discussion?) but [here](https://www.youtube.com/watch?v=bI-FS7aZJpY)'s a great explanation of the linear algebra behind the BaseUnits of different UnitSystems.

PR #710, opened 2019-10-09
AreaMomentOfInertia operator / overload

lwestfall:

PR #711, opened 2019-10-09
Add company "Harrington Hoists, Inc." to readme

lwestfall:


angularsen:
Awesome! The change looks good, but I see you got the extra commits here too so you need to fix that first. See comment in #710 .

If you have local commits on your fork that you wish to keep, then you may want to create a separate branch for those instead of working on your fork's `master`, to avoid confusing with the "official" `master` branch.

lwestfall:
Thanks @angularsen , I usually work in a single developer environment so I'm not too keen on some of the git / github etiquette.

I think I fixed this PR and #710. If there's anything else I need to do, please let me know.

angularsen:
Perfect

PR #712, opened 2019-10-17
New unit: TorquePerLength and fix for array serialization & deserialization

dschuermans:
Added a new unit: TorquePerLength
This unit is used in the structural analysis field when expressing the rotational stiffness resistance around an axis

Also added support for serialization & deserialization of an array of units in JSON


dschuermans:
@angularsen ping
If you don't want to merge this, is there any way i can create a pre-release package and host it on our own nuget?

dschuermans:
Okay, I added the TorquePerLength on a copy from old/v3, bumped minor & suffix (3.112.0-2) and uploaded the nuget package to our nuget feed.
So i'm all set for now 👍

angularsen:
I'm unavailable all this week but glad you found a way.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
```diff
@@ -41,9 +42,36 @@ public class UnitsNetJsonConverter : JsonConverter
 
             object obj = TryDeserializeIComparable(reader, serializer);
             // A null System.Nullable value or a comparable type was deserialized so return this
-            if (!(obj is ValueUnit vu))
+            if (!(obj is ValueUnit) && !(obj is Array))
```
angularsen: Minor, but this could benefit from being rewritten to `if (obj is ValueUnit valueUnit) {} else if (obj is Array array) { }` instead.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
```diff
@@ -41,9 +42,36 @@ public class UnitsNetJsonConverter : JsonConverter
...
                 return obj;
 
+            if (obj is Array)
+            {
+                object[] values = (object[]) obj;
+
+                List<object> results = new List<object>();
+                
+                foreach (var value in values)
```
angularsen: This foreach can be changed to LINQ.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
```diff
@@ -41,9 +42,36 @@ public class UnitsNetJsonConverter : JsonConverter
...
+                {
+                    object result = ParseValueUnit(value as ValueUnit);
+                    results.Add(result);
+                }
+                
+                Type unitType = objectType.GetElementType();
+                Array typedArray = Array.CreateInstance(unitType, results.Count);
```
angularsen: Is this all actually necessary? I think if you replace the foreach and this below with a LINQ like this, I believe it should work: 
```c#
var results = values.Select(value => ParseValueUnit(value as ValueUnit)).ToArray();
```

UnitsNet.Tests/CustomCode/TorquePerLengthTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        protected override double MegapoundForceFeetPerFootInOneNewtonMeterPerMeter => 0.224808943099E-6;
+        protected override double MegapoundForceInchesPerFootInOneNewtonMeterPerMeter => 2.69770722235E-6;
+        protected override double NewtonCentimetersPerMeterInOneNewtonMeterPerMeter => 1E2;
+        protected override double NewtonMetersPerMeterInOneNewtonMeterPerMeter => 1;
+        protected override double NewtonMillimetersPerMeterInOneNewtonMeterPerMeter => 1E3;
+        protected override double PoundForceFeetPerFootInOneNewtonMeterPerMeter => 0.224808943099;
+        protected override double PoundForceInchesPerFootInOneNewtonMeterPerMeter => 2.69770722235;
+        protected override double TonneForceCentimetersPerMeterInOneNewtonMeterPerMeter => 1.01971621298e-2;
+        protected override double TonneForceMetersPerMeterInOneNewtonMeterPerMeter => 1.01971621298e-4;
+        protected override double TonneForceMillimetersPerMeterInOneNewtonMeterPerMeter => 1.01971621298e-1;
```
angularsen: Verified these.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
dschuermans: True, i copied it from my initial attempt and it was laid out like this for easier debugging while developing.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
dschuermans: I think you'll run into runtime exceptions, saying it can't convert object to array or whatever...
This is how I've always converted an "object" into an type specific array (eg string[] or double[] or whatever)

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
dschuermans: Now that I think about it, "obj is Array" might be cutting it a bit short too?
What if it's an array of strings? or any other type?
Not entirely sure if we can even end up in that scenario though...

dschuermans:
> Looks good. Good job fixing the compatibility tests 👍
> For next time, it would be better to create two separate pull requests for adding a quantity and fixing JSON serialization. I'll look at addressing the syntax suggestions after merging this PR.

Cheers.

Yea, having to do the merge from v3 to v4 it was easier to also incorporate my fixes for the JSON stuff...
Sorry about that one!

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
angularsen: You were absolutely right, it wasn't quite as simple as I first imagined. I did make some simplifications in #715 if you want to take a look.

UnitsNet.Serialization.JsonNet/UnitsNetJsonConverter.cs
angularsen: `is Array` should cover `string[]` I believe. It's the base type all arrays implement if I understand it right.

dschuermans: Yea, but shouldn't we only be expecting an array of Units there?
When it's an array of e.g. strings, we should return that array as is and not try to parse them as units?

As per the comment there: '// A null System.Nullable value or a comparable type was deserialized so return this'
I'd imagine it applies to a "array of system.nullable values or an array of comparable types"

Just thinking out loud here :)

angularsen: Good questions, I don't really recall how this all works anymore and I didn't write the implementation. I would have to read some more unit tests and dig a bit more into how this is actually used.

For instance, does this JSON deserialize properly or does it trip up on the other stuff?
```json
{
"somestring": "foo",
"somenumber": 5,
"some_quantity": { "Value": 5, "Unit": "LengthUnit.Meter" },
"some_quantity_array": [ { "Value": 5, "Unit": "LengthUnit.Meter" } ],
"some_other_array": [ 5, "abc", ["def" ] ]
}
```

angularsen:
I seem to have forgot to release the nuget. Here it is.

[Release UnitsNet/4.40.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.40.0)

Issue #422, opened 2018-03-22
Proposal : Convertion b/w Pressure Measurement References


gratestas:
Hi,
I have thought whether conversion feature of one pressure measurement reference to another (*absolute, gauge, vacuum, barometric*) will fit the UnitsNet's standards and enrich it.

Regarding the matter, it is widely used feature in fluid dynamics. So, in case it is valued as worth to think about, I'd like to discuss the situation in details.

gojanpaolo:
I believe this will be a good addition to the library. What will be your specific use cases for this feature? 

I'm thinking this will be a new quantity class (e.g. `ReferencePressure` or something like that) since the calculations to convert between the different pressure references will not really fit inside the `Pressure` class. We can then edit the [CustomCode](https://github.com/angularsen/UnitsNet/tree/master/UnitsNet/CustomCode) of both the classes to convert to and from each other if needed. Or extension methods?

angularsen:
@gratestas Waiting for your input to continue this discussino.

gratestas:
Only for the sake of avoiding misunderstandings, I go into the details:

 Pressure is a state function, for its measurement depends on environmental factors such as ambient pressure, elevation above sea level and local weather conditions. For two persons located in the different environment to speak of the same pressure, one must relate it to a reference. There are three basis reference: *absolute, gauge, vacuum*. 
  - **Absolute** is zero-referenced to the total vacuum. 
  - **Gauge** refers to a level of the local atmospheric pressure. 
  - **Vacuum** is the negative of the gauge.

Therefore, to obtain consistent and qualitative data, the reference measurement is crucial.

Here it is **why**. Tanks, pumps, compressors and turbines are sensitive to pressure changes and the choice upon the relative measurement reference.

**Tanks/containers** contain pressurised fluids **surface** of which is referenced to atmospheric pressure. 
- The implication is that it is much more convenient and sufficient to work with maximum two-digits gauges. The _gauge_ will start reading from **0 atm**. Reservoirs are just the same except that it is open to the atmosphere and its surface will always be 0 atm relative to gauge. 
- Another aspect to consider is the fact that during the year, the atmospheric pressure of the specific location changes enough to damage a working system. 

    - So if it is desired to isolate internal process of a system from the external pressure disturbances, absolute reference is the preference.
    - However, if a certain pressure level needs to be instanteneously maintained relative to atmospheric pressure, gauge reference is better choice.

[...]

Issue #724, opened 2019-12-04
Standard and normal flow rates, SCFH and Nm^3/h


sequc82:
First-time contributor here!

It looks like this feature idea has been addressed to some degree already in #212 and #713.

It is already clear that standard and normal flow rates are missing in UnitsNet. Units such as SCCM (standard cubic centimeters/min), SCFM (standard cubic feet/min), SCFH (standard cubic feet/hour), Nm<sup>3</sup>/h (normal cubic meters/hour), etc. are best described as the type MassFlowUnit rather than VolumeFlowUnit. This is due to the gas flow being constrained to Normal or Standard pressure and temperature conditions as previously mentioned in #212. As mentioned in #713, these units are typically used with mass flow meters and controllers.

Conversions from standard and normal flow rates to more conventional mass flow rates are then just constant conversions like any other **with the exception** that the specific gravity of the gas must also be accounted for.

For example, lb/h gas to SCFH gas: 
SCFH gas = lb/h*13.1/S.G.
S.G.: Gas Specific Gravity (unitless)

Converting from mass flow in SCFH to volume flow in ACFH (actual cubic feet per hour) of course is more complex as has been discovered in #212 and #713.

How can I help?

Some references:
[Fisher Control Valve Handbook Chapter 15: Conversions and Equivalents](https://www.emerson.com/documents/automation/control-valve-handbook-en-3661206.pdf)
[Wikipedia: SCFH](https://en.wikipedia.org/wiki/Standard_cubic_feet_per_minute)
[The Engineering Toolbox: Gas Specific Gravity](https://www.engineeringtoolbox.com/specific-gravities-gases-d_334.html)

angularsen:
Hi there, thanks for doing all the research on the existing issues and creating a summary here.

I don't know this domain at all and I currently don't what exactly is lacking in Units.NET or what needs to be added or changed. My time is also very limited these days with family and other projects, so if you could create a short bullet list of what concrete changes you'd like to see in Units.NET, or maybe you need some help from me to decide how to fit a new concept into Units.NET, then it's a lot easier for me to comment on it and help move this forward. 

Should we add some new quantities? New units? Should some quantities be renamed/repurposed because they are wrong/ambiguous? Should we add some wrapper types that help with conversions by adding extra information that our current quantities don't provide?

Things like that. The more specific/concrete the better. Pseudocode examples of intended usage is also very nice.

[...]

PR #726, opened 2019-12-06
Conversion between pressure measurement references

sequc82:
Based off the proposal in #422, and building upon the work of @gratestas in Units by @YektaMirkan.

Adds a wrapper class called `ReferencePressure` and an enum `PressureReference` to provide a reference for pressure measurements.

```C#
public enum PressureReference
    {
        Undefined = 0,
        Absolute,
        Gauge,
        Vacuum,
    }
```

Utilizes two constructors for implementation:
```C#
ReferencePressure refPressure = new ReferencePressure(Pressure.fromAtmospheres(3));
ReferencePressure refPressure = new ReferencePressure(Pressure.fromAtmospheres(3), PressureReference.Gauge);
```
The single-parameter constructor sets the `PressureReference` parameter to a default of `PressureReference.Absolute`

As @gratestas mentioned in #422, implementation uses the example:
```C#
double pressure = refPressure.Gauge.Atmosphere;
```

Extended the work of @gratestas with further testing, and handling of exceptions regarding pressures less than absolute zero.


CodeGen/CodeGen.csproj
```diff
@@ -3,13 +3,14 @@
...
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
-    <PackageReference Include="Serilog" Version="2.8.0" />
-    <PackageReference Include="Serilog.Sinks.Console" Version="3.1.1" />
-    <PackageReference Include="System.CommandLine.DragonFruit" Version="0.2.0-alpha.19174.3" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+    <PackageReference Include="Serilog" Version="2.9.1-dev-01154" />
+    <PackageReference Include="Serilog.Sinks.Console" Version="4.0.0-dev-00834" />
+    <PackageReference Include="System.CommandLine.DragonFruit" Version="0.3.0-alpha.19577.1" />
```
angularsen: I'm fine with running preview nugets in the codegen, but is there a particular reason you upgraded to preview versions? Was it needed for anything?

UnitsNet.Serialization.JsonNet/UnitsNet.Serialization.JsonNet.csproj
```diff
@@ -42,8 +42,8 @@
   <!-- NuGet references that work for both signed and unsigned -->
   <ItemGroup>
     <PackageReference Include="JetBrains.Annotations" Version="2019.1.3" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
```
angularsen: Nuget version upgrades could benefit from its own pull request so it is more visible in the git history and searchable in the github issue list. Would you mind moving it to a separate PR?

UnitsNet.Tests/CustomCode/PressureTests.cs
```diff
@@ -117,5 +120,102 @@ public void PressureDividedByLengthEqualsSpecificWeight()
...
             Assert.Equal(SpecificWeight.FromNewtonsPerCubicMeter(10), specificWeight);
         }
+        // Pressure Measurement References
+        [Fact]
+        public void ReferenceConversion_WithDefaultPressureReference_PressureReferenceIsAbsolute()
+        {
+            ReferencePressure refPressure = new ReferencePressure(Pressure.FromAtmospheres(3));
+            AssertEx.Equals(PressureReference.Absolute, refPressure.Reference);
+        }
+        [Fact]
```
angularsen: Minor, but if you sorted these methods alphabetically then it would be easier to read and verify that you are verifying all three properties Absolute, Gauge and Vacuum for each of the the different reference pressures.

UnitsNet.Tests/CustomCode/PressureTests.cs
```diff
@@ -117,5 +120,102 @@ public void PressureDividedByLengthEqualsSpecificWeight()
             SpecificWeight specificWeight = Pressure.FromPascals(20) / Length.FromMeters(2);
             Assert.Equal(SpecificWeight.FromNewtonsPerCubicMeter(10), specificWeight);
         }
+        // Pressure Measurement References
+        [Fact]
+        public void ReferenceConversion_WithDefaultPressureReference_PressureReferenceIsAbsolute()
```
angularsen: May I suggest the following naming scheme to be consistent with existing tests: `WhatIsTested_WhatAreTheInputs_WhatIsExpected`

Following that, I would suggest something like `Reference_WithDefaultPressureReference_IsAbsolute` since we are testing the `Reference` property and we did not pass in an explicit pressure reference in the ctor.

The below method would then be `Gauge_WithDefaultPressureReference_IsOneLessAtmosphere` or something.

UnitsNet.Tests/CustomCode/PressureTests.cs
```diff
@@ -117,5 +120,102 @@ public void PressureDividedByLengthEqualsSpecificWeight()
...
+            AssertEx.EqualTolerance(3, refPressure.Absolute.Atmospheres, AtmospheresTolerance);
+        }
+        [Fact]
+        public void ReferenceConversion_WithNegativeAbsolutePressureReferenceToAbsolute_AbsoluteCannotBeLessThanZero()
+        {
+            ReferencePressure refPressure = new ReferencePressure(Pressure.FromAtmospheres(-3), PressureReference.Absolute);
+            Assert.Throws<ArgumentOutOfRangeException>(() => refPressure.Absolute.Atmospheres);
+        }
+        [Fact]
+        public void ReferenceConversion_WithNegativeGaugePressureReferenceToAbsolute_AbsoluteCannotBeLessThanZero()
```
angularsen: I propose renaming to `Absolute_WithNegativeGaugePressureReference_ThrowsArgumentOutOfRangeException`. Similar for other exception assert tests.

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
```diff
@@ -0,0 +1,129 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using UnitsNet.Units;
+using UnitsNet.CustomCode.Units;
+
+namespace UnitsNet.CustomCode.Wrappers
+{
+    /// <summary>
+    /// From gratestas in #422
```
angularsen: `From gratestas in #422`
This does not belong in summary I think. At best it could be in `<remarks>`, but I think this is information that does not belong in source code. The pull request has this information and I think that is enough.

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// Pressure is a state function, for its measurement depends on environmental factors such as ambient pressure, elevation above sea level and local weather conditions. For two persons located in the different environment to speak of the same pressure, one must relate it to a reference. There are three basis reference: absolute, gauge, vacuum
```
angularsen: Consider changing to `located in different environments to speak of the same pressure`.

angularsen: Add xmldoc list formatting here:
```xml
<list type="bullet">
<item><description>Absolute is...</description></item>
<item>...
</list>
```

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// </summary>
+        /// <param name="pressure"></param>
+        public ReferencePressure(Pressure pressure)
+        {
+            _reference = BaseReference;
+            Pressure = pressure;
+        }
+
+        /// <summary>
+        ///     The public repersentation of the measured reference this quantity was constructed with.
```
angularsen: Typo `representation`

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
```diff
@@ -0,0 +1,129 @@
...
+    ///    Gauge refers to a level of the local atmospheric pressure.
+    ///    Vacuum is the negative of the gauge.
+    ///Therefore, to obtain consistent and qualitative data, the reference measurement is crucial.
+    /// </summary>
+    public struct ReferencePressure
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="pressure"></param>
```
angularsen: I would add a copy of the summaries for each property to the parameters here for convenience.

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        /// <summary>
+        /// List property of reference options: Gauge, Absolute, and Vacuum
+        /// </summary>
+        public static List<PressureReference> References { get; } = Enum.GetValues(typeof(PressureReference)).Cast<PressureReference>().Except(new[] { PressureReference.Undefined }).ToList();
+
+        /// <summary>
+        ///     The base reference representation of this quantity for the numeric value stored internally. All conversions go via this value.
+        /// </summary>
+        public static PressureReference BaseReference { get; } = PressureReference.Absolute;
```
angularsen: Can this be a `const` field instead? I don't see any assignments to it.

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        /// <summary>
+        ///     The public repersentation of the measured reference this quantity was constructed with.
+        /// </summary>
+        public PressureReference Reference => _reference.GetValueOrDefault(BaseReference);
+
+        /// <summary>
+        ///     The measured reference this quantity was constructed with.
+        /// </summary>
+        private readonly PressureReference? _reference;
```
angularsen: Does this need to be nullable? It seems to me it's always getting a value assigned in the ctors. I feel it is more natural to assign it `BaseReference` in the ctor that does not take a parameter, instead of doing `GetValueOrDefault()` in the getter property.

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
```diff
@@ -0,0 +1,129 @@
...
+        /// List property of reference options: Gauge, Absolute, and Vacuum
+        /// </summary>
+        public static List<PressureReference> References { get; } = Enum.GetValues(typeof(PressureReference)).Cast<PressureReference>().Except(new[] { PressureReference.Undefined }).ToList();
+
+        /// <summary>
+        ///     The base reference representation of this quantity for the numeric value stored internally. All conversions go via this value.
+        /// </summary>
+        public static PressureReference BaseReference { get; } = PressureReference.Absolute;
+
+        private Pressure Pressure { get; }
```
angularsen: This is a matter of preference, but I personally favor private readonly fields over private getter properties. It seems more typical C#.

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
```diff
@@ -0,0 +1,129 @@
...
+                default:
+                    throw new NotImplementedException($"Can not convert {Reference} to {reference}.");
+            }
+        }
+
+        private double AsBaseReference()
+        {
+            switch (Reference)
+            {
+                case PressureReference.Absolute:
```
angularsen: I would suggest `{}` braces around multi-line statements inside switch cases. Single line cases don't need it.
You can also omit `else` when the `if` throws an exception.

UnitsNet/GeneratedCode/Quantities/Pressure.g.cs
```diff
@@ -155,6 +155,11 @@ public Pressure(double value, UnitSystem unitSystem)
         /// </summary>
         public static Pressure MinValue { get; } = new Pressure(double.MinValue, BaseUnit);
 
+        /// <summary>
+        /// Represents the pressure at which Pressure is referenced (1 atm default)
+        /// </summary>
+        public static Pressure Reference { get; set; } = new Pressure(1, PressureUnit.Atmosphere);
```
angularsen: I don't think a static property should be settable. It can cause too much problems if 3rd party code start manipulating this value and expect this or that. Can it be made readonly getter property?

CodeGen/CodeGen.csproj
sequc82: No worries, just old habits to keep packages updated. Will revert back to original versions.

UnitsNet.Serialization.JsonNet/UnitsNet.Serialization.JsonNet.csproj
sequc82: No worries, just old habits to keep packages updated. Will revert back to original versions and add new PR for stable NuGet package updates,

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
sequc82: To confirm, you are suggesting a copy of the summary of pressure and reference?

angularsen:
Yes, not necessarily word by word, just describe the parameters and what
they will be used for.

>

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
sequc82: Refactored as non-nullable property

UnitsNet/GeneratedCode/Quantities/Pressure.g.cs
sequc82: Reviewing this property, it is not referenced anywhere, and was added to generated code. Considering `ReferencePressure` wraps `Pressure` to tie it to a pressure reference, I would expect implementation to access `ReferencePressure` and `Pressure` would be referenced for other uses of pressure, such as material stress, which unless I'm mistaken, does not require a reference pressure.
That said, I think this property can be removed as redundant.

sequc82:
I added a ctor overload for a third parameter: the atmospheric pressure referenced for conversion between Absolute, Gauge, and Vacuum. Otherwise, `ReferencePressure` would always assume 1 atm.

angularsen:
I'll try to get a look at this tomorrow.

sequc82:
@angularsen, based on your comment in #724, would it be best to move the code in `Pressure.Wrapper.cs` and the `PressureReference` enum into `Pressure.extra.cs`?

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
```diff
@@ -0,0 +1,194 @@
...
+                    throw new NotImplementedException($"Can not convert {Reference} to base reference.");
+            }
+        }
+
+        /// <summary>
+        ///     Represents the pressure at which _pressure is referenced (1 atm default)
+        /// </summary>
+        public static Pressure AtmosphericPressure { get; set; }
+
+        private static readonly Pressure _defaultAtmosphericPressure = new Pressure(1, PressureUnit.Atmosphere);
```
angularsen: Nitpick, but please move static fields to the top of the class.

We try to follow the common C# conventions pretty much as described here, but we're not super strict about it.
https://stackoverflow.com/a/310967/134761

Also, static fields should use `PascalCase` even if private.

UnitsNet/CustomCode/Wrappers/Pressure.Wrapper.cs
```diff
@@ -0,0 +1,194 @@
...
+    ///         </item>
+    ///         <item>
+    ///             <description>Gauge references the local atmospheric pressure.</description>
+    ///         </item>
+    ///         <item>
+    ///             <description>Vacuum is the negative of the gauge.</description>
+    ///         </item>
+    ///     </list>
+    /// </summary>
+    public struct ReferencePressure
```
angularsen: I would rename this file to `ReferencePressure.cs`. The `Pressure.xxx.cs` naming convention is typically used for partial types, when two files both hold code for a single type `Pressure`.

CodeGen/Generators/QuantityJsonFilesParser.cs
```diff
@@ -110,25 +110,17 @@ private static Localization[] GetLocalizationForPrefixUnit(IEnumerable<Localizat
...
                 // kilo ("k") + meter ("m") => kilometer ("km")
                 var prefix = prefixInfo.GetPrefixForCultureOrSiPrefix(loc.Culture);
                 unitAbbreviationsForPrefix = loc.Abbreviations.Select(unitAbbreviation => $"{prefix}{unitAbbreviation}").ToArray();
 
-                return new Localization
-                {
-                    Culture = loc.Culture,
-                    Abbreviations = unitAbbreviationsForPrefix
-                };
+                return new Localization {Culture = loc.Culture, Abbreviations = unitAbbreviationsForPrefix};
```
angularsen: You need to revert ffa1d05 with R# cleanup of entire solution :)

The intention was good, but it makes it pretty much impossible to review this PR and it should at any rate be a separate PR. Heck, Chrome browser is so heavily loaded trying to display all the file diffs that I can barely type into this text box 😆 

It's a very good practice to keep PRs small and do changes in small isolated chunks, so we can much more efficiently review them.

Besides, the R# rules are not perfectly configured and results in some changes to formatting I'm not a fan of. We can discuss further if you do a separate PR and maybe we can tweak some rules to make it behave better.

angularsen: I'll review further after the revert, because my browser is really struggling to open up diffs etc.

sequc82: Haha sorry, saw it was set up in the wiki, figured the rules were configured, and my contributions were all that needed cleanup... Will revert and correct!

UnitsNet/CustomCode/Wrappers/ReferencePressure.cs
```diff
@@ -25,6 +25,21 @@ namespace UnitsNet.CustomCode.Wrappers
...
+        /// </summary>
+        public static Pressure AtmosphericPressure { get; set; }
+
+        private static readonly Pressure DefaultAtmosphericPressure = new Pressure(1, PressureUnit.Atmosphere);
+
+        /// <summary>
+        ///     Gets a list of <see cref="PressureReference" /> options: <see cref="PressureReference.Gauge" />,
+        ///     <see cref="PressureReference.Absolute" />, and <see cref="PressureReference.Vacuum" />
+        /// </summary>
+        public static List<PressureReference> References { get; } =
```
angularsen: This should perhaps be `IReadOnlyList<PressureReference>` or similar. At the very least an array, to indicate people shouldn't try to manipulate it.

UnitsNet/CustomCode/Wrappers/ReferencePressure.cs
```diff
@@ -25,6 +25,21 @@ namespace UnitsNet.CustomCode.Wrappers
     /// </summary>
     public struct ReferencePressure
     {
+        /// <summary>
+        ///     Represents the pressure at which _pressure is referenced (1 atm default)
+        /// </summary>
+        public static Pressure AtmosphericPressure { get; set; }
```
angularsen: I don't think this is supposed to be static?

UnitsNet/CustomCode/Wrappers/ReferencePressure.cs
```diff
@@ -25,6 +25,21 @@ namespace UnitsNet.CustomCode.Wrappers
     /// </summary>
     public struct ReferencePressure
     {
+        /// <summary>
+        ///     Represents the pressure at which _pressure is referenced (1 atm default)
```
angularsen: It feels awkward that only pressure is a private field and the other info are public getter properties. I think we can promote `_pressure` to a public getter property and rename it `Pressure`. What do you think?

UnitsNet/CustomCode/Wrappers/ReferencePressure.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        /// <summary>
+        ///     Gets the <see cref="PressureReference" /> of the <see cref="ReferencePressure" />
+        /// </summary>
+        public PressureReference Reference { get; }
+
+        /// <summary>
+        ///     The base reference representation of <see cref="ReferencePressure" /> for the numeric value stored internally. All
+        ///     conversions go via
+        ///     this value.
```
angularsen: Nitpick, can remove a linebreak here.

UnitsNet/CustomCode/Wrappers/ReferencePressure.cs
```diff
@@ -0,0 +1,195 @@
...
+        /// </summary>
+        public const PressureReference BaseReference = PressureReference.Absolute;
+
+        private readonly Pressure _pressure;
+
+        /// <summary>
+        ///     Get Gauge <see cref="Pressure" />.
+        ///     It refers pressure level above Reference pressure.
```
angularsen: It refers to ?

UnitsNet/CustomCode/Wrappers/ReferencePressure.cs
sequc82: I would agree.

UnitsNet/CustomCode/Wrappers/ReferencePressure.cs
sequc82: That is a good point. Keeping this private requires processing through the reference conversion to reference pressure under the original measurement reference. Opening it up as a public getter should simplify access to the `Pressure` when simply retrieving the value under the original measurement reference.

UnitsNet/CustomCode/Wrappers/ReferencePressure.cs
sequc82: Apparently not compatible with .NET Framework 4.0. submitting a PR for an update to .NET Framework 4.8 first.

UnitsNet/CustomCode/Wrappers/ReferencePressure.cs
```diff
@@ -36,7 +36,7 @@ public struct ReferencePressure
         ///     Gets a list of <see cref="PressureReference" /> options: <see cref="PressureReference.Gauge" />,
         ///     <see cref="PressureReference.Absolute" />, and <see cref="PressureReference.Vacuum" />
         /// </summary>
-        public static IReadOnlyList<PressureReference> References { get; } =
+        public static List<PressureReference> References { get; } =
```
angularsen: Sorry I forgot about net40. May I suggest using `PressureReference[]` instead of List? At least that prevents anyone from adding/removing/clearing the list, but the contents of the array can still be manipulated.

sequc82: That's a fair consideration. Since there should be no changes to this anyway, a `List` would not really offer much benefit that I can think of.

angularsen:
Awesome, great work!

Nuget is on the way out.

[Release UnitsNet/4.42.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.42.0)

PR #727, opened 2019-12-10
Updated NuGet packages to latest stable

sequc82:
Prerelease versions skipped. All available stable updates included

### DIFF
--- a/CodeGen/CodeGen.csproj
+++ b/CodeGen/CodeGen.csproj
@@ -6,8 +6,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
-    <PackageReference Include="Serilog" Version="2.8.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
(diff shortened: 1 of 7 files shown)


angularsen:
Huh! Seems like som test coverage failed.
https://ci.appveyor.com/project/angularsen/unitsnet/builds/29444674#L328

I can take a look later if you don't figure it out.

sequc82:
That was unusual. It seems there is an SDK that the test projects are referencing that held .NET Core from updating with the rest of the NuGet packages. An update to the .csproj files and restoration of NuGet packages seems to have resolved the issue.

sequc82:
@angularsen Well it looks like the tests passed, but there seems to be an issue with codecov. Any ideas?

https://ci.appveyor.com/project/angularsen/unitsnet/builds/29447339

angularsen:
The codecov.io API seems to be a bit unstable, or perphaps the connection from the build agent. It doesn't happen very often, but sometimes it fails on this. I simply re-queued the build and it went through 😄

angularsen:
Great stuff, thanks!

PR #728, opened 2019-12-11
Update .NET Framework to 4.8

sequc82:
.NET 4.0 is not compatible with IReadOnlyList<T>. interested in updating .NET Framework to make compatible.

### DIFF
--- a/CodeGen/CodeGen.csproj
+++ b/CodeGen/CodeGen.csproj
@@ -6,8 +6,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
-    <PackageReference Include="Serilog" Version="2.8.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+    <PackageReference Include="Serilog" Version="2.9.0" />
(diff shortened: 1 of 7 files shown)


angularsen:
I hate to kill your initiative :-) But this is unfortunately a breaking change for users out there stuck on any versions below .NET 4.8. 

I'm totally open to discuss moving the lowest version up, but we would have to target v5 for a change like this. #563 

One one hand, we can write cleaner and better code with newer .NET versions and really the only target we typically need is netstandard2.0. That supports pretty much all commonly used .NET runtimes and versions, including the .NET framework down to 4.6.1. https://github.com/dotnet/standard/blob/master/docs/versions.md

On the other hand, it costs us very little to support consumers who actually are stuck on older .NET versions and all the legwork of supporting net40 is already done. I wouldn't make this change only to get types like `IReadOnlyList`, but I'm sure there are some real benefits to dropping net40. I'd have to research it a bit more.

I wish there was a way to get statistics on what targets our consumers use, but I don't know of anything like that.

If you want, you can start a discussion on #563 or at least add it to the list of things to consider. There isn't a lot of momentum for v5 yet, but the new UnitSystem was one of the things we wanted to get out on that version.

sequc82:
I appreciate the feedback! I'll look into #563 and close this for now.

Issue #410, opened 2018-02-13
Generate TypeScript or JavaScript source code


angularsen:
I think it would be awesome to reuse the code generator scripts to generate TypeScript and/or JavaScript code and get that published on NPM. I don't currently have a need for this myself and haven't looked into what already exists, but I imagine we have an opportunity to contribute something valuable here.

As a start, someone should figure out the structure of the generated code and what makes sense in those languages, and draft up a proposal in bullet form.

They key features are listed on the README front page. Static typing is obviously a win for TypeScript and other typed languages. Quantity representations (value+unit) and unit conversions seems obvious.
Quantity and unit parsing should be useful. The [WPF sample app](https://github.com/angularsen/UnitsNet/tree/master/Samples/UnitConverter.Wpf) probably best resembles how this would be used in front-end web development.

Difficulty hard as it requires drafting up the architecture for an entirely new code base. You probably also need experience with modern front-end build tools like webpack and publishing proper packages to NPM, as well as figure out how our current code base and features can best be translated to a new language. Finally you need to modify our existing code generator scripts to output entirely new code.

Hard - but fun! :-)

Related: #409

angularsen:
Sounds awesome! Will look into it when I get back from vacation.

angularsen:
This is really cool. I've only skimmed through the web pages and wiki, so if I understand it right it's basically translating your code to JS and shimming the parts of the .NET framework you are actually using with JS equivalent functions. 

I can't help think you will run into issues when you are using less common parts of the .NET framework, and the JS dependency must be pretty huge to cover non-trivial parts of the .NET framework. I can't seem to find any information about the binary size you get when using this library and what limitations there are in what parts of the .NET framework you can use. I don't see any mention about webassembly, so I'm assuming they are not using anything similar to Blazor's mono/c# in the browser.

Do you know anything about how this would all work? I don't have much time to play with this these days, but it would be fun to see how much works and what walls we would run into.

angularsen:
Solved by @haimkastner with https://www.npmjs.com/package/unitsnet-js

PR #730, opened 2019-12-28
Change unit definition JSON from UTF-8-BOM to UTF-8

haimkastner:
When I developed a similar generated library for JavaScript (called [unitsnet-js](https://www.npmjs.com/package/unitsnet-js)) from those units definition files, the JavaScript JSON parser failed because most of the schemas start with an unknown character.

The unknown character is shown only in the live web version, when I used git to clone the repo, the character has gone (Tested on Windows 10 OS). this why the fix is separated into each file one commit.

This is how it's look in the GitHub web version:
image

### DIFF
--- a/Common/UnitDefinitions/BitRate.json
+++ b/Common/UnitDefinitions/BitRate.json
@@ -1,4 +1,4 @@
-﻿{
+{
   "Name": "BitRate",
   "BaseUnit": "BitPerSecond",
   "BaseType": "decimal",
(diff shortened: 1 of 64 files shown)


angularsen:
This is probably the UTF-8 BOM characters you are seeing. I believe it causes problems on the web, where UTF-8 files are expected to _not_ include the BOM character. I'll try your change and if it doesn't break anything, I think we can merge it in. Will get back to you.

angularsen:
Regenerating code locally does not seem to do any changes except a couple of innocent whitespace changes, so this change should be safe 👍 Famous last words!

angularsen:
Confirmed files are now encoded with utf-8 and not utf-8-bom.

angularsen:
Thanks! Also, I dig that you made a javascript equivalent. We had an issue #410 on this before, but I never had the time or incentive to put the time into something like that. Very nice!

PR #733, opened 2020-01-10
add Area cubic yard per foot

gojanpaolo:


angularsen:
Huh, is this really a commonly used unit? Although the dimensions match an area, it doesn't quite sound like one. I also can't find any good references on google. Do you have any references?

gojanpaolo:
@angularsen in our case (dredging), we use it for calculating the amount/volume of soil required to fill up the beach.

I found this online which is a similar use case (i.e. filling up something) https://www.cityofsignalhill.org/DocumentCenter/View/1987/PW-Standard-Plans2Street200-242?bidId=

The usual term for the unit is "Cubic Yards per Linear Foot".

angularsen:
I found the beach filling example too. But are you sure it makes senses to add this as a unit to `Area` quantity?

I think maybe this fits better in https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/VolumePerLength.json . Then add the appropriate overloads for multiplication and division to support the calculations you need.

gojanpaolo:
Ah, I didn't know that unit exists, nice catch! Will update PR later. Thanks!

PR #736, opened 2020-01-16
New Unit added: Nominal Power

McGum:


McGum:
I'm confused: this is auto generated code, so is there somthing what I have todo for finishing this pull request?

angularsen:
Hi, no sorry, this is only a consequence of our code generator producing less code coverage than the existing code base :-) Not your fault so you can disregard this. Will review soon.

angularsen:
It looks good to me, a few comments
- Does it make sense to add any arithmetic or conversion methods while at it? See https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/CustomCode/Quantities/Power.extra.cs as an example.

McGum:
> It looks good to me, a few comments
> 
> * Does it make sense to add any arithmetic or conversion methods while at it?

I think about - but its only in combination with a lot of other new Units nessesary [in normal this Unit is used to describe the PV module, not for calculating].

So: NonimalPower/Area or Currency/NominalPower is used ... But as we dont have Currency and NominalPower/Area is used very rarely I think for now its OK

angularsen:
Ok, I'm fine with adding it then.

I tried to fix the merge conflict for you, because it is my fault having changed the folder structure in the code generator output recently, but I can't push commits to your master branch to fix it.

Two options:
1. Create a new branch `add-nominal-power` based on latest `angularsen/master`, copy the JSON file and NominalPowerTests.cs to it, regenerate the code and create a new pull request from this branch.
2. Or I can merge it for you locally, but your name would not be on the PR.

Let me know how you would like to proceed or if you need any assistance on option 1.

PR #738, opened 2020-01-19
[WIP] Test all target frameworks

angularsen:
Adding `net472` target framework to test project in the hopes of increasing test coverage. When running locally in Rider editor, it helped because before this change only the netstandard version of UnitsNet got any test coverage attributed to it.


angularsen:
Turns out it didn't impact the coverage, yet led to 2-3x the build time 😁

Issue #732, opened 2020-01-08
Rework of UnitsNetJsonConverter


dschuermans:
I've bumped into a few issue while trying to use the provided UnitsNetJsonConverter:

1. Initially, there was no support for arrays of units (Fixed in #712)
2. The UnitsNetJsonConverter does not take into account any settings defined by the user (e.g. custom contract resolver being used, additional converters, date time handling, null value handling etc... Pretty much any of the JsonSerializerSettings that can be provided the JsonConvert.Serialize() or JsonConvert.DeserializeObject() methods. (see #568)
3. The UnitsNetJsonConverter does not respect the configured naming conventions (see also #568)

I've attempted to create a more simplified and straight forward converter, since the current one also breaks the single responsiblity principle because it handles both ValueUnit types and IComparable.

In my opinion, providing support for deserializing IComparable is tricky, due the generic nature of IComparable. There are so many types that also implement IComparable that I feel it's a dangerous path to walk and it could potentionally interfere with any existing converter(s) the user might have for IComparable types.

So what I propose is the following:

* 2 `JsonConverter` implementations instead of one.
* Each converter inherits from `JsonConverter<T>` instead of `JsonConverter`
    * 1 converter that inherits from `JsonConverter<IQuantity>` - This will deal with serializing & deserializing any UnitsNet unit
    * 1 converter that inherits from `JsonConverter<IComparable>` - This will deal with **deserializing** an IComparable into a UnitsNet unit.

By inheriting from `JsonConverter<T>` the code for type checking and all of the custom Array functionality becomes obsolete, which in turn simplifies the implementation.
If the user is indeed using IComparable as types in their code base, they can now opt-in to use the `UnitsNetIComparableJsonConverter`.
But i'm sure that in 99% of the cases it will be sufficient to only register the `UnitsNetIQuantityJsonConverter`

**Mind you, the order in which the converters are registered is important: First the converter for IQuantity and then optionally the IComparable converter**

Here's my working implementation, tested against the existing UT's for the current V4 converter:

```
public abstract class UnitsNetBaseJsonConverter<T> : JsonConverter<T>
    {
        /// <summary>
        /// Reads the "Unit" and "Value" properties from a JSON string
        /// </summary>
        /// <param name="jsonToken">The JSON data to read from</param>
        /// <returns>A <see cref="ValueUnit"/></returns>
        protected ValueUnit ReadValueUnit(JToken jsonToken)
        {
            if (!jsonToken.HasValues)
            {
[...]

PR #692, opened 2019-08-16
 UnitMath extension methods (Min, Max, Sum, Average)

lipchev:
Added a set of extension methods for some of the most common Math operations, such as Min, Max, Sum and Average (with selector overloads)

PR #739, opened 2020-01-21
Extendable json serialisation

dandm1:
Extensions to the Quantity class to allow the registration of classes implementing IQuantity from outside the UnitsNet library, and to support JSON serialisation and Quantity.From with this classes. 

I am a developer on v4 of the CumulusMX weather station software.  The weather measurements on this are a combination of measures that have units and simple scalars. To simplify the code, we have decided to implement the scalars as a custom IQuantity - but were finding that this prevented us effectively serialising and deserialising the state. By registering the custom type using the new functionality in this pull request we should resolve that.

### DIFF
--- a/UnitsNet.Serialization.JsonNet.CompatibilityTests/UnitsNet.Serialization.JsonNet.CompatibilityTests.csproj
+++ b/UnitsNet.Serialization.JsonNet.CompatibilityTests/UnitsNet.Serialization.JsonNet.CompatibilityTests.csproj
@@ -25,6 +25,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\UnitsNet.Tests\UnitsNet.Tests.csproj" />
     <ProjectReference Include="..\UnitsNet\UnitsNet.csproj" />
   </ItemGroup>
 </Project>
--- a/UnitsNet.Serialization.JsonNet.Tests/CustomQuantities/HowMuch.cs
+++ b/UnitsNet.Serialization.JsonNet.Tests/CustomQuantities/HowMuch.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Security.Cryptography;
 using Newtonsoft.Json;
 
 namespace UnitsNet.Serialization.JsonNet.Tests.CustomQuantities
@@ -22,7 +21,7 @@ namespace UnitsNet.Serialization.JsonNet.Tests.CustomQuantities
             Info = new QuantityInfo<HowMuchUnit>
             (
                 QuantityType.Ratio,
-                new UnitInfo<HowMuchUnit>[]
+                new[]
                 {
                     new UnitInfo<HowMuchUnit>(HowMuchUnit.Some, BaseUnits.Undefined),
                     new UnitInfo<HowMuchUnit>(HowMuchUnit.ATon, BaseUnits.Undefined),
@@ -35,11 +34,11 @@ namespace UnitsNet.Serialization.JsonNet.Tests.CustomQuantities
 
         }
 
-        public double As(HowMuchUnit unit)
+        double IQuantity<HowMuchUnit>.As(HowMuchUnit unit)
         {
             return As(unit);
         }
-        
+
         QuantityInfo<HowMuchUnit> IQuantity<HowMuchUnit>.QuantityInfo => Info;
 
         public HowMuchUnit Unit { get; }
@@ -50,9 +49,9 @@ namespace UnitsNet.Serialization.JsonNet.Tests.CustomQuantities
             return new HowMuch((double)value,fromUnit);
         }
 
-        public static HowMuchUnit BaseUnit { get; } 
+        public static HowMuchUnit BaseUnit { get; }
         public static HowMuch Zero { get; } = new HowMuch(0, BaseUnit);
-        
+
 
         public QuantityType Type => QuantityType.Ratio;
         public static BaseDimensions BaseDimensions => BaseDimensions.Dimensionless;
--- a/UnitsNet.Serialization.JsonNet.Tests/CustomQuantities/HowMuch.cs
+++ b/UnitsNet.Serialization.JsonNet.Tests/CustomQuantities/HowMuch.cs
@@ -1,4 +1,6 @@
 ﻿using System;
+using System.Security.Cryptography;
+using Newtonsoft.Json;
 
 namespace UnitsNet.Serialization.JsonNet.Tests.CustomQuantities
 {
@@ -6,40 +8,76 @@ namespace UnitsNet.Serialization.JsonNet.Tests.CustomQuantities
     /// <summary>
     /// Example of a custom/third-party quantity implementation, for plugging in quantities and units at runtime.
     /// </summary>
-    public struct HowMuch : IQuantity
+    public struct HowMuch : IQuantity<HowMuchUnit>
     {
         public HowMuch(double value, Enum unit) : this()
         {
-            Unit = unit;
+            Unit = (HowMuchUnit)unit;
             Value = value;
         }
 
-        public Enum Unit { get; }
+        static HowMuch()
+        {
+            BaseUnit = HowMuchUnit.Some;
+            Info = new QuantityInfo<HowMuchUnit>
+            (
+                QuantityType.Ratio,
+                new UnitInfo<HowMuchUnit>[]
+                {
+                    new UnitInfo<HowMuchUnit>(HowMuchUnit.Some, BaseUnits.Undefined),
+                    new UnitInfo<HowMuchUnit>(HowMuchUnit.ATon, BaseUnits.Undefined),
+                    new UnitInfo<HowMuchUnit>(HowMuchUnit.AShitTon, BaseUnits.Undefined),
+                },
+                BaseUnit,
+                Zero,
+                BaseDimensions
+            );
+
+        }
+
+        public double As(HowMuchUnit unit)
+        {
+            return As(unit);
+        }
+        
+        QuantityInfo<HowMuchUnit> IQuantity<HowMuchUnit>.QuantityInfo => Info;
+
+        public HowMuchUnit Unit { get; }
         public double Value { get; }
 
-        #region Crud to satisfy IQuantity, but not really used for anything
+        public static HowMuch From(QuantityValue value, HowMuchUnit fromUnit)
+        {
+            return new HowMuch((double)value,fromUnit);
+        }
 
-        private static readonly HowMuch Zero = new HowMuch(0, HowMuchUnit.Some);
+        public static HowMuchUnit BaseUnit { get; } 
+        public static HowMuch Zero { get; } = new HowMuch(0, BaseUnit);
+        
 
         public QuantityType Type => QuantityType.Ratio;
-        public BaseDimensions Dimensions => BaseDimensions.Dimensionless;
+        public static BaseDimensions BaseDimensions => BaseDimensions.Dimensionless;
+        public BaseDimensions Dimensions => BaseDimensions;
+        [JsonIgnore]
+        public QuantityInfo QuantityInfo => Info;
 
-        public QuantityInfo QuantityInfo => new QuantityInfo(Type,
-            new UnitInfo[]
-            {
-                new UnitInfo<HowMuchUnit>(HowMuchUnit.Some, BaseUnits.Undefined),
-                new UnitInfo<HowMuchUnit>(HowMuchUnit.ATon, BaseUnits.Undefined),
-                new UnitInfo<HowMuchUnit>(HowMuchUnit.AShitTon, BaseUnits.Undefined),
-            },
-            HowMuchUnit.Some,
-            Zero,
-            BaseDimensions.Dimensionless);
+        public static QuantityInfo<HowMuchUnit> Info { get; }
 
         public double As(Enum unit) => Convert.ToDouble(unit);
 
+        #region Crud to satisfy IQuantity, but not really used for anything
         public double As(UnitSystem unitSystem) => throw new NotImplementedException();
+        Enum IQuantity.Unit => Unit;
 
         public IQuantity ToUnit(Enum unit) => new HowMuch(As(unit), unit);
+        IQuantity<HowMuchUnit> IQuantity<HowMuchUnit>.ToUnit(UnitSystem unitSystem)
+        {
+            throw new NotImplementedException();
+        }
+
+        public IQuantity<HowMuchUnit> ToUnit(HowMuchUnit unit)
+        {
+            throw new NotImplementedException();
+        }
 
         public IQuantity ToUnit(UnitSystem unitSystem) => throw new NotImplementedException();
 
--- a/UnitsNet.Serialization.JsonNet.Tests/CustomQuantities/HowMuch.cs
+++ b/UnitsNet.Serialization.JsonNet.Tests/CustomQuantities/HowMuch.cs
@@ -1,0 +1,53 @@
+﻿using System;
+
+namespace UnitsNet.Serialization.JsonNet.Tests.CustomQuantities
+{
+    /// <inheritdoc cref="IQuantity"/>
+    /// <summary>
+    /// Example of a custom/third-party quantity implementation, for plugging in quantities and units at runtime.
+    /// </summary>
+    public struct HowMuch : IQuantity
+    {
+        public HowMuch(double value, Enum unit) : this()
+        {
+            Unit = unit;
+            Value = value;
+        }
+
+        public Enum Unit { get; }
+        public double Value { get; }
+
+        #region Crud to satisfy IQuantity, but not really used for anything
+
+        private static readonly HowMuch Zero = new HowMuch(0, HowMuchUnit.Some);
+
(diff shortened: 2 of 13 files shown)


angularsen:
Hi and thanks for sharing your approach with us. I like the idea of better supporting third-party quantities, but some comments below on how we can maybe fit it better into the library.

#### 1. `Quantity.AddUnit()` and `Quantity.ExternalQuantities` cannot be static. 
As an example, in your tests you are adding a unit mapping in one test, and that mapping lives on to the next tests that run. It is prone to cause bugs like that.

From your test case
https://github.com/angularsen/UnitsNet/pull/739/files#diff-4593c2e3fa9c32725a300f36f3951db4R47-R53

it seems you want this in order to be able to do `IQuantity q = Quantity.From(1, myCustomUnit)`.

I propose we do the same as we did for `UnitConverter` in https://github.com/angularsen/UnitsNet/blob/25ff826de0d7c2cc58bfa573362295fd05470626/UnitsNet/UnitConverter.cs#L51-L64.

https://github.com/angularsen/UnitsNet/blob/25ff826de0d7c2cc58bfa573362295fd05470626/UnitsNet.Tests/UnitConverterTest.cs#L95-L111

Use singleton pattern to make it testable, yet easy to manipulate and use via the `Default` property.

Maybe something like this?
```c#
public class QuantityFactory
{
    public static QuantityFactory Default { get; } = new QuantityFactory();

    public void SetCreateFunction<TQuantity>(Enum unit, Func<QuantityValue, TQuantity>)
		where TQuantity : IQuantity
    { /* ...magic here... */ }

    public TQuantity Create<TQuantity>(QuantityValue value, Enum unit)
		where TQuantity : IQuantity
    { /* ...magic here... */ }
}
```

Note that `QuantityValue` is a wrapper type that implicitly casts `decimal` or `double` values. We introduced this to reduce the binary size, because 1000+ units with twice the method count actually became significantly larger.

With this, I don't think we need `Quantity.From()` anymore, but we could make that work too if we wanted to.

#### 2. Take `QuantityFactory` as parameter in the JsonConverter
Now that we have a configurable and testable way to construct quantities we can pass in an `QuantityFactory` instance to the converter. `QuantityFactory.Default` would then be the default instance passed unless otherwise specified.

Thoughts?

dandm1:
Thanks for your feedback.

 

I agree that the static functions are a weak architecture – they came out of a desire to minimise the refactoring to the remainder of the code necessary to implement the JSON serialisation. The same applies to the support for Quantity.From – hoping to have the custom quantities perform like standard ones in as many contexts as possible.

 

For what it is worth, the biggest blocker I hit on this was the QuantityType enum – as I was trying to avoid changing the code generation and I don’t believe it is possible to add values to an enum at runtime. At the moment where these are relevant I am returning QuantityType.Undefined, but I find this unsatisfactory / potentially confusing.

 

The upside of the static approach is that I could register a custom type in my project just by calling Quantity.AddUnit once at startup – and then everything subsequent doesn’t have to know or care about the custom type.  It does however make unit tests difficult.  It would also make a horrible mess of any use cases where code wants to redefine a unit on the fly; you would hope no one would want to do this, but people are very creative.

 

I think I may struggle to get sufficiently on top of the codebase to wholesale strip out a factory class in the time I have available. I will have a go and see how that looks – comparing it to just having a factory class that knows about the custom types and leaving a broader refactor for later work.

 

Cheers,

 

David

dandm1:
OK - much wider changes to the code now, but a nicer solution without the static list.

I didn't quite follow the suggestion on implementation approach, preferring instead to mostly change leave the existing signatures alone. I moved the object generation functions into QuantityInfo - where they can be created by reflection when the type is built. I then created a dictionary of types in a QuantityFactory type as suggested and moved the quantity generation logic from both Quantity.cs and the JsonSerialiser class into this. 

As a result of this there is a greatly reduced need for Quantity.g.cs - it is now only used in generating the Infos list.

I expected changing to reflection approaches to have a bad impact on performance, but I'm not actually seeing that. My profiling seems to suggest this approach is no slower, though would be worth verifying that.

dschuermans:
How will this affect #732? I've created the PR today (see #746)

dandm1:
I think this an #732 fit pretty well together.  Biggest change necessary to #732 to accommodate this is that your UnitsNetBaseJsonConverter<T>.ConvertValueUnit function just becomes a call to QuantityFactory.FromValueAndUnit - as well as adding a constructor to take the QuantityFactory.

dschuermans:
Well, i'm not a big fan of adding a QuantityFactory parameter to the ctors of the JsonConverters because it adds an explicit dependency on the QuantityFactor and it might become troublesome when used in a dependency injection setup

If i understand correctly, the problem you are running into is that you are unable to deserialize a json
e.g `{ "unit": "HowMuchUnit.Some", "value": 123 }` 
into your custom "HowMuch" type defined in an assembly outside UnitsNet, because the ConvertValueUnit code builds the `UnitsNet.Units.HowMuch,UnitsNet` fully qualified type name and it can't find that?

If i were to go this approach, I would expect some kind of service / factory like your QuantityFactory to exist in UnitsNet, which already knows about all the available units shipped with UnitsNet.
This factory can be used to construct any unit, given its type, unit and value (e.g. typeof(Power), PowerUnit.Watts, 20)

It is then up to **your application**, during its bootstrapping phase to tell UnitsNet that you have defined some additional units in your assemblies and where UnitsNet can find them and how they should be constructed.

Then the UnitsNetBaseJsonConverter can simply access the QuantityFactory directly (without it having to be provided through the ctor) and attempt to construct any type from the data it extracted from the JSON string.

Psudo example:
image


These are my 2 cents 😄

angularsen:
@dschuermans I'm short on time, but I think maybe we are getting best of both worlds now?
If I understood correctly; 
1. JSON converter will use the default (global singleton instance) of `QuantityFactory` by default. You can do `QuantityFactory.Default.AddUnit()` at application startup to configure this globally, if you so wish.
2. Alternatively, you can pass in your own instance of `QuantityFactory` to the JSON converter ctor to be sure the instance has not been tampered with by other nugets or code you have running.

> If i were to go this approach, I would expect some kind of service / factory like your QuantityFactory to exist in UnitsNet [..]  to construct any unit, given its type, unit and value (e.g. typeof(Power), PowerUnit.Watts, 20)

I think this is what we are getting with this design? You can now do 
```c#
IQuantity oneMeter = QuantityFactory.Default.From(1, LengthUnit.Meter);
IQuantity oneSillyTon = QuantityFactory.Default.Parse(typeof(HowMuch), "1 sillyton");
```

Let me know if I got anything wrong, I kinda rushed a bit to get through this PR tonight.

angularsen:
I see your concern, but I really don't think it's all that bad. 

We could very well let `Quantity.From()` and `Quantity.Parse()` methods be a facade for `QuantityFactory.Default.From()` and similar methods, if that reads better.

At least this way, we have the opportunity to fully control the configuration at runtime - but more importantly we can write tests that don't affect other tests that happen to use the same static values.

Global static values is a painful design in my opinion, singletons offers a good compromise of being convenient yet testable.

dandm1:
I think the pattern of using a QuantityFactory.Default in almost all cases here, but having the ability to construct alternatives, is the right balance here.  I struggle to think of any use cases outside of unit testing where you would want to have different quantity factories - but I can't think how you'd do unit testing well without it.

I don't really agree with the philosophy of moving away from reflection in these cases - though I suspect some of the reflection can be done more cleanly. It feels to me like there is a lot of non-obvious boilerplate code required to make a custom instance of IQuantity work well. I was working to a philosophy of minimising the amount of code needed in IQuantity - even if that made the QuantityInfo class more complex.   The AddUnit call in QuantityFactory is a case in point - I see virtue in being able to call it from external code using type names and leaving the AddUnit call to do the heavy lifting of creating an instance to get the QuantityInfo data - rather than requiring anyone calling AddUnit to create an instance.  Supporting both might be an option?

angularsen:
Great input, I'll try to take another look tonight. The design for
supporting third-party quantities is absolutely still up in the air and
kind of introducing itself organically now to solve specific needs. It is
not a holistic design yet.
I didn't quite get a handle on why reflection made the integration easier,
but I'll give another read later. I think we should maybe take a step back
and consider exactly what should be in IQuantity vs QuantityInfo vs
Quantity/QuantityFactory and how to best support third-party quantities
that way.
If we can avoid reflection, I'd prefer so, but if it adds real value then
yes, we'll do that.

angularsen:
@dschuermans A follow up:

> I think it shouldn't be the JsonConverter's respsonsibility to determine which instance of QuantityFactory to use.

I agree, and I believe that is exactly what we achieve by having the default ctor pick the singleton and the more specific ctor take an explicit factory object, which is entirely optional to use. The converter does not have to decide anything on that regard. 

The whole reason for adding this factory was to support third-party quantities. The reason for making it possible to inject the factory into the converter was to make it testable.

We already had support for converting between third-party quantities/units, now we also have support for deserializing and serializing third-party quantities.

> Maybe it's a better idea to allow overriding of the custom methods

Maybe, but I generally try to avoid inheritance if I can. It tends to create more complexity than composition and injecting dependencies.

dschuermans:
It's all good. Are you first going to merge this or my rework?

angularsen:
@dschuermans I'm not sure, depends on which one feels more ready to merge I guess. I have yet to look at yours, but I plan to in a few moments if I get enough time.

angularsen:
@dandm1 
> It feels to me like there is a lot of non-obvious boilerplate code required to make a custom instance of IQuantity work well. 

I agree, but please mind that `IQuantity` was never designed for the purpose of third-party quantities. We are now trying to retrofit support for that via this type.

I'm all for reducing the boilerplate needed to add support for new quantities, but I really want to avoid reflection to achieve that if we can. It's just too prone to bugs in the future when we refactor and change method signatures and if we missed a test case.

### A holistic approach to third-party quantities
Okay, I just had to ramble down some thoughts to clear my own mind a bit.

It's out of scope for this PR, but I think we should at least give thought to how we can best solve support for third-party quantities as a whole.

These are the things I can think of that we need to solve (or have solved):
1. Represent quantities generically (solved with `IQuantity` and `IQuantity<TUnitEnum>`)
	- Value members - `Value`, `Unit`, `Type`, `Dimensions`
    - Metadata - `QuantityInfo`
	- Formatting strings - `ToString()` with various overloads
	- Conversion members - `As(Enum)`

2. Construct quantities given value/unit
	- `new Length(double, LengthUnit)` for strongly typed construction, only built-in types
	- `Quantity.From(double, Enum)` currently only supports built-in types, this PR adds support for third-party types via `QuantityFactory.Default.AddUnit()`.

3. Convert from value/unit to unit
	- `new Length(1, LengthUnit.Meter).As(LengthUnit.Centimeter)` for strongly typed conversion, only for built-in types
	- `UnitConverter.Convert(QuantityValue, Enum, Enum)` for any two units, even from different quantities. Use `UnitConverter.Default.SetConversionFunction()` to configure third-party units, the same is used for built-in units.

4. Parse quantities
	- `Length Length.Parse(string)` for strongly typed parsing, only built-in types
	- `IQuantity Quantity.Parse(Type, string)` currently only supports built-in types, this PR adds support for third-party types via `QuantityFactory.Default.AddUnit()`.

5. JSON serialization
	- Previously only built-in types supported, this PR adds support for third-party quantities via `QuantityFactory.Default.AddUnit()`

#### Some challenges
- `QuantityType` enum is used by several methods and can't support third-party quantities.
- `IQuantity` is very thick, maybe we could slim it down to the absolute bare minimum needed to support use cases like serialization/deserialization, 
- Extension points seems a bit "spread out" and the design feels inconsistent
	- Use `UnitConverter` to add conversion functions
	- Use `QuantityFactory` to add create/parse functions
	- `QuantityParser` exists, but instead of adding third-party parsing capabilities it is only used internally for the actual parsing logic. This feels inconsistent with the above two.

#### Idea - Introduce a facade type to configure Units.NET
To make the API more consistent for built-in types vs third-party types and to gather everything related to global configuration of Units.NET, I think we should consider adding a facade configuration type.

1. Configure built-in types the exact same way as third-party types (inspired by `UnitConverter`), at least strive to where possible. 
Some examples in pseudo code:
	a. `UnitsNetConfig.Default.SetConversionFunction((value, fromUnit, toUnit) => {})`
    b. `UnitsNetConfig.Default.SetCreateFunction((value, unit) => {})`
    c. `UnitsNetConfig.Default.SetParseFunction((string, quantityType) => {})`
    d. `UnitsNetConfig.Default.AddQuantity(QuantityInfo)` that maybe calls the above methods to set everything up?

angularsen:
For this PR though, to move this forward, I think we are not too far off. I still favor configuring convert/parse/create functions like in the proposal above, instead of reflection.

To reduce boilerplate for this PR, we can start by moving the create/parse functions from `QuantityFactory` and `QuantityInfo` to a new `UnitsNetConfig` type - as per the proposal. `HowMuch` will still have to implement all the stuff in `IQuantity`, we can't easily change that right now, but we can later consider introducing facets of the interface that are more minimal, such as `ISimpleQuantity` with only value/unit and `IFormattableQuantity` with all the ToString() methods - etc. I haven't thought this through yet, but the idea is to be able to serialize/deserialize third-party quantities without having to implement unnecessary stuff.

Thoughts?

dschuermans:
KISS :wink

angularsen:
That is a fair point, it is easy to get overboard on this stuff. 

Supporting third-party stuff is not trivial though, requires some thinking to fit it into all the existing code and make it intuitive to use. We also need to consider how many will actually use Units.NET for third party quantities and whether it is really worth too big changes.

dandm1:
I definitely agree that there is a risk of over engineering here in the pursuit of a low yield use case.  The units in Units.NET are quite comprehensive already, so the places where something new would be required are quite limited. The use case in Cumulus MX is almost embarrassingly trivial for example - simply trying to be able to work with a single IStatistics<IQuantity> implementation rather than needing a special case for IStatistics<double>. 

The thinking in @angularsen's notes on add-in support are all very good.  A few observations:
- Custom types can of course support a call to `new HowMuch(double,HowMuchUnit` and `.as(HowMuchUnit)` - but support can't be guaranteed.
- The QuantityType enum kept causing me problems. At the least it would be good to add a `QuantityType.Custom` to avoid the need to overload one of the existing members. Biggest use I can currently see for this is in enumerating the defined types on construction of Quantity.  I considered changing to using reflection to get rid of this use case, but didn't seem necessary.
- Bringing all the config points together seems like a positive approach - but probably a backlog item.

There might be a viable middle way of avoiding reflection in the base code, but having a (suggested?) CustomQuantityBase class that takes a minimalist interface and uses it to reflect the other details necessary to create a full IQuantity.

angularsen:
@dandm1 Could you please elaborate a bit on your usecase? It's very useful to have a concrete example to discuss around. What is `IStatistics`, what does it represent and how do you use it? Why does it benefit you to treat it as an `IQuantity` (if you do)?

dandm1:
In the CumulusMX context we are using various IQuantities to capture the observations from a weather station. There is an IStatistics<> class that calculates statistics based on those observations - maximums, minimums, averages, rates of change, etc over daily, monthly, annual and rolling horizons. We/I then save the results by serialising the whole structure as JSON.

So far so good, except that some of the observable quantities are unitless -  doubles and integers - so we ended up with two different implementations of IStatistics  :
`StatisticsQuantity<TQuantity> : IStatistics<TQuantity>`
`StatisticsDouble : IStatistics<double>`

And a whole lot of branching based on the generic type. This became too much duplication, so in the name of simplicity I implemented `Number : IQuantity` which simply wrapped a double into an IQuantity so all of the measures could use the StatisticsQuantity implementation. But then the JSON Serialisation failed - so we ended up with this PR.

As I said, an almost embarrassingly simple use case - and certainly there are other approaches:
- Go back to forked code.
- Add a unitless double equivalent to the core of UnitsNet.
- Wrap the half-dozen units we need in wrapper classes with a common interface which also has a 'double' implementation.
- Unseal the classes and extend them with a common interface which also extends 'double' (this was my first approach).

angularsen:
Thanks, will get back to you when I find time.

angularsen:
Ok, a couple of thoughts.

1. Min/max/avg of `IQuantity` is already supported, added in #692. Would that solve your need?

https://dotnetfiddle.net/K11USx
```c#
		var masses = new IQuantity[]{ Mass.FromKilograms(50), Mass.FromKilograms(100) };
		var lengths = new IQuantity[]{ Length.FromCentimeters(50), Length.FromCentimeters(100) };
		IQuantity avgMass = masses.Average(MassUnit.Gram);
		IQuantity sumLength = lengths.Sum(LengthUnit.Meter);
		IQuantity minLength = lengths.Min(LengthUnit.Meter);
		IQuantity maxLength = lengths.Max(LengthUnit.Meter);

		Console.WriteLine("Avg: {0}", avgMass); // Avg: 75,000 g
		Console.WriteLine("Sum: {0}", sumLength); // Sum: 1.5 m
		Console.WriteLine("Min: {0}", minLength); // Min: 0.5 m
		Console.WriteLine("Max: {0}", maxLength); // Max: 1 m
```

2. Could you do with [Ratio](https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/Ratio.json)? It is unitless.

PR #745, opened 2020-02-04
Add pound/foot-second in DynamicViscosity

bigmic:
Add pound-foot-second in DynamicViscosity

Numbers are based on https://www.convertunits.com/info/lb/ft-s

### DIFF
--- a/Common/UnitDefinitions/DynamicViscosity.json
+++ b/Common/UnitDefinitions/DynamicViscosity.json
@@ -17,7 +17,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "Ns/m²" ]
+          "Abbreviations": ["Ns/m²"]
         }
       ]
     },
@@ -26,11 +26,11 @@
       "PluralName": "PascalSeconds",
       "FromUnitToBaseFunc": "x",
       "FromBaseToUnitFunc": "x",
-      "Prefixes": [ "Milli", "Micro" ],
+      "Prefixes": ["Milli", "Micro"],
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "Pa·s", "PaS" ]
+          "Abbreviations": ["Pa·s", "PaS"]
         }
       ]
     },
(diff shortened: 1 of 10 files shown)


Common/UnitDefinitions/DynamicViscosity.json
```diff
@@ -79,7 +79,19 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "lbf·s/ft²" ]
+          "Abbreviations": ["lbf·s/ft²"]
+        }
+      ]
+    },
+    {
+      "SingularName": "PoundFootSecond",
```
angularsen: Should this perhaps be named PoundPerFootSecond?
Ref: https://www.aqua-calc.com/what-is/dynamic-viscosity/pound-per-foot-second

Common/UnitDefinitions/DynamicViscosity.json
```diff
@@ -79,7 +79,19 @@
...
+      "PluralName": "PoundsFootSecond",
+      "FromUnitToBaseFunc": "x * 1.4881639",
+      "FromBaseToUnitFunc": "x / 1.4881639",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "lb/ft-s" ]
```
angularsen: This is debatable, but the convention in Units.NET is to favor ft·s instead of ft-s, for consistency across units. We can make an exception if this other form is clearly more widely used, let me know.

We can also add both abbrevation forms if it adds value, but the first item in the array is the one used by ToString() by default so we need to choose a primary abbrevation in that case.

UnitsNet.Tests/CustomCode/DynamicViscosityTests.cs
```diff
@@ -14,6 +14,7 @@ public class DynamicViscosityTests : DynamicViscosityTestsBase
         protected override double NewtonSecondsPerMeterSquaredInOneNewtonSecondPerMeterSquared => 1;
         protected override double PascalSecondsInOneNewtonSecondPerMeterSquared => 1;
         protected override double PoiseInOneNewtonSecondPerMeterSquared => 10;
+        protected override double PoundsFootSecondInOneNewtonSecondPerMeterSquared => 0.671968994813;
```
angularsen: Verified. 
https://www.convertunits.com/from/pound/foot-second/to/newton+second/square+meter

PR #746, opened 2020-02-05
Rework of UnitsNetJsonConverter

### LINKED ISSUE #732: Rework of UnitsNetJsonConverter

dschuermans:
Fixes #732.

Reworked UnitsNetJsonConverter according to #732. 
Also split up serialization and deserialization tests into seperate files for better readability and extracted test objects into their own classes. Consider this an overall clean-up of the test project


dschuermans:
@angularsen If you want better code coverage, I could always update the unit tests to also test the actual converters.
The current tests are actually integration tests if you look at it 😮

dschuermans:
Okay, i'm adding tests for the converters themselves at the moment 😄

dschuermans:
Okay, i'm done now 👍

angularsen:
Perfect, I'll try to take a look later today.

UnitsNet.Serialization.JsonNet.Tests/UnitsNetBaseJsonConverterTest.cs
```diff
@@ -0,0 +1,216 @@
...
+
+            Assert.NotNull(result);
+            Assert.Equal(unitName, result?.Unit);
+            Assert.Equal(unitValue, result?.Value);
+        }
+
+        /// <summary>
+        /// Dummy converter, used to access protected methods on abstract UnitsNetBaseJsonConverter{T}
+        /// </summary>
+        private class TestConverter : UnitsNetBaseJsonConverter<string>
```
angularsen: Hm, why do we need to `access protected methods on abstract UnitsNetBaseJsonConverter`?
I love tests, but is it really useful to test the implementation details of an abstract class?
Isn't it enough to test the public members of the derived types?

I mean, in the end, we really just care about it being serialized and deserialized correctly.

UnitsNet.Serialization.JsonNet.Tests/UnitsNetIComparableJsonConverterTest.cs
```diff
@@ -0,0 +1,123 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Newtonsoft.Json;
+using UnitsNet.Serialization.JsonNet.Tests.Infrastructure;
+using Xunit;
+
+namespace UnitsNet.Serialization.JsonNet.Tests
+{
+    public sealed class UnitsNetIComparableJsonConverterTest
```
angularsen: These tests are awesome ❤️

UnitsNet.Serialization.JsonNet.Tests/UnitsNetIQuantityJsonConverterTest.cs
```diff
@@ -0,0 +1,125 @@
...
+            using (var jsonReader = new JsonTextReader(stringReader))
+            {
+                var result = sut.ReadJson(jsonReader, typeof(IQuantity), null, false, JsonSerializer.CreateDefault());
+
+                Assert.Null(result);
+            }
+        }
+
+        [Fact]
+        public void UnitsNetIQuantityJsonConverter_ReadJson_works_as_expected()
```
angularsen: Should we add a test case about not supporting deserialization of `IComparable` anymore?

UnitsNet.Serialization.JsonNet/Internal/ReflectionHelper.cs
```diff
@@ -7,8 +7,10 @@ namespace UnitsNet.Serialization.JsonNet.Internal
     /// <summary>
     ///     Helper for dealing with reflection, abstracting API differences between old and new .NET framework.
     /// </summary>
+    [Obsolete("The JsonConverter(s) for UnitsNet no longer relies on reflection")]
```
angularsen: Nice

UnitsNet.Serialization.JsonNet/UnitsNetIComparableJsonConverter.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace UnitsNet.Serialization.JsonNet
+{
+    /// <inheritdoc />
+    /// <summary>
+    /// JSON.Net converter for IComparable properties that are actually UnitsNet units.
+    /// Use with caution, as this might interfere with any other converters for IComparable.
+    /// Must be registered in the <see cref="JsonSerializerSettings.Converters"/> collection -after- <see cref="UnitsNetIQuantityJsonConverter"/>
+    /// Should only be used when UnitsNet types are assigned to properties of type IComparable.
+    /// Requires TypeNameHandling on <see cref="JsonSerializerSettings"/> to be set to something other than <see cref="TypeNameHandling.None"/> so that it outputs $type when serializing.
```
angularsen: Should we clarify this bit? This implementation doesn't support writes, so I guess what you are saying here is that the JSON needs to include $type in order to deserialize correctly and that this option must be used by those who produce the JSON string.

This, however, begs a question. Maybe I am forgetting something, but, is this serializer really useful if it doesn't support writes? Is it really useful to only deserialize? Who will produce the JSON then?

Below I see
> "Serialization of IComparable is handled by default serialization"

Are you saying that consumers should use the old serializer for writes, and this new serializer for reads?

UnitsNet.Serialization.JsonNet/UnitsNetIQuantityJsonConverter.cs
```diff
@@ -0,0 +1,70 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace UnitsNet.Serialization.JsonNet
+{
+    /// <inheritdoc />
+    /// <summary>
+    /// JSON.net converter for IQuantity types (e.g. all units in UnitsNet)
+    /// Use this converter to serialize and deserialize UnitsNet types to and from JSON
+    /// </summary>
+    public sealed class UnitsNetIQuantityJsonConverter : UnitsNetBaseJsonConverter<IQuantity>
```
angularsen: Nice, short and simple.

UnitsNet.Serialization.JsonNet.Tests/UnitsNetBaseJsonConverterTest.cs
dschuermans: what you are referring to are the integration tests imo. 
These are actual unit tests.
This way, if anything changes inside the base converter we'll notice straight away.
Otherwise, i'd have to test the shared functions on either both or one of the converters which makes it unclear.
Not explicitly testing the functionality is a no-no

UnitsNet.Serialization.JsonNet.Tests/UnitsNetIQuantityJsonConverterTest.cs
dschuermans: we are supporting it, but in a different converter?
The obsolete message on the old converter should reflect this

UnitsNet.Serialization.JsonNet/UnitsNetIComparableJsonConverter.cs
dschuermans: If i recall correctly, when serializing an IComparable then JsonNet is capable of simply doing this, following the standard serialization conventions.
The "WriteJson" is never called in our case, always the "WriteJson" of the IQuantityConverter (because it's registered before this one).

Irregardless of our implementation, deserializing **any** interface requires either a $type or custom converter.

Its simply not meant to use this converter on its own, always in combination with the other one...

angularsen: Aha, gotcha, that was the missing piece for me. Could you update the xmldoc to more specifically call this out? It was not obvious to me.

UnitsNet.Serialization.JsonNet.Tests/UnitsNetBaseJsonConverterTest.cs
```diff
@@ -0,0 +1,216 @@
...
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
+using Xunit;
+
+namespace UnitsNet.Serialization.JsonNet.Tests
+{
+    public sealed class UnitsNetBaseJsonConverterTest
+    {
+        private readonly string unitName;
+        private readonly double unitValue;
```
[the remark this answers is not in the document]
dschuermans: It's your code base, but i'll call you on to change the values everywhere if it ever needs changing 😋

angularsen:
Haha, that is fine by me 😆
In my experience, it is extremely rare that such constants in test code
change. When they do, either it is a matter of simple search/replace or
there is a bigger conceptual change that requires rewriting the tests
anyway. Time spent reading code vs writing code is probably some tens of
times more, so I always favor readability.

UnitsNet.Serialization.JsonNet/UnitsNetIComparableJsonConverter.cs
dschuermans: sure thing

dschuermans: I've updated the XmlDoc of the WriteJson method

UnitsNet.Serialization.JsonNet.Tests/UnitsNetBaseJsonConverterTest.cs
```diff
@@ -0,0 +1,216 @@
...
+using Newtonsoft.Json.Converters;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
+using Xunit;
+
+namespace UnitsNet.Serialization.JsonNet.Tests
+{
+    public sealed class UnitsNetBaseJsonConverterTest
+    {
+        private readonly string unitName;
```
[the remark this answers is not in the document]
dschuermans: Done

UnitsNet.Serialization.JsonNet/UnitsNetIComparableJsonConverter.cs
angularsen: Looks good, thanks!

UnitsNet.Serialization.JsonNet.Tests/UnitsNetBaseJsonConverterTest.cs
angularsen: Alright, we'll keep it

angularsen:
Great job, thanks!

angularsen:

PR #747, opened 2020-02-05
Added kilopound-force imperial variant to force

dschuermans:
We were missing the imperial variant of kilonewtons, being Kilopound-force.
Up untill now, we've worked around it in our code using an extension method:

image

So i'm adding it to the library so others can use it too.

Due to the difference in abbreviation (lbf vs kipf), i couldn't simply add the "Kilo" prefix to Kilopound-force (which I initially tried)

Converter used: https://www.unitconverters.net/force/kilopound-force-to-newton.htm

### DIFF
--- a/Common/UnitDefinitions/Force.json
+++ b/Common/UnitDefinitions/Force.json
@@ -147,6 +147,22 @@
           "Abbreviations": [ "ozf" ]
         }
       ]
+    },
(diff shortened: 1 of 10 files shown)


angularsen:
Looks perfect to me!

angularsen:

PR #748, opened 2020-02-05
Added KilopoundsForcePerFoot (kipf/ft) imperial variant to ForcePerLength unit

dschuermans:
Added Kilopounds-force per foot imperial variant (kipf/ft)

PR #749, opened 2020-02-05
Added KilopoundForceFeetPerDegrees variant to RotationalStiffness

dschuermans:
Added KilopoundForceFeetPerDegrees variant to RotationalStiffness, along with PoundForceFeetPerDegrees.

Used to do it through extension methods:

image

Figured i add it to the library, for the benefit of others looking for an imperial variant.

Not much to find in terms of converters, so I used google's:

image


Common/UnitDefinitions/RotationalStiffness.json
```diff
@@ -20,6 +20,30 @@
...
+    },
+    {
+      "SingularName": "KilopoundForceFootPerDegrees",
+      "PluralName": "KilopoundForceFeetPerDegrees",
+      "FromUnitToBaseFunc": "x*77682.6",
+      "FromBaseToUnitFunc": "x/77682.6",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "kipf·ft/deg" ]
```
angularsen: Awaiting decision in #748 on kipf vs kip.

angularsen: Should we use degree symbol in the abbreviation? Below ref uses `kips.ft.s/°`.

Ref: http://webcache.googleusercontent.com/search?q=cache:BzUwiIqqit4J:52.202.177.230/units_reference_guide.html&hl=en&gl=no&strip=1&vwsrc=0

UnitsNet.Tests/CustomCode/RotationalStiffnessTests.cs
```diff
@@ -28,6 +28,8 @@ namespace UnitsNet.Tests.CustomCode
     public class RotationalStiffnessTests : RotationalStiffnessTestsBase
     {
         protected override double NewtonMetersPerRadianInOneNewtonMeterPerRadian => 1;
+        protected override double PoundForceFeetPerDegreesInOneNewtonMeterPerRadian => 0.0128729;
```
angularsen: I couldn't find a second source to verify, but your google result looks good to me.

Common/UnitDefinitions/RotationalStiffness.json
dschuermans: Personally I think it's best to stick with "deg", to be consistent with "rad" if it's in radians.

But we could define all of the following?

1. kipf·ft/deg
2. kip·ft/deg
3. k·ft/deg
4. kipf·ft/°
5. kip·ft/°
6. k·ft/°

angularsen: I'm fine with deg, but I think we should add both as we have done for other quantities. We seem to have consistently used degree symbol as primary and "deg" as secondary.

https://github.com/angularsen/UnitsNet/search?l=JSON&q=deg

angularsen: So yes, I agree with your proposal if we choose this ordering:
1. Degree symbol, "deg"
2. kipf, kip, k (as discussed in #748 )

angularsen:

PR #752, opened 2020-02-15
Add prefixes for electron volt

0xF6:
```
1keV = 1 000                eV
1MeV = 1 000 000            eV
1GeV = 1 000 000 000        eV
1TeV = 1 000 000 000 000    eV 
```


UnitsNet.Tests/CustomCode/EnergyTests.cs
```diff
@@ -37,6 +38,7 @@ public class EnergyTests : EnergyTestsBase
         protected override double ErgsInOneJoule => 10000000;
 
         protected override double MillijoulesInOneJoule => 1000;
+        protected override double TeraelectronVoltsInOneJoule => 6241509.34326018;
```
angularsen: Are there different values for this? I see Kilo, Mega, Giga and Tera all have different digits, while I expected only the power value to be different (1e3, 1e6, 1e9 etc).

Also, as an example, your value of `TeraelectronVoltsInOneJoule  => 6241509.34326018` is different than `6241506.479963235` from https://www.convertunits.com/from/TeV/to/joule.

I'd like these to be consistent, if there exists one accurate constant to use for all of them.

angularsen: `6.241509343260179e18` is the constant used for `ElectronVoltsInOneJoule`. Unless this one is wrong too, I suggest using it as basis for all the others.
Also, please use scientific notation 1.12345678e15 for all these numbers so it is easier to read for large/small numbers.

0xF6: Is that better? :(
Or you need to record explicit results?

UnitsNet.Tests/CustomCode/EnergyTests.cs
```diff
@@ -63,6 +58,17 @@ public class EnergyTests : EnergyTestsBase
 
         protected override double WattHoursInOneJoule => 0.000277777778;
 
+        protected override double ElectronVoltsInOneJoule => 6.2415093433e+18;
+
+        protected override double KiloelectronVoltsInOneJoule => ElectronVoltsInOneJoule / 1_000;
```
angularsen: Sorry, I was maybe not clear, these values still need to be constants and not computations. The idea is to double check the computations in the conversion functions, preferably from a reference like a website or conversion calculator.

Please change these back to how they were, but use the same constant value as in electron volts and use scientific notation: `KiloelectronVoltsInOneJoule => 6.2415093433e+15` etc.

angularsen: Here are the conventions we use:
https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#conventions

angularsen:
Perfect, thanks! Nuget is out.

[Release UnitsNet/4.46.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.46.0)

Issue #753, opened 2020-02-18
Cells/s Unit


WalkerCodeRanger:
We have a need for a unit for a number of cells per second. We generally call this cell throughput.

I understand that this is probably not a unit that should be added to the library as it is too specific to us. However, perhaps there could be a unit for a generic rate? This is very similar to `BitRate`. I know there is a closed issue asking for a unit for the amount/number of items. This seems like a case where that is needed. The rate is the number of items per unit of time.

Alternatively, if support for [Extending with Custom Units](https://github.com/angularsen/UnitsNet/wiki/Extending-with-Custom-Units) was better I'd be happy to create a custom unit. However, as it stands the directions for that are too vague for me to feel comfortable doing that.

Please recommend the best way for me to proceed on this. As this is for our company, I may have some bandwidth to contribute a PR if that is what is needed.

angularsen:
Hi, 

We do have `Frequency` with units like Hertz and KiloHertz. 
It also has other units like BeatsPerMinute RadiansPerSecond.

https://en.wikipedia.org/wiki/Rate_(mathematics)

> Often rate is a synonym of rhythm or frequency, a count per second (i.e., Hertz); e.g., radio frequencies or heart rate or sample rate.

I think perhaps `Frequency` is a better place to keep these units, rather than adding `Rate` quantity in addition.

Would it work for you to add more units to `Frequency`? 
Maybe `UnitsPerSecond`, `UnitsPerHour` etc. 
Alternatively name it `CountsPerSecond`.

WalkerCodeRanger:
Thanks for pointing us in the right direction. After doing some thinking and researching, we like the idea of using `Frequency`. What we have is what Wikipedia calls an [aperiodic frequency](https://en.wikipedia.org/wiki/Aperiodic_frequency). That is normally measured in [inverse seconds](https://en.wikipedia.org/wiki/Inverse_second) (s⁻¹). Wikipedia offers several names for that unit one of which is just "per second". That seems to fit best with the naming scheme of UnitsNet.

I'll put together a PR adding a "PerSecond" unit to `Frequency` so it would be `Frequency.FromPerSecond()` with the symbol "s⁻¹".

angularsen:
Alright, sounds good to me!

WalkerCodeRanger:
We have integrated the updated version with the `PerSecond` unit on `Frequency` and it is working well for us.

angularsen:
Great! I'm glad to hear it works well for you.

>

PR #762, opened 2020-02-27
Add 'Per Second' Unit to Frequency

WalkerCodeRanger:
This PR adds a 'Per Second' unit to `Frequency` to support [inverse seconds](https://en.wikipedia.org/wiki/Inverse_second). This PR closes issue #753.

### DIFF
--- a/Common/UnitDefinitions/Frequency.json
+++ b/Common/UnitDefinitions/Frequency.json
@@ -67,6 +67,18 @@
           "Abbreviations": [ "bpm" ]
         }
       ]
+    },
+    {
+      "SingularName": "PerSecond",
+      "PluralName": "PerSecond",
+      "FromUnitToBaseFunc": "x",
+      "FromBaseToUnitFunc": "x",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "s⁻¹" ]
+        }
+      ]
     }
   ]
 }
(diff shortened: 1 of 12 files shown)


Common/UnitDefinitions/Frequency.json
```diff
@@ -67,6 +67,18 @@
...
+    },
+    {
+      "SingularName": "PerSecond",
+      "PluralName": "PerSecond",
+      "FromUnitToBaseFunc": "x",
+      "FromBaseToUnitFunc": "x",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "s⁻¹" ]
```
angularsen: Encoding error? Doesn't look right in Chrome on Android at least.

WalkerCodeRanger: I saw the same thing on Andriod, but looks fine on Chrome on Windows. I intentionally saved this file as UTF-8 without BOM. Let me know if it should be a different encoding

angularsen: Looks good here too

angularsen:
Nuget on the way out.
[Release UnitsNet/4.50.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.50.0)

PR #771, opened 2020-03-29
Add MillidegreeCelsius and add prefixes to TemperatureDelta

hwaien:
Seems like Millidegree Celsius is currently missing. This pull request adds support for it.


angularsen:
Looks good, except you are missing the new properties in the test code. See https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit.

hwaien:
Seems like degree Celsius cannot be treated the same way as Meter because the conversion between degree Celsius and the base unit of degree Kelvin is not proportional.

Is the recommended approach to declare millidegree Celsius as its own unit (with its own conversion functions from and to the base unit) side-by-side to degree Celsius?

angularsen:
Huh, good question. Temperature units have time and time again proven to be difficult due how they don't share the same zero value and things like arithmetic don't behave like all the other quantities.

I think the solution you came up with is the best compromise, I don't see any other way of getting it added. Good job figuring that out. For me, this is ready to merge. Agreed?

angularsen:
Just a control question, are you sure millidegrees Celsius is used for absolute temperature measurements? My googling seems to indicate it is used for delta/relative temperature measurements (1 millidegree Celsius hotter than X), but I can't immediately find references to absolute measurements (the measured temperature was 10 millidegrees Celsius).

hwaien:
Good question @angularsen. I am not sure how widespread the practice is, but at my company, millidegrees Celsius is the default unit used by our stakeholders, for both absolute temperature and relative temperature. For now, we maintain our own conversion functions in conjunction with UnitsNet to meet stakeholder needs.

On a different note: This is my first time contributing to UnitsNet. (What a wonderful project. Thanks to all your hard work!) What's the release process (and timeline) from code changes proposed in a pull request like this one, to the changes being available for consumption from NuGet?

angularsen:
Thanks. Just making sure this will be actually used by someone and that it
makes sense in someone's domain.

The nuget can be out later this evening when I'm back on the computer.

>

angularsen:
Nuget on the way
[Release UnitsNet/4.55.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.55.0)

hwaien:
Thanks @angularsen !

PR #779, opened 2020-04-24
Added "Deca", "Deci" and "Hecto" prefixes for "UsGallon" unit with tests

DInozemtsev:


angularsen:
Hi, thanks for contributing.

I have to ask though, are you sure these units are actually widely used?

I can't seem to find much on Google and it seems a bit weird to me that metric prefixes (deci, kilo, deca) etc are used for US units.

It would be great to see some URLs pointing to references or actual usage of these units before merging this as I suspect it is maybe not used much. Do you use these units yourself? In what domain/context?

Best,
Andreas

DInozemtsev:
Hi, @angularsen 

First of all, thank you so much for creation of such amazing library! I can not express well enough how much I love it.

2nd, I work as an developer of Energy and Utility Management system that targets US market. A plenty of US utility companies use 'cGal'/'100 Gal' and 'kGal'/'1000 Gal' units in theirs reports and DBs for utility types as 'Water', 'Diesel Fuel', 'Oil' and so on.

More than that, Google seems to be aware of this units and looks like people search for such conversions on regular basis.

1GalConv
 search for theirs conversions regularly.  
 
100Gal

1000Gal

However I totally agree that 'DecaGallon' it rather something exotic, so plz let me know if you would like to remove it and I'll get rid of this in no time.

UnitsNet.Tests/CustomCode/VolumeTests.cs
```diff
@@ -32,12 +32,17 @@ public class VolumeTests : VolumeTestsBase
...
+        protected override double DecausGallonsInOneCubicMeter => 2.6417217e+1;
+
         protected override double DecilitersInOneCubicMeter => 1E4;
+        protected override double DeciusGallonsInOneCubicMeter => 2.6417217e+3;
         protected override double HectocubicFeetInOneCubicMeter => 3.531472e-1;
         protected override double HectocubicMetersInOneCubicMeter => 0.01;
 
         protected override double HectolitersInOneCubicMeter => 1E1;
 
+        protected override double HectousGallonsInOneCubicMeter => 2.6417217;
```
angularsen, line 44: Verified

Common/UnitDefinitions/Volume.json
```diff
@@ -242,7 +242,7 @@
       "PluralName": "UsGallons",
       "FromUnitToBaseFunc": "x*0.00378541",
       "FromBaseToUnitFunc": "x/0.00378541",
-      "Prefixes": [ "Kilo", "Mega" ],
+      "Prefixes": [ "Deca", "Deci", "Hecto", "Kilo", "Mega" ],
```
angularsen, line 245: You mentioned reports using 'cGal'/'100 Gal' and 'kGal'/'1000 Gal'.

Please note:
1. The abbreviations used by `ToString()` will in this PR will be `dgal (U.S.)` and `hgal (U.S.)` etc. 
2. `Volume.Parse("5 cGal")` will throw an exception, because parsing is currently case sensitive.

If you really want `ToString()` and `Parse()` to support using uppercase `Gal`, then you need to add that `Gal (U.S.)` abbreviation to the end of the `Abbreviations` array below. It would still need the `(U.S.)` suffix to be consistent with our naming conventions though..

To use a specific abbreviation in `ToString()`, take a look at [custom string formatting](https://github.com/angularsen/UnitsNet/wiki/String-Formatting).

If you are fine with the existing lowercase `gal (U.S.)` abbreviation, then I recommend not adding `Gal (U.S.)` unless it will actually be used.

angularsen:
I've had this nagging in the back of my mind for some time, so I finally got a PR out to support case insensitive parsing that I mentioned above. If that is merged, then it should be able to parse `cGal (U.S.)` without having to add that variant explicitly in the JSON file.

https://github.com/angularsen/UnitsNet/pull/781

Common/UnitDefinitions/Volume.json
angularsen: #781 is now merged, so this is point is no longer relevant :-) This PR is then good to go!

angularsen:
Nuget is on the way out.
[Release UnitsNet/4.59.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.59.0)

Common/UnitDefinitions/Volume.json
DInozemtsev: @angularsen Sorry for delay with my reply. You see, in our system exists a few different names for the same unit. First company names it "kGal", the second - "100 Gal", the third - "Hundred Gallons" and so on. Thus I don't think that there is any standard name of those units in the States, and if there is actually one, looks like no one really respects it. So I would leave it as it is now. :-)

angularsen:
Excellent, thanks for the update.

PR #781, opened 2020-04-27
Support case insensitive parsing of quantities

angularsen:
## Motivation
Make it easier to define unit abbreviations in order for `Parse()` to work for all variations of upper/lowercase in the unit abbreviation.

This was already supported for `UnitParser` and even though `QuantityParser` builds on top of `UnitParser`, it failed due to a regex it computes based on the list of unit abbreviations.

## Design
- Case insensitive parsing should "just work" if it matches exactly one abbreviation
- If case insensitive parsing matches multiple units, then only an exact match on casing will succeed. Otherwise it already throws `UnitNotFoundException` in `UnitParser`.

## Examples
- `Volume.Parse("1 L")` currently fails, because only lowercase `l` is defined in JSON. There are no other volume units with the same lowercase abbreviation.

## Changes
- Add ignore case flag to regex in `QuantityParser`
- Add tests for happy and sad cases

### DIFF
--- a/UnitsNet.Tests/QuantityParserTests.cs
+++ b/UnitsNet.Tests/QuantityParserTests.cs
@@ -9,6 +9,53 @@ namespace UnitsNet.Tests
     public class QuantityParserTests
     {
         [Fact]
+        public void Parse_WithSingleCaseInsensitiveMatch_ParsesWithMatchedUnit()
+        {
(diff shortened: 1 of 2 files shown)


angularsen:
[Release UnitsNet/4.59.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.59.0)

Issue #786, opened 2020-05-21
Quantity formatting: Trailing zeros


pgrawehr:
The documentation of `QuantityFormatter.Format` specifies that (and it also behaves like that):
```
/// "s1", "s2", ..., "sN": The value with N significant digits after the radix followed by the unit abbreviation. For example,
/// "s4" would return "1.2345 m" if <see cref="IQuantity.Value" /> is 1.2345678. Trailing zeros are omitted.
```

Is there a reason trailing zeros are ommitted? Besides that this constantly changes the length of the output string (if the value changes slightly), it may present a value different from the expectation. For instance, if I have a temperature sensor from which I know it is accurate to 0.01 °C, I would want to use "s2" to print its value with two digits. The output now prints either "20.02 °C" or "20 °C", just because the measurement has slightly changed. Reading the later, I don't know that the measurement was actually "20.00 °C". 

For a scientist, it does make a difference whether a reading says "20 °C" or "20.00 °C". I'd like to see an option to force the number of digits to be fixed (much like the "F" default number format specifier does).

tmilnthorp:
It becomes difficult to support all the permutations for custom format strings on a single object which holds multiple internal values. For the most flexibility, you will want to use the built-in string formatting. For your case, I would suggest something as follows:

```csharp
var temperature = UnitsNet.Temperature.FromDegreesCelsius( 20.02 );
var output = string.Format( "{0:F3} {1:a}", temperature.Value, temperature );
```

This will result in the string "20.020 °C"

pgrawehr:
I could also make sense to allow `string.Format( "{0:F3 a}", temperature );` That would allow for maximum flexibility. Idealy, `string.Format( "{0:F3 a2}", temperature );` would then return a converted value (`72.223 °F`), since the value in celcius with another unit name is meaningless. 

Supporting all the standard format strings would be a good idea. The overlap with the "g" isn't so big of a deal I think, since "G" is still available and (unless you are into astrophysics, maybe) we usually use the unit that best suits the number, so that scientific notation is not needed.

tmilnthorp:
I agree, we just don't support compound format strings currently. I created #788 to get us a bit closer to what you need.

FYI ```"{0:a2}"``` would not give °F for a °C value. It's only for when there multiple abbreviations for a unit. This will throw a ```FormatException```:
```csharp
var temperature = UnitsNet.Temperature.FromDegreesCelsius( 20.02 );
[...]

PR #788, opened 2020-05-22
Add support for standard numeric format strings

tmilnthorp:
Add support for standard double numeric format strings. Also allow upper case format specifiers.

Work for #786


UnitsNet/QuantityFormatter.cs
```diff
@@ -13,6 +13,32 @@ namespace UnitsNet
...
+        /// "a0", "a1", ..., "aN": The Nth unit abbreviation for <see cref="IQuantity{TUnitType}.Unit" />. "a0" is the same as "a".
+        /// A <see cref="FormatException"/> will be thrown if the requested abbreviation index does not exist.
+        /// "V" or "v": String representation of <see cref="IQuantity.Value" />.
+        /// "U" or "u": The enum name of <see cref="IQuantity{TUnitType}.Unit" />, such as "Meter".
+        /// "Q" or "q": The quantity name, such as "Length".
+        /// "s1", "s2", ..., "sN": The value with N significant digits after the radix followed by the unit abbreviation. For example,
+        /// "s4" would return "1.2345 m" if <see cref="IQuantity.Value" /> is 1.2345678. Trailing zeros are omitted.
+        /// </remarks>
+        /// <returns>The string representation.</returns>
+        public static string Format<TUnitType>(IQuantity<TUnitType> quantity, string format)
```
tmilnthorp, line 36: Just adding a convenience helper to avoid null passing

tmilnthorp:
@angularsen I suggest we deprecate the following format strings: "V", "v", "S", "s".

With this PR I think it makes more sense to use the standard format strings for these, or a custom string.Format on ```Value``` for fine-grained control.

tmilnthorp:
I can't help but think when given a standard numeric specifier that you should get the value in that format plus the default abbreviation instead of just the value?

UnitsNet/QuantityFormatter.cs
```diff
@@ -64,7 +64,7 @@ public static string Format<TUnitType>(IQuantity<TUnitType> quantity, string for
         public static string Format<TUnitType>(IQuantity<TUnitType> quantity, string format, IFormatProvider formatProvider)
             where TUnitType : Enum
         {
-            formatProvider ??= CultureInfo.CurrentUICulture;
+            formatProvider = formatProvider ?? CultureInfo.CurrentUICulture;
```
angularsen, line 72: You can keep the fancy formatting, I'm still getting used to it too but I don't see why we shouldn't use newer syntax in C#. It is more terse.

tmilnthorp: It wouldn't compile. I'm guessing it's a C# 8.0 feature - same reason the nullable reference types PR won't compile.

angularsen:
@tmilnthorp Hm, I think maybe you are right. I have honestly never used the string format stuff much myself, but I can see how it doesn't really add much value to do `myQuantity.ToString("v")` rather than `myQuantity.Value.ToString()`. It's better to rely on the standard numeric formatting than reinventing our own.

UnitsNet/QuantityFormatter.cs
angularsen: As per your comment in the other PR, the c#8 syntax might work now that I upgraded to VS2019 image. Maybe. It seems netstandard2.0 defaults to c# 7.3 though, so might require explicitly configuring language version.

angularsen:
Getting late so I'll have to revisit this, but are we proposing to support standard *numeric* formats on our quantities? I mean, we could, but I have a feeling this is not commonly used or a best practice. Why can't the consumer just do `$"{myLength.Value:f3} {myLength:a01}"` to get `1.001 cm`? Are there maybe use cases I am not remembering, that warrants supporting this many custom formats?

lipchev:
Being in the category of people who are doing the dual string formatting- I would much prefer the simple "f3" format. My reasons are:

1.  Avoid MultiBinding for XAML
2.  "aX" is very much _not standard_ - having it around in my code is both annoying and to some degree implies that I actually care about the abbreviation- when 99% of the time- I don't.
3. If we somehow made the abbreviations dependent on a UnitSystem / RegionalSettings / CustomConfiguration- would I have to go over my existing code looking for those `a02`  ?

UnitsNet/QuantityFormatter.cs
tmilnthorp: True. Can do that another day though

tmilnthorp:
- ToString should generally convey the information about the "whole" object. In our case, both the unit and the value are important in representing the entirety of the quantity.
- In general, most string formatting cares about how the numeric part is formatted and less with the abbreviation (the default abbreviation is probably fine for the majority of use cases). 
- This also helps simplify formatting code from the client's perspective (using well-known formats) while also maintaining the flexibility to pass the individual parts to ```string.Format```

pgrawehr:
Note that nothing prevents us from allowing `$"{temperature:F3 v}"` or $"{tempeature:g2 a1}", where you would get standard formatting and (optional) unit name in just one call. I think it is convenient to format one object in one call, as this prevents misstakes such as `$"{temperature1:F3} {temperature2:v}"`.

angularsen:
@tmilnthorp @pgrawehr Ok, I am convinced. I like the argument that `ToString should generally convey the information about the "whole" object.` and that the standard numeric formats simply only have effect on the value part.

@lipchev 
> "aX" is very much not standard - having it around in my code is both annoying and to some degree implies that I actually care about the abbreviation- when 99% of the time- I don't.

I forgot to mention, but `a` also works. It is the same as `a0`.

UnitsNet/QuantityFormatter.cs
```diff
@@ -13,6 +13,37 @@ namespace UnitsNet
     /// </summary>
     public class QuantityFormatter
     {
+        /// <summary>
+        /// The available UnitsNet custom format specifiers.
+        /// </summary>
+        private static readonly char[] UnitsNetFormatSpecifiers = { 'A', 'a', 'G', 'g', 'Q', 'q', 'S', 's', 'U', 'u', 'V', 'v' };
+
+        /// <summary>
+        /// Formats the given quantity using the given format string. Uses the <see cref="CultureInfo.CurrentUICulture" />.
```
angularsen, line 22: CurrentUICulture or CurrentCulture?

The former would be consistent with the current implementation, the latter would be more correct for the number formatting as discussed in #795.

UnitsNet/QuantityFormatter.cs
```diff
@@ -13,6 +13,37 @@ namespace UnitsNet
...
+        /// </summary>
+        /// <typeparam name="TUnitType">The quantity's unit type, for example <see cref="LengthUnit"/>.</typeparam>
+        /// <param name="quantity">The quantity to format.</param>
+        /// <param name="format">The format string.</param>
+        /// <remarks>
+        /// The valid format strings are as follows:
+        /// Any of the standard numeric format strings for double values except for "G" or "g" ("C" or "c", "E" or "e", "F" or "f", "N" or "n", "P" or "p", "R" or "r").
+        /// "G" or "g": The value with 2 significant digits after the radix followed by the unit abbreviation, such as "1.23 m".
```
angularsen, line 30: Very nice docs, but it would probably not read so well in Visual Studio / Rider intellisense.

Try to add a list structure here, definition list probably works well for this:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/list

UnitsNet/QuantityFormatter.cs
```diff
@@ -23,13 +54,14 @@ public class QuantityFormatter
         /// <see cref="CultureInfo.CurrentUICulture" /> if null.</param>
         /// <remarks>
         /// The valid format strings are as follows:
-        /// "g": The value with 2 significant digits after the radix followed by the unit abbreviation, such as "1.23 m".
-        /// "a": The default unit abbreviation for <see cref="IQuantity{TUnitType}.Unit" />, such as "m".
+        /// Any of the standard numeric format strings for double values except for "G" or "g" ("C" or "c", "E" or "e", "F" or "f", "N" or "n", "P" or "p", "R" or "r").
```
angularsen, line 57: I would suggest list structure here too.

UnitsNet/QuantityFormatter.cs
```diff
@@ -39,41 +71,62 @@ public static string Format<TUnitType>(IQuantity<TUnitType> quantity, string for
...
+            if(UnitsNetFormatSpecifiers.Any(c => c == format[0]))
             {
-                if(formatString.Length > 1 && !int.TryParse(formatString.Substring(1), out number))
-                    throw new FormatException($"The {format} format string is not supported.");
+                // UnitsNet custom format string
 
-                formatString = formatString.Substring(0, 1);
-            }
+                int precisionSpecifier = 0;
+                char formatSpecifier = format[0];
```
angularsen, line 82: Nitpick, can reuse this variable in the above if.

UnitsNet/QuantityFormatter.cs
tmilnthorp: Yea, for consistency

UnitsNet/QuantityFormatter.cs
tmilnthorp: 👍

UnitsNet/QuantityFormatter.cs
tmilnthorp: I'll add that tomorrow!

UnitsNet/QuantityFormatter.cs
tmilnthorp: Added it back when I was updating based on PR :)

UnitsNet/QuantityFormatter.cs
tmilnthorp: Added!

angularsen:
Fixes #786

angularsen:

Issue #688, opened 2019-08-09
Use C#8 nullable reference types


angularsen:
C#8 is due in the fall of 2019 and one interesting feature we can take advantage of is nullable reference types, instead of JetBrains' `[NotNull]` and `[CanBeNull]` attributes that don't really give much value to the consumers of the library.

With nullable reference types I believe we can communicate to consumers whether a returned object can be null or not. We use struct value types for all quantities today so I'm not sure how much impact it will make, but we are discussing moving to class instead and at that point it will probably give a lot more value.

Anyway, it's a cool new language feature and if anyone wants to do a PR to play with it, this can be a good start. If no one else does, I'll probably take a stab at it sometime.

ebfortin:
I'm looking at using Nullable Reference Types as well. From what I understand though it's a compiler only thing. It will not validate at runtime. So it's only usable at compile time with static analysis.

angularsen:
Yes that's right. It only has an effect at compile time. Also, it is an optional feature that consumers of the nuget will have to enable in their projects to benefit from. Still, the idea of communicating nullability to nuget consumers is pretty great.

angularsen:
Keepiung this one up for a bit in case some one wants to pick it up and run with it.

lipchev:
@angularsen Are the jet-brains annotations still used anywhere: I cannot find anything, but I may not be searching right...

angularsen:
If you search for `JetBrains.Annotations` in the codebase, you should find `using` statements or similar if we do use it. Either way, we can just remove the nuget and remove any lingering usages. I see no point in keeping it around.

angularsen:
Fixed in #789

PR #789, opened 2020-05-22
Utilizing nullable reference types

tmilnthorp:
Per #688

### DIFF
--- a/CodeGen/Generators/UnitsNetGen/StaticQuantityGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/StaticQuantityGenerator.cs
@@ -22,6 +22,8 @@ using JetBrains.Annotations;
 using UnitsNet.InternalHelpers;
 using UnitsNet.Units;
 
+#nullable enable
+
 namespace UnitsNet
 {
     /// <summary>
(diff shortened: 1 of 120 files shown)


tmilnthorp:
@angularsen I am not sure how to fix this build

angularsen:
Well, the error says
> CSC : error CS8630: Invalid 'nullable' value: 'Enable' for C# 7.3. Please use language version 'preview' or greater. 

I'll try pushing a language version fix and see if it helps.

angularsen:
It builds locally here, so must be environment specific.

angularsen:
Weird, I just upgraded to VS2019 image and it.. worked. I tried earlier and it.. well, didn't work :-) Maybe the appveyor images have been improved since then.

tmilnthorp:
Glad it works! Everything ok to merge? I just did the main UnitsNet project for now. We can work on projects in parts.

I'm also guessing that component runtime doesn't support this.

CodeGen/Generators/UnitsNetGen/StaticQuantityGenerator.cs
```diff
@@ -22,6 +22,8 @@ public override string Generate()
 using UnitsNet.InternalHelpers;
 using UnitsNet.Units;
 
+#nullable enable
```
angularsen, line 25: Shouldn't we set this on the project instead?
https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/nullable-reference-types

Or are there so many issues to fix that we will have to introduced this in bits and pieces?

angularsen: Other than that, this looks pretty good to me and ready to merge.

tmilnthorp: It is set on the project. However without this in the generated files, I receive a compiler error:

```
Error CS8669 The annotation for nullable reference types should only be used in code within a '#nullable' annotations context. Auto-generated code requires an explicit '#nullable' directive in source.
```

I'm guessing the .g. annotation tells the compiler it's generated?

angularsen: Aha, yes it probably does. Today I learned.

angularsen:

PR #794, opened 2020-05-26
Fix TryParseFeetInches when current locale uses ' as number separator

pgrawehr:
My current locale (de-CH) uses ' (an apostrophe) as thousands separator. This broke the unit test for the `TryParseFeetInches` method, because the input string `1'1"` is ambiguous. Therefore I now removed the separator as valid input in a feet-inches combination. It's pretty unlikely that someone will need to parse `1,345'21"`.

### DIFF
--- a/UnitsNet.Tests/QuantityIFormattableTests.cs
+++ b/UnitsNet.Tests/QuantityIFormattableTests.cs
@@ -65,6 +65,9 @@ namespace UnitsNet.Tests
             Assert.Equal(expected, length.ToString(sFormatString, NumberFormatInfo.InvariantInfo));
         }
 
+        /// <summary>
+        /// This verifies that the culture is correctly considered when formatting objects with an explicit culture.
+        /// </summary>
         [Fact]
         public void FormattingUsesSuppliedLocale()
         {
@@ -72,6 +75,23 @@ namespace UnitsNet.Tests
             CultureInfo c = CultureInfo.CreateSpecificCulture("de-CH");
             string formatted = string.Format(c, "{0:g}", t);
             Assert.Equal("2'012.12 °C", formatted);
+        }
+
+        /// <summary>
+        /// This verifies that the culture is correctly considered when using <see cref="FormattableString.ToString(IFormatProvider)"/>
+        /// </summary>
+        [Fact]
+        public void FormatStringWorksWithSuppliedLocale()
+        {
+            Temperature t = Temperature.FromDegreesCelsius(2012.1234);
+            CultureInfo c = CultureInfo.CreateSpecificCulture("de-CH");
+
+            FormattableString f = $"{t:g}";
+            Assert.Equal("2'012.12 °C", f.ToString(c));
+
+            // This does not work. Looks like a compiler bug to me.
+            // string f2 = $"{t:g}".ToString(c);
+            // Assert.Equal("2'012.12 °C", f2.ToString(c)); // Actual value is formatted according to CurrentUiCulture.
         }
 
         [Fact]
--- a/UnitsNet.Tests/QuantityIFormattableTests.cs
+++ b/UnitsNet.Tests/QuantityIFormattableTests.cs
@@ -72,9 +72,10 @@ namespace UnitsNet.Tests
         public void FormattingUsesSuppliedLocale()
         {
             Temperature t = Temperature.FromDegreesCelsius(2012.1234);
-            CultureInfo c = CultureInfo.CreateSpecificCulture("de-CH");
+            CultureInfo c = new CultureInfo("de-CH", false);
             string formatted = string.Format(c, "{0:g}", t);
-            Assert.Equal("2'012.12 °C", formatted);
+            // Let's be very explicit here
+            Assert.Equal("2" + c.NumberFormat.NumberGroupSeparator + "012" + c.NumberFormat.NumberDecimalSeparator + "12 °C", formatted);
         }
 
         /// <summary>
@@ -84,10 +85,10 @@ namespace UnitsNet.Tests
         public void FormatStringWorksWithSuppliedLocale()
         {
             Temperature t = Temperature.FromDegreesCelsius(2012.1234);
-            CultureInfo c = CultureInfo.CreateSpecificCulture("de-CH");
+            CultureInfo c = new CultureInfo("de-CH", false);
 
             FormattableString f = $"{t:g}";
-            Assert.Equal("2'012.12 °C", f.ToString(c));
+            Assert.Equal("2" + c.NumberFormat.NumberGroupSeparator + "012" + c.NumberFormat.NumberDecimalSeparator + "12 °C", f.ToString(c));
 
             // This does not work. Looks like a compiler bug to me.
             // string f2 = $"{t:g}".ToString(c);
--- a/UnitsNet.Tests/QuantityIFormattableTests.cs
+++ b/UnitsNet.Tests/QuantityIFormattableTests.cs
@@ -66,6 +66,15 @@ namespace UnitsNet.Tests
         }
 
         [Fact]
+        public void FormattingUsesSuppliedLocale()
+        {
+            Temperature t = Temperature.FromDegreesCelsius(2012.1234);
+            CultureInfo c = CultureInfo.CreateSpecificCulture("de-CH");
+            string formatted = string.Format(c, "{0:g}", t);
+            Assert.Equal("2'012.12 °C", formatted);
+        }
+
+        [Fact]
         public void UFormatEqualsUnitToString()
         {
             Assert.Equal(length.Unit.ToString(), length.ToString("u"));
--- a/UnitsNet/CustomCode/Quantities/Length.extra.cs
+++ b/UnitsNet/CustomCode/Quantities/Length.extra.cs
@@ -60,6 +60,27 @@ namespace UnitsNet
         }
 
         /// <summary>
+        ///     Try to parse a string with one or two quantities of the format "&lt;quantity&gt; &lt;unit&gt;".
+        /// </summary>
+        /// <param name="str">String to parse. Typically in the form: {number} {unit}</param>
+        /// <param name="provider">Format to use when parsing number and unit. Defaults to <see cref="CultureInfo.CurrentUICulture" /> if null.</param>
+        /// <param name="allowedNumberStyles">Allowed number styles</param>
+        /// <param name="result">Resulting unit quantity if successful.</param>
+        /// <returns>True if successful, otherwise false.</returns>
+        /// <example>
+        ///     Length.Parse("5.5 m", new CultureInfo("en-US"));
+        /// </example>
+        private static bool TryParse([CanBeNull] string str, [CanBeNull] IFormatProvider provider, NumberStyles allowedNumberStyles, out Length result)
+        {
+            return QuantityParser.Default.TryParse<Length, LengthUnit>(
+                str,
+                provider,
+                From,
+                allowedNumberStyles,
+                out result);
+        }
+
+        /// <summary>
         /// Special parsing of feet/inches strings, commonly used.
         /// 2 feet 4 inches is sometimes denoted as 2′−4″, 2′ 4″, 2′4″, 2 ft 4 in.
         /// The apostrophe can be ′ and '.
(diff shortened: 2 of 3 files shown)


UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -80,7 +101,8 @@ public static bool TryParseFeetInches([CanBeNull] string str, out Length result,
             str = str.Trim();
 
             // This succeeds if only feet or inches are given, not both
-            if (TryParse(str, formatProvider, out result))
+            // Do not allow thousands separator here, since it may be equal to the unit abbreviation (').
+            if (TryParse(str, formatProvider, NumberStyles.AllowDecimalPoint | NumberStyles.AllowExponent | NumberStyles.AllowLeadingSign, out result))
```
angularsen, line 105: It seems from [docs](https://docs.microsoft.com/en-us/dotnet/api/system.double.tryparse?view=netcore-3.1#System_Double_TryParse_System_String_System_Double__) that `double.Parse()` defaults to `NumberStyles.Float` and `NumberStyles.AllowThousands`. Could we then simply use [NumberStyles.Float](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.numberstyles?view=netcore-3.1#System_Globalization_NumberStyles_Float
) here?

> The s parameter is interpreted by using a combination of the NumberStyles.Float and NumberStyles.AllowThousands flags. This means that white space and thousands separators are allowed but currency symbols are not.

UnitsNet.Tests/QuantityIFormattableTests.cs
```diff
@@ -65,6 +65,15 @@ public void SFormatEqualsSignificantDigits(string sFormatString, string expected
             Assert.Equal(expected, length.ToString(sFormatString, NumberFormatInfo.InvariantInfo));
         }
 
+        [Fact]
+        public void FormattingUsesSuppliedLocale()
```
angularsen, line 69: A short comment would help on why this choice of culture was used. Perhaps that it was to demonstrate a different thousands separator than is used in US English and invariant culture, which many developers also have configured when running tests locally.

tmilnthorp: Shouldn't the test be doing a parse on foot inches?

angularsen: Oh yeah, good point. I thought the generic implementation was fixed, but it was only feetinch parsing. Good catch.

tmilnthorp:
I'm not so certain that just ignoring it is a good idea. Looks like we're missing a couple tests though - for your case as well as a number separated by a ```'```

UnitsNet/CustomCode/Quantities/Length.extra.cs
pgrawehr: Yea, that's probably the better choice.

UnitsNet.Tests/QuantityIFormattableTests.cs
pgrawehr: This test is actually not directly related to the other issue. This was to check that the specified culture argument really ends up being passed to the `ToString(string, IFormatInfo)` method - and is correctly used there. 

I didn't add new tests for the FootInches stuff, as that test already existed (and failed in the first place when I was initialy building the library).

pgrawehr: Background: I was initially trying to do `$"{temperature}".ToString(myCulture)`, but for reasons I don't quite understand yet, this didn't work and the Temperature.ToString method would always get null as culture argument.

tmilnthorp: Yes, but we should add tests that show the failure for everyone / Appveyor builds. Pass an explicit culture such as your own.

pgrawehr: I've clarified the test case. Looks to me as if I ran into a compiler bug :-/ 

@tmilnthorp : Unfortunatelly, I don't fully understand what I should be doing. The build did pass (we'll, before running into a merge issue that is - i'll fix that).

pgrawehr:
> 
> 
> I'm not so certain that just ignoring it is a good idea. Looks like we're missing a couple tests though - for your case as well as a number separated by a `'`

What test cases are you thinking of? If the number is i.e. embedded in quotes, then I think it's ok to have parsing fail, as quotes are not part of a number (except in the case where the quote is one of the valid separators).

tmilnthorp:
I added an example here: #799. I'm confused now though because it passes :)

pgrawehr:
@tmilnthorp : And I'm confused now because my updated test fails :-/ 

Output:
```
Error Message:
   Assert.Equal() Failure
            (pos 1)
Expected: 2'012.12 øC
Actual:   2'012.12 øC
            (pos 1)
  Stack Trace:
     at UnitsNet.Tests.QuantityIFormattableTests.FormatStringWorksWithSuppliedLocale() in C:\projects\unitsnet\UnitsNet.Tests\QuantityIFormattableTests.cs:line 90
```

The strings are exactly the same to me...

I'll check what's going on with your PR.

angularsen:
Looks like an encoding issue. There are different apostrophe characters in
different code sets.

tmilnthorp:
Oh man. There's the right single quotation mark ```U+2019``` and an apostrophe ```U+0027```.

Sometimes, I feel like I know nothing 😆

angularsen:
To be fair, xUnit COULD help identify this by providing the character values when there is a single character off.

UnitsNet.Tests/QuantityIFormattableTests.cs
```diff
@@ -84,10 +85,10 @@ public void FormattingUsesSuppliedLocale()
         public void FormatStringWorksWithSuppliedLocale()
         {
             Temperature t = Temperature.FromDegreesCelsius(2012.1234);
-            CultureInfo c = CultureInfo.CreateSpecificCulture("de-CH");
+            CultureInfo c = new CultureInfo("de-CH", false);
 
             FormattableString f = $"{t:g}";
-            Assert.Equal("2'012.12 °C", f.ToString(c));
+            Assert.Equal("2" + c.NumberFormat.NumberGroupSeparator + "012" + c.NumberFormat.NumberDecimalSeparator + "12 °C", f.ToString(c));
```
angularsen, line 91: I would hard code the expected string here instead of dynamically looking it up. Same above.

UnitsNet.Tests/QuantityIFormattableTests.cs
angularsen: @pgrawehr 
> $"{temperature}".ToString(myCulture) 
> but for reasons I don't quite understand yet, this didn't work and the Temperature.ToString method would always get null as culture argument.

Won't the string interpolation call the parameterless `ToString()`, and that's maybe why you are seeing null?

tmilnthorp: That's just calling ToString on a string literal. Temperature was already converted to a string at construction time. Equivalent to temperature.ToString().ToString(myCulture)

tmilnthorp: And a string literal when passed a format provider does nothing

tmilnthorp: https://docs.microsoft.com/en-us/dotnet/api/system.string.tostring?view=netcore-3.1#System_String_ToString_System_IFormatProvider_

pgrawehr:
> $"{temperature}".ToString(myCulture)

The reason is: It's defined that way: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#interpolated-strings

The type of `$"..."` is string, _unless_ you immediatelly assign it to FormattableString. So the type of the expression is derived from the left-hand-side of the assigment.

UnitsNet.Tests/QuantityIFormattableTests.cs
pgrawehr: I changed this just because of the problem that showed up yesterday: The characters Windows uses for these members change. And so any Windows Update would break the test. It could also be that the values are different on another platform (i.e when running the tests under linux).

angularsen: Ok, although I don't expect Windows to change this back and forth frequently, it's fine to keep it like this too.

UnitsNet.Tests/QuantityIFormattableTests.cs
```diff
@@ -65,6 +65,36 @@ public void SFormatEqualsSignificantDigits(string sFormatString, string expected
...
+        {
+            Temperature t = Temperature.FromDegreesCelsius(2012.1234);
+            CultureInfo c = new CultureInfo("de-CH", false);
+
+            FormattableString f = $"{t:g}";
+            Assert.Equal("2" + c.NumberFormat.NumberGroupSeparator + "012" + c.NumberFormat.NumberDecimalSeparator + "12 °C", f.ToString(c));
+
+            // This does not work. Looks like a compiler bug to me.
+            // string f2 = $"{t:g}".ToString(c);
+            // Assert.Equal("2'012.12 °C", f2.ToString(c)); // Actual value is formatted according to CurrentUiCulture.
```
angularsen, line 95: The commented code does not work, because these two statements are equivalent:
```c#
string f2 = $"{t:g}".ToString(c);
string f2 = t.ToString("g").ToString(c);
```

### LINQPad sample
when CurrentUICulture is Norwegian and thousand separator is comma `,`.
```c#
Temperature t = Temperature.FromDegreesCelsius(2012.1234);
CultureInfo c = new CultureInfo("de-CH", false);

FormattableString f = $"{t:g}";
string f2 = $"{t:g}".ToString(c);
string f3 = t.ToString("g").ToString(c);

var expected = "2" + c.NumberFormat.NumberGroupSeparator + "012" + c.NumberFormat.NumberDecimalSeparator + "12 °C";
(expected == f.ToString(c).Dump()).Dump();
(expected == f2.ToString(c).Dump()).Dump(); // Actual value is formatted according to CurrentUiCulture.
(expected == f3.ToString(c).Dump()).Dump(); // Actual value is formatted according to CurrentUiCulture.
```

Output:
```
2’012.12 °C
True
2,012.12 °C
False
2,012.12 °C
False
```

angularsen: Remove the commented code from this PR.

UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -80,7 +101,8 @@ public static bool TryParseFeetInches(string? str, out Length result, IFormatPro
             str = str.Trim();
 
             // This succeeds if only feet or inches are given, not both
-            if (TryParse(str, formatProvider, out result))
+            // Do not allow thousands separator here, since it may be equal to the unit abbreviation (').
+            if (TryParse(str, formatProvider, NumberStyles.Float, out result))
```
angularsen, line 105: As @tmilnthorp pointed out earlier, we should have unit tests that covers the new behavior for feet-inches. It should test what happens if you try to parse strings like these, with a culture that has apostrophe character as thousand separator. We could also test cultures with other thousand separators, like `nb-NO` Norwegian that has comma `,` as separator. 
```
1'500' 4"
1'5002′−4″
1'500 ft 4 in
1'5002′ 4″
``` 

My intuition tells me we should get an exception when parsing these, because the thousand separator is no longer allowed and in the first case the feet symbol is even listed twice.

UnitsNet.Tests/QuantityIFormattableTests.cs
pgrawehr: As you've seen 
```
FormattableString f = $"{t:g}";
string result = f.ToString(c);
```
is not equivalent to
```
string result = $"{t:g}".ToString(c);
```
which is what confused me (as this would be the case for every other expression).

I'll remove it, since the problem has nothing to do with this lib, but is a (weird but documented) feature of the CLR.

UnitsNet/CustomCode/Quantities/Length.extra.cs
pgrawehr: I'll do that, but I feel we should somehow merge @tmilnthorp s PR with this one, otherwise it gets confusing.

angularsen: I guess we can wait for #799 to merge first to reuse the combinatorial parameter refactoring that is about to be added there.

angularsen: #799 was abandonend, please see #803 for supporting thousand separators in regex instead of not allowing them. I think that may be a better solution to this problem.

PR #795, opened 2020-05-26
Use CurrentCulture rather than CurrentUICulture

### LINKED ISSUE #787: Default number format should be CultureInfo.CurrentCulture, not CurrentUICulture
The default formatting of numbers (in QuantityFormatter.cs), when no explicit culture is given, should use `CultureInfo.CurrentCulture` not `CultureInfo.CurrentUICulture`. On Windows, the first corresponds to the selection from the control panel "Regional Settings" and defines the locale settings of the user. `CurrentUICulture` represents the language of the UI, which will be english for everyone that uses Windows in english. 

In my case, CurrentUICulture is "de-DE", since my Windows is German, but CurrentCulture is "de-CH", since I use the number format for Switzerland (which is quite different).

tmilnthorp:
Fixes #787


angularsen:
Looks good to me, but as you said. I too think this should be treated as a breaking change. It's quite likely that this will break something for someone.

I have no problems with bumping to major version, other than that there are other things we want to include when first start doing a bump. 

#563

CodeGen/Generators/UnitsNetGen/StaticQuantityGenerator.cs
```diff
@@ -87,7 +87,7 @@ public static bool TryFrom(QuantityValue value, Enum unit, out IQuantity quantit
         /// <summary>
         ///     Try to dynamically parse a quantity string representation.
         /// </summary>
-        /// <param name=""formatProvider"">The format provider to use for lookup. Defaults to <see cref=""CultureInfo.CurrentUICulture"" /> if null.</param>
+        /// <param name=""formatProvider"">The format provider to use for lookup. Defaults to <see cref=""CultureInfo.CurrentCulture"" /> if null.</param>
```
angularsen, line 90: Hm. What is correct here, if we can only specify a single culture?

My understanding:
CurrentUICulture is used for display language in the UI, which affects our abbreviation parsing.
CurrentCulture is used for date/number formatting, which affects our number parsing.
Also,  (but not for the abbreviation, right?).

tmilnthorp: From my understanding, CurrentUICulture is for use with resource manager only. It's an odd one, as CurrentUICulture would be en-US even for someone in Britain (where I'd expect en-GB). Important if we ever want to distinguish meter from metre for example?

angularsen: Well. Thought experiment:
- I am Norwegian.
- I configure my Windows region to Norway, so that Windows Store shows me content available in my region. I believe this is not exposed in .NET.
- I configure my Windows regional format to `English (Sweden)` to get ISO dates instead of Norwegian dates, because I'm weird that way. I believe this configures `CurrentCulture` to `en-SE`.
- I configure Windows display language to `English (United States)`, because I prefer reading English. I believe this configures `CurrentUICulture` to `en-US`.

This is my actual current setup:
image

image

image

Interestingly, my country/region Norway is not visible in RegionInfo. In fact, it seems [this choice is not available in .NET framework, only via WinSDK or P/Invoke](https://stackoverflow.com/a/46213603/134761).

So. In my particular Windows configuration. I want to read English, but I want date/numbers formatted in ISO (Sweden). 

How should UnitsNet format the number and what language should it choose abbreviations from? I would prefer Swedish numbers and English abbreviations, but it seems I may be stuck with all Swedish in this PR proposal, right?

tmilnthorp: Interesting. They had to make it difficult didn't they 😂

tmilnthorp: What does DateTime.Now.ToString("F") give you?

angularsen: Friday, 29 May 2020 01:26:19

angularsen: This is the clock in Windows, ISO formatted
image

tmilnthorp: Presumably you get "Wednesday" because of en and the date because of the SE (en-SE)? 

sv-SE would give you Wednesday in Swedish?

Morbid curiosity what sv-EN gives you :)

tmilnthorp:
```csharp
			var dt = DateTime.Now;

			Console.WriteLine( string.Format( new CultureInfo( "en-US", false ), "{0:F}", dt ) );
			Console.WriteLine( string.Format( new CultureInfo( "en-SE", false ), "{0:F}", dt ) );
			Console.WriteLine( string.Format( new CultureInfo( "sv-SE", false ), "{0:F}", dt ) );
```
Result:
```
Thursday, May 28, 2020 9:30:26 PM
Thursday, 28 May 2020 21:30:26
den 28 maj 2020 21:30:26
```

tmilnthorp: From [MSDN](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.tostring?view=netcore-3.1#System_DateTime_ToString_System_String_):

> The ToString(String) method returns the string representation of a date and time value in a specific format that uses the formatting conventions of the current culture; for more information, see CultureInfo.CurrentCulture.

pgrawehr: > 
> Result:
> 
> ```
> Thursday, May 28, 2020 9:30:26 PM
> Thursday, 28 May 2020 21:30:26
> den 28 maj 2020 21:30:26
> ```

That looks like what I'd expect. So everything is formatted according to the given culture. [This](https://docs.microsoft.com/de-de/visualstudio/code-quality/ca1305?view=vs-2019) suggests to always use CurrentCulture, and I think we can agree that this is what should be used for formatting the number part. For the units, it's a bit more complicated and the docs are a bit less clear.  

Now technically, the unit names are resource-lookups and should use CurrentUiCulture, but apparently, its not used that way (otherwise, the day and month names in the above example would have been english, regardless of the explicitly provided culture, because @tmilnthorp uses en-US as CurrentUiCulture [I assume]). It's not possible to provide two cultures to a ToString call. And it would be confusing to use one language's string formatting with another languages unit names. 

IMHO, the correct behavior is to always use the provided culture for the whole formatting operation (number and unit name) and use CurrentCulture as the default.

angularsen: I think you are right. We can only pass a single culture, this is the norm in .NET.

Seeing that .NET DateTime also uses the passed in culture, instead of trying to use CurrentUICulture for the text parts, we should probably do the same.
In a way, it makes sense that whatever culture you pass in, is used consistently for all the parts of the string formatting. It is easier to reason about, even if it leaves my Windows configuration edge case hanging in the cold. I'm fine with that.

lipchev: Not sure if this has any relevance- but for my localized WPF applications - I tend to setup 4 things  on login (culture is part of the user profile):
         
            CultureInfo.DefaultThreadCurrentCulture = cultureInfo;
            CultureInfo.DefaultThreadCurrentUICulture = cultureInfo;

            Thread.CurrentThread.CurrentCulture = cultureInfo;
            Thread.CurrentThread.CurrentUICulture = cultureInfo;
..thinking that the UI sets the default culture for the UI threads- with the other one concerning the culture of any background threads. And from what I understand- you're saying that actually one is for the language-localization and the other for the formatting?
PS I'm in the exact same situation as @angularsen in terms of the regional settings (only the input is different)
image

pgrawehr: @lipchev : You would normally do that, if your application supports localization. If it does not, you could still use localized number formatting. The CurrentUiCulture will have a fallback (usually english) if it doesn't exist, while the CurrentCulture is always completely defined. Using applications (Windows itself, but also Office or Visual Studio) in english while still using the locale's preferred number/date formatting is a very common scenario.

tmilnthorp:
Perhaps we should move the abbreviations into a resource file.

angularsen:
Replaced by #795, merged in v5.

PR #802, opened 2020-05-29
Add board foot

danypellerin:


angularsen:
Good job adding this perfectly on the first go!

When you think you've heard just about most units for a quantity, then something like this comes along 😆 I have never seen this one before.

angularsen:
Nuget on the way out.

Issue #777, opened 2020-04-06
Temperature delta base dimensions?


lipchev:
As a prelude to #709 I figure I'd run a test of the currently defined operators and see if there is mismatch between the dimensions of the operands and the result. To my great satisfaction- it found the faulty operator for Density / Mass (which I had previously annotated as obsolete- and completely forgotten about). The only other _problem_ was with the TemperatureDelta which has no defined BaseDimensions- thus failing the dimensions multiplication (though I assume the operators are correct). Was there a reason for it not having the Temperature dimension defined?
Here is the test (I'd put it in the GeneratedQuantityCodeTests):

```

            [Fact]
            public void HasMultiplicationOperator_GivenMassAndVolume_ReturnsFalse()
            {
                Assert.False(HasMultiplicationOperator(typeof(Mass), typeof(Volume)));
                Assert.DoesNotContain(typeof(Volume), GetMultipliers(typeof(Mass)));
            }

            [Fact]
            public void HasMultiplicationOperator_GivenDensityAndVolume_ReturnsTrue()
            {
                Assert.True(HasMultiplicationOperator(typeof(Density), typeof(Volume)));
                Assert.Contains(typeof(Volume), GetMultipliers(typeof(Density)));
                Assert.Equal(typeof(Mass), GetMultiplicationResult(typeof(Density), typeof(Volume)));
            }

            [Fact]
            public void HasDivisionOperator_GivenDensityAndVolume_ReturnsFalse()
            {
                Assert.False(HasDivisionOperator(typeof(Density), typeof(Volume)));
                Assert.DoesNotContain(typeof(Volume), GetDivisors(typeof(Density)));
            }

            [Fact]
            public void HasDivisionOperator_GivenMassAndVolume_ReturnsTrue()
            {
                Assert.True(HasDivisionOperator(typeof(Mass), typeof(Volume)));
                Assert.Contains(typeof(Volume), GetDivisors(typeof(Mass)));
                Assert.Equal(typeof(Density), GetDivisionResult(typeof(Mass), typeof(Volume)));
            }

            [Fact]
            public void HasMultiplicationOperator_GivenTwoQuantities_ReturnsTrueIfDimensionsMultiplicationIsValid()
            {
                var failed = new List<Tuple<QuantityInfo, QuantityInfo, QuantityInfo>>();
                foreach (QuantityInfo firstQuantity in Quantity.Infos)
                {
[...]

PR #805, opened 2020-06-02
Add formula to compute electric power

pgrawehr:
Electric power is defined as P = U * I (Power
equals potential times current)

### DIFF
--- a/UnitsNet.Tests/CustomCode/PowerTests.cs
+++ b/UnitsNet.Tests/CustomCode/PowerTests.cs
@@ -127,5 +127,19 @@ namespace UnitsNet.Tests.CustomCode
             MassFlow massFlow = Power.FromWatts(15.0) / SpecificEnergy.FromJoulesPerKilogram(3);
             Assert.Equal(massFlow, MassFlow.FromKilogramsPerSecond(5));
         }
+
+        [Fact]
+        public void CurrentMultipliedByPotentialEqualsPower()
+        {
+            ElectricCurrent i = ElectricCurrent.FromAmperes(5);
+            ElectricPotential u = ElectricPotential.FromVolts(10);
+            Power p = Power.FromWatts(50);
+            Assert.Equal(p, i * u);
+            Assert.Equal(p, u * i);
+
+            // Validate the inverse is true as well
+            Assert.Equal(i, p / u);
+            Assert.Equal(u, p / i);
+        }
     }
 }
--- a/UnitsNet.Tests/CustomCode/PowerTests.cs
+++ b/UnitsNet.Tests/CustomCode/PowerTests.cs
@@ -129,17 +129,17 @@ namespace UnitsNet.Tests.CustomCode
         }
 
         [Fact]
-        public void CurrentMultipliedByPotentialEqualsPower()
+        public void PowerDividedByElectricCurrentEqualsElectricPotential()
         {
-            ElectricCurrent i = ElectricCurrent.FromAmperes(5);
-            ElectricPotential u = ElectricPotential.FromVolts(10);
-            Power p = Power.FromWatts(50);
-            Assert.Equal(p, i * u);
-            Assert.Equal(p, u * i);
+            ElectricPotential u = Power.FromWatts(10) / ElectricCurrent.FromAmperes(2);
+            Assert.Equal(5, u.Volts);
+        }
 
-            // Validate the inverse is true as well
-            Assert.Equal(i, p / u);
-            Assert.Equal(u, p / i);
+        [Fact]
+        public void PowerDividedByElectricPotentialEqualsElectricCurrent()
+        {
+            ElectricCurrent i = Power.FromWatts(20) / ElectricPotential.FromVolts(5);
+            Assert.Equal(4, i.Amperes);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


UnitsNet.Tests/CustomCode/PowerTests.cs
```diff
@@ -127,5 +127,19 @@ public void PowerDividedBySpecificEnergyEqualsMassFlow()
...
         }
+
+        [Fact]
+        public void CurrentMultipliedByPotentialEqualsPower()
+        {
+            ElectricCurrent i = ElectricCurrent.FromAmperes(5);
+            ElectricPotential u = ElectricPotential.FromVolts(10);
+            Power p = Power.FromWatts(50);
+            Assert.Equal(p, i * u);
+            Assert.Equal(p, u * i);
```
angularsen, line 138: These two test casees are already covered by the above two U\*I and I\*U tests. I think they can be removed here.

UnitsNet.Tests/CustomCode/PowerTests.cs
```diff
@@ -127,5 +127,19 @@ public void PowerDividedBySpecificEnergyEqualsMassFlow()
...
+
+            // Validate the inverse is true as well
+            Assert.Equal(i, p / u);
+            Assert.Equal(u, p / i);
```
angularsen, line 142: Split these two division tests into two separate test cases as you did for multiplication.
A single test case should preferably only test one behavior.

pgrawehr: I purposedly did one test that "tests them all", because that tests that the four methods are consistent against each other. Otherwise, a small typo in one of the tests could hide a bug (i.e. accidentally testing the same method twice, missing another, or "fixing" a problem in the test by adjusting the implementation).

These are simple methods (and simple tests as well), but there's no guarantee that if something is wrong, it can't be in the test.

UnitsNet.Tests/CustomCode/PowerTests.cs
pgrawehr: See below

UnitsNet.Tests/CustomCode/PowerTests.cs
angularsen: Your intention is good, but I still believe this is redundant.
In the end, I don't think we should write tests to ensure the other tests are correct :-) It should test the implementation.

pgrawehr: Yea, it's redundant, but I think that's a good thing for tests. Maybe tests for  overall combined behavior should be (in addition to the others) placed in a special place, but they do help to prevent problems as noted above and they test the logic, not the implementation. Personally, I'm not so much a fan of "one method" - "one test" just because of this - and it tempts people to just rebuild the implementation in the test (including its bugs).

angularsen: I suppose this is to some extent a matter of preference, but I must respectfully disagree. 

I think the premise of adding tests to help catch mistakes in tests that should have caught the implementation error to begin with, is not a good practice. We should instead try to make the tests (and the implementation) correct to begin with. PR reviews help catch mistakes like that, but it will never be 100% of course.

We test that the implementation for the operator overloads do their thing. I really don't think we should make a habit of sprinkling additional combination tests just to make doubly sure.

lipchev: May I also remark that using the tests that I propose in #777  (rough draft) , it should be possible to test the validity of _all_ operators- with the inverse-equality test being just one of the tested behaviors. 
And also to the comment about _`testing tests`_- that was **_exactly_** what I was thinking while writing one of the tests in #777:
"This looks like a test for testing a _testing functionality_ used in other tests."
I mean- the _test-tests_ should obviously be apart from the other tests. 
I guess this is a sort of a round-about counter-example- which in the end appears to still be in line with what  @angularsen is saying.. :)

pgrawehr: @lipchev: Yea, if something like that could be done (where a test i.e. globaly tests that operators are inversible) that would even be better than the single implementation I proposed here. 

That isn't _testing the tests_ IMHO, but testing that the implementation is complete and consistent (and no test case was forgotten - something that quite easily slips a review).

I'll change this here for now, but I still think something with a more global view should be done.

angularsen: Hm, I see you removed the two tests on division? That implementation is now untested.

I think maybe you misunderstood what I meant for you to change. This is my proposal a bit more clearly explained:

1. Remove `CurrentMultipliedByPotentialEqualsPower` method
2. Add `PowerDividedByElectricCurrentEqualsElectricPotential`
3. Add `PowerDividedByElectricPotentialEqualsElectricCurrent`

This way all the new implementations (multiplication and division) are fully tested.

Best,
Andreas

pgrawehr: Oh, stupid me. I thought I had separate implementations for those. So I removed the wrong ones :-/

angularsen:
Thanks a lot. Nuget is on the way.

[Release UnitsNet/4.65.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.65.0)

PR #821, opened 2020-08-07
New unit: Force.ShortTonForce

### LINKED ISSUE #819: Thoughts regarding new ForceUnit "ShortTonForce" / "USTonForce" / whatever?
I was surprised to see that there isn't yet a ForceUnit.ShortTonForce (or equivalent) implemented (though it is just another imperial oddity... I digress).

Is there a specific reason this hasn't been implemented yet, or was there just no demand? If not I can make the contribution, in that case what is the preferred name? ShortTonForce makes the most sense in my mind and nicely follows the already present mass unit of "ShortTon," but I'm open to suggestions.

Here's more info on the unit for those unfamiliar: https://en.wikipedia.org/wiki/Ton-force#Short_ton-force

lwestfall:
Fixes #819


Common/UnitDefinitions/Force.json
```diff
@@ -163,6 +163,20 @@
...
+      "SingularName": "ShortTonForce",
+      "PluralName": "ShortTonsForce",
+      "FromUnitToBaseFunc": "x*8.896443230521e3",
+      "FromBaseToUnitFunc": "x/8.896443230521e3",
+      "XmlDocSummary": "The short ton-force is a unit of force equal to 2,000 pounds-force (907.18474 kgf), that is most commonly used in the United States – known there simply as the ton or US ton.",
+      "XmlDocRemarks": "https://en.wikipedia.org/wiki/Ton-force#Short_ton-force",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "sh tn", "T" ]
```
angularsen, line 177: Are these abbreviations widely used?
In my googling, I find:
- [t (US)f](http://www.kylesconverter.com/force/short-tons--force-to-newtons)
- [short tf](https://www.translatorscafe.com/unit-converter/en-US/force/23-25/ton-force%20(short)-ton-force%20(metric)/) or [tf (short)](https://www.translatorscafe.com/unit-converter/en-US/force/23-25/ton-force%20(short)-ton-force%20(metric)/), with metric ton force as `tf`

In [Mass.json](https://github.com/angularsen/UnitsNet/blob/d684f04d17c80087b649f65e72409594a346df01/Common/UnitDefinitions/Mass.json#L50-L69) we use:
- `"Abbreviations": [ "short tn", "ST" ]`

I don't know what is the most commonly used here, but please consider what should be the primary abbreviation (the first in the array) and whether to add some of those I found (primarily for the purpose of supporting `Parse()`, or `ToString("a2")` string format where in this example the abbreviation index 2 is specified.

UnitsNet.Tests/CustomCode/ForceTests.cs
```diff
@@ -32,6 +32,8 @@ public class ForceTests : ForceTestsBase
 
         protected override double OunceForceInOneNewton => 3.596943089595368;
 
+        protected override double ShortTonsForceInOneNewton => 1.12404471549816e-4;
```
angularsen, line 35: Verified in http://www.kylesconverter.com/force/short-tons--force-to-newtons

Common/UnitDefinitions/Force.json
lwestfall: Yes, I struggled with this as well. In my industry of material handling (I know, I know, this essentially means nothing) we frequently use capital T for US/Short Ton and lower case t for metric tonne, with no qualifier between force and mass (we use weight aka force by default for most purposes). With that said I was unable to find any definitive source as to any standard abbreviation so I agree that some adjustment could be good. I was also unaware that there is preference between the first abbreviation in the array vs others so I'll take that into account in the future.

After thinking about it more, perhaps capitalization should be left out of this decision since people might frequently use something like .ToUpper() or .ToLower() depending on use case which would make it a moot point.

Here's what I propose after viewing your links/suggestions. If you agree then I'll make the changes and commit on top of this PR, but I'm certainly open to further discussion if necessary:

```
          "Culture": "en-US",
          "Abbreviations": [ "t (US)f" , "tf (short)"]
```

angularsen: > use something like .ToUpper() or .ToLower() 

We recently introduced case insensitive parsing, unless two units of the same quantity differ only by casing (rare). https://github.com/angularsen/UnitsNet/pull/781
So we don't have to define both upper/lower variants in JSON, but the one we define will be used by ToString() though.

A few more observations:
- [Ton-force wiki](https://en.wikipedia.org/wiki/Ton-force) uses `1 short tons-force`
- [Force.TonneForce](https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/Force.json#L49-L57) uses `Ton` (seems wrong to me? [Ton-force wiki](https://en.wikipedia.org/wiki/Ton-force) suggests `tf`)

I also would prefer to try to stay consistent with `Mass` and `Force` abbreviations unless it goes against conventions in the domain.

How about something like this:
```
# ShortTonneForce (I personally prefer "tf (short)" as the primary, but no strong opinion other than consistency with other quantities)
"Abbreviations": [ "tf (short)", "t (US)f", "short tons-force" ]

# TonneForce (add "tf", keep Ton to not break backwards compatibility in v4.x)
"Abbreviations": [ "tf", "Ton" ]

# Mass (add "t (short)" and make it primary)
"Abbreviations": [ "t (short)", "short tn", "ST" ]
```

lwestfall:
Sounds good! I'll make changes tomorrow morning.

angularsen:
Perfect! Nuget is on the way out.

[Release UnitsNet/4.69.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.69.0)

Issue #822, opened 2020-08-11
Is it possible to add new Unit Humidity?


ilchdeveloper:
[Humitidy.json.txt](https://github.com/angularsen/UnitsNet/files/5059739/Humitidy.json.txt)

Hello Andreas,

We liked your project and work. We want to use UnisNet library in a feature. I try to add the Humidity Unit pre your instructions - https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit. The codegen run well, but when I run build.bat I get a lot of errors:
![Capture](https://user-images.githubusercontent.com/20879796/89954250-bf8bae80-dbfe-11ea-9b87-d11486729473.PNG)

We don't use the VS 2019. Could you please give us the advice about how we can do it in our environment:  VS2017 and .Net Framework 4.7.1

Thank you!!!

angularsen:
Hi, this project should build just fine on VS2017 (or even without Visual Studio at all).

I believe the only requirement is [.NET Core SDK 2.x](https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#requirements).

You could also try creating a pull request and see if the automated build runs OK. But it's easier to experiment locally of course.

angularsen:
I see you added the PR already, I assume you figured it out. Closing this issue.

krwq:
@angularsen I'd also prefer to have this unit, please reconsider (although I would call it RelativeHumidity).

The reasoning is that we'd like to have strongly typed units in https://github.com/dotnet/iot which have lots of benefits: i.e. our [WeatherHelper](https://github.com/dotnet/iot/blob/master/src/devices/Common/Iot/Device/Common/WeatherHelper.cs) could be implemented partially as extension methods. We could also use the strongly typed unit also to generate the code but since sensors returning relative humidity return the ration instead of RelativeHumidity that makes it a bit harder.

I understand if you push back on that (in such case we'd likely just add similar unit in our repo but I think this repo is a good place to have them all in one place rather than spread across the ecosystem)

angularsen:
@krwq Thanks for chiming in. It helps to have multiple points of view. If several people ask for a quantity, then I think that helps justify adding it.

There is a PR #824 to add quantity `Humidity` with unit `Percent` `%Rh`. Does that cover your needs, or would you need any modifications?

krwq:
@angularsen yes, looks good! Additional fraction unit (0..1) for humidity would be useful as well but percent is sufficient to cover most of the needs.

[...]

PR #824, opened 2020-08-12
Added new Humidity unit

ilchdeveloper:


Common/UnitDefinitions/Humitidy.json
```diff
@@ -0,0 +1,19 @@
+{
+  "Name": "Humidity",
```
angularsen, line 2: I'm not so sure about adding this quantity, it seems to overlap a lot with `Ratio` percentage. Is that an option for you to use? You would lose the `%Rh` abbreviation out of the box though.

We generally want to avoid adding quantities that are not convertible - meaning they only have a single unit (no units to convert between) and there are no other quantities to convert to (like `Speed = Length / Duration`). We are simply trying to avoid bloat, because the number of measurable quantities is very large.

If you can think of additional units or quantities to convert to, that would help justify adding humidity as its own quantity type.

angularsen (changes requested):

angularsen:
@ilchdeveloper There was renewed interest in this quantity in #822 , I'm reopening the PR so it can be merged.

If you agree that it is useful to add the unit Fraction as unit, then it would be great if you could add that too before merging.

https://github.com/angularsen/UnitsNet/issues/822#issuecomment-737518802

Common/UnitDefinitions/Humitidy.json
krwq, line 2: Consider changing name to RelativeHumidity

Common/UnitDefinitions/Humitidy.json
```diff
@@ -0,0 +1,19 @@
...
+  "Units": [
+    {
+      "SingularName": "Percent",
+      "PluralName": "Percents",
+      "FromUnitToBaseFunc": "x",
+      "FromBaseToUnitFunc": "x",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "%Rh" ]
```
krwq, line 14: I believe h should be capital letter as well

krwq:
@ilchdeveloper let me know if you want me to redo this PR

ilchdeveloper:
@krwq go-ahead

angularsen:
Closed in favor of #863

PR #833, opened 2020-09-08
Fix overridden abbrev

### LINKED ISSUE #832: Overrides of unit abbreviations in non-default cultures causes Quantity.Parse to fail
**Describe the bug**
UnitsNet offers a public method that allows consumers of the NuGet package to override the default abbreviations used for different units.

For example, one can write

`UnitAbbreviationsCache.Default.MapUnitToDefaultAbbreviation(typeof(MassFlowUnit), (int)MassFlowUnit.GramPerSecond, CultureInfo.CurrentCulture, "grams/second")`

in order to use `"grams/second"` instead of `"g/s"` as the displayed abbreviation.

However, if the current culture is something other than en-US, some problems arise when attempting to call

`UnitsNet.Quantity.Parse(Type quantityType, string quantityString)`

after performing the above override.

If we had an override as in the example above, attempting to parse any mass flow `string` that doesn't use the `"grams/second"` abbreviation would fail in a non-English culture.

For example,

`Quantity.Parse(typeof(MassFlow), "10 lb/s")`

would fail if the current culture were de-DE.

**To Reproduce**
(1) Add an override similar to the one in the description, using a culture other than en-US.
(2) Call `Quantity.Parse` in the culture that was used specified in your override, and make this call on a quantity `string` that uses a unit abbreviation other than the one you specified in the override.
The parse should fail.
(Bug was found in UnitsNet 4.65.0 but hasn't been fixed to my knowledge.)

**Expected behavior**
Calls to `Quantity.Parse` that succeed without overrides to unit abbreviations should behave the same way with overrides.
Without overrides, UnitsNet falls back to using en-US for non-English cultures (in most cases). The override is causing UnitsNet to ignore the fallback abbreviations.

**Screenshots**
The problem appears to lie in how the following method in UnitAbbreviationsCache.cs is handled. A return value of `true` is given even when the lookup that ends up getting stored in `unitToAbbreviations` may not include any abbreviations other than ones that have been specified in an override. At other points in this file, a `true` return value from `TryGetUnitValueAbbreviationLookup` will cause UnitsNet to use the incomplete lookup and ignore the fallback culture.
image

**Additional context**
I encountered this bug when overriding unit abbreviations, but it would seem to apply also to cases where one simply adds new abbreviations as well.

tmilnthorp:
Fixes #832

### DIFF
--- a/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
+++ b/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
@@ -300,7 +300,7 @@ namespace UnitsNet.Tests
         public void MapUnitOverrideTest()
         {
             UnitAbbreviationsCache.Default.MapUnitToDefaultAbbreviation(MassFlowUnit.GramPerSecond, NorwegianCulture, "grams/second");
-            Assert.Equal(MassFlow.FromPoundsPerSecond(10), Quantity.Parse( NorwegianCulture, typeof( MassFlow ), "10 lb/s" ));
+            Assert.Equal(MassFlow.FromPoundsPerSecond(10), Quantity.Parse(NorwegianCulture, typeof(MassFlow), "10 lb/s"));
         }
 
         [Fact]
--- a/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
+++ b/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
@@ -297,10 +297,13 @@ namespace UnitsNet.Tests
         }
 
         [Fact]
-        public void MapUnitOverrideTest()
+        public void MapUnitToDefaultAbbreviation_ParsesUsingDefaultString()
         {
-            UnitAbbreviationsCache.Default.MapUnitToDefaultAbbreviation(MassFlowUnit.GramPerSecond, NorwegianCulture, "grams/second");
-            Assert.Equal(MassFlow.FromPoundsPerSecond(10), Quantity.Parse(NorwegianCulture, typeof(MassFlow), "10 lb/s"));
+            var cache = new UnitAbbreviationsCache();
+            cache.MapUnitToDefaultAbbreviation(MassFlowUnit.GramPerSecond, NorwegianCulture, "grams/second");
+
+            var parser = new QuantityParser(cache);
+            Assert.Equal(MassFlow.FromPoundsPerSecond(10), parser.Parse<MassFlow, MassFlowUnit>("10 lb/s", NorwegianCulture, MassFlow.From));
         }
 
         [Fact]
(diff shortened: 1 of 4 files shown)


UnitsNet/CustomCode/UnitAbbreviationsCache.cs
```diff
@@ -276,9 +276,11 @@ public string[] GetUnitAbbreviations(Type unitType, int unitValue, IFormatProvid
...
                 return formatProvider != FallbackCulture ? GetUnitAbbreviations(unitType, unitValue, FallbackCulture) : new string[] { };
 
-            var abbreviations = lookup!.GetAbbreviationsForUnit(unitValue);
-            if(abbreviations.Count == 0)
-                return formatProvider != FallbackCulture ? GetUnitAbbreviations(unitType, unitValue, FallbackCulture) : new string[] { };
+            var abbreviations = new HashSet<string>();
+            abbreviations.UnionWith(lookup!.GetAbbreviationsForUnit(unitValue));
+
+            if(formatProvider != FallbackCulture)
+                abbreviations.UnionWith(GetUnitAbbreviations(unitType, unitValue, FallbackCulture));
```
angularsen, line 283: At first glance, I'm not sure about this union.

If you call `GetUnitAbbreviations(typeof(LengthUnit), 1, russianCulture)` then from this public method signature, isn't it unexpected to also receive English abbreviations?

tmilnthorp: A good question :) Should it? Currently you can get English abbreviations from GetUnitAbbreviations, but only if no Russian ones exist.

The parser obviously will not fall back without it. It's either that, or the parser logic will need to include the fallback logic. I wanted to keep everything contained.

UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
```diff
@@ -296,6 +296,25 @@ public void MapUnitToDefaultAbbreviation_GivenCustomAbbreviation_SetsAbbreviatio
             Assert.Equal("1 m^2", Area.FromSquareMeters(1).ToString(newZealandCulture));
         }
 
+        [Fact]
+        public void MapUnitOverrideTest()
+        {
+            UnitAbbreviationsCache.Default.MapUnitToDefaultAbbreviation(MassFlowUnit.GramPerSecond, NorwegianCulture, "grams/second");
```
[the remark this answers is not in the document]
tmilnthorp: Fixed

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
angularsen: I'm really not sure. How about adding a new parameter `bool includeFallbackCulture = false` and call it with `true` from our own usages where we need to?

angularsen: I still think we should not do a union here indiscriminately. What do you think of my proposal of a new argument?

Issue #717, opened 2019-10-30
Question: Default equality precision


dschuermans:
Hey,

I'm wondering if there's anything done concerning default equality functionality...
From the docs i found that precision is being tested up to 1E-5.

However, when doing the following:

```Length unit1 = Length.FromMeters(10D)
Length unit2 = Length.FromFeet(10D * 3.28084D)

Assert.AreEqual(unit1, unit2)
Assert.IsTrue(unit1.Equals(unit2)
```

the test fails.

However:
`unit1.Equals(unit2, 1E-5, ComparisonType.Absolute)`
will return true.

The value in meters for unit2 is 10.00000032

Should we always use the overloaded Equals methods or is there a way we can configure the default Equals?

angularsen:
I believe we discussed using ComparisonType.Relative as the default one
with a sane value, but I don't recall if we found some use cases where that
would not work as expected or not.
If you want to look into overriding the default Equals(), start with
searching through old issues to see if it's been discussed before. If we
can find a behavior that works for a range of small and large quantities
and some unit tests on it, then I'm all for changing it.

Best,
Andreas

tmilnthorp:
I would argue that as a value type, Equals should work similar to how you would expect with say [Double.Equals](https://docs.microsoft.com/en-us/dotnet/api/system.double.equals?view=netframework-4.8) does not return true for anything other than absolute equality. 

Looking at our Equals(object) implementation, it checks of the object is the same quantity, converts it to the same unit, and checks for absolute equality. I would argue that as a value type, we shouldn't even do the conversion. It should check that the value and the unit are exactly the same.

lipchev:
I agree with @tmilnthorp - this is also how the GetHashCode function is currently implemented (I meant to open issue on the subject)- we could alternatively devise an implementation of GetHashCode that is consistent- by converting (and possibly rounding to tolerance- should this become the default equality contract) - but I think it should go into the UnitsNet.Comparison class rather than as the default implementation.
It is then up to the domain class that holds the measure units to re-define Equals/HashCode, should this be deemed appropriate: one example (from my experience) is the chemical solution- still a value object(no identity) but considered equal to another solution with the same component(s) and concentration(s) within a margin of error- specified by the domain experts.

angularsen:
[...]

PR #838, opened 2020-09-21
Add round-tripping Equality/IComparable contract generation

lipchev:
Addresses issues posed in #717 

- CompareTo implemented in terms of the two-way comparison between  'this.As(other)' & 'other.As(this)'
- Equals changed to CompareTo(other) == 0
- Comparison operators re-written in terms of CompareTo values
- GetHashCode generation modified such that it is consistent with the Equals function- i.e. not including the UnitType
- Added tests (QuantityComparisonTests) for the IComparable, Equality contracts, including the (previously failing) comparisons with problematic values such as 0.001


CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -792,7 +794,8 @@ public bool Equals({_quantity.Name} other, double tolerance, ComparisonType comp
         /// <returns>A hash code for the current {_quantity.Name}.</returns>
         public override int GetHashCode()
         {{
-            return new {{ QuantityType, Value, Unit }}.GetHashCode();
+            var roundedBaseValue = Math.Round(GetValueInBaseUnit(), 5);
+            return new {{ QuantityType, roundedBaseValue }}.GetHashCode();
```
angularsen, line 798: I don't fully grok this one.
I tried reading up on he last comments in #717, but I still don't quite see why this works and whether it is a good idea or if it has some edge-cases to it. Could you try to ELI5 ?

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -715,7 +715,9 @@ public int CompareTo(object obj)
         /// <inheritdoc />
         public int CompareTo({_quantity.Name} other)
         {{
-            return _value.CompareTo(other.GetValueAs(this.Unit));
+            var asFirstUnit = other.GetValueAs(this.Unit);
+            var asSecondUnit = GetValueAs(other.Unit);
+            return (_value.CompareTo(asFirstUnit) - other.Value.CompareTo(asSecondUnit)) / 2;
```
angularsen, line 720: I don't fully grok this one.
I tried reading up on he last comments in #717, but I still don't quite see why this works and whether it is a good idea or if it has some edge-cases to it. Could you try to ELI5 ?

I get that we are doing an average of comparing A and B in both A's unit and in B's unit.

What exactly does this solve? Is it when A's and B's units are really different and trying to represent either quantity in the other's unit results in big rounding errors? Why does averaging help?

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
angularsen, line 798: This deserves some comments explaining the rationale for the GetHashCode. It is not intuitive to me, had I not known about the discussion.

Why is `5` a good choice here?

UnitsNet.Tests/QuantityComparisonTests.cs
```diff
@@ -0,0 +1,114 @@
...
+    public class QuantityComparisonTests
+    {
+        [Fact]
+        public void CompareTo_WithSameQuantityInAnotherUnit_ReturnsZero()
+        {
+            var firstMass = Mass.FromGrams(0.001);
+            var secondMass = firstMass.ToUnit(MassUnit.Microgram);
+
+            Assert.Equal(0, firstMass.CompareTo(secondMass));
+            Assert.Equal(0, secondMass.CompareTo(firstMass));
```
angularsen, line 17: Thes resulted in `-1` and `0` before this PR, so that seems like an improvement.

UnitsNet.Tests/QuantityComparisonTests.cs
```diff
@@ -0,0 +1,114 @@
...
+        public void GetHasCode_WithSameQuantityInAnotherUnit_ReturnsSameValue()
+        {
+            var firstMass = Mass.FromGrams(0.001);
+            var secondMass = firstMass.ToUnit(MassUnit.Microgram);
+
+            Assert.Equal(firstMass.GetHashCode(), secondMass.GetHashCode());
+        }
+
+        [Fact]
+        public void GetHasCode_WithVeryCloseQuantities_ReturnsSameValue()
```
angularsen, line 60: Some typos on `GetHasCode`

UnitsNet.Tests/QuantityComparisonTests.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            var firstMass = Mass.FromGrams(0);
+            var secondMass = Mass.FromGrams(double.Epsilon);
+
+            Assert.Equal(firstMass.GetHashCode(), secondMass.GetHashCode());
```
angularsen, line 65: I believe you mentioned this earlier from the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=netcore-3.1#notes-to-inheritors)

> If two objects compare as equal, the GetHashCode() method for each object must return the same value. However, if two objects do not compare as equal, the GetHashCode() methods for the two objects do not have to return different values.

UnitsNet.Tests/QuantityComparisonTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            var thirdMass = Mass.FromGrams(0.001);
+            var fourthMass = thirdMass.ToUnit(MassUnit.Microgram);
+
+            var collection = new[] {thirdMass, firstMass, fourthMass, secondMass};
+            Array.Sort(collection);
+
+            Assert.Equal(firstMass, collection[0]);
+            Assert.Equal(secondMass, collection[1]);
+            Assert.Equal(thirdMass, collection[2]);
+            Assert.Equal(fourthMass, collection[3]);
```
angularsen, line 111: Is it co-incidental that third and fourth place are ordered this way, since they are equal?

angularsen: Oh, right, we are simply comparing equality so we don't actually know which value got inserted into 3rd or 4th position. We can't use `ReferenceEquals` either since these are struct values.

UnitsNet.Tests/QuantityComparisonTests.cs
```diff
@@ -0,0 +1,114 @@
+﻿using System;
+using System.Collections.Generic;
+using UnitsNet.Units;
+using Xunit;
+
+namespace UnitsNet.Tests
+{
+    public class QuantityComparisonTests
```
angularsen, line 8: The implementation made a whole lot more sense after I found these tests, thank you!
I still don't fully understand the implementation and whether there can be any issues with it, but I am way more comfortable after seeing the excellent work on these test cases 👍

angularsen:
I think this looks pretty good. Just some loose questions/comments above.

@tmilnthorp you were involved in #717 too, maybe take a look when you have the time?

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
tmilnthorp: QuantityType was added as you could have two different units (say Length and Area) with identical values. If they also had the same unit enum value (treated as int), they would have the same hash code.

angularsen: Is it problematic that different quantity types would share hash code though? They would still not be equal.
As long two equal values produce the same hash code, we should be fine. There is no requirement that in-equal values should produce different hash codes, as mentioned in https://github.com/angularsen/UnitsNet/pull/838#discussion_r492287743.

lipchev: Honestly- it's an over-estimation of the max rounding error that we could get by converting to base (it's so stated in the Precision chapter of the wiki).

UnitsNet.Tests/QuantityComparisonTests.cs
lipchev: I'm pretty sure it's in the order of insertion- but this might depend on the sorting algorithm.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
angularsen: https://github.com/angularsen/UnitsNet/issues/717#issuecomment-548425378
> HashSet may have been a better example - for set operations. It too supports an IEqualityComparer to be passed in.

So one impact would be that
```c#
var hs = new HashSet<IQuantity>();
hs.Add(Mass.FromKilograms(1));
hs.Add(Mass.FromKilograms(1));
hs.Add(Length.FromMeters(1));
hs.Dump();

/* Today
5HashSet<IQuantity> (2 items)
--
1 kg
1 m
*/

/* This PR would change it to this?
5HashSet<IQuantity> (2 items)
--
1 m
*/

```

UnitsNet.Tests/QuantityComparisonTests.cs
lipchev: Before considering this failure of the implementation of the IComparable interface I was actually leaning more towards the Equals contract that also included the current Unit (without doing any conversions) - however coming up with rounding-error-proof (hopefully) implementation of the CompareTo tipped the scales for me in the other direction. 
In any case- for the Equality contract both options are still on the table..

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
tmilnthorp: Ok I'm not sure what I'm reading but QuantityValue is still there hah. Ignore that.

Just for our own learning: it does work! I thought it would disallow it.
The difference is without the quantity type it has to call Equals to check if it's a real collision or not. Given how easy it is to avoid, without a significant hash cost, it should be included for performance.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
lipchev: Sure- so basically the HashCode is used in some collections for fast lookup of objects. It is part of the 'Equality Contract' - which states that 'Equals(x,y) => Equals(x.GetHashCode(), y.GetHashCode()))', yet at the same time the result of GetHashCode is not considered as uniquely identifying (it is possible to have the same HashCode for two objects for which Equals(x,y) == false). When creating hashing functions one tries to make the result as unique as possible- but not such that it violates the first condition- typically we take the result of combining the hashes of the properties that make up the Equals function. 
In practice all collections that use the hash function by definition compare both the result of the hash comparison, followed by a full Equals check- the first condition skipping 99.99% of the non-equal entities, while the second one ensures that we are not being tricked by a hash-collision.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -668,25 +668,25 @@ private void GenerateEqualityAndComparison()
...
         {{
-            return left.Value < right.GetValueAs(left.Unit);
+            return left.CompareTo(right) == -1;
         }}
 
         /// <summary>Returns true if greater than.</summary>
         public static bool operator >({_quantity.Name} left, {_quantity.Name} right)
         {{
-            return left.Value > right.GetValueAs(left.Unit);
+            return left.CompareTo(right) == 1;
```
tmilnthorp, line 689: I like this notation better

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
tmilnthorp: Not sure why we wouldn't want this compared to raw double comparison in equivalent units.

UnitsNet.Tests/QuantityComparisonTests.cs
tmilnthorp: They don't have to, but it can save a lot of equality checking.

```csharp
double v1 = 0.0;
double v2 = double.Epsilon;
		
Console.WriteLine(v1.GetHashCode());
Console.WriteLine(v2.GetHashCode());
```
```
0
1
```

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
lipchev: By the way (I didn't know that) but renaming any of the variables forming the anonymous class `new { QuantityType, roundedBaseValue }`, would produce a wholly different result from the GetHashCode that follows it :)

angularsen: > Ok I'm not sure what I'm reading but QuantityValue is still there hah. Ignore that.

LOL! I read it that way too! 😆

lipchev:
Hey guys, ignore this if you're busy- just pinging out to ask if there is something I should do to move this forward.

angularsen:
I tried to refresh my memory by looking back at the comments.
I think for me, the only think I'm still unsure about is the hard coded constant of `5`. Can we run into any issues by this choice?

https://github.com/angularsen/UnitsNet/pull/838#discussion_r492282925

> This deserves some comments explaining the rationale for the GetHashCode. It is not intuitive to me, had I not known about the discussion.
> Why is 5 a good choice here?

> Honestly- it's an over-estimation of the max rounding error that we could get by converting to base (it's so stated in the Precision chapter of the wiki)

lipchev:
I gave it some more thought tonight- and I no longer think this is a good idea. Not that it doesn't do what it was supposed to (`q == q.ToUnit(X)`)  but it just doesn't make any sense to actually use it- given the limiting constraint that `q.ToUnit(X).ToUnit(Y)` is not necessarily equal to `q` (the precision lost on the way is non-recoverable). Any method with a quantity as input cannot thus be allowed to make any sane assumptions of equality beyond "**both unit and value match**" (you'd just be _one more_ `ToUnit()` away from ruining your day).
I therefore propose that we revert to using the `GetValueInBaseUnit()` for the `Comparator` and the _strict Equals/HashCode equality contract_ (using both Unit & Value)

angularsen:
I agree it sounds like the safest option. I still don't feel this equality is all that useful though, for any non-trivial unit conversion there will be loss of conversion and two very similar quantities will still be considered in-equal. And even worse, it is an implementation detail where some quantities use decimal.

I think custom equality comparison methods adds more value than trying to fit the quantities into `IEquatable` and `IComparable`, but if we should offer an implementation I would vote for the most conservative/safest option.

lipchev:
The problem is that the "safest" option in this case {unit, value} would be quite a breaking change. Anyone who has ever used something like x == Mass.Zero or default(Mass) could be in trouble if 'x' does not come with the default unit (kg in this case).
Also there are some edge cases that might come as a surprise: like having X<=Y & X >= Y but X != Y.

angularsen:
Good point. Changing equality from base unit to value+unit would break things like Zero comparison, I hadn't thought of that. So. Hm.

Actually. I think we should honor this contract, not value+unit:
```
Length.FromCentimeters(0).Equals(Length.Zero)
```

For me, that is the most intuitive. I _know_, equality won't work anything trivial except `Length.FromCentimeters(100).Equals(Length.FromMeters(1))` etc, but I just don't see a better compromise given this very limited contract.

lipchev:
So, I've reverted the `Equals` method (same as before) and made the `GetHashCode()` return only the `QuantityType`'s hash-code- as such it would match the contract requirements. However, I'm pretty sure there is 'reflexivity' requirement that the Equals method does not currently satisfy (or at least not _always_). I've adjusted the mixed-unit tests such as to demonstrate the problematic behavior. 

As far as the Comparable interface- there I've kept the proposed scheme (that _is_ reflexive). I think this is the safer option, even if there could be cases where A == B && A.CompareTo(B) != 0, the inverse is always true, i.e. A != B => A.CompareTo(B) != 0.

All in all, I'm not terribly happy with the whole thing- it would have been so much neater with a decimal-like value representation. In any case, even if we kept the original implementation- it wouldn't hurt to grab the tests from this branch (and make a note about the limitations in the wiki).

PR #839, opened 2020-09-21
QuantityFormatterTests: Formatting the 'expected' values using the CurrentUICulture

lipchev:
- var expected = string.Format(CultureInfo.CurrentUICulture, ...) as per the defaultCulture of the [QuantityFormatter.Format](https://github.com/angularsen/UnitsNet/blob/ca373f75d51604f7e8918f2aafb5f2fc380b757e/UnitsNet/QuantityFormatter.cs#L72) method overload

PS This might have to get reverted after #795 is merged, but I suggest updating this so that it is consistent with the currently tested behavior (there are currently 17 tests failing on my machine)


angularsen:
Thanks! We'll revisit in #795.

angularsen:
[Release UnitsNet/4.71.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.71.0)

PR #842, opened 2020-09-22
Add Turbidity Unit

### LINKED ISSUE #825: Add Unit NTU (Turbidity)
Add new Quantity => Turbidity + Unit => [NTU]

New Quantity Turbidity with unit NTU

**Additional context**
I already implemented this in my fork and could build/generate custom code without issues.

But since NTU is not an SI unit, I would like to ask if I should make a PR for this, i.e. add the quantity to the library, instead of creating my own copy.

This is the json for `Turbidity`

```json
{
  "Name": "Turbidity",
  "BaseUnit": "NTU",
  "XmlDoc": "Turbidity is the cloudiness or haziness of a fluid caused by large numbers of individual particles that are generally invisible to the naked eye, similar to smoke in air. The measurement of turbidity is a key test of water quality.",
  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Turbidity",
  "BaseDimensions": {
    "M": 0,
    "T": 0,
    "I": 0
  },
  "Units": [
    {
      "SingularName": "NTU",
      "PluralName": "NTU",
      "FromUnitToBaseFunc": "x",
      "FromBaseToUnitFunc": "x",
      "Localization": [
        {
          "Culture": "en-US",
          "Abbreviations": ["NTU"]
        }
      ]
    }
  ]
}
```
My question is about `BaseDimensions`...Due NTU not being SI, there are no `BaseDimensions` for it. Can I do like this and set all to `0`?

If OK, then I would make a PR for this.

rgomez90:
Added Turbidity Unit and conversions

* NTU
* mg/l
* ppm

Fixes #825


rgomez90:
@angularsen I am having issues with the conversions tests.

I suppose I did sth wrong, but I tried to follow the wiki, by giving the conversion values a fixed value.

lipchev:
Hello,
I'm not at all an expert on the subject- but from what I gather- it is not really possible to relate NTU to concentrations. In fact I'm not really sure what concentration you are referring to here- as the Formazin mixture (I'm told) is made up of two different solutions. 
Do you have some other standard in mind?

UnitsNet.Tests/CustomCode/TurbidityTests.cs
```diff
@@ -0,0 +1,33 @@
...
+// Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
+
+using System;
+
+namespace UnitsNet.Tests.CustomCode
+{
+    public class TurbidityTests : TurbidityTestsBase
+    {
+        // Override properties in base class here
+        protected override double MiligramsPerLiterInOneNTU => 0.333333333;
```
angularsen, line 27: How do I verify this?

If I understand correctly, turbidity needs to refer to a specific liquid in order to convert to milligrams per liter or PPM?

https://sciencing.com/definition-automated-titration-5833327.html

angularsen (changes requested):

rgomez90:
@angularsen @lipchev I think you're right. I am also not a specialist in this type of unity...But yeah, it seems, the conversion depends on a calibration. 

So I suppose, we could add NTU without conversions?

angularsen:
Right, the only way to support these conversions is probably to add
something like `myTurbidity.ToMilligramsPerLiter(someReference)` or create
a wrapper type `new ReferenceTurbidity(Turbidity.FromNTU(100),
someReference).ToMilligramsPerLiter();` but you lose out on much of the
generated code benefits.

We had similar issues for relative/absolute/gauge pressure conversions.

Please see ReferencePressure wrapper type introduced in
https://github.com/angularsen/UnitsNet/pull/726

rgomez90:
@angularsen  I've modified the code and removed the conversions to ppm and mg/l.

I think we could merge it already just adding the new NTU unit.

For the conversion, I think we could go the `myTurbidity.ToMiligramsPerLiter(someReference)' way...

They will be just the 2 conversions, to ppm and to mg/l. So manually adding both methods and tests should not be a headache.

angularsen:
Ok, merged. Do you plan on adding the conversions later?

angularsen:
[Release UnitsNet/4.71.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.71.0)

PR #843, opened 2020-09-30
Add WarpingMomentOfInertia (m^6)

dschuermans:
We found a missing unit, which again is pretty tied against the field of structural engineering...

Its pretty much the same as AreaMomentOfInertia , except it's to the sixth instead of to the fourth.
So i copied AreaMomentOfInertia and adjusted the values.

There isn't much to go on about either on the internet in terms of conversion verification, so I double checked with one of our engineers to make sure the values are okay.
Warping moment of inertia is also known as the Warping constant


dschuermans:
@angularsen did I miss something? 
Why is it complaining about the code coverage dropping? 
I followed the guidelines on the wiki...

angularsen:
It's likely that the generated code has less code coverage than the sum of
generated + custom code, so new PRs adding new units with mostly generated
code will reduce the coverage.

dschuermans:
Alright, so there's nothing I need to do besides wait for the PR to be merged 👍 

Cheers!

angularsen:
Ok, this quantity is pretty obscure to me, but at least I find stuff on google and structural engineering as a domain is familiar enough. There are also multiple units to convert between here that makes this quantity more useful in my mind. 

Is there by chance other quantities this could be useful to convert to/from?

I can't really verify the conversions, but they seem reasonable.

angularsen:
[Release UnitsNet/4.72.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.72.0)

dschuermans:
> Ok, this quantity is pretty obscure to me, but at least I find stuff on google and structural engineering as a domain is familiar enough. There are also multiple units to convert between here that makes this quantity more useful in my mind.
> 
> Is there by chance other quantities this could be useful to convert to/from?
> 
> I can't really verify the conversions, but they seem reasonable.

It's pretty obscure to me as well, I'm not a structural engineer, I'm just a software engineer working in a company that provides structural engineering software 😄 

In any case, I can ask one of the engineers if it makes sense, being able to convert that one to other units.
If there are any use cases, I'll gladly create a new PR and add them.

Thanks for merging this one though, now I can continue my work 👍

PR #846, opened 2020-10-12
Add poundal-foot unit to Torque quantity

NickHeinz-Personal:
Add missing poundal-foot unit to torque quantity.

### DIFF
--- a/UnitsNet.Tests/CustomCode/TorqueTests.cs
+++ b/UnitsNet.Tests/CustomCode/TorqueTests.cs
@@ -49,6 +49,8 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double TonneForceMillimetersInOneNewtonMeter => 1.01972e-1;
 
+        protected override double PoundalFeetInOneNewtonMeter => 23.730360404231938;
+
         [Fact]
         public void TorqueDividedByForceEqualsLength()
         {
--- a/Common/UnitDefinitions/Torque.json
+++ b/Common/UnitDefinitions/Torque.json
@@ -52,6 +52,18 @@
       ]
     },
     {
+      "SingularName": "PoundalFoot",
+      "PluralName": "PoundalFeet",
+      "FromUnitToBaseFunc": "x*4.21401100938048e-2",
+      "FromBaseToUnitFunc": "x/4.21401100938048e-2",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "pdl·ft" ]
+        }
+      ]
+    },
+    {
       "SingularName": "PoundForceInch",
       "PluralName": "PoundForceInches",
       "FromUnitToBaseFunc": "x*1.129848290276167e-1",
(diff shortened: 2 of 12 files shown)


UnitsNet.Tests/CustomCode/TorqueTests.cs
```diff
@@ -49,6 +49,8 @@ public class TorqueTests : TorqueTestsBase
 
         protected override double TonneForceMillimetersInOneNewtonMeter => 1.01972e-1;
 
+        protected override double PoundalFeetInOneNewtonMeter => 23.730360404231938;
```
angularsen, line 52: Verified https://www.google.com/search?sxsrf=ALeKk021uyzz2lY7xp-4PkOig305bWestg%3A1602702638134&ei=Lk2HX5bcB4TxrgTvuaeACg&q=PoundalFoot+In+One+NewtonMeter&oq=PoundalFoot+In+One+NewtonMeter&gs_lcp=CgZwc3ktYWIQAzoECCMQJ1D9DljcMGCbMmgBcAB4AIABoAKIAdILkgEFMC44LjGYAQCgAQGqAQdnd3Mtd2l6wAEB&sclient=psy-ab&ved=0ahUKEwiWq7TL5LTsAhWEuIsKHe_cCaAQ4dUDCA0&uact=5

angularsen:
Is this equivalent to https://en.wikipedia.org/wiki/Pound-foot_(torque) ?

If so, should it be named PoundFoot?

angularsen:
It seems poundal foot is about pound-force applied over a linear distance of 1 foot.

https://en.wikipedia.org/wiki/Foot-poundal#:~:text=The%20foot%2Dpoundal%20(symbol%3A,commonly%20used%20foot%2Dpound%20force.

> The foot-poundal (symbol: ft-pdl) is a unit of energy that is part of the foot-pound-second system of units

http://convert-to.com/conversion/energy/convert-pdl-ft-to-ft-lb.html#:~:text=1%20poundal%20foot%20(pdl%20ft,0.031%20foot%20pounds%20(ft%20lb)

NickHeinz-Personal:
Yes, apparently a Foot-poundal is a unit of energy, but a Poundal-foot is a unit of torque.  Same way that there's some confusion between Pound-foot (torque) and Foot-pound (energy).  Doesn't help that commonly people use both names to refer to the torque (especially in the lbf-ft case).  [This is the site I grabbed the conversion from](http://www.kylesconverter.com/torque/foot--poundals-to-newton-meters) - note the name convention (force-length) here is wrong.

angularsen:
Thank you for the clarification. Including Google as a source of conversion factor, I'm happy.

angularsen:
What do you think of PoundFoot [wiki](https://en.wikipedia.org/wiki/Pound-foot_(torque)) vs PoundalFoot?

image
image

NickHeinz-Personal:
PoundFoot is different than PoundalFoot, it needs to be PoundalFoot.  A Pound (lbf) isn't the same as a Poundal (pdl).  1 pdl = 0.138254954376 N exactly ([wiki](https://en.wikipedia.org/wiki/Poundal)), while a Pound (force) is equal to 4.4482216152605 N.  So in other words a Poundal is about 1/32 of a Pound (lbf).  The background section of the Poundal wiki tries to clarify this a bit more, but it's still confusing.  Got to love Imperial units!!

angularsen:
Ahh, so that was what I was missing. Thanks. This should be ready to go then 👍

angularsen:
Nuget on the way out
[Release UnitsNet/4.73.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.73.0)

PR #857, opened 2020-11-13
Add prefixes to VolumeFlow.LitersPerSecond

ntpincus:
I'd like to add some standard unit support to volume flow.
nL/s, uL/s, mL/s, cL/s, dL/s and KL/s

The pumps we use are much closer to the nano and micro scale, but I tried to add enough units for other people's more general use.


ntpincus (commented):
Not sure why generate-code.bat says it's an empty file...
When I view it it has a script inside :thinking:

angularsen:
Hi, currently you can't remove/replace unit abbereviations - you can only add new ones.

See some examples here:
https://github.com/angularsen/UnitsNet/wiki/Extending-with-Custom-Units

If you need remove functionality and are willing to do a pull request, I think it could be added without too much trouble.

angularsen:
- `Tools/reportgenerator` binary file was added, you should probably remove this

angularsen (changes requested):
This looks good to me, just a few minor things.

Please revert these changes:
- `generate-code.bat`
- `Tools/reportgenerator`
- Changed sort order of `Molarity`, not sure why this was different on your system?

lipchev:
> 
> 
> Hi, currently you can't remove/replace unit abbereviations at runtime - you can only add new ones.
> 
> See some examples here:
> https://github.com/angularsen/UnitsNet/wiki/Extending-with-Custom-Units
> 
> If you need remove functionality and are willing to do a pull request, I think it could be added without too much trouble.

Either I'm misunderstanding the question or you are forgetting the [UnitAbbreviationsCache](https://github.com/angularsen/UnitsNet/blob/5719cb1c650e99dc46196eaa9934071f3e024b01/UnitsNet/CustomCode/UnitAbbreviationsCache.cs#L21)
Example:
`UnitAbbreviationsCache.MapUnitToAbbreviation(MassUnit.Microgram, "ug");`

angularsen:
@lipchev I probably read the question wrong.
@ntpincus Please see examples on using `MapUnitToAbbreviation` in https://github.com/angularsen/UnitsNet/wiki/Extending-with-Custom-Units

angularsen:
Nuget is on the way out
[Release UnitsNet/4.74.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.74.0)

PR #863, opened 2020-12-03
Add RelativeHumidity unit

krwq:
Fixes: https://github.com/angularsen/UnitsNet/issues/822
Replaces PR: https://github.com/angularsen/UnitsNet/pull/824

cc: @angularsen @ilchdeveloper


Common/UnitDefinitions/RelativeHumidity.json
```diff
@@ -0,0 +1,19 @@
+{
+  "Name": "RelativeHumidity",
+  "BaseUnit": "Percent",
+  "XmlDoc": "Relative humidity is a ratio of the actual water vapor present in the air to the maximum water vapor in the air at the given temeperature.",
+  "Units": [
+    {
+      "SingularName": "Percent",
```
angularsen, line 7: You mentioned adding Fraction. Will you not be adding that?

angularsen (approved):

Common/UnitDefinitions/RelativeHumidity.json
krwq: I've considered that initially but I'm not 100% it will be that useful. I think might make sense to wait until people start complaining about lack of it. ("0.3 RH" didn't make much sense to me so it would probably be just confusing if someone tries to serialize it)

krwq:
@angularsen is there any work left here for me?

lipchev:
UnitsNet/4.77.0 should be on the way. 
Thank you for your contribution 👍 (and sorry for the delay)

Issue #858, opened 2020-11-22
QuantityType is not extensible


generateui:
I'm currently adding some quantities with units for a specific domain. I miss a few things which I ideally like to see, most of these are convenience features which I can happily work around.

I have added `.json` files with specification into my own project. In there, I have defined quantities. I then wrote a script to copy these json files to the `Common/UnitDefinitions` folder of UnitsNet project. I subsequently ran the CodeGen assembly.

Changes made to the UnitsNet project make mostly sense. 

One thing which does not make sense to me is the use of QuantityType as enum. I can't just add a QuantityType member - enums are not extensible. Note that [the guide](https://github.com/angularsen/UnitsNet/wiki/Extending-with-Custom-Units) refers to [HowMuch](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet.Tests/CustomQuantities/HowMuch.cs) quantity which [returns undefined](https://github.com/angularsen/UnitsNet/blob/c308026489eed508cb5a48b35abf8fc16ea76ae7/UnitsNet.Tests/CustomQuantities/HowMuch.cs#L26).

An alternative would be to make `QuantityType` an interface `IQuantityType` with static instances. 

1. To be sure: am I expected to return `QuantityType.Undefined` like `HowMuch`?
2. Can we make QuantityType extensible in some way?

angularsen:
Hi and thanks for the detailed description.

Custom units is the least mature part of this library, and as you found out the enum is not extensible.
I think your suggestion of using classes is a good one.

If you are interested in attempting a pull request then I'd be happy to assist!

generateui:
I'd like to, but do not have the time currently. I need to dive in it for a day or so to have a good PR. It'd be good to discuss the design upfront, that may save some effort when prototyping/implementing.

I see the `System.Enum` type is used as surrogate supertype for Unit enums. Is this done for performance? 

To represent QuantityType, I'd use an interface. Then have a class implement it with static instances. Do you avoid trying to instantiate static instances for performance? 

I do not see performance listed as a specific feature. If it is not a feature, why not simply merge `{Name}Type` and `{Name}` for both Quantity and Unit into a class? A wrapper for QuantityValue seems logical.

angularsen:
> I see the System.Enum type is used as surrogate supertype for Unit enums. Is this done for performance?
> Do you avoid trying to instantiate static instances for performance?

[...]

Issue #815, opened 2020-07-11
Drop support for Windows Runtime Component


angularsen:
I just wanted to bring this up again. 
We added support for this in 2016 with #150, mainly because it was requested by a Microsoft team working on providing IoT nuget packages.

It is clear that this effort has stopped and the only two dependencies I can track via
[fuget.org](https://www.fuget.org/packages/UnitsNet.WindowsRuntimeComponent) are:
- https://www.fuget.org/packages/Microsoft.IoT.DeviceCore
- https://www.fuget.org/packages/Zen.IoT.Devices

Both of these stopped getting updates in 2017.

We visited this topic in 2018 in #576 and the Microsoft member that originally requested this feature pretty much confirmed that the Windows Runtime Component effort is dead in its tracks on Microsoft's side of things.

Nothing new has happened since then really, it's just been nagging at the back of my mind and I don't like having crud if it is not actually being used by anyone or it's like 5 people in the world who care.

### Proposal
- Deprecate the nuget package as no longer being maintained with a notice to contact us
- Stop publishing WRC nugets as part of our normal workflow
- Listen for a few months if anyone complains, then reconsider
- If no one complains, then remove all WRC code and build scripts and stop publishing new nugets

Pros:
- Simplifies UnitsNet code base and build pipeline, the wtf-factor goes down a few notches.

Cons:
- The [nuget is still being downloaded](https://www.nuget.org/packages/UnitsNet.WindowsRuntimeComponent) although it is hard to distinguish robots from actual usage.
- We made an effort to make WRC codegen its own separate thing, so it does not really cost us much to keep it around and update it with new units.

angularsen:
Thank you for your opinion on this.

I 66% agree with you 😄 there is a cost however, in on-boarding new contributors having to know what WRC is.

I'm perfectly open to keep it, but I do think it should be removed at some point in the future if we can determine that pretty much no one relies on it anymore. My hunch is that very, very few use this already today.

I can't say for sure that no one is using it, which is why I created the issue hoping someone would notice and chime in saying they were actually using this.

angularsen:
A couple new sources of usage since last time:
- [GitHub search - only forks](https://github.com/search?q=unitsnet.windowsruntimecomponent&type=Code)
- [nuget.org](https://www.nuget.org/packages/UnitsNet.WindowsRuntimeComponent) - 0 listed dependents, 72 downloads of latest release 20 days ago
[...]

PR #864, opened 2020-12-04
Deprecate QuantityType to better support custom quantities

generateui:
See #858. Turns out, QuantityInfo is already sufficient to represent a type of quantity. Hence, the solution is simple: remove QuantityType.

Tests pass, but I have not done a thorough check to actually merge the PR. Let's discuss first if this solution is desirable. To me, it solves the problem - QuantityType is gone, extensibility on QuantityType is now irrelevant.

Furthermore, users of all quantities can now use `Quantity.ByName` to enumerate all quantities.

Please let me know if this is the direction this library want to go into. If so, I'll publish follow up changes to ready up for a merge.


UnitsNet.Tests/GeneratedCode/IQuantityTests.g.cs
```diff
@@ -252,111 +252,111 @@ public void QuantityInfo_IsSameAsStaticInfoProperty()
         [Fact]
         public void Type_EqualsStaticQuantityTypeProperty()
         {
-            void Assertion(QuantityType expected, IQuantity quantity) => Assert.Equal(expected, quantity.Type);
```
angularsen, line 255: I believe we should duplicate the assertions to cover both old and new functionality, until removed. Here and other places.

UnitsNet/QuantityTypeConverter.cs
```diff
@@ -146,12 +146,13 @@ public override bool CanConvertFrom(ITypeDescriptorContext context, Type sourceT
             TAttribute? attribute = (TAttribute)context.PropertyDescriptor.Attributes[typeof(TAttribute)];
             if (attribute != null)
             {
-                QuantityType expected = default(TQuantity).Type;
-                QuantityType actual = QuantityType.Undefined;
+                // TODO
```
angularsen, line 149: TODO to address

angularsen (changes requested):
I think you made an excellent observation that we can reuse `QuantityInfo` instead of `QuantityType` for enumerating quantities, looking up one statically via `Length.Info` or dynamically via `Quantity.Infos`.

As in my comments, I propose we revert to keep old functionality and keep it backwards compatible, but mark it obsolete. Unit tests should test both old and new, until removed. 

Then update #180 with what to remove and cleanup when we bump the major version.
There is a bunch of other stuff waiting there already.

What I didn't immediately see, is how do we achieve more extensibility than before?

[Wiki](https://github.com/angularsen/UnitsNet/wiki/Extending-with-Custom-Units) describes how to extend with custom units, for converting or parsing, but it doesn't describe how to add custom quantities. One specific thing I believe is missing is how to add `HowMuch` to `Quantity.Infos`, which is currently an array. I can't immediately think of other scenarios where extensibility is lacking, but I'm sure there is something.

I'm totally fine with attacking that point later, just wanted to bring it up to hear if you had any thoughts of it.

generateui:
Right, I forgot UnitsNet supports graceful removal of features. Obsoleting first makes total sense. Agreed that only when actual removal is done, tests should be removed.

Would you please update #180, when the time is there?

> What I didn't immediately see, is how do we achieve more extensibility than before?

We don't. We just remove a blocker on extensibility by changing the fact that the library relies on QuantityType. The PR effectively moves QuantityType to a role where it is not required to use any more, which is needed for extended quantities. Extended quantities can't provide a correct QuantityType value, and as such using api's where QuantityType is used will fail. That is what this PR fixes.

I have opened #865 to discuss extensibility features in a tracking issue.

generateui:
> I will update the PR to support obsoletion and initial addition of the QuantityInfo apis.

Done. Could you re-review? Thanks!

UnitsNet.Tests/CustomQuantities/HowMuch.cs
```diff
@@ -26,7 +26,7 @@ public HowMuch(double value, HowMuchUnit unit)
         public QuantityType Type => QuantityType.Undefined;
         public BaseDimensions Dimensions => BaseDimensions.Dimensionless;
 
-        public QuantityInfo QuantityInfo => new QuantityInfo(Type,
+        public QuantityInfo QuantityInfo => new QuantityInfo("HowMuch",
```
angularsen, line 29: This is a good improvement on its own, third party quantities get an actual name instead of Undefined quantity type.

UnitsNet.Tests/GeneratedCode/IQuantityTests.g.cs
```diff
@@ -252,111 +252,111 @@ public void QuantityInfo_IsSameAsStaticInfoProperty()
         [Fact]
         public void Type_EqualsStaticQuantityTypeProperty()
```
angularsen, line 253: Duplicate this Fact method and name it accordingly, in order to keep the old assertion on QuantityType property, and add the new assertion on Info property.

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
```diff
@@ -127,7 +127,6 @@ public void Acceleration_QuantityInfo_ReturnsQuantityInfoDescribingQuantity()
 
             Assert.Equal(Acceleration.Zero, quantityInfo.Zero);
             Assert.Equal("Acceleration", quantityInfo.Name);
-            Assert.Equal(QuantityType.Acceleration, quantityInfo.QuantityType);
```
angularsen, line 130: Keep this

UnitsNet.Tests/QuantityInfoTest.cs
```diff
@@ -57,8 +60,8 @@ public void GenericsConstructor_AssignsProperties()
             Assert.Equal(expectedZero, info.Zero);
             Assert.Equal("Length", info.Name);
             Assert.Equal(expectedUnitInfos, info.UnitInfos);
-            Assert.Equal(expectedQuantityType, info.QuantityType);
```
angularsen, line 60: Keep

UnitsNet.Tests/QuantityTest.cs
```diff
@@ -42,68 +43,6 @@ public void From_GivenValueAndUnit_ReturnsQuantity()
             Assert.Equal(Pressure.FromMegabars(3), Quantity.From(3, PressureUnit.Megabar));
         }
 
-        [Fact]
```
angularsen, line 45: Keep these?

UnitsNet.WindowsRuntimeComponent/GeneratedCode/Quantity.g.cs
```diff
@@ -30,6 +31,665 @@ namespace UnitsNet
     /// </summary>
     internal static partial class Quantity
     {
+        public static readonly IDictionary<string, QuantityInfo> ByName = new Dictionary<string, QuantityInfo>
```
angularsen, line 34: This is not obvious, but UnitsNet.WindowsRuntimeComponent is deprecated: #815 

I see the [AppVeyor build failed](https://ci.appveyor.com/project/angularsen/unitsnet/builds/36676346), so it is probably best to revert and not touch anything under that namespace in order to avoid problems. WRC is currently receiving new quantities and units via changes to JSON files, but we stopped maintaining its code generator some time ago.

UnitsNet/CustomCode/Quantity.cs
```diff
@@ -13,23 +13,16 @@ public partial class Quantity
...
-                .Select(quantityType => FromQuantityType(quantityType, 0.0).QuantityInfo)
+            InfosLazy = new Lazy<QuantityInfo[]>(() => quantityTypes
                 .OrderBy(quantityInfo => quantityInfo.Name)
                 .ToArray());
         }
 
         /// <summary>
-        /// All enum values of <see cref="QuantityType"/>, such as <see cref="QuantityType.Length"/> and <see cref="QuantityType.Mass"/>.
-        /// </summary>
-        public static QuantityType[] Types { get; }
```
angularsen, line 29: Keep

UnitsNet/CustomCode/Quantity.cs
```diff
@@ -94,16 +87,6 @@ public static IQuantity Parse(IFormatProvider? formatProvider, Type quantityType
         public static bool TryParse(Type quantityType, string quantityString, out IQuantity? quantity) =>
             TryParse(null, quantityType, quantityString, out quantity);
 
-        /// <summary>
-        ///     Get information about the given quantity type.
-        /// </summary>
-        /// <param name="quantityType">The quantity type enum value.</param>
-        /// <returns>Information about the quantity and its units.</returns>
-        public static QuantityInfo GetInfo(QuantityType quantityType)
```
angularsen, line 102: Keep

UnitsNet/IQuantity.cs
```diff
@@ -16,6 +16,7 @@ public interface IQuantity : IFormattable
         /// <summary>
         ///     The <see cref="QuantityType" /> of this quantity.
         /// </summary>
+        [Obsolete("QuantityType will be removed in the future. Use QuantityInfo instead.")]
```
angularsen, line 19: I see the [AppVeyor build failed](https://ci.appveyor.com/project/angularsen/unitsnet/builds/36676346) on obsolete errors. We probably need to add this to `UnitsNet.Tests.csproj`

```
<WarningsNotAsErrors>612,618</WarningsNotAsErrors>
```

Even better, put it in `Directory.Build.props` so it affects all projects that have `<MSBuildTreatWarningsAsErrors>true</MSBuildTreatWarningsAsErrors>` in them.

Or use pragma statements as you already have some places.

UnitsNet/QuantityTypeConverter.cs
angularsen: This still need to be fixed.

angularsen (changes requested):
Good progress, some findings below.

UnitsNet.Tests/CustomQuantities/HowMuch.cs
generateui: 👍

UnitsNet/QuantityTypeConverter.cs
generateui: close this duplicate feedback item (see below)

UnitsNet.Tests/GeneratedCode/IQuantityTests.g.cs
generateui: done?

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
generateui: done

UnitsNet.Tests/QuantityTest.cs
generateui: done

UnitsNet.WindowsRuntimeComponent/GeneratedCode/Quantity.g.cs
generateui: done, reverted

UnitsNet/CustomCode/Quantity.cs
generateui: reinstated

UnitsNet/CustomCode/Quantity.cs
generateui: reinstated

UnitsNet/IQuantity.cs
generateui: done, works like a charm. Restart Visual Studio though.

UnitsNet.Tests/QuantityInfoTest.cs
generateui: done

UnitsNet.Tests/GeneratedCode/IQuantityTests.g.cs
angularsen: Maybe I am getting lost in the diffs here, but it seems there are two methods in this test class:
- QuantityInfo_IsSameAsStaticInfoProperty
- Type_EqualsStaticQuantityTypeProperty

I believe this PR incorrectly changed `Type_EqualsStaticQuantityTypeProperty` to be an exact copy of `QuantityInfo_IsSameAsStaticInfoProperty`, they both test the same thing. I think this file should simply be reverted.

angularsen: My bad, I did not see diff for latest commit. Already reverted.

UnitsNet/QuantityTypeConverter.cs
angularsen: Maybe tomorrow

angularsen:
Seems like tests fail on `NullReferenceException` in the build:

https://ci.appveyor.com/project/angularsen/unitsnet/builds/36677657

```
Failed UnitsNet.Tests.CustomCode.ThermalConductivityTests.ToString_ReturnsValueAndUnitAbbreviationInCurrentCulture [1 ms]
  Error Message:
   System.TypeInitializationException : The type initializer for 'UnitsNet.ThermalConductivity' threw an exception.
---- System.TypeInitializationException : The type initializer for 'UnitsNet.Quantity' threw an exception.
-------- System.TypeInitializationException : The type initializer for 'UnitsNet.Acceleration' threw an exception.
------------ System.NullReferenceException : Object reference not set to an instance of an object.
  Stack Trace:
     at UnitsNet.ThermalConductivity..ctor(Double value, ThermalConductivityUnit unit) in C:\projects\unitsnet\UnitsNet\GeneratedCode\Quantities\ThermalConductivity.g.cs:line 71
   at UnitsNet.Tests.ThermalConductivityTestsBase.ToString_ReturnsValueAndUnitAbbreviationInCurrentCulture() in C:\projects\unitsnet\UnitsNet.Tests\GeneratedCode\TestsBase\ThermalConductivityTestsBase.g.cs:line 360
----- Inner Stack Trace -----
   at UnitsNet.QuantityInfo..ctor(String name, UnitInfo[] unitInfos, Enum baseUnit, IQuantity zero, BaseDimensions baseDimensions) in C:\projects\unitsnet\UnitsNet\QuantityInfo.cs:line 99
   at UnitsNet.QuantityInfo`1..ctor(String name, UnitInfo`1[] unitInfos, TUnit baseUnit, IQuantity`1 zero, BaseDimensions baseDimensions) in C:\projects\unitsnet\UnitsNet\QuantityInfo.cs:line 238
   at UnitsNet.ThermalConductivity..cctor() in C:\projects\unitsnet\UnitsNet\GeneratedCode\Quantities\ThermalConductivity.g.cs:line 56
----- Inner Stack Trace -----
   at UnitsNet.Acceleration.FromMetersPerSecondSquared(QuantityValue meterspersecondsquared) in C:\projects\unitsnet\UnitsNet\GeneratedCode\Quantities\Acceleration.g.cs:line 358
   at UnitsNet.Tests.DecimalOverloadTests.CreatingQuantityWithDoubleBackingFieldFromDecimalReturnsCorrectValue() in C:\projects\unitsnet\UnitsNet.Tests\DecimalOverloadTests.cs:line 13
----- Inner Stack Trace -----
   at UnitsNet.QuantityInfo..ctor(String name, UnitInfo[] unitInfos, Enum baseUnit, IQuantity zero, BaseDimensions baseDimensions) in C:\projects\unitsnet\UnitsNet\QuantityInfo.cs:line 99
   at UnitsNet.QuantityInfo`1..ctor(String name, UnitInfo`1[] unitInfos, TUnit baseUnit, IQuantity`1 zero, BaseDimensions baseDimensions) in C:\projects\unitsnet\UnitsNet\QuantityInfo.cs:line 238
   at UnitsNet.Acceleration..cctor() in C:\projects\unitsnet\UnitsNet\GeneratedCode\Quantities\Acceleration.g.cs:line 53
  Failed UnitsNet.Tests.CustomCode.ThermalConductivityTests.ConversionRoundTrip [1 ms]
  Error Message:
```

generateui:
I have fixed some warnings and fought a little fight with Visual Studio, resulting in #pragma littering. That seems to be fixed now (I still wonder why previous week the build succeeded and now a reboot and restart reports all these warnings as errors. The ways of VS move in mysterious ways).

I can't reproduce the error. Tracing it back points to 
```csharp
            if (Quantity.QuantityTypeByName.TryGetValue(name, out var quantityType))
                QuantityType = quantityType;
            else
                QuantityType = QuantityType.Undefined;
```

in the ctor of `public QuantityInfo([NotNull] string name, [NotNull] UnitInfo[] unitInfos, [NotNull] Enum baseUnit, [NotNull] IQuantity zero, [NotNull] BaseDimensions baseDimensions)`. Yet, this dictionary is initialized and all tests pass locally. Can you take a look?

angularsen:
I pushed some fixes on a cyclic static ctor call, and fixed some test cases.
It builds and all tests pass locally with `build.bat` now.

angularsen:
AppVeyor is happy, I'll try to re-review this soon.

generateui:
How come the cyclic ctor call did not explode on my machine? I ran all tests.

angularsen:
Try running `build.bat`. That excludes any quirks by Visual Studio and Rider, using dotnet CLI to run tests.

One thing is that the order of tests. If any test in the same app domain (test project) called `Quantity` type, then Quantity static ctor would successfully init.
If a test called `Temperature.Info` first, then it would cause problems.

generateui:
Gotcha, thanks. Is there anything you want me to do to move forward on this PR?

angularsen:
No, I just need to read through the whole PR again. Maybe tomorrow.

angularsen:
@generateui I think this is ready to merge! Just waiting for AppVeyor to do its thing. It has been awfully slow lately.

angularsen:
I was able to sort out the pragma stuff, see #872 .

angularsen:
Quantity.g.cs had 30% lower test coverage, but we can sort that out later. I'd rather merge this beast now :D 

If I read it right, the method is not covered by tests:
`FromQuantityInfo`

https://codecov.io/gh/angularsen/UnitsNet/pull/864/diff?src=pr&el=tree#diff-VW5pdHNOZXQvR2VuZXJhdGVkQ29kZS9RdWFudGl0eS5nLmNz

angularsen:
Nuget is on the way out, thank you for the contribution!

[Release UnitsNet/4.79.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.79.0)

generateui:
This is awesome! Thanks for your close support.

PR #867, opened 2020-12-15
Add chinese abbreviations for area,length,mass units

chinasqzl:
Add chinese abbreviations for area,length,mass units


Common/UnitDefinitions/Length.json
```diff
@@ -23,6 +23,10 @@
         {
           "Culture": "ru-RU",
           "Abbreviations": [ "м" ]
+        },
+        {
+          "Culture": "zh-CN",
+          "Abbreviations": [ "米" ]
```
angularsen, line 29: Can you please verify that the prefixes work as expected?
For example, 

- Does 毫米 read correctly as millimeter? 
- Does 千克 read correctly as kilogram?
- Does 兆吨 read correctly as megatonne?

Same for all the other units where "Prefixes" are used.
Google seems to agree, but it's best you verify.

If there are any issues with some units, where prepending the prefix does not read correctly, you can customize the abbreviation for a particular unit and prefix with "AbbreviationsForPrefixes", see examples here:

https://github.com/angularsen/UnitsNet/search?l=JSON&q=AbbreviationsForPrefixes

angularsen (approved):
This looks good to me, I just need you to verify the prefixes work as intended.

Update:
You need to run `generate-code.bat` to generate all the new unit abbreviations.

Common/UnitDefinitions/Length.json
angularsen: You need to run generate-code.bat to generate all the new unit abbreviations.

chinasqzl: image

angularsen: I take it you have reviewed these abbreviations then 👍

angularsen:
Nuget is on the way out, thank you for the contribution!

[Release UnitsNet/4.79.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.79.0)

angularsen:
@chinasqzl Hi, there is some recent discussion on reverting `厘米` back to `cm` and instead localizing `厘米` for the unit enum name "Centimeter".

Assuming you know Chinese, could you please help determine what is correct?

#1171 Localization related Issues

Issue #285, opened 2017-09-08
Proposal: Alternative nuget with decimal as internal representation


angularsen:
### Proposal
* New nugets: `UnitsNet.Decimal`, `UnitsNet.Decimal.Signed`, `UnitsNet.Serialization.JsonNet.Decimal`, `UnitsNet.Serialization.JsonNet.Decimal.Signed`
* Change internal representation from `double` to `decimal`

### Why
Decimal has some benefits, such as fixed precision (predictable rounding) and high precision.
However, duplicating all types would nearly double the nuget size, so we propose to create a new nuget using `decimal` as internal representation as well as the output type.

Example:
```
double meters = Length.FromMeters(10).Meters; // currently
decimal meters = Length.FromMeters(10).Meters; // new nuget
```

### Notes
`UnitsNet.WindowsRuntimeComponent` nuget is left out due to [decimal not supported by type constraints](https://msdn.microsoft.com/en-us/library/windows/apps/xaml/br230301.aspx?f=255&MSPPError=-2147217396#PassingToManaged).

ferittuncer:
I think we can use generics to support decimal internal representation without doubling source-code size. 

Perhaps with a usage like this:

`double meters = Length<double>.FromMeters(10).Meters;` // double internal representation
`decimal meters = Length<decimal>.FromMeters(10).Meters;` // decimal internal representation

angularsen:
We could go this route, but I don't want to force usage to include the `<double>` part all over the place. Unfortunately there are many cases when it can't infer the type. Often when numbers are coded, they are integers and that would have to be explicitly specified as `1d` or `1m`. The nullable overloads are also particularly bad at type inference.

Even if we accepted this, it is not trivial to work with generic types internally. You can't readily multiply or add two generic numbers together. You either need to check the type and handle the calculation separately (which does not fix the size problem), or use some arithmetic abstraction code such as [this](https://www.codeproject.com/Articles/33617/Arithmetic-in-Generic-Classes-in-C). But.. it adds overhead and feels complicated. I'm not a fan, but it might be a way do it and maybe it works better than my prejudice tells me.

I've used Math.NET and it has the same thing with matrices, that can be either double, decimal, long and possibly more. They solved it by having both a `Matrix<double>` type as well as a separate "double" namespace with `Matrix`, `Vector` and other types written assuming `double`.

It worked quite well, but the same approach would cost us almost N times the nuget size.

So.. the way I see it we have these options:
[...]

PR #868, opened 2020-12-15
Fix JSON serialization of quantities with decimal values

rohahn:
Add IDecimalQuantity interface to expose the decimal value
Serialize decimal values as string to keep number of decimal places

### DIFF
--- a/UnitsNet.Serialization.JsonNet/UnitsNetBaseJsonConverter.cs
+++ b/UnitsNet.Serialization.JsonNet/UnitsNetBaseJsonConverter.cs
@@ -2,6 +2,7 @@
 // Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
 
 using System;
+using System.Globalization;
 using System.Linq;
 using JetBrains.Annotations;
 using Newtonsoft.Json;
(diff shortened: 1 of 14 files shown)


angularsen:
Solid work on this 👏 Love the tests!

I think it is clever to use string to support `decimal`, but I'm concerned we are breaking backwards compatibility on the JSON schema for those expecting to read `Value` as a number.

How about this:
- Keep `Unit` and `Value` exactly as before, the `double` representation of the number matching our common public interface for quantities.
- If value is not `double`, then add two more fields
  - `ValueString` with textual representation of number
  - `ValueType` with string values like `"decimal"`, `"float"`, `"SomeLibrary.BigInt"`

This makes the schema more extensible to other types of numbers with generics, as have been discussed a number of times:
#180
#285
#666
#714

rohahn:
I implemented your proposal.

UnitsNet.Serialization.JsonNet/UnitsNetBaseJsonConverter.cs
```diff
@@ -53,17 +73,28 @@ protected ValueUnit ReadValueUnit(JToken jsonToken)
...
-            if (valueUnit == null || string.IsNullOrWhiteSpace(valueUnit.Unit))
+            if (string.IsNullOrWhiteSpace(valueUnit?.Unit))
             {
                 return null;
             }
 
-            var unitParts = valueUnit.Unit.Split('.');
+            var unit = GetUnit(valueUnit.Unit);
+
+            return valueUnit switch
```
angularsen, line 83: Nice

angularsen (approved):
Looks perfect to me, great work on this!

angularsen:
Nuget is on the way out, I'd appreciate it if you gave it a spin to ensure it works as expected and comment back here! 🙇 
[Release JsonNet/4.3.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/JsonNet%2F4.3.0)

rohahn:
Why is the version number 4.3.0?

angularsen:
JsonNet nuget is versioned separately from UnitsNet nuget, but they share major version.

Issue #879, opened 2020-12-26
Oktas


MarkCiliaVincenti:
I want to add oktas to UnitsNet. The problem is that they're usually expressed in integers. What approach do you recommend? I was thinking internally saving a Ratio and then express in octa integers when converted.

https://en.wikipedia.org/wiki/Okta

ebfortin:
An Okta is always an integer value? You can't have fraction of an Okta? If that's the case then use the current backing store with a max of 9 and a min of 9 and always round the value at creation.

MarkCiliaVincenti:
> An Okta is always an integer value? You can't have fraction of an Okta? If that's the case then use the current backing store with a max of 9 and a min of 9 and always round the value at creation.


But what if someone does an Okta.FromPercent and gets an instance of Okta which is saving an integer okta value, and the developer wants to convert it back to the original percentage value? This is what I mean.

angularsen:
Hi, how about something like this:

- Add quantity `CloudCover` with units `Okta` [0..8], `Percent` [0..100] and `Fraction` [0..1] and the latter as the base unit (`double` numeric value).
- Okta `FromUnitToBaseFunc`: `x * 0.125`
  - 0 Okta => 0 fractions
  - 1 Okta => 0.125 fractions
  - 2 Okta => 0.25 fractions
  - etc..
- Okta `FromBaseToUnitFunc`: `System.Math.Round(x / 0.125)`
  - 0 fractions => 0 Okta
  - 0.15 fractions => 1.2 => 1 Okta
  - 0.2 fractions => 1.6 => 2 Okta
  - etc

This way it rounds to the nearest Okta integer, while quantity supports values like 35% cloud cover or fractions.

Source: https://en.wikipedia.org/wiki/Cloud_cover

MarkCiliaVincenti:
Brilliant, I have this CloudCover.json so far, not sure what else I need. Some notes:

1) I used `System.Math.Round(x / 0.125, MidpointRounding.AwayFromZero)` as I prefer it to the default rounding method
2) What about unit tests?
3) How do we define the ranges [0..8] etc?

````
{
  "Name": "CloudCover",
  "BaseUnit": "Fraction",
  "XmlDoc": "Cloud cover (also known as cloudiness, cloudage, or cloud amount) refers to the fraction of the sky obscured by clouds when observed from a particular location.",
  "Units": [
    {
      "SingularName": "Okta",
      "PluralName": "Oktas",
[...]

PR #880, opened 2020-12-29
Rename DecimalFraction units to Decimal

angularsen:
Decimal fraction is not a correct term.

> Definition of Decimal Fraction
> A fraction where the denominator (the bottom number) is a power of ten (such as 10, 100, 1000, etc).
[Source](https://www.mathsisfun.com/definitions/decimal-fraction.html).

> So, to convert a fraction to a decimal, divide the numerator by the denominator.
[Source](bbc.co.uk/bitesize/topics/znmtsbk/articles/z4ymtv4).

### Changes
- Add `Ratio.Decimal` unit, mark `DecimalFraction` unit as obsolete
- Add `RatioChangeRate.DecimalPerSecond`, mark `DecimalFractionPerSecond` unit as obsolete
- Change base unit of each quantity to the new unit

Changing the base unit should not be a breaking change for most users and the conversion precision is exactly the same as before for all units.


angularsen:
Related to #879

lipchev:
I'm not sure about this- the linked definition emphasizes the distinction between 1/2 (fraction) and 5/10 (decimal fraction). However what we're representing is the right hand side of the equation: e.g. 0.53 which is the [decimal](https://www.mathsisfun.com/decimal-fraction-percentage.html) _value representation_ of a fraction (i.e. using a decimal symbol).

angularsen:
@lipchev Good point, some more googling and I believe _decimal_ is the correct term?

> So, to convert a fraction to a decimal, divide the numerator by the denominator.
https://www.bbc.co.uk/bitesize/topics/znmtsbk/articles/z4ymtv4

angularsen:
Updated PR to use `Decimal` and updated both `Ratio` and `RatioChangeRate`.

MarkCiliaVincenti:
The problem with Ratio.Decimal is that decimal is a type in C#, so it could create confusion.

angularsen:
@MarkCiliaVincenti Right, but since its type is `double` I think maybe not a big problem.
After reading up on the terms I believe DecimalFraction is not a good choice either.

Are we happy with Decimal? Any other suggestions?
@lipchev @MarkCiliaVincenti

lipchev:
I personally prefer `DecimalFraction` over `Decimal`: we're represents a `Ratio` in terms of a [Decimal Fraction](https://en.wikipedia.org/wiki/Decimal#Decimal_fractions) expressed (visually) using the 

> modern European decimal notation

I also think that renaming this would have quite a visible effect on client code (having obsolete warnings popping all over the place), that it simply wouldn't be worth the effort..

angularsen:
@lipchev Ok, I had not considered the "visual representation" part. I guess it makes sense when reading the wiki section.

https://en.wikipedia.org/wiki/Decimal#Decimal_fractions

> The numbers that are represented by decimal numerals are the decimal fractions (sometimes called decimal numbers),
> For example, the numerals 
image
> represent the fractions 
image

I guess we _could_ call it `DecimalNumeral`, but that is a very unfamiliar term to me at least.
I agree, not worth the hassle. Let's stick with `DecimalFraction`.

PR #881, opened 2020-12-29
Fix RatioChangeRate conversion functions

angularsen:
Seems like the wrong base unit was assumed.
The tests still pass after changing it, which probably explains why it was not noticed during PR review.
It also seems like a non-breaking change, the values from the consumer still appeared correct both before and after this PR.

```c#
RatioChangeRate.FromDecimalFractionsPerSecond(1).As(RatioChangeRate.BaseUnit) // 1
RatioChangeRate.FromDecimalFractionsPerSecond(1).PercentsPerSecond // 100
```

### DIFF
--- a/Common/UnitDefinitions/RatioChangeRate.json
+++ b/Common/UnitDefinitions/RatioChangeRate.json
@@ -6,8 +6,8 @@
     {
       "SingularName": "PercentPerSecond",
       "PluralName": "PercentsPerSecond",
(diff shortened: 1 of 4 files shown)


angularsen:
[Release UnitsNet/4.82.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.82.0)

PR #883, opened 2020-12-29
Add quantity CloudCover

### LINKED ISSUE #879: Oktas

MarkCiliaVincenti:
Fixes #879


Common/UnitDefinitions/CloudCover.json
```diff
@@ -0,0 +1,43 @@
+{
+  "Name": "CloudCover",
+  "BaseUnit": "Fraction",
```
angularsen, line 3: I know I suggested `Fraction`, but let's await the decision in #880 on whether we want to use `DecimalFraction` or `Decimal` here.

Common/UnitDefinitions/CloudCover.json
```diff
@@ -0,0 +1,43 @@
...
+  "Units": [
+    {
+      "SingularName": "Okta",
+      "PluralName": "Oktas",
+      "FromUnitToBaseFunc": "x * 0.125",
+      "FromBaseToUnitFunc": "System.Math.Round(x / 0.125, MidpointRounding.AwayFromZero)",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "oktas" ]
```
angularsen, line 14: Apparently there are unicode symbols for the okta values, maybe we should offer a method to get the unicode string representation for the cloud cover?

https://en.wikipedia.org/wiki/Okta

This can be added to `UnitsNet/CustomCode/Quantities/CloudCover.extra.cs` as an additional method
```c#
public string ToOktaSymbol() 
{
    // 50% cloud cover => ◑
    // etc.
}
```

UnitsNet.Tests/CustomCode/CloudCoverTests.cs
```diff
@@ -0,0 +1,34 @@
...
+// Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
+
+using System;
+
+namespace UnitsNet.Tests.CustomCode
+{
+    public class CloudCoverTests : CloudCoverTestsBase
+    {
+        protected override double FractionsInOneFraction => 1;
+        protected override double FractionsTolerance => base.FractionsTolerance;
```
angularsen, line 27: Should not be necessary to override tolerance unless changing them to something else.

Common/UnitDefinitions/CloudCover.json
MarkCiliaVincenti: I was working on this, actually doing it as a char property with a getter but then I noticed there aren't Unicode characters for all oktas; 3, 5 and 7 are missing. Not sure we should go there. The full symbols can be found at https://en.wikipedia.org/wiki/Okta#/media/File:Cloud_Cover_in_Oktas.svg but as you see we don't have Unicode values for 3, 5 and 7.

UnitsNet.Tests/CustomCode/CloudCoverTests.cs
MarkCiliaVincenti: Fixed.

MarkCiliaVincenti:
Not sure what the problem is with regards to the AppVeyor build failing. The code compiles for me and all cloud cover unit tests pass.

Common/UnitDefinitions/CloudCover.json
angularsen: We could just round to the nearest available symbol, but I understand if this would perhaps not satisfy real use.

angularsen:
> Not sure what the problem is with regards to the AppVeyor build failing. The code compiles for me and all cloud cover unit tests pass.

It fails on

```
C:\projects\unitsnet\UnitsNet.WindowsRuntimeComponent\GeneratedCode\Quantities\CloudCover.g.cs(526,53): error CS0103: The name 'UnitsNetMath' does not exist in the current context
```

angularsen:
Ah, it's the windows runtime component code generator.
You added a new extension method that is currently not added to the WRC code base.

This is a target that is largely deprecated, but kept around for new units via JSON changes.

Could you please duplicate the extension method to `UnitsNet.WindowsRuntimeComponent` so that its code also compiles?

Common/UnitDefinitions/CloudCover.json
MarkCiliaVincenti: OK, is this the only thing left? Everything else LGTY?

MarkCiliaVincenti:
> Ah, it's the windows runtime component code generator.
> You added a new extension method that is currently not added to the WRC code base.
> 
> This is a target that is largely deprecated, but kept around for new units via JSON changes.
> 
> Could you please duplicate the extension method to `UnitsNet.WindowsRuntimeComponent` so that its code also compiles?

Done.

Common/UnitDefinitions/CloudCover.json
MarkCiliaVincenti: No, it won't. if you have 3 oktas, you don't want to see the symbol of either 2 or 4. Best we could do would be output the SVG for displaying the symbol, but I think this would go way out of the scope of UnitsNet.

MarkCiliaVincenti:
Bump.

Common/UnitDefinitions/CloudCover.json
angularsen: Sorry yes, I bumped the conversion in #880 to try to move this along.

angularsen:
Waiting on #880 to decide on naming. Bumped that thread.

UnitsNet.WindowsRuntimeComponent/UnitsNet.WindowsRuntimeComponent.csproj
```diff
@@ -11,25 +11,22 @@
     <AssemblyName>UnitsNet</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.16299.0</TargetPlatformVersion>
-    <TargetPlatformMinVersion>10.0.10586.0</TargetPlatformMinVersion>
+    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.18362.0</TargetPlatformVersion>
+    <TargetPlatformMinVersion>10.0.18362.0</TargetPlatformMinVersion>
```
lipchev, line 15: Is there a reason to bump up the MinVersion?

UnitsNet/InternalHelpers/UnitsNetMath.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        /// <summary>
+        /// Clamps a numeric value to a min and max, similar to Math.Clamp.
+        /// </summary>
+        /// <typeparam name="T">The type</typeparam>
+        /// <param name="val">The value to be clamped</param>
+        /// <param name="min">The minimum value allowed</param>
+        /// <param name="max">The maximum value allowed</param>
+        /// <returns>The clamped value</returns>
+        public static T Clamp<T>(this T val, T min, T max) where T : IComparable<T>
```
lipchev, line 18: This would also work for any IQuantity - we could maybe expose it using a more restrictive (IQuantity) version in [UnitMath](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/UnitMath.cs)

UnitsNet.WindowsRuntimeComponent/UnitsNet.WindowsRuntimeComponent.csproj
angularsen: I think Visual Studio makes this change automatically. But yes, best to revert this.

angularsen:
As discussed in #880, let's go with `DecimalFraction`.

lipchev:
Sorry I haven't read up much on the Oktas but doesn't it make more sense to have Clamp as a method on the quantity itself, rather than the internally stored value? Currently it seems that we cannot ensure proper addition/subtraction operations give the _lossy_ nature of the internal round. For example isn't the result from (A + B)/2 likely to be (significantly) different from A/2 +B/2? 
Anyway, If this is the desired behavior then I would think it is worth a test or two, demonstrating the specific (discrete?) nature of the quantity (like with the [PowerRatioTests](https://github.com/angularsen/UnitsNet/blob/df6eafb593c1db074a465d8a015b8a04ab432e22/UnitsNet.Tests/CustomCode/PowerRatioTests.cs#L16)).
 
Also, maybe we should think about flagging the json file somehow, to indicate that it is non-linear (like with the Logarithmic flag)

angularsen:
> Currently it seems that we cannot ensure proper addition/subtraction operations give the lossy nature of the internal round. 

@lipchev Good point, but then I propose we rather disable arithmetic codegen?
I don't think it makes a whole lot sense to do quarter cloud cover + half cloud cover anyway.

```json
"GenerateArithmetic": false,
```
See [Temperature](https://github.com/angularsen/UnitsNet/blob/77941e424db92305397e855b60b139501a701fb3/Common/UnitDefinitions/Temperature.json) as an example.

angularsen:
@lipchev As for clamping, a better solution might be to add support for clamping of a quantity - so you could define that CloudCover should be clamped to say [0,1] of base unit - or some selected unit. Then it would inject clamping where appropriately. However, since this is the first quantity that requires it, I'm fine with doing it per unit for now. I think it should do what we want, to not allow less than 0% and not more than 100% cloud cover.

angularsen:
@MarkCiliaVincenti I believe this is the current status of the PR feedback:

- Rename to `DecimalFraction`
- Don't generate code for arithmetic with `"GenerateArithmetic": false,` in JSON
- Move clamp to `UnitMath` and if possible make it reusable for any `IQuantity`
- Revert version change in `UnitsNet.WindowsRuntimeComponent.csproj`

MarkCiliaVincenti:
> @MarkCiliaVincenti I believe this is the current status of the PR feedback:
> 
> * Rename to `DecimalFraction`
> * Don't generate code for arithmetic with `"GenerateArithmetic": false,` in JSON
> * Move clamp to `UnitMath` and if possible make it reusable for any `IQuantity`
> * Revert version change in `UnitsNet.WindowsRuntimeComponent.csproj`

Sorry, I saw this almost 2 years late :)

What did you mean by renaming to `DecimalFraction`? Instead of CloudCover?

angularsen:
In JSON, rename SingularName/PluralName Fraction/Fractions to DecimalFraction/DecimalFractions to be consistent with `Ratio.DecimalFraction`.

Issue #848, opened 2020-10-14
QuantityInfo.MinValue and QuantityInfo.MaxValue are missing


bitbonk:
**Is your feature request related to a problem? Please describe.**
For a generic (WPF) UI where you can input any type of quantity, we need the ability to configure the maximum and minimum bounds that the user is currently allowed to enter for a given value. If the user can enter any value (unbounded), the bounds should be set to the minimum and maximum that the given quantity type generally allows. Unfortunately there is no generic way to determine the minimum and maximum of a given quantity type. Although all quantities have the concrete properties like `Pressure.MinValue` and `Pressure.MaxValue`, there is no generic/polymorph way to determine these values without knowing the quantity type at compile time. 

**Describe the solution you'd like**
Introduce the properties `QuantityInfo.MinValue` and `QuantityInfo.MaxValue`, just like you have already introduced the properety `QuantityInfo.Zero`:

```csharp
    /// <summary>
    ///     Minimum value of quantity, such as <see cref="P:UnitsNet.Length.MinValue" />.
    /// </summary>
    public IQuantity MinValue { get; }

    /// <summary>
    ///     Maximum value of quantity, such as <see cref="P:UnitsNet.Length.MaxValue" />.
    /// </summary>
    public IQuantity MaxValue { get; }
```

**Describe alternatives you've considered**
We are currently using the following code:

```chsarp
(IQuantity)quantityInfo.ValueType.GetProperty("MinValue").GetValue(null)
```

angularsen:
I think this would be useful. Would you be interested in attempting a pull request? I'm happy to assist.

angularsen:
A few notes:

1. Some quantities use `decimal`, but most use `double` as their internal value. This implementation should reflect that.
2. A quantity is a pair of `value` and `unit`. Length units span from nanometers to kilometers. So `MinValue` should probably be `double.MinValue` with unit `Nanometer` and `MaxValue` be `double.MaxValue` with unit `Kilometer`.

rohahn:
It was rather simple to expose the `MinValue` and `MaxValue` from the specific quantity classes through the `QuantityInfo` class. 

But unfortunately I noticed that the current min/max values are not very useful since they always use the base unit. They neither represent a physical limit (e.g. light speed or absolute zero temperature) nor a lossless conversion limit of the datatype.

Regarding your notes:

[...]

PR #884, opened 2020-12-30
Deprecate MaxValue, MinValue

angularsen:
See discussion in #848.

There is no good choice for min/max value that don't risk running into overflow exception and the value may change if larger/smaller units are added later.

For range checks, we recommend specifying an application-specific value or using nullability for unbounded upper/lower value.

### Changes
- Mark `MinValue` and `MaxValue` as obsolete for all quantities


bitbonk:
LGTM ?

angularsen:
Thanks, yes I'll just merge this. It is safe enough to revert if we change our minds later.

angularsen:
[Release UnitsNet/4.82.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.82.0)

PR #888, opened 2021-01-03
Use NumberGroupSeparator and NumberDecimalSeparator

rohahn:
Use NumberFormat.NumberGroupSeparator and NumberFormat.NumberDecimalSeparator
instead of hardcoded separators in unit tests since the values for some cultures
changed between different versions of .NET

### DIFF
--- a/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
+++ b/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
@@ -50,70 +50,50 @@ namespace UnitsNet.Tests
             // ReSharper restore UnusedMember.Local
         }
 
-        // These cultures all use a comma for the radix point
         [Theory]
         [InlineData("de-DE")]
         [InlineData("da-DK")]
         [InlineData("es-AR")]
         [InlineData("es-ES")]
         [InlineData("it-IT")]
-        public void CommaRadixPointCultureFormatting(string culture)
-        {
-            Assert.Equal("0,12 m", Length.FromMeters(0.12).ToUnit(LengthUnit.Meter).ToString(GetCulture(culture)));
-        }
-
-        // These cultures all use a decimal point for the radix point
-        [Theory]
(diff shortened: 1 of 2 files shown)


rohahn:
I noticed the problems with .NET 5 and cultures fr-FR and ar-EG.

UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
```diff
@@ -50,70 +50,55 @@ private enum CustomUnit
...
-        public void DecimalPointDigitGroupingCultureFormatting(string culture)
+        public void DigitGroupingCultureFormatting(string cultureName)
         {
-            Assert.Equal("1.111 m", Length.FromMeters(1111).ToUnit(LengthUnit.Meter).ToString(GetCulture(culture)));
+            CultureInfo culture = GetCulture(cultureName);
+            string gs = culture.NumberFormat.NumberGroupSeparator;
+
+            Assert.Equal($"1{gs}111 m", Length.FromMeters(1111).ToUnit(LengthUnit.Meter).ToString(culture));
+
+            if (!cultureName.StartsWith("en"))
```
angularsen, line 91: Do we need to exclude non-English cultures from footinch/stone/pounds test cases?
If someone wants to print stone pounds in Italian, even if the abbrevaitions are not currently translated, then why not?

angularsen (changes requested):

UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
rohahn: I remove the check. The comment about the English speaking countries confused me.

angularsen (approved):

angularsen:
Nuget is on the way
[Release UnitsNet/4.83.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.83.0)

PR #896, opened 2021-01-18
Fix custom quantity support in QuantityInfo

### LINKED ISSUE #895: QuantityInfo can not be initialized for a custom quantity
**Describe the bug**
A custom unit that is not located in the _UnitsNet_ assembly can not initialize a _QuantityInfo_ instance which is required to implement the _IQuantity_ interface.

The following code in _QuantityInfo_ assumes that all unit enumerations must be located in the same assembly as the _Length_ type (UnitsNet assembly), which is not valid for custom quantities.

```
        private static readonly Type[] UnitEnumTypes = typeof(Length)
            .Wrap()
            .Assembly
            .GetExportedTypes()
            .Where(t => t.Wrap().IsEnum && t.Namespace == UnitEnumNamespace && t.Name.EndsWith("Unit"))
            .ToArray(); 
```

Since _HowMuch.QuantityInfo_ is not called by any unit test, the problem is not exposed.

**To Reproduce**
New unit test for QuantityInfoTests:

        [Fact]
        public void Constructor_AssignsPropertiesForExternalQuantity()
        {
            var expectedZero = new HowMuch(10, HowMuchUnit.Some);
            var expectedUnitInfos = new UnitInfo[]
            {
                new UnitInfo(HowMuchUnit.Some, BaseUnits.Undefined),
                new UnitInfo(HowMuchUnit.ATon, BaseUnits.Undefined),
                new UnitInfo(HowMuchUnit.AShitTon, BaseUnits.Undefined)
            };
            var expectedBaseUnit = HowMuchUnit.Some;
            var expectedQuantityType = QuantityType.Undefined;
            var expectedBaseDimensions = BaseDimensions.Dimensionless;

            var info = new QuantityInfo(nameof(HowMuch), expectedUnitInfos,
                expectedBaseUnit, expectedZero, expectedBaseDimensions);

            Assert.Equal(expectedZero, info.Zero);
            Assert.Equal(nameof(HowMuch), info.Name);
            Assert.Equal(expectedUnitInfos, info.UnitInfos);
            Assert.Equal(expectedQuantityType, info.QuantityType);
            Assert.Equal(expectedBaseDimensions, info.BaseDimensions);

            // Obsolete members
            Assert.Equal( expectedBaseUnit, info.BaseUnit );
            Assert.Equal(new[] {nameof(HowMuchUnit.Some), nameof(HowMuchUnit.ATon), nameof(HowMuchUnit.AShitTon)}, info.UnitNames);
        }

**Expected behavior**
_QuantityInfo_ instances can be created for custom units.
_QuantityInfo_ should have a constructor with a Type parameter for the units enumeration, so the _UnitEnumTypes_ field and following line can be removed:
`UnitType = UnitEnumTypes.First(t => t.Name == $"{name}Unit");`

rohahn:
Fixes #895 

Replace cache of internal unit enum types with constructor parameter
Remove obsolete constructors (which did not work for custom quantities anyway)


rohahn:
This is a proposal for a fix for issue #895.
The enum type can be passed to the constructor now.
I also removed the obsolete constructors which were not used internally anymore, since any external usage for custom types would have resulted in an exception anyway.

UnitsNet/QuantityInfo.cs
```diff
@@ -21,52 +20,28 @@ namespace UnitsNet
...
-        /// </summary>
-        /// <param name="quantityType">The quantity enum value.</param>
-        /// <param name="unitInfos">The information about the units for this quantity.</param>
-        /// <param name="baseUnit">The base unit enum value.</param>
-        /// <param name="zero">The zero quantity.</param>
-        /// <param name="baseDimensions">The base dimensions of the quantity.</param>
-        /// <exception cref="ArgumentException">Quantity type can not be undefined.</exception>
-        /// <exception cref="ArgumentNullException">If units -or- baseUnit -or- zero -or- baseDimensions is null.</exception>
-        [Obsolete("QuantityType will be removed in the future. Use QuantityInfo(string, UnitInfo[], Enum, IQuantity, BaseDimensions) instead.")]
-        public QuantityInfo(QuantityType quantityType, [NotNull] UnitInfo[] unitInfos, [NotNull] Enum baseUnit, [NotNull] IQuantity zero, [NotNull] BaseDimensions baseDimensions)
```
angularsen, line 44: We can't remove this constructor, it would be a breaking change since it's been around for a long time.

Whenever me or another maintainer finds the time and energy to do a major version bump, then we can remove all the obsolete code.

UnitsNet/QuantityInfo.cs
```diff
@@ -21,52 +20,28 @@ namespace UnitsNet
...
         /// <param name="name">Name of the quantity.</param>
+        /// <param name="unitType">The unit enum type, such as <see cref="LengthUnit" />.</param>
         /// <param name="unitInfos">The information about the units for this quantity.</param>
         /// <param name="baseUnit">The base unit enum value.</param>
         /// <param name="zero">The zero quantity.</param>
         /// <param name="baseDimensions">The base dimensions of the quantity.</param>
         /// <param name="quantityType">The the quantity type. Defaults to Undefined.</param>
         /// <exception cref="ArgumentException">Quantity type can not be undefined.</exception>
         /// <exception cref="ArgumentNullException">If units -or- baseUnit -or- zero -or- baseDimensions is null.</exception>
-        public QuantityInfo([NotNull] string name, [NotNull] UnitInfo[] unitInfos, [NotNull] Enum baseUnit, [NotNull] IQuantity zero, [NotNull] BaseDimensions baseDimensions,
```
angularsen, line 60: This would also be a breaking change, however since it was recently introduced and it didn't actually work as intended, then I'm ok with changing it.

UnitsNet/QuantityInfo.cs
```diff
@@ -193,17 +168,10 @@ public IEnumerable<UnitInfo> GetUnitInfosFor(BaseUnits baseUnits)
     public class QuantityInfo<TUnit> : QuantityInfo
         where TUnit : Enum
     {
-        /// <inheritdoc cref="QuantityInfo{TUnit}(string,UnitsNet.UnitInfo{TUnit}[],TUnit,UnitsNet.IQuantity{TUnit},UnitsNet.BaseDimensions,UnitsNet.QuantityType)" />
-        [Obsolete("QuantityType will be removed in the future. Use QuantityInfo(QuantityType, string, UnitInfo{TUnit}[], TUnit, IQuantity{TUnit}, BaseDimensions) instead.")]
-        public QuantityInfo(QuantityType quantityType, UnitInfo<TUnit>[] unitInfos, TUnit baseUnit, IQuantity<TUnit> zero, BaseDimensions baseDimensions)
```
angularsen, line 198: Same here, can't remove this.

angularsen (changes requested):
I believe you are introducing breaking changes here, see comments.

UnitsNet/QuantityInfo.cs
rohahn: In theory it is a breaking change. But this constructor did not work for custom quantities and why should any external code manually create a QuantityInfo instance for one of the internal quantities? But I will restore the removed constructors.

UnitsNet/QuantityInfo.cs
rohahn: I restored it.

angularsen:
[Release UnitsNet/4.87.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.87.0)

PR #900, opened 2021-02-04
Units in support of reaction kinetics

thehoglet:
Added unit def for molar flow
Added unit def for maximum rate of reaction
Added prefixes to LitersPerHour in VolumeFlow*

* seems necessary in order to support e.g. mg/h


lipchev (commented):
I'm a little confused about the MolarMaximumRateOfReaction: if we're representing Vmax then shouldn't the units be in the form _mol m−3 s−1_ (here the word _Molar_ being used to represent the [MolarConcentration unit](https://github.com/angularsen/UnitsNet/blob/3558d59b47205db775822a826b4894dfdfd5c387/Common/UnitDefinitions/Molarity.json#L41)). To quote [this article](https://www.sciencedirect.com/topics/engineering/michaelis-menten-equation):

> rA is the volumetric rate of reaction with respect to reactant A, CA is the concentration of A, vmax is the maximum rate of reaction, and Km is the Michaelis constant. vmax has the same dimensions as rA; Km has the same dimensions as CA. Typical units for vmax are mol m−3 s−1; typical units for Km are mol m−3. 

Maybe this is a different thing altogether- something like "MassDependentMolarFlow"? You haven't provided the operators for it- but I presume that multiplying by Mass should give us MolarFlow, right?

Otherwise, if this is indeed meant to represent Vmax as described in the above article, then as a quantity- should this be different from something like "MolarReactionRate", which I would assume to be the special case of _change of concentration over time_ (mol/l/s), not to be confused with the MolarFlow which is _change of amount of substance over time_ (mol/s).

UnitsNet/UnitsNet.csproj
```diff
@@ -2,7 +2,7 @@
   <!-- NuGet properties -->
   <PropertyGroup>
     <PackageId>UnitsNet</PackageId>
-    <Version>4.83.0</Version>
+    <Version>4.86.0</Version>
```
lipchev, line 5: I don't think 4.86.0 is out yet :)

thehoglet:
@lipchev As regards _MolarMaximumRateOfReaction_, on reflection, your confusion is justified. To confirm I consulted a colleague, one who is better qualified and still practicing chemistry. He wrote:

> ...I have never actually had to look at MM kinetics or PBPK really, so I’m afraid I don’t  know what units are “standard”.

>  I can see why it’s confused. I think the mmol/min/mg looks like you should be able to say

> “The quantity mmol/mg is mol/g which is 1/ the molar mass.”

> But the two “measures of stuff” are actually measuring different materials.

> And it doesn’t help that the language is inconsistent in this field. (Molar doesn’t always  mean of a mole (as in molar mass) but also means 1 mole per litre). 

> So MolarMaximumRateOfReaction is ambiguous.

> Is it the max rate for 1 gram molecular weight (to use an old but specific term) or is it the max rate when you have a solution containing 1 gram molecular weight per litre?

> If the issue is one of “what should we call this thing” I invite you to consider MolarityMaximumRateOfReaction or MoleMaximumRateOfReaction though they both look wrong.

> Frankly, you can call it Snardlepingle  as long as it works.

> Units for reaction kinetics were always screwy – square litres per square mole per second, if I remember rightly.

Anyway, rather than bend a coherent, consistent, general-purpose library to support such vague concepts, I can withdraw this request and continue building my app using a local nuget package built from the fork. Or if your suggestion of _MolarReactionRate_ is acceptable, I'll resubmit the request with a name change. The bottom line is I need to be able to parse user units with dimensionality N.T<sup>-1</sup>.M<sup>-1</sup>.

lipchev:
@thehoglet  I'm not sure: I personally think the name _MolarReactionRate_ better suits the molar concentration version of this quantity, where for this I would suggest a non-specialist name like "_MassDependentMolarFlow_" (although I'm not at all sure that there are other possible applications, while the keyword "_Dependent_" could become a gateway into the library for (m)any types fractions). There is also the third option of [Extending with Custom Units](https://github.com/angularsen/UnitsNet/wiki/Extending-with-Custom-Units)

@angularsen What do you think?

angularsen:
@lipchev I mentally tapped out a long time ago 😅 
I really have no experience in this domain or any preference here, sorry!

thehoglet:
@lipchev I like your suggestion of _MassDependentMolarFlow_. Nice and generic, and, for my application (reaction kinetics), it's applicable because the 'flow' is dependent on the mass of substrate present. I will resubmit the request with all your suggested changes.

PR #906, opened 2021-02-26
Fix BUnits

bplett-wgtcorp:
Wouldn't you know it!  I got the units reversed. :(

### DIFF
--- a/Common/UnitDefinitions/Frequency.json
+++ b/Common/UnitDefinitions/Frequency.json
@@ -95,8 +95,8 @@
     {
       "SingularName": "BUnit",
       "PluralName": "BUnits",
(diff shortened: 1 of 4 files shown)


angularsen:
No problem, that's why we try to look up the test case value from a references as a double-check we got the conversion functions right.

[Release UnitsNet/4.85.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.85.0)

bplett-wgtcorp:
I thought I had double and triple checked it.  Well, I did but in the reverse order.  Anyway, thanks again!!!

Brad.

PR #907, opened 2021-03-08
Added DataContract annotations to all Quantities {Value, Unit}

lipchev:
..also
- added tests covering the default [DataContractSerializer](https://github.com/lipchev/UnitsNet/blob/data-contract-annotations/UnitsNet.Tests/Serialization/DataContractSerializerTests/DataContractSerializerTests.cs) (xml) & [DataContractJsonSerializer](https://github.com/lipchev/UnitsNet/blob/data-contract-annotations/UnitsNet.Tests/Serialization/DataContractJsonSerializerTests/DefaultDataContractJsonSerializerTests.cs) (json)
- added [DataContractSurrogates](https://github.com/lipchev/UnitsNet/tree/data-contract-annotations/UnitsNet/Serialization/Surrogates) (n/a in [.netcore](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.idatacontractsurrogate?view=netframework-4.8&viewFallbackFrom=netcore-2.0#applies-to)) that can be used in the same way as the JsonConverter (i.e. offer an alternative json representation):
a) [GenericQuantityDataContractSurrogate](https://github.com/lipchev/UnitsNet/blob/data-contract-annotations/UnitsNet/Serialization/Surrogates/GenericQuantityDataContractSurrogate.cs): [{"1.20", "InformationUnit.Exabyte"}](https://github.com/lipchev/UnitsNet/blob/cd1fc52e904defd37e708e478893f7e990200d86/UnitsNet.Tests/Serialization/DataContractJsonSerializerTests/Surrogates/GenericDataContractJsonSerializerTests.cs#L20)  // **output should be quoted**- numeric input should be handled automatically, quoted input- parsed
b) ~~[ExtendedQuantityDataContractSurrogate](https://github.com/lipchev/UnitsNet/blob/data-contract-annotations/UnitsNet/Serialization/Surrogates/ExtendedQuantityDataContractSurrogate.cs): [{"1.2", "InformationUnit.Exabyte", ""1.20", "decimal"}](https://github.com/lipchev/UnitsNet/blob/cd1fc52e904defd37e708e478893f7e990200d86/UnitsNet.Tests/Serialization/DataContractJsonSerializerTests/Surrogates/ExtendedDataContractJsonSerializerTests.cs#L22)  // same format as the one used with the UnitsNetIQuantityJsonConverter~~
c) ~~[BasicQuantityContractSurrogate](https://github.com/lipchev/UnitsNet/blob/data-contract-annotations/UnitsNet/Serialization/Surrogates/BasicQuantityContractSurrogate.cs): {"1.2", "InformationUnit.Exabyte"}  // same format as the one used with the UnitsNetJsonConverter (obsolete)~~
d) [QuantityWithAbbreviationContractSurrogate](https://github.com/lipchev/UnitsNet/blob/data-contract-annotations/UnitsNet/Serialization/Surrogates/QuantityWithAbbreviationContractSurrogate.cs): [{"1", "kg", "Mass"}](https://github.com/lipchev/UnitsNet/blob/cd1fc52e904defd37e708e478893f7e990200d86/UnitsNet.Tests/Serialization/DataContractJsonSerializerTests/Surrogates/QuantityWithAbbreviationJsonDataContractSerializerTests.cs#L20) // **output should be quoted**- numeric input should be handled automatically, quoted input- parsed
- ~~added two-way compatibility tests with both [UnitsNetJsonConverter](https://github.com/lipchev/UnitsNet/tree/data-contract-annotations/UnitsNet.Serialization.JsonNet.CompatibilityTests) & [UnitsNetIQuantityJsonConverter](https://github.com/lipchev/UnitsNet/tree/data-contract-annotations/UnitsNet.Serialization.JsonNet.Tests/Compatibility) (in their respective test-projects- only applicable to **net48**, which I had to [include](https://github.com/lipchev/UnitsNet/blob/cd1fc52e904defd37e708e478893f7e990200d86/UnitsNet.Serialization.JsonNet.Tests/UnitsNet.Serialization.JsonNet.Tests.csproj#L4))~~ (the DC <-> JsonNet compatibility would be broken- unless I also include their respective converters)

Finally [here](https://stackoverflow.com/a/25368029) is some background info on how you can setup data contract surrogates as a service behavior (pretty much the same way you fix the nasty date format, only they've removed the [property](https://github.com/lipchev/UnitsNet/blob/cd1fc52e904defd37e708e478893f7e990200d86/UnitsNet.Tests/Serialization/DataContractJsonSerializerTests/Surrogates/GenericDataContractJsonSerializerTests.cs#L12), along with the whole [interface](https://github.com/lipchev/UnitsNet/blob/cd1fc52e904defd37e708e478893f7e990200d86/UnitsNet/Serialization/Surrogates/QuantityDataContractSurrogateBase.cs#L16) after net48).


lipchev:
Right, so I did a couple of tests and turns out that adding a second framework to any of the tests projects causes the build to freeze (tested on the main repository).
I guess I'll move the Surrogate tests to a new dedicated project (say, UnitsNet.Serialization.DataContract.Tests). And then another one for the compatibility tests- *DataContract.Compatability.JsonNet.Tests- both targeting net48 (or earlier if we want)..
Unless of course you have some other suggestion- in truth the data contract surrogates may arguably be outside the scope of the project altogether- but I would argue that leaving them out could induce many people to simply opt in for the simpler solution- that is just use the default DataContractJsonSerializer's schema, which would work- but would be incompatible with the one produced by the JsonNet nuget.

lipchev:
Hi, 
This has been up for a while- and while there aren't any stoppers for me in here- I think at least the DataContract annotations would be nice-to-have (unless I'm mistaken - the previously mentioned issues regarding the between-version compatibility have been addressed).  I think those should be plug-and-play for your standard WCF communication (not requiring interoperability with json.net).

As for the Surrogates - I really don't like how they introduced all the conditionals (which I failed to properly configure on the project level) - but I had no idea they were removed in core... Anyway- since they don't add any dependencies - I figured I'd keep them, allowing me to have the json-serializers-compatibility-tests (as a kind of a baseline contract for whatever other serialization library support we add- like System.Text.Json).

If you want- I can remove the Surrogates / Compatibility Tests (leaving only the 'default' DataContract xml/json serialization/deserialization tests).

Honestly, my goal was to make another PR after this one introducing a json.net Converter using the abbreviation-based schema (which I personally prefer, and regardless of preference, would be forced to use soon):
> d) QuantityWithAbbreviationContractSurrogate: {"1", "kg", "Mass"}

UnitsNet.Serialization.DataContract.Tests/DataContractJsonSerializerTests/Surrogates/BasicQuantityContractSurrogateTests.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Runtime.Serialization.Json;
+using UnitsNet.Serialization.Surrogates;
+
+namespace UnitsNet.Serialization.DataContract.Tests.DataContractJsonSerializerTests.Surrogates
+{
+    /// <summary>
+    ///     There are no special tests here: uses the <i>standard</i> format of encoding the Unit & Value (with no support for
```
angularsen, line 7: Need to use `&amp;` I think. Or just `and`.

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -34,16 +35,19 @@ namespace UnitsNet
     /// <summary>
     ///     Acceleration, in physics, is the rate at which the velocity of an object changes over time. An object's acceleration is the net result of any and all forces acting on the object, as described by Newton's Second Law. The SI unit for acceleration is the Meter per second squared (m/s²). Accelerations are vector quantities (they have magnitude and direction) and add according to the parallelogram law. As a vector, the calculated net force is equal to the product of the object's mass (a scalar quantity) and the acceleration.
     /// </summary>
-    public partial struct Acceleration : IQuantity<AccelerationUnit>, IEquatable<Acceleration>, IComparable, IComparable<Acceleration>, IConvertible, IFormattable
+        [DataContract]
+        public partial struct Acceleration : IQuantity<AccelerationUnit>, IEquatable<Acceleration>, IComparable, IComparable<Acceleration>, IConvertible, IFormattable
```
angularsen, line 39: It seems the indentation increased here.

UnitsNet/Serialization/Contracts/ExtendedQuantityValueContract.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        [DataMember(Order = 3)]
+        public string ValueString { get; protected set; }
+
+        /// <summary>
+        /// The value type associated with this given string representation.
+        /// </summary>
+        [DataMember(Order = 4)]
+        public TValueType ValueType { get; protected set; }
+    }
```
angularsen, line 39: Codecov points out some of these data contract members may be lacking test coverage. If that is correct, I think we should add some basic serialization tests for these.

UnitsNet/Serialization/Contracts/QuantityValueContract.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <typeparam name="TUnit">The type used to represent the quantity unit (e.g. enum / int / string) </typeparam>
+    [DataContract]
+    public class QuantityValueContract<TValue, TUnit>
+    {
+        /// <summary>
+        ///     Initializes a new quantity with the given unit and value
+        /// </summary>
+        /// <param name="value">The value</param>
+        /// <param name="unit">The associated unit</param>
+        public QuantityValueContract(TValue value, TUnit unit)
```
angularsen, line 19: Isn't a default empty constructor required for WCF contracts?

UnitsNet/Serialization/Contracts/QuantityValueContract.cs
```diff
@@ -0,0 +1,37 @@
...
+        ///     The value.
+        /// </summary>
+        [DataMember(Order = 1)]
+        public TValue Value { get; protected set; }
+
+        /// <summary>
+        ///     The unit of the value.
+        /// </summary>
+        [DataMember(Order = 2)]
+        public TUnit Unit { get; protected set;}
```
angularsen, line 35: Codecov points out some of these data contract members may be lacking test coverage. If that is correct, I think we should add some basic serialization tests for these.

UnitsNet/Serialization/Contracts/QuantityWithAbbreviationContract.cs
```diff
@@ -0,0 +1,36 @@
...
+        public QuantityWithAbbreviationContract(TQuantity quantityType, TValue value, string unitAbbreviation) : base(value, unitAbbreviation)
+        {
+            QuantityType = quantityType;
+        }
+
+        /// <summary>
+        /// The quantity type identifier.
+        /// </summary>
+        [DataMember(Order = 3)]
+        public TQuantity QuantityType { get; protected set; }
```
angularsen, line 34: Codecov points out some of these data contract members may be lacking test coverage. If that is correct, I think we should add some basic serialization tests for these.

angularsen (changes requested):
I think this is solid work 👏

Although WCF data contracts are no longer part of .NET future, there are tons of projects still on .NET 4 and on WCF stack for sure. I'm positive to include this for those targeting `net40` of `UnitsNet` nuget.

Just some minor observations. I only skimmed the implementation and tests, but it looks really good to me. I trust that you who actually will be consuming this will be in a good position to verify it works as intended and it's easy to consume.

One thing, the README serialization section should be updated to reflect the new data contract support and the surrogates we provide out of the box. You might want to link to a wiki page if it gets too detailed to include on the front page.

UnitsNet/Serialization/Contracts/QuantityValueContract.cs
lipchev: To be honest, I had never seen a constructor on a DataContract before - but the DC serializers don't seem to mind it.
I haven't tried generating any proxies - so there is still a chance this could fail. (I know you definitely need an empty constructor on the service)

UnitsNet/Serialization/Contracts/ExtendedQuantityValueContract.cs
lipchev: I tried a few ways of "covering" those. My final attempt has both the [QuantityValueContract_Constructor_AssignsAllProperties](https://github.com/lipchev/UnitsNet/blob/a7e0bc60f9d70b0d320f88f11b8fe868972e1158/UnitsNet.Serialization.DataContract.Tests/QuantityValueDataContractTests.cs#L12) and [QuantityValueContract_SerializationRoundTrips](https://github.com/lipchev/UnitsNet/blob/a7e0bc60f9d70b0d320f88f11b8fe868972e1158/UnitsNet.Serialization.DataContract.Tests/DataContractSerializerTests/DataContractSerializerTests.cs#L80) (one for each DC serializer).
However now that I've also pushed the generated quantities - the code-coverage fails because of the merge conflicts.. :(

lipchev:
Ok here's a cheat-sheet of all the serialization schemas currently supported: the last 3 I have stashed locally as I wasn't sure which one(s) to select. Note that both json serializers can handle quoted numbers (so do [System.Text](https://github.com/dotnet/docs/pull/21108)), and in fact- this appears to be the only way for handling decimals with JsonNet (and possibly for System.Text [as well](https://github.com/dotnet/runtime/issues/1366)). No out-of-the-box support in protobuf [either](https://github.com/protocolbuffers/protobuf/issues/4406) :)
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/galin/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/galin/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">



Name | Dependency | Schema | Description | Decimal Output | Decimal Input | Arrays | IEnumerable | Tuples | 1.2 mg | 1.20 mW
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
DefaultDataContractJsonSerializer | None | XML | The default WCF serializer | Supported | Supported | Supported | Supported | Supported | <Mass xmlns=\http://schemas.datacontract.org/2004/07/UnitsNet\"   xmlns:i=\"http://www.w3.org/2001/XMLSchema-instance\"><Value>1.2</Value><Unit>Milligram</Unit></Mass>" | <Power xmlns=\http://schemas.datacontract.org/2004/07/UnitsNet\"   xmlns:i=\"http://www.w3.org/2001/XMLSchema-instance\"><Value>1.20</Value><Unit>Milliwatt</Unit></Power>"
DataContractJsonSerializer | None | {Number, Integer} | The default WCF json serializer | Supported | Supported | Supported | Supported | Supported | {\Value\":1.2,\"Unit\":16}" | {\Value\":1.20,\"Unit\":20}"
BasicQuantityContractSurrogate | .Net Framework | {Double, Enum} | The default DataContractJsonSerializer configured with the   BasicQuantityContractSurrogate: double precision with string-based unit   serialization (e.g. MassUnit.Milligram) | Double Precision | Double Precision | Supported | Supported | Supported | {\Value\":1.2,\"Unit\":\"MassUnit.Milligram\"}" | {\Value\":1.2,\"Unit\":\"PowerUnit.Milliwatt\"}"
GenericDataContractJsonSerializer | .Net Framework | {Number, Enum} | The default DataContractJsonSerializer configured with the   GenericQuantityDataContractSurrogate: double/decimal precision with   string-based unit serialization (e.g. MassUnit.Milligram) | Supported | Supported | Supported | Supported | Supported | {\Value\":1.2,\"Unit\":\"MassUnit.Milligram\"}" | {\Value\":1.20,\"Unit\":\"PowerUnit.Milliwatt\"}"
ExtendedQuantityDataContractSurrogate | .Net Framework | {Double, Enum, Decimal, Type} | The default DataContractJsonSerializer configured with the   ExtendedQuantityDataContractSurrogate: custom (e.g. decimal) values types   stored as quoted numbers (two additional fields) | Supported | Supported | Supported | Supported | Supported | {\Value\":1.2,\"Unit\":\"MassUnit.Milligram\"}" | {\Value\":1.2,\"Unit\":\"PowerUnit.Milliwatt\",\"ValueString\":\"1.20\",\"ValueType\":\"decimal\"}"
QuantityWithAbbreviationContractSurrogate | .Net Framework | {Number, Abbreviation, Type} | The default DataContractJsonSerializer configured with the   QuantityWithAbbreviationContractSurrogate: double/decimal values with   quantity type and unit abbreviation | Supported | Supported | Supported | Supported | Supported | {\Value\":1.2,\"Unit\":\"mg\",\"QuantityType\":\"Mass\"}" | {\Value\":1.20,\"Unit\":\"mW\",\"QuantityType\":\"Power\"}"
UnitsNetIQuantityJsonConverter | JsonNet | {Double, Enum, Decimal, Type} | The currently recommended UnitsNet.JsonNet converter: decimals supported   using the extended schema | Supported | Supported | Supported | Supported | Not Supported | {\"Unit\":\"MassUnit.Milligram\",\"Value\":1.2} | {\Unit\":\"PowerUnit.Milliwatt\",\"Value\":1.2,\"ValueString\":\"1.20\",\"ValueType\":\"decimal\"}"
UnitsNetJsonConverter | JsonNet | {Double, Enum} | The obsolete UnitsNet.JsonConverter | Double Precision | Double Precision | Not Supported | Not Supported | Not Supported | {\Unit\":\"MassUnit.Milligram\",\"Value\":1.2}" | {\Unit\":\"PowerUnit.Milliwatt\",\"Value\":1.2}"
QuantityWithAbbreviationJsonConverter | JsonNet | {Double, Abbreviation, Type} | Double values with quantity type and unit abbreviation | Double Precision | Double Precision | Supported | Supported | Not Supported | {\Value\":1.2,\"Unit\":\"mg\",\"QuantityType\":\"Mass\"}" | {\Value\":1.2,\"Unit\":\"mW\",\"QuantityType\":\"Power\"}"
PreciseQuantityWithAbbreviationJsonConverter(quoted:false) | JsonNet | {Number, Abbreviation, Type} | Serialization preserves the correct precision; unquoted values converted   to double during deserialization | Supported | Double Precision | Supported | Supported | Not Supported | {\Value\":1.2,\"Unit\":\"mg\",\"QuantityType\":\"Mass\"}" | {\Value\":1.20,\"Unit\":\"mW\",\"QuantityType\":\"Power\"}"
PreciseQuantityWithAbbreviationJsonConverter(quoted:true) | JsonNet | {Number, Abbreviation, Type} | Decimal values serialized as string (quoted numbers are supported by both   JsonNet & the DataContractJsonSerializer) | Supported | Supported | Supported | Supported | Not Supported | {\Value\":1.2,\"Unit\":\"mg\",\"QuantityType\":\"Mass\"}" | {\Value\":\"1.20\",\"Unit\":\"mW\",\"QuantityType\":\"Power\"}"



</body>

</html>

angularsen:
That's one impressive list @lipchev 😅 

If decimals must be represented as strings, then I suppose it is simpler and more consistent to serialize all quantity values regardless of number type.

Beyond that, let me know when this is getting ready to review or if you are waiting for feedback on anything specific 👍

lipchev:
> If decimals must be represented as strings, then I suppose it is simpler and more consistent to serialize all quantity values regardless of number type.

I did a quick test today with the System.Text serializer and it looks like it is correctly handling the unquote decimal precision. That means that JsonNet is the only one that actually requires the use of strings (and the additional properties of the ExtendedContract). 
I would therefore suggest that we recommend the use of the following two schemas:

1. For use with applications referencing UnitsNet: [QuantityValueContract](https://github.com/lipchev/UnitsNet/blob/data-contract-annotations/UnitsNet/Serialization/Contracts/QuantityValueContract.cs)= `{\Value":1.20,"Unit":"PowerUnit.Milliwatt"}"`
2. For use within a broader environment: [QuantityWithAbbreviationContract ](https://github.com/lipchev/UnitsNet/blob/data-contract-annotations/UnitsNet/Serialization/Contracts/QuantityWithAbbreviationContract.cs)= ` {\Value":1.20,"Unit":"mW","QuantityType":"Power"}"`

These are fully supported by the `GenericQuantityDataContractSurrogate` and `QuantityWithAbbreviationContractSurrogate`, and will be natively supported the (future) System.Text converter.
In both cases quoted strings are accepted as input - but the output is always a Number: e.g. 1.20.

For JsonNet users there are two scenarios w.r.t. the value type:

1. The precision (number of zeros) associated with the decimal is not important- use the `GenericQuantityJsonConverter `(using the generic `QuantityValueContract<double | decimal, string>`) or the version currently named `PreciseQuantityWithAbbreviationJsonConverter(quoted:false)` 
2. Preserving the exact precision during de-serialization is important (you've probably got some custom quantity implementation, as most UnitsNet operations are double-based):  you can use the `UnitsNetIQuantityJsonConverter `(the current implementation using the extended format), `PreciseQuantityJsonConverter `(serialized using the `QuantityValueContract<double | string, string>`) or the `PreciseQuantityWithAbbreviationJsonConverter(quoted:false)`

Unless there is some significant difference in performance between the _UnitsNetIQuantityJsonConverter_ and the (_not yet implemented version_) of the _PreciseQuantityJsonConverter_, I think I would favor the latter.

Finally, for people only interested in the IQuantity (double) interface it is trivial to provide slightly faster (and cleaner) implementations: e.g. `BasicQuantityContractSurrogate `& `QuantityWithAbbreviationJsonConverter` that completely ignore the IDecimalQuantity interface, but I'm not sure if it is actually worth the extra bytes (_unless of course, you think bytes are fun_). 

So the main question is - HowMany converter-versions do we provide? 

Bonus question: what do you think about "QuantityType" as property name- maybe we should use something shorter?

angularsen:
First, I need to understand something. Is this PR primarily about supporting WCF xml + json?

System.text.json and json.net is discussed here also. 

There is #966, won't that cover system.text.json?
Is the idea here to also support system.text.json via data contracts?
If so, should we still go ahead with custom json converters?


> I think I would favor the latter
> Finally, for people only interested in the IQuantity (double) interface it is trivial to provide slightly faster (and cleaner) implementations:

Yes! The fewest number of options that we absolutely need to provide, please.

lipchev:
We are going about it backwards: from the existing schemas (implementations of JsonNet) that are out there we've got the corresponding DataContracts (1:1 mapping of a particular schema). Those can then be used by any of the serializers/converters (e.g. [ValueUnit](https://github.com/lipchev/UnitsNet/blob/a7e0bc60f9d70b0d320f88f11b8fe868972e1158/UnitsNet.Serialization.JsonNet/UnitsNetBaseJsonConverter.cs#L194) is the [QuantityValueContract](https://github.com/lipchev/UnitsNet/blob/data-contract-annotations/UnitsNet/Serialization/Contracts/QuantityValueContract.cs). Same for the [ExtendedValueUnit ](https://github.com/lipchev/UnitsNet/blob/a7e0bc60f9d70b0d320f88f11b8fe868972e1158/UnitsNet.Serialization.JsonNet/UnitsNetBaseJsonConverter.cs#L218) and [ExtendedQuantityValueContract](https://github.com/lipchev/UnitsNet/blob/data-contract-annotations/UnitsNet/Serialization/Contracts/ExtendedQuantityValueContract.cs). #966 is also based on the ExtendedQuantityValue[Contract]. Those can both be refactored using the common (generic) contract- but that is not the point of this PR.

The point was to add the [DataContract] annotations and to provide reference schema(s), implemented (as an example) by the DataContractSurrogates. At the same time, I was making a point about having both an internal and external schema (our Delphi application cannot already has a table with standard units - and conversions, but knows nothing about MassUnit.Milligram). 

The problem is then further divided into two parts by the (JsonNet's) existing ExtendedQuantityValue schema- which I argue should be deprecated in favor of the GenericQuantityDataContract schema (with the unquoted number- example implementation with GenericQuantityDataContractSurrogate).

While this PR should be good to go as it (not sure I finally managed to make codecov happy- but I think things are otherwise _covered_), I started work on the JsonNet-based implementations of the QuantityWithAbbreviationSchema (QuantityWithAbbreviationJsonConverter and  PreciseQuantityWithAbbreviationJsonConverter)- trying out the capabilities of the different serializers (producing the cheat-sheet). 

The two implementations above are ready- and mostly covered, but I haven't pushed them as they add to the existing JsonConverter/Tests projects. If you would rather have it all in one- I can push them as well (giving the PR some broader name)...

lipchev:
>  If so, should we still go ahead with custom json converters in #966?

That is my question as well- we can easily have both the Generic & Extended schemas implemented- but should we and which one would _you_ recommended?

UnitsNet.NanoFramework/GeneratedCode/AmountOfSubstance/packages.config
```diff
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="nanoFramework.CoreLibrary" version="1.10.4-preview.11" targetFramework="netnanoframework10" />
```
angularsen, line 3: I don't think this PR should have to touch these nanoframework files? 
Make sure you have merged in latest origin/master and then maybe revert these.

UnitsNet/Serialization/Contracts/QuantityValueContract.cs
angularsen: As long as it works. The tests seems to work at least.

angularsen (changes requested):

UnitsNet.NanoFramework/GeneratedCode/AmountOfSubstance/packages.config
angularsen: > That is my question as well- we can easily have both the Generic & Extended schemas implemented- but should we and which one would you recommend?

I don't like the extended schema. If some quantities need quoted numbers, then I think it would be simpler to just quote all numbers.

I have to admit I am struggling to grasp all the contracts, and surrogates and converters here and in #966.
That table alone is mind-boggling to put it mildly 😅 so bear with me if I am not quite getting it.

How about these two contracts:
- **Shape A**: Quoted number + full enum name
- **Shape B**: Quoted number + unit abbreviation

It could then look like this:
- WCF and its `DataContractSerializer` (xml) or `DataContractJsonSerializer` (json)
  - Shape A: `GenericQuantityDataContractSurrogate`
  - Shape B: `QuantityWithAbbreviationContractSurrogate`
- Json.NET and its `JsonConverter`
  - Shape A: `GenericQuantityJsonConverter` (but with quoted numbers)
  - Shape B: `QuantityWithAbbreviationJsonConverter`
- System.Text.Json and its `JsonSerializer`
  - Shape A: #966 ? 
  - Shape B: #966 ?
- Protobuf and other data contract compatible serializers
  - I assume similar converters as above must be created for each? That pesky decimal seems to be the main problem.

lipchev: Those got auto-incremented on the last commit (after running the build script)- I didn't follow much of the discussion around the nano-framework, so I thought this was intentional.

angularsen: Yes, sorry, that was a flaw in the setup earlier. It should have been fixed by now.

lipchev: I've updated the summary: 

1. From what I gather there should be no backwards support for the Basic & Extended schemas (will remove the corresponding Surrogates). 
2. All values should be quoted: will update the contracts & tests
3. This would break the compatibility tests- so those should be removed as well 
4. Should I add the new JsonNet converters? System.Text?
5. If (4) then should I keep the DC <-> JsonNet compatibility tests? Maybe change them something about them? I first wanted to make it work with strings and quantities- but the format is never exactly the same : there is a difference in serializing slashes- like in "mg/l", so I went with the serialize-with-one -deserialize-with-another workflow..
6. Nano packages should be reverted

angularsen: 1. I think breaking backwards compatibility for new converters and new contracts is fine.
2. Yes
3. Yes
4. json.net and system.text converters can be added in separate PRs, if it makes sense to scope this PR can to only the data contracts. You decide. Also I don't know PR 966 well enough, if you should reuse that work or not.
5. If I understand correctly, Shape A and Shape B should have compatible JSON formats regardless of DC, JsonNet or SystemText being used. If so, then yes, compatibility tests should be kept.
6. Yes

lipchev:
I'm going to start this over: this time without the DataContractSurrogates - there is simply too much overhead (with the dependency to .NET Framework) - and I can't imagine anybody is ever going to upgrade to *that* particular setup. 
Otherwise, I've been testing the `[DataContract]`s along with the basic abbreviation schema (unquoted) for the past few weeks and so far it's going pretty smoothly. A ton of auto-mapping code went away.

angularsen:
@lipchev Ok, sounds good to me. It did feel a bit overwhelming with all different ways to serialize, and I believe surrogates and WCF is largely a thing of the past. I can imagine a very small minority would be interested in that.

Issue #847, opened 2020-10-14
JSON deserialization of quantities that use decimal internally may fail


bitbonk:
**Describe the bug**
When serializing and deserializing quantities that use the `decimal` type for storage interally, JSON deserialization using the converters from _UnitsNet.Serialization.JsonNet_ may fail with an `OverflowException` if the value is too big or too large.

**To Reproduce**
Steps to reproduce the behavior (just an example):
1. Add nuget UnitsNet 4.72.0, UnitsNet.Serialization.JsonNet 4.2.0, Xunit, and FluentAssertions to a .NET Core 3.1 class library project
2. Build and run the test code below.
3. See 6 failing tests.

```csharp
    using System.Collections.Generic;
    using System.Linq;
    using FluentAssertions;
    using Newtonsoft.Json;
    using UnitsNet;
    using UnitsNet.Serialization.JsonNet;
    using Xunit;

    public class QuantitySerializationTests
    {
        public static IEnumerable<object[]> QuantityZeros { get; } =
            Quantity.Infos.Select(
                info => new object[]
                {
                    (IQuantity)info.Zero
                });
        
        public static IEnumerable<object[]> QuantityMinValues { get; } =
            Quantity.Infos.Select(
                info => new object[]
                {
                    (IQuantity)info.ValueType.GetProperty("MinValue").GetValue(null)
                });
        
        public static IEnumerable<object[]> QuantityMaxValues { get; } =
            Quantity.Infos.Select(
                info => new object[]
                {
                    (IQuantity)info.ValueType.GetProperty("MaxValue").GetValue(null)
                });

        [Theory]
        [MemberData(nameof(QuantityZeros))]
        public void Zero_IsMessagePackSerializable(IQuantity value)
        {
            var sutClone =
                JsonConvert.DeserializeObject(
                    JsonConvert.SerializeObject(value, Settings),
                    value.GetType(), Settings);

            sutClone.Should().NotBeSameAs(value);
            sutClone.Should().BeEquivalentTo(value, o => o.RespectingRuntimeTypes());
        }

        [Theory]
        [MemberData(nameof(QuantityMinValues))]
        public void MinValue_IsMessagePackSerializable(IQuantity value)
        {
            var sutClone =
                JsonConvert.DeserializeObject(
                    JsonConvert.SerializeObject(value, Settings),
                    value.GetType(), Settings);

            sutClone.Should().NotBeSameAs(value);
[...]

PR #911, opened 2021-03-19
Fix deserializing decimal quantities regardless of local culture

angularsen:
Fixes parsing JSON for decimal quantities like Power on machines with cultures like Norwegian, where a comma is decimal separator is used. Several tests were red on my machine, but green on build agent.

Related to #847, #868

### DIFF
--- a/UnitsNet.Serialization.JsonNet.Tests/UnitsNetBaseJsonConverterTest.cs
+++ b/UnitsNet.Serialization.JsonNet.Tests/UnitsNetBaseJsonConverterTest.cs
@@ -239,7 +239,7 @@ namespace UnitsNet.Serialization.JsonNet.Tests
                 var result = ConvertIQuantity(value);
                 if (result is ExtendedValueUnit {ValueType: "decimal"} decimalResult)
                 {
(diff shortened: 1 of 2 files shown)


lipchev (approved):
As per the JSON _specs_ the separator is indeed "." (InvariantCulture).

However, given that nobody has complained so far, I would suggest that for v5 we simply replace the `Value` with the `ExtendedValue` (i.e. the string representation of the number) and remove the `ExtendedValueUnit` as it's CLR type could easily be determined from the `Unit` (once parsed, we can check for the `IDecimalQuantity`).

angularsen:
Great, yes, that sounds reasonable.

As for v5, the wishlist is piling up and there are several big ideas for changes that are still up in the air; #714, #709, class vs struct.
 
If you or @tmilnthorp want to start nudging the v5 snowball, I'm happy to help, but work and family has been pretty much consuming all life the past 9 months and it's likely to continue.

Baby steps is probably a good idea if we are ever to get any of the breaking changes out the door.

angularsen:
@lipchev forgot to mention you. ☝️

angularsen:

PR #912, opened 2021-03-19
Trim ToString() for units without abbreviation

angularsen:
Scalar, Ratio and other quantities with units that don't have an abbreviation,
would result in a trailing whitespace.

Example: `Scalar.FromAmount(1).ToString() == "1 "`

### DIFF
--- a/UnitsNet/QuantityFormatter.cs
+++ b/UnitsNet/QuantityFormatter.cs
@@ -125,6 +125,12 @@ namespace UnitsNet
         public static string Format<TUnitType>(IQuantity<TUnitType> quantity, string format, IFormatProvider? formatProvider)
             where TUnitType : Enum
         {
+            return FormatUntrimmed(quantity, format, formatProvider).Trim();
+        }
+
+        private static string FormatUntrimmed<TUnitType>(IQuantity<TUnitType> quantity, string format, IFormatProvider? formatProvider)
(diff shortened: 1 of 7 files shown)


UnitsNet/QuantityFormatter.cs
```diff
@@ -124,6 +124,12 @@ public static string Format<TUnitType>(IQuantity<TUnitType> quantity, string for
         /// <returns>The string representation.</returns>
         public static string Format<TUnitType>(IQuantity<TUnitType> quantity, string format, IFormatProvider? formatProvider)
             where TUnitType : Enum
+        {
+            return FormatUntrimmed(quantity, format, formatProvider).Trim();
```
lipchev, line 128: Not a huge difference, but I think this could be simplified to `TrimEnd`

lipchev (commented):
It is quite unlikely but not impossible that someone out there is doing something like string.split(x)[1]

UnitsNet/QuantityFormatter.cs
angularsen: Fixed

angularsen:
> It is quite unlikely but not impossible that someone out there is doing something like string.split(x)[1]

Right, but I still think the improvement outweighs the downside in that scenario.

angularsen:

PR #918, opened 2021-04-09
Add Tilt to Angle

bplett-wgtcorp:
"Tilt" is commonly used in Tilt Meters, and it is simply the Sin of the angle.

### DIFF
--- a/UnitsNet.Tests/CustomCode/AngleTests.cs
+++ b/UnitsNet.Tests/CustomCode/AngleTests.cs
@@ -38,6 +38,8 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double RevolutionsInOneDegree => 2.777777777777777e-3;
 
+        protected override double TiltInOneDegree => 0.01745240643728351281941897851632;
+
         [Fact]
         public void AngleDividedByDurationEqualsRotationalSpeed()
         {
--- a/Common/UnitDefinitions/Angle.json
+++ b/Common/UnitDefinitions/Angle.json
@@ -92,6 +92,18 @@
           "Abbreviations": [ "r" ]
         }
       ]
+    },
+    {
+      "SingularName": "Tilt",
+      "PluralName": "Tilt",
+      "FromUnitToBaseFunc": "Math.Asin(x)*180/Math.PI",
+      "FromBaseToUnitFunc": "Math.Sin(x/180*Math.PI)",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "sin(θ)" ]
+        }
+      ]
     }
   ]
 }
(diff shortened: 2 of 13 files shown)


UnitsNet.Tests/CustomCode/AngleTests.cs
```diff
@@ -38,6 +38,8 @@ public class AngleTests : AngleTestsBase
 
         protected override double RevolutionsInOneDegree => 2.777777777777777e-3;
 
+        protected override double TiltInOneDegree => 0.01745240643728351281941897851632;
```
angularsen, line 41: Could you point me to a source reference where I can verify this, besides your calculation? I can't seem to find any.

I found these:
https://en.wikipedia.org/wiki/Inclinometer
https://en.wikipedia.org/wiki/Grade_(slope)

bplett-wgtcorp:
Verifying the constant is trivial.  In fact, had you not previously advised me against it, I wouldn’t have used a constant.  I’d just have used:
protected override double TiltInOneDegree => Math.Sin(RadiansInOneDegree);

Hope that helps!
Brad.

UnitsNet.Tests/CustomCode/AngleTests.cs
angularsen: @bplett-wgtcorp Right, but what I meant is that I can't find any source of reference that says the measurement "tilt" equals Sin(radians) of an angle. A wiki or unit converter or any kind of digital source would be helpful to verify this unit.

bplett-wgtcorp:
Ahh, sorry.  I actually get lots of “hits”.  Try this search:  https://www.google.com/search?q=%2Btilt+meter+output+%2Bsine+angle

Using just the top three on the results I get:
From https://www.geosense.co.uk/media/BlockAttributeValueFile/152/file/2TiltMeterBeamApplicationGuideV1.0.pdf I get “The readings generated by Digital Tilt Sensors are in ‘Sine of the Angle’.”
From https://www.encardio.com/blog/tiltmeter-how-does-it-work-installation-how-to-read/ I get “The tilt sensor thus provides a bipolar DC voltage output proportional to the sine of tilt angle measured by the tiltmeter.”
From https://www.althensensors.com/landing-pages/tilt-angle-measurement/ I get “…provides an output signal of +/- 5 volts proportional to the sine wave”

I have nothing to do with any of those companies.  😊  Is that sufficient?  Would you like more?

Thanks!
Brad.

UnitsNet.Tests/CustomCode/AngleTests.cs
angularsen: @bplett-wgtcorp Thanks, no those are plenty!

angularsen:
Thanks!

bplett-wgtcorp:
Fantastic!  Thanks a lot!

Brad.

PR #923, opened 2021-04-30
Adding benchmark workflows

### LINKED ISSUE #920: GitHub Action for Continuous Benchmarking
I was waiting for [the PR](https://github.com/rhysd/github-action-benchmark/pull/34) for [this](https://github.com/marketplace/actions/continuous-benchmark-net) to get merged, but as it is still open after nearly a year- I suggest we try out the forked action. 

Granted, we don't have many performance affecting PRs- so we could maybe simply enabled it manually for some commits?

I'm quite a novice on the whole git-hub workflow actions thing, but if there aren't any gotchas this would seem like a cool thing to have.

lipchev:
- updated the UnitsNet.Benchmark project (updated nugets & added frameworks)
- added BenchmarkCategories to the existing benchmarks
- added a few scripts for local testing (individual, multi-platform + rplots)
- added an automatic benchmark workflow on pushing to [master] (some folders), running for ["netcoreapp50", "netcoreapp21", "net472"]
- added workflows triggered on workflow_dispatch for running a benchmark (with options for comparing against a baseline)

Closes #920 

First a few ground rules

1. As of this writing, manually triggered workflows (a.k.a. `workflow_dispatch`), [only work on the master branch](https://github.community/t/how-to-trigger-repository-dispatch-event-for-non-default-branch/14470). 
2. Composite actions have some nasty [limitations](https://github.com/actions/runner/issues/862)
3. The Continuous Benchmark [needs to be in a context](https://github.com/rhysd/github-action-benchmark/issues/24) with a commit in order to work 
4.  It publishes [commits](https://github.com/lipchev/UnitsNet-Benchmarks/commits/gh-pages/benchmarks/netcoreapp50) (with optional comments and alerts) targeting a given folder on a separate branch (typically gh-pages/benchmark/)- [here](https://lipchev.github.io/UnitsNet-Benchmarks/benchmarks/netcoreapp50/) is an example (all but the last benchmarks were run using a very-low iteration count- a sort of DryRun, so the wild fluctuations are expected). Here is [another](https://lipchev.github.io/UnitsNet-Benchmarks/benchmarks/net472/) (that's consistently slower).
5. It [does not support](https://github.com/rhysd/github-action-benchmark/issues/18) multi-line charts (out-of-the-box)
6. I read somewhere that the average performance of the virtual machines vary within about 10-15%
7. We can run the benchmarks for [any framework](https://benchmarkdotnet.org/articles/configs/toolchains.html#multiple-frameworks-support) from 472 onward (windows required)
8. We can easily configure all aspects of a benchmark using the [command line arguments](https://benchmarkdotnet.org/articles/guides/console-args.html) (there is also a `dotnet benchmark tool` but that seems to be doing the exact same things- although I suspect it's also going to get some 'compare results' functionality at some point which would likely remove the need for the next step)
9. There are a few result-comparing extensions out there- but I most liked the one from [dotnet/performance](https://github.com/dotnet/performance/tree/main/src/tools/ResultsComparer)
10. Just a few days ago, I knew almost none of that.

The workflows are well documented, I think you would easily figure out what's going on. Here are the highlights:

1. The automatic workflow runs the benchmarks (multiple frameworks), for each one exporting the result to the [gh-pages/benchmarks](https://github.com/lipchev/UnitsNet-Benchmarks/tree/gh-pages/benchmarks)- one framework one sub-folder
2. It is only supposed to run on the master branch & only on certain occasions (see [here](https://github.com/lipchev/UnitsNet/blob/dd071c6255424049401ad1d2ea6c7c85383eb38c/.github/workflows/continious-benchmarking.yml#L2))
2. For [each folder](https://github.com/lipchev/UnitsNet-Benchmarks/tree/gh-pages/benchmarks/net472) (framework) we have the chart data: `data.js & index.html` (here we append results) and the [results](https://github.com/lipchev/UnitsNet-Benchmarks/tree/gh-pages/benchmarks/net472/results) folder containing the latest benchmark results, as exported by BenchmarkDotNet (we override these each time the script is executed)
3. We can then reference / compare results across branches & repositories (a fork doesn't have to have gh-pages in order to compare against upstream/gh-pages/benchmark/xxx/)
4. This can be done using the second action: which can be used to generate a report (the typical results folder) which can then be compared against a baseline (this step only runs if the [baseline parameter](https://github.com/lipchev/UnitsNet/blob/dd071c6255424049401ad1d2ea6c7c85383eb38c/.github/workflows/run-benchmarks.yml#L29) points to a valid "*-report-full.json")

PS Some of the links point to lipchev/UnitsNet-Benchmarks which is the repository I used for testing. Unfortunately something got messed up during the rebase so I made the PR in a new branch (of my main fork) - this (as expected) did not trigger a benchmark-run as I pushed on a non master branch. As to why the manual flow isn't showing up in the actions menu- see 1). 

PS2 You should still have access to the test repository (lipchev/UnitsNet-Benchmarks) if you want to test out the [workflows](https://github.com/lipchev/UnitsNet-Benchmarks/actions)- feel free to commit on it if you want, I plan to delete it after this is merged.

### DIFF
--- a/.github/actions/run-benchmarks/action.yaml
+++ b/.github/actions/run-benchmarks/action.yaml
@@ -1,0 +1,43 @@
+name: 'Run benchmark'
+inputs:
+  framework:
+    description: 'The runtime version to use (e.g. net5.0)'
+    required: false
+    default: 'net5.0'
+  runtimes:
+    description: 'The runtime version to use (e.g. netcoreapp31, net5.0)'
+    required: false
+    default: 'net5.0'
+  output-folder:
+    description: 'The output folder for the benchmark (a results folder is created inside)'
+    required: false
+    default: 'Artifacts/Benchmark'
+  exporters:
+    description: 'The exporter(s) used for this run (GitHub/StackOverflow/RPlot/CSV/JSON/HTML/XML)'
+    required: false
+    default: fulljson rplot
+  filter:
+    description: 'An optional class filter to apply'
+    required: false
+    default: '*'
+  categories:
+    description: 'An optional categories filter to apply'
+    required: false
+    default: ''
+  execution-options:
+    description: 'Any additional parameters passed to the benchmark'
+    required: false
+    default: ''
+runs:
+  using: "composite"
+  steps:
+    - name: Generating benchmark results
+      run: >
+        dotnet run --project UnitsNet.Benchmark -c Release 
+        --framework ${{ inputs.framework }} 
+        --runtimes ${{ inputs.runtimes }} 
+        --artifacts '${{ inputs.output-folder }}' 
+        --exporters ${{ inputs.exporters }} 
+        --filter '${{ inputs.filter }}' 
+        ${{ inputs.categories }} ${{ inputs.execution-options }}
+      shell: bash
--- a/.github/workflows/continious-benchmarking.yml
+++ b/.github/workflows/continious-benchmarking.yml
@@ -1,0 +1,101 @@
+name: UnitsNet Benchmarks (auto)
+on:
+  push:
+    branches: [master]    
+    paths:
+      - "UnitsNet/*"
+      - "UnitsNet.Benchmark/*"
+      - ".github/workflows/**"
+      - ".github/actions/**"
+      
+env:
+  FRAMEWORK: net5.0 
+  EXECUTION_OPTIONS: --iterationTime 500 --disableLogFile # see https://benchmarkdotnet.org/articles/guides/console-args.html
+  BENCHMARK_PAGES_BRANCH: gh-pages
+  BENCHMARK_DATA_FOLDER: benchmarks
+  
+jobs:
+  benchmark:
+    runs-on: windows-latest # required by the older frameworks
+    strategy:
+      # max-parallel: 1 # is it better to avoid running in parallel?
+      matrix: 
+        runtime: ["netcoreapp50", "netcoreapp21", "net472"]        
+    steps:
+      - run: echo Starting benchmarks for ${{ matrix.runtime }} 
+      
+      # checkout the current branch
+      - uses: actions/checkout@v2 
+      
+      # we need all frameworks (even if only running one target at a time)
+      - uses: actions/setup-dotnet@v1
+        with: 
+          dotnet-version: '2.1.x'
+            
+      - uses: actions/setup-dotnet@v1
+        with: 
+          dotnet-version: '3.1.x'
+            
+      - uses: actions/setup-dotnet@v1
+        with: 
+          dotnet-version: '5.0.x'
+            
+      # executing the benchmark for the current framework, placing the result in a corresponding sub-folder              
+      - uses: ./.github/actions/run-benchmarks
+        with:
+          framework: ${{ env.FRAMEWORK }}
+          runtimes: ${{ matrix.runtime }}
+          output-folder: Artifacts/Benchmark/${{ matrix.runtime }}
+          execution-options: ${{ env.EXECUTION_OPTIONS }}
+            
+      # saving the current artifact (downloadable until the expiration date of this action)        
+      - name: Store benchmark artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: UnitsNet Benchmarks (${{ matrix.runtime }})
+          path: Artifacts/Benchmark/${{ matrix.runtime }}/results/
+          
+  publish-results:
+    needs: [benchmark]
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 1 # cannot commit on the same branch in parallel
+      matrix: 
+        runtime: ["netcoreapp50", "netcoreapp21", "net472"]
+    steps:        
+      - name: Initializing git folder ️
+        uses: actions/checkout@v2.3.1
+
+      - name: Download Artifacts # The benchmark results are downloaded into a 'runtime' folder.
+        uses: actions/download-artifact@v1
+        with:
+          name: UnitsNet Benchmarks (${{ matrix.runtime }})
+          path: ${{ matrix.runtime }}        
+        
+      # publishing the current results to the benchmark-pages branch (overriding the previous result)
+      - name: Saving benchmark results to ${{ env.BENCHMARK_PAGES_BRANCH }}/${{ env.BENCHMARK_DATA_FOLDER }}/${{ matrix.runtime }}/results
+        uses: JamesIves/github-pages-deploy-action@4.1.1
+        with:
+          folder: ${{ matrix.runtime }}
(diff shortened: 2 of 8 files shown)


angularsen:
What the hell did you do in these commits? 😆 
image

angularsen (approved):

angularsen:
I mean, wow, great job and great outline! I love it ❤️ 
I say, pull the trigger! 

Letting you merge it, in case there is anything else to do before it merges.

The only thing I'm curious about, is how do we best use this tool to monitor regressions over time? 
If I understand correctly, this only runs on `master` so we can't easily use it when reviewin PRs. Is there any way to alert us to give a heads up if it regresses significantly, say 50% slower on some tests?

Also, are there any limits on the free plan for actions we need to know about?

lipchev:
There it's on- the action [is running](https://github.com/angularsen/UnitsNet/actions/runs/800789157)

lipchev:
Here is the free plan limits:

> Job execution time - Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete.
> 
> Workflow run time - Each workflow run is limited to 72 hours. If a workflow run reaches this limit, the workflow run is cancelled.

I'm currently running the matrix in parallel- so no way we could reach 6 hours. I'm not yet sure if running the benchmarks in parallel is any worse (I assume we get a random machine assignment).

lipchev:
As for testing the PR's- we wouldn't want to run the benchmarks, for something like adding a new quantity. As for the performance affecting PR's - the person proposing the modification (like you and me, or anyone else- even without privilages) can run the 'run-benchmarks' workflow on his branch (targeting one of framework at time*) using the baseline, that's available on your gh-pages branch (which the person would not need to pull or anything).

I'll try to give an example in a few minutes.

angularsen:
Sounds perfect

lipchev:
Ok, here are the initial results:
https://angularsen.github.io/UnitsNet/benchmarks/net472/
https://angularsen.github.io/UnitsNet/benchmarks/netcoreapp21/
https://angularsen.github.io/UnitsNet/benchmarks/netcoreapp50/

And the detailed results:
https://github.com/angularsen/UnitsNet/tree/gh-pages/benchmarks/net472/results
https://github.com/angularsen/UnitsNet/tree/gh-pages/benchmarks/netcoreapp21/results
https://github.com/angularsen/UnitsNet/tree/gh-pages/benchmarks/netcoreapp50/results

lipchev:
Ok, here is an example run from my newly created fork that returns a new list for each unknown abbreviation:
image

[Here](https://github.com/lipchev/UnitsNet/actions/runs/801146843) is the action log (net50)

And here are the results:

Runtime=.NET Core 5.0  Toolchain=netcoreapp50  

```
|          Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|---------------- |---------:|---------:|---------:|-------:|-------:|------:|----------:|
|           Parse | 49.42 μs | 0.980 μs | 1.048 μs | 1.7700 | 0.0610 |     - |  32.56 KB |
|   TryParseValid | 49.98 μs | 0.736 μs | 0.653 μs | 1.7700 | 0.0610 |     - |  32.54 KB |
| TryParseInvalid | 54.05 μs | 1.061 μs | 1.090 μs | 1.7090 | 0.0610 |     - |  32.16 KB |

..and the results of the comparison to the baseline
* (not comparing the allocations, but if you open the base-line from the link above you should be able to spot the difference)

summary:
worse: 3, geomean: 1.072
total diff: 3

| Slower                                                | diff/base | Base Median (ns) | Diff Median (ns) | Modality|
| ----------------------------------------------------- | ---------:| ----------------:| ----------------:| --------:|
| UnitsNet.Benchmark.UnitsNetBenchmarks.TryParseInvalid |      1.10 |         48927.15 |         53800.22 |         |
| UnitsNet.Benchmark.UnitsNetBenchmarks.TryParseValid   |      1.08 |         46153.61 |         49879.57 |         |
| UnitsNet.Benchmark.UnitsNetBenchmarks.Parse           |      1.04 |         47511.59 |         49251.15 |         |

No Faster results for the provided threshold = 0.001% and noise filter = 0.3ns.

lipchev:
Ok I only needed to fix [one little thing](https://github.com/angularsen/UnitsNet/commit/3bca803d0beb9c6e5da35b250d25c3aa09ddc807). This in turn triggered a re-run, so now we should get two points on the chart.

Overall the benchmarks appear to be quite noisy- some of the operations seem to be too short. For those cases microsoft is doing a [bunch of operations in a loop](https://github.com/dotnet/performance/blob/d7dac8a7ca12a28d099192f8a901cf8e30361384/src/benchmarks/micro/runtime/Math/Functions/Double/MinDouble.cs). I'm thinking of stealing this test, and putting it side by side with ours (sort of like a baseline for the IComparable interface, for which there is currently no benchmark).

lipchev:
It's worse than what I expected- I re-ran the workflow to see what a third point (of the same code-base) would look like, and it kinda looks like "the machine" was (almost) consistently slower/faster on each run (except for ToUnit on net472 where we seem to have an inflection, but that's probably due to the benchmark itself). 

It would probably smooth out if I run it a couple of more times (at different times), but I'm not sure how useful it would be for comparing results between branches. Maybe if I run the tests for a longer period, we would get some multi-modal distribution that when averaged out would give us a sample of the performance on a _typical_ VM. 

Anyways, the workflows seem to be doing what their supposed to. I'm sure an automatic comparison workflow can be devised that executes [`on: pull_request`] but I don't think it worth it for now (the manual-way doesn't take that many steps anyway).

This would be pretty cool to see run inside a stable self-hosted runner.

angularsen:
I guess it might be tricky to get stable numbers from a shared pool of worker VMs.

This one observed 10-20% fluctuation.
https://github.com/rhysd/github-action-benchmark#stability-of-virtual-environment

lipchev:
Yeah, pretty useless for any between-commit comparisons (might look ok in the long run, though I think I'm going to disable the auto-comment feature).. 

Also- I hadn't foreseen it triggering on release, but yeah- I guess the project file is inside the UnitsNet folder :)

Anyways, we got lucky this time- [Intel Xeon CPU E5-2673 v3 2.40GHz appears to have almost the same performance as Intel Xeon Platinum 8171M CPU 2.60GHz](benchmarks/netcoreapp50/results/UnitsNet.Benchmark.UnitsNetBenchmarks-report-github.md)

PS I'm playing with a little _hack_ for re-scaling the results (only for the charts)- that would probably smooth out the lines a little. 

PS1 There is a bug in the QuantityFrom benchmark (I've fixed it locally).

PS2 There are at least two performance improvements I'm going to PR for soon (so that this wouldn't have been a complete waste of time..).

angularsen:
You are on 🔥 🤩

PR #924, opened 2021-05-01
fixing the (possible) concurrent modifications

lipchev:
Hasn't been benchmark-ed yet (or tested for that matter :) ).

Don't merge this just yet, let me try to fix the benchmarks first..


Jagailo:
Faced this problem.
I'm using the library at different endpoints on the server.
When calling endpoints at the same time after starting the server, I get an exception.
`System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.`
image
image

angularsen:
Merged in #1159

PR #927, opened 2021-05-17
Update nanoFramework mscorlib

josesimoes:


angularsen:
[Release UnitsNet/4.91.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.91.0)

angularsen:
@josesimoes @Ellerbach I changed the nanoFramework codegen to not update nuget versions by default, you have to opt in with a new flag. It is documented here.

https://github.com/angularsen/UnitsNet/wiki/nanoFramework

The reason is that pull requests unrelated to nanoFramework often got spammed with updates to nanoFramework versions, which confused the contributors and made the PR harder to review.

josesimoes:
OK. That makes sense.

I just have a question : from now on, how will the updates be handled? I mean when a new release of those Nugets comes out, should a manual build be triggered? Guess there is no need to submit PRs... 🤔

angularsen:
Update via PR as before. When you run the command in the wiki, it will update the project files with the latest mscorlib/system.math versions. I did one update just now, but I skipped doing a pull requests since I didn't need a review on the change.

https://github.com/angularsen/UnitsNet/commit/ff0fa59a4bf36b4e20353d3ba30c9c62b52109c7

Ellerbach:
> Update via PR as before. When you run the command in the wiki, it will update the project files with the latest mscorlib/system.math versions. I did one update just now, but I skipped doing a pull requests since I didn't need a review on the change.
> 
> [ff0fa59](https://github.com/angularsen/UnitsNet/commit/ff0fa59a4bf36b4e20353d3ba30c9c62b52109c7)

So in other words, you decoupled the "normal" version with then "nano" version. which makes sense anyway. Most of the time the nano version is updated because of MS CoreLib or Math as they are based on the units only.

So that makes sense for me! Thanks!

PR #930, opened 2021-05-24
Added nuspec generation for nanoFramework UnitsNet

rschiefer:
Added code to automatically generate the nuspec file for UnitsNet.nanoFramework units.  Used placeholder ($version$) for version number in nuspec so will probably need to pass that in the build pipeline (unless thats a builtin variable, not sure).  nanoFramework.IoT.Device generates their nuget packages in the same way.


rschiefer:
Just realized I didn't catchup my branch of the main repo before creating my branch.  WOrking on that now.

rschiefer:
Sorry I've never setup GitVersion before.  I hardcoded the version number for now to buy us some time.

CodeGen/Generators/NanoFrameworkGenerator.cs
```diff
@@ -126,6 +120,22 @@ public static void Generate(string rootDir, Quantity[] quantities)
...
+                // Convert decimal based units to floats; decimals are not supported by nanoFramework
+                if (quantity.BaseType == "decimal")
+                {
+                    var replacements = new Dictionary<string, string>
+                    {
+                        //{ "(\\)sdecimal(\\s)", "$1float$2" }
+                        { "(\\d)m", "$1f" },
+                        { "(\\d)M", "$1f" },
+                        { " decimal ", " float " },
+                        { "(decimal ", "(float " }
```
angularsen, line 133: Why is float better than double? Decimal was originally chosen for a handful few quantities, to support higher precision of conversions between very large and very small units.

angularsen: Hardcoded version is fine for now, we can work more on this later.

rschiefer:
Honestly I don't remember why I went with float.  I agree double would be better.

CodeGen/Generators/NanoFrameworkGenerator.cs
rschiefer: Honestly I don't remember why I went with float. I agree double would be better.  Changed to use double.

angularsen:
@rschiefer Thanks a lot. Nugets 4.92 are now out.

rschiefer:
Thats great, thanks!

rschiefer:
@angularsen I'm not seeing the unitsnet.nanoframework.power nuget package.  This was one of the decimal conversions so it wasn't previously available.  Anything special we need to do to get it up to nuget?

angularsen:
@rschiefer Yes the published nugets are hand-picked for now. 
https://github.com/angularsen/UnitsNet/blob/master/Build/build-pack-nano-nugets.psm1

We didn't want to barf all 100+ packages out unless there was demand for them.

rschiefer:
ok, i'll update with the missing units we need.  thx

PR #939, opened 2021-06-09
Added Specific Range Unit

mcgougwm:


Common/UnitDefinitions/SpecificRange.json
```diff
@@ -0,0 +1,49 @@
+{
+  "Name": "SpecificRange",
+  "BaseUnit": "MeterPerKilogramForce",
+  "XmlDoc": "The range of the aircraft per unit of fuel.",
```
angularsen, line 4: I'm sorry, but I think this is too domain specific to include in the main UnitsNet library. We try to only add units that are widely used.

Maybe a separate aviation library could be added for quantities unique to that domain?

angularsen (changes requested):

Common/UnitDefinitions/SpecificRange.json
mcgougwm: I understand the desire to keep the UnitsNet library manageable. It's not clear to me what you mean by a separate aviation library. I was hoping I'd be able to get these added here so we wouldn't have to maintain our own version of your package or be responsible for building/distributing new versions.

angularsen: There are a couple of options for including units outside the UnitsNet nuget:

- [Extending with Custom Units · angularsen/UnitsNet Wiki](https://github.com/angularsen/UnitsNet/wiki/Extending-with-Custom-Units) and register custom units on startup. Optionally, these can be published in separate nugets for reuse by others.
- Fork UnitsNet repo and customize the JSON files. Would have to manually keep it in sync.

Hope this helps. I understand if it's not the ideal solution, but at this point I think it's the best we can do.

PR #940, opened 2021-06-09
Added Inverse Angle Units

mcgougwm:
Added Inverse Angle Units


Common/UnitDefinitions/InverseAngle.json
```diff
@@ -0,0 +1,39 @@
+{
+  "Name": "InverseAngle",
```
angularsen, line 2: What are some typical use cases for this quantity?

Just trying to make sure the quantities are widely used and avoid bloat, we have a small issue with binary size. I could not find much references by a quick google.

mcgougwm: Probably not very widely used, but a lot of our stability and control derivatives are in inverse degrees or inverse radians.

angularsen: Since we have 100+ quantities and 1000+ units, I am hesitant to start adding the inverse variant of all our quantities and units. 

We are adding #954 , but that one does seem like a widely used quantity with both wiki article and online converters for.

Are you OK with closing this PR, or do you still see value for others in adding inverse angle?

angularsen:
@mcgougwm I see you deleted the branch so maybe this is a bit late, but I've learned in #954 and #955 that reciprocal angle is also commonly used. So if you are still interested in adding it, I'm happy to merge it.

PR #941, opened 2021-06-10
Added Inverse Speed Unit

mcgougwm:
Added Inverse Speed Unit


Common/UnitDefinitions/InverseSpeed.json
```diff
@@ -0,0 +1,187 @@
+{
+  "Name": "InverseSpeed",
```
angularsen, line 2: Hm, I believe the term is _pace_, and I'm not sure this is really a different quantity than _speed_?
https://www.aqua-calc.com/what-is/speed/second-per-meter

Aren't these just different units of speed, and could be added to `Speed.json` instead?

Also, please only add units that are widely used. I don't know of many use cases for pace other than running, where minutes per kilometer is used. I suspect only a handful of the units you added here are going to be used by anyone.

angularsen (changes requested):

Common/UnitDefinitions/InverseSpeed.json
mcgougwm: I guess pace would be another way of looking at it. It's the inverse of speed, and I acknowledge that not everything I added is widely used. I'm not sure if it would make sense adding these to Speed.json or not. The conversions between units would be easy enough

angularsen: FYI, we are adding #954 .
It seems related, but I am still not sure whether `InverseSpeed` warrants its own quantity or if we could instead add some pace-related units to `Speed`.

I can't find any wiki or similar sources on `reciprocal speed` or `inverse speed`, so I still think this is maybe not its own quantity but rather extra units of `Speed`?

PR #945, opened 2021-06-17
Improve generation on nuspecs for .NET nanoFramework

josesimoes:
- Mscorlib version is now set to the latest available version.

(Math version was already being properly handled, for some reason mscorlib was wrongly left with a fixed version)

### DIFF
--- a/CodeGen/Generators/NanoFrameworkGen/NuspecGenerator.cs
+++ b/CodeGen/Generators/NanoFrameworkGen/NuspecGenerator.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Text;
 using CodeGen.JsonTypes;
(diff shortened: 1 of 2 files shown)


angularsen:
Great thanks. One thing I've observed is that the nanoFramework versions are bumped for unrelated pull requests causing a lot of spam, so I think maybe we should make it a separate script to explicitly run in order to fetch the latest versions - instead of making it part of the default codegen.

I might look into it, but wanted to mention it in case you would like to take a look at it.

josesimoes:
> Great thanks. One thing I've observed is that the nanoFramework versions are bumped for unrelated pull requests causing a lot of spam, so I think maybe we should make it a separate script to explicitly run in order to fetch the latest versions - instead of making it part of the default codegen.
> 
> I might look into it, but wanted to mention it in case you would like to take a look at it.

Yeap. As it is now all versions (actually the only two: math and mscorlib) are now fully automated. They are comming from the latest version available in nuget. So this should be working as you're describing.

BTW, this release cadence is not usual. It just happened that we had a lot of fixes comming in after adding unit tests. 😉

Usually there are months between releases of mscorlib.

PR #950, opened 2021-07-05
Fix nanoFramework assembly file version

josesimoes:
- Add call to Set-AssemblyInfoVersion because it wasn't bumping the version of nanoFramework AssemblyInfo.cs file as it should.
- Remove app.config files from 3 projects. These should never, ever be there, otherwise the references won't work properly.


josesimoes:
@angularsen found the above when debugging an issue with the Pressure nuget package. The current version can't be updated with nuget because of that app config and the fact that the assembly and file versions have been set to 1.0.0.0 since the 1st release.

After reviewing this if you could publish a release bumping the lowest possible value, that would be very helpful to get past this. 😉

angularsen:
Done, bumped patch version to
[Release UnitsNet/4.97.1 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.97.1)

josesimoes:
> Done, bumped patch version to
> [Release UnitsNet/4.97.1 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.97.1)

Super!! Thank you very much! 😃

PR #955, opened 2021-07-21
Add ReciprocalArea

bc913:
Added Reciprocal (Inverse) Area quantity and the corresponding units. Check this [link](https://en.wikipedia.org/wiki/Inverse-square_law) for further info. It is also used in fluid dynamics.

- Implemented for squaremeters, squareacentimeters, squareamillimeters, squarekilometers, squaredecimeters, SquareMicrometers, squaremiles, squareyards, squarefeet, squareInches and UsSurveySquareFeet. These units are selected to match with existing Area units.
-  Some arithmetic operator overloads are provided for Force, Area, Pressure and ReciprocalLength  to interact with the ReciprocalArea quantity for convenience
- Inverse method is implemented for Area quantity for convenience.


angularsen:
@bc913 I see the natural connection with ReciprocalLength and the wiki helps me see some value in adding this, but I am concerned about not adding too many inverse/reciprocal versions of quantities too.

How many more quantities do you expect to add reciprocal versions of? 
Is Volume a candidate? 
https://unitconverterapp.com/quantity/reciprocal%20volume

bc913:
@angularsen I totally understand your concern about bloating the repo but based on my experiences throughout several engineering disciplines, the usages of the reciprocal length and the reciprocal area  quantities are not so uncommon. They are useful quantities to represent some direct or derived physical representations. I think these are useful.

The reciprocal angle can also be considered as part of my argument I mentioned above but I'm not planning to add it or any other reciprocal quantity in near future.  The reciprocal area is the last reciprocal quantity I'd propose.

No, volume is not a candidate.

tmilnthorp:
We should probably figure out a more generic way in the future. Perhaps a struct that wraps an IQuantity and can somehow invert the units automatically? We'd have to change a few things around rather than purely string based (probably an exponent per unit).

But not to derail this PR :)

UnitsNet.Tests/CustomCode/AreaTests.cs
```diff
@@ -118,5 +118,26 @@ public void ToUnit_GivenSIUnitSystem_ReturnsSIQuantity()
...
+        [InlineData(-2.0, -0.5)]
+        [InlineData(0.0, 0.0)]
+        [InlineData(3.0, 3.0)]
+        [InlineData(2.0, 0.5)]
+        public void InverseReturnsReciprocalArea(double value, double expected)
+        {
+            var area = new Area(value, AreaUnit.SquareMeter);
+            var inverseArea = area.Inverse();
+
+            AssertEx.Equals(expected, inverseArea.InverseSquareMeters);
```
angularsen: I couldn't for the life of me understand why these tests passed.

Turns out `AssertEx.Equals` does not exist, so this wound up calling `System.Object.Equals()` and this naturally does not throw an assertion exception. 😄

Pushed a fix now.

angularsen:
Nuget is on the way out.

By the way, #940 on inverse angle was closed earlier to avoid bloat, but I've since learned that reciprocal length/area/angle are maybe the most common ones worth adding so if you or the PR author wants to add it I'm happy to merge it.

[Release UnitsNet/4.100.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.100.0)

Issue #959, opened 2021-08-09
UnitsNet installs System.ValueTuple into net47 or higher targets


tmilnthorp:
**Describe the bug**
UnitsNet installs System.ValueTuple into net47 or higher targets

**To Reproduce**
Steps to reproduce the behavior (just an example):
1. Add nuget UnitsNet 4.9.0 to a .NET 4.7 or higher project
2. See it installs System.ValueTuple

**Expected behavior**
.NET Framework 4.7 and higher has ValueTuples [include in mscorlib](https://docs.microsoft.com/en-us/dotnet/api/system.valuetuple?view=netframework-4.7)

PR #960, opened 2021-08-09
Only reference System.ValueTuple in net40

tmilnthorp:
Fix for #959. Only reference System.ValueTuple for net40 target. Add net47 target that doesn't need to reference it.


angularsen:
[Release UnitsNet/4.101.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.101.0)

PR #964, opened 2021-08-15
Rework .NET nanoFramework code generator and references update

### LINKED ISSUE #962: Buid failing for .NET nanoFramework
**Describe the bug**
- Build is failing after #958 was merged.

josesimoes:
- Updates to mscorlib and math NuGet packages are now performed using nuget CLI.
- NanoFrameworkGenerator fields for mscorlib and math are now self-updating.
- Rework project, packages and nuspec generators to use class fields.
- Closes #962.


CodeGen/Generators/NanoFrameworkGenerator.cs
```diff
@@ -126,85 +122,183 @@ public static void Generate(string rootDir, Quantity[] quantities, bool updateNa
...
+                    Arguments = $"update {path}\\UnitsNet.nanoFramework.sln -PreRelease",
+                    UseShellExecute = false,
+                    CreateNoWindow = true
+                }
+            };
+
+            nugetCLI.Start();
+
+            // start nuget CLI and wait for exit
+            if (!nugetCLI.Start())
```
angularsen, line 155: Aren't you calling Start twice here?

CodeGen/Program.cs
```diff
@@ -77,7 +77,19 @@ public static int Main(bool verbose = false, DirectoryInfo? repositoryRoot = nul
                 if (!skipNanoFramework)
                 {
                     Log.Information("Generate nanoFramework projects\n---");
-                    NanoFrameworkGenerator.Generate(rootDir, quantities, updateNanoFrameworkDependencies);
+                    NanoFrameworkGenerator.Generate(rootDir, quantities);
+
+                    // after generating code, force a dependencies update to make sure it's using the latest NuGet packages
+                    updateNanoFrameworkDependencies = true;
```
angularsen, line 83: Not sure about this line. The default behavior when running `generate-code.sh` is to also generate code for nanoFramework. If I understand it right, it will then also bump the version numbers for nanoFramework in the codegen if newer versions are available.

When someone creates a PR on a new quantity or unit, this will just cause confusion as this change is entirely unrelated to the PR.

CodeGen/Generators/NanoFrameworkGenerator.cs
```diff
@@ -126,85 +122,183 @@ public static void Generate(string rootDir, Quantity[] quantities, bool updateNa
...
-                // making an assumption here that the available version is referencing the correct mscolib
-                var mathVersion = packageVersions.OrderByDescending(v => v).First();
-                MathVersion = mathVersion.Version.ToString();
-                MathNuGetVersion = mathVersion.ToNormalizedString();
-
-                logger.LogInformation($"Referencing nanoFramework.CoreLibrary {MscorlibNuGetVersion}");
-                logger.LogInformation($"Referencing nanoFramework.System.Math {MathNuGetVersion}");
+                StartInfo = new ProcessStartInfo
+                {
+                    FileName = "nuget.exe",
```
angularsen, line 145: I don't have nuget.exe in my PATH on Windows so this fails to run for me.
There is one in `Tools\NuGet.exe` we could use though. If it's too old, just update it.

image

CodeGen/Generators/NanoFrameworkGenerator.cs
```diff
@@ -126,85 +122,183 @@ public static void Generate(string rootDir, Quantity[] quantities, bool updateNa
...
+        public static void UpdateNanoFrameworkDependencies(
+            string rootDir,
+            Quantity[] quantities)
         {
-            if (updateNanoFrameworkDependencies)
+            // working path
+            string path = Path.Combine(rootDir, "UnitsNet.NanoFramework\\GeneratedCode");
+
+            Log.Information("");
+            Log.Information("Updating .NET nanoFramework references using nuget CLI");
```
angularsen, line 138: I ran `generate-code.sh` locally, and it seems to downgrade from stable to preview package?

image

angularsen (changes requested):
Found some issues in my testing

CodeGen/Generators/NanoFrameworkGenerator.cs
angularsen: Seems related to Tools/NuGet.exe version being too old and `nuget update` failed silently in codegen.
I pushed some commits updating it from 4.9.3 to 5.10.0 and then it was able to update the nuget versions.

CodeGen/Generators/NanoFrameworkGenerator.cs
angularsen: Pushed a commit with updated nuget.exe and using this from codegen.

CodeGen/Program.cs
angularsen: Also, while this works, it seems a bit meta for the codegen to update its own source code.
Can't we look up the current versions from .nfproj or .nuspec files instead of having it stored in codegen source code?

CodeGen/Generators/NanoFrameworkGenerator.cs
angularsen: Actually, I can't seem to push any changes to your fork for some reason. Have you locked it?
I pushed the commits to a branch on my repo.
https://github.com/angularsen/UnitsNet/commits/work-962

CodeGen/Generators/NanoFrameworkGenerator.cs
angularsen: See above, I couldn't push commits to this fork.

CodeGen/Generators/NanoFrameworkGenerator.cs
josesimoes: Ooops... Yes. My bad. This poped after a final refactor and slipped my review...

CodeGen/Program.cs
josesimoes: Yes, when adding a new quantity it will run the update references as part of the process. 
This intent was to have it automatically checked and updated to be sure it was using the latest available version.

As for the auto-update, yes again, it was an intended to have the class updating itself to always have the latest version available when the generated code step was run.

And yes, as it is now it grabs the "current" base version strings from the nfproj in order to fill in the nuspec and packages files.
When pulling the NuGet versions from nuget.org we only have access to the package version. The assembly version (that's used on the nfproj file here: _<Reference Include="mscorlib, Version=**1.10.5.4** ..._  can't be pulled from anywhere except from the DLL file itself. The previous update strategy had this flaw, that's why it ended up failing like it did. Because the automated update was only updating the nuget version. Leaving it up to the human doing it the job of updating the version (which wasn't being done).

But nuget CLI is smart enough to handle that and it does update both versions: assembly and nuget. It's therefore a superior way of dealing with the automated updates. 

If you prefer, I can remove this part of the self update of the base versions. Or we can have it not run automatically when generating code. 
Let me know. 😄

CodeGen/Generators/NanoFrameworkGenerator.cs
josesimoes: Just pulled your changes.

CodeGen/Generators/NanoFrameworkGenerator.cs
josesimoes: Got you! Merged.

CodeGen/Program.cs
angularsen: @josesimoes 
Thanks for clarifying. I may not have been clear in what I was trying to point out.

1. You are now hard coding `internal static string MscorlibVersion = "1.10.5.0";` etc. instead of looking it up from `Angle.nfproj`.
2. Codegen is run, using hardcoded values. Project files, nuspec files, packages.config files, etc.
3. The new version is determined by `nuget.exe update` instead of nuget.org APIs.
4. The new version is stored in `NanoFrameworkGenerator.cs`, replacing the hard coded value. (Presumably used on next codegen).
5. Nuspec files are written (again) looking up versions from `packages.config` after the `nuget.exe update`.

I believe step 1 & 4 could be dropped. 

I pushed a suggestion to this in my branch: d610dfe4b


Just to comment on the `nuget update` approach vs nuget API. Overall, I do like that it is simpler. One caveat is that it is much slower and takes about 30-40 seconds on my machine. It seems to spend the time updating N projects in sequence. As long as this does not run on every codegen, I think it's fine. Another option is to go the nuget API route and download the nuget and inspect the .dll to find the assembly version as you described. I'm fine with either way.

josesimoes: Glad that we're getting there! 

> 1. You are now hard coding `internal static string MscorlibVersion = "1.10.5.0";` etc. instead of looking it up from `Angle.nfproj`. ✅ 

> 2. Codegen is run, using hardcoded values. Project files, nuspec files, packages.config files, etc.
✅  

> 3. The new version is determined by `nuget.exe update` instead of nuget.org APIs. ✅ 

> 4. The new version is stored in `NanoFrameworkGenerator.cs`, replacing the hard coded value. (Presumably used on next codegen). ✅ 

> 5. Project and nuspec files are written to (again) by looking up versions from `packages.config` after the `nuget.exe update`. ✅ 

 
> I believe step 1 & 4 could be dropped.

By dropping this "memory" field we are left with parse one of the existing nfproj files to grab the "last" version and use that to generate the various files for the current iteration. It seems a bit more inefficient, but I get it. You don't want the code changing every time the version changes. 

> I pushed a suggestion to this in my branch: [d610dfe](https://github.com/angularsen/UnitsNet/commit/d610dfe4b03b9a114e00bcc8cc5ee6f7f9c1b7cf)

Got you! Looks good to me. Let's go with it. 


> Just to comment on the `nuget update` approach vs nuget API. Overall, I do like that it is simpler. One caveat is that it is much slower and takes about 30-40 seconds on my machine. It seems to spend the time updating N projects in sequence. As long as this does not run on every codegen, I think it's fine. Another option is to go the nuget API route and download the nuget and inspect the .dll to find the assembly version as you described. I'm fine with either way.

As it is, it will take the "long run" (to perform the nuget CLI run) only if the `updateNanoFrameworkDependencies` is passed. Which I believe it's OK.

And despite the longest execution time using nuget CLI I believe it's easier and requiring less code maintenance, when compared with the approach of having to call the nuget API, download the package, extract it, peek the version from the dll file! 😅

angularsen: Perfect, I think we agree then 🤝 😄

PR #905, opened 2021-02-20
WIP: Add serialization support for System.Text.Json

rgomez90:
Add support for serialization using System.Json.Text

Closes #901

PR #966, opened 2021-08-30
Add serialization support for System.Text.Json (Attempt #2)

### LINKED ISSUE #901: Add serialization support for System.Text.Json

There is no serialization support via `System.Text.Json` so users using this library instead of  `Newtonsoft Json` will need to write their own converters. It would be nice if the library would have built-in support for this.

Add Serialization support for System.Text.Json

**Describe alternatives you've considered**

Use the built-in feature with Newtonsoft

akaegi:
Second attempt for System.Text.Json support.
Should close #901. 

Tried to adress issues mentioned by @chucker in #905 (thanks!).
The problem lies in non-polymorphic deserialization IMO. The original deserialization failed when the constructor had an IQuantity subtype as argument (e.g. Information) but the converter converted to IQuantity. I addressed that by returning specific converters in the factory as described in https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-converters-how-to?pivots=dotnet-5-0#support-round-trip-for-stackt.

Further notes:
- Carefully tried to extract common serialization/deserialization of JsonNet and System.Text.Json in a common project "UnitsNet.Serialization"
- UnitsNetIComparableJsonConverter is still missing -> time is running out...

Hope it works and helps :-)


chucker:
(Haven't tried this yet, but thanks for giving it a shot!)

angularsen:
FYI, #972 was merged.

Also, there was some discussion that reveals our current JSON schema is maybe not optimal.
`ValueType` and `ValueString` properties are likely not necessary in order to preserve `decimal` precision:
https://github.com/angularsen/UnitsNet/pull/972#discussion_r730490511

dmytro-gokun:
@angularsen Is there any ETA on this? Maybe some help is required? We miss this feature a lot, it's the only thing that holds us back from migrating a big project to System.Text.Json :)

angularsen:
@dmytro-gokun 
ETA is whenever someone implements it 😄 

If you want to help out, then great! I am not actively working on it myself, but @lipchev has done quite a bit of work on json lately.

Some relevant background:

https://github.com/angularsen/UnitsNet/pull/972#discussion_r730490511
https://github.com/angularsen/UnitsNet/pull/985
https://github.com/angularsen/UnitsNet/issues/990

In particular, we see that maybe the current JSON schema can be simplified a bit.

@lipchev Do you plan to attack system.text.json anytime soon?

maxreb:
Okay I found another solution how this will work with STJ. Install NuGet package [ZCS.DataContractResolver](https://github.com/zcsizmadia/ZCS.DataContractResolver) and then replace the default `TypeInfoResolver` with the `DataContractResolver`. JST is now using the `DataContract`s:
```
//LevelSensorData is a class which implements Units.NET quantities
var opt = new JsonSerializerOptions() { TypeInfoResolver = System.Text.Json.Serialization.Metadata.DataContractResolver.Default };
var str = JsonSerializer.Serialize(originalObject, opt);
var obj = JsonSerializer.Deserialize<LevelSensorData>(str, opt);
```

angularsen:
Thanks for sharing @maxreb . Yes, DataContract should be one way to achieve this.

Another option is to implement custom converters, as this PR originally tried to. I believe that gives more control on the serialized format.

Issue #831, opened 2020-09-04
Dynamic Unit Selection


AndyNetDuma:
**Is your feature request related to a problem? Please describe.**
I would like to take a unit and write it with the simplest unit available. I.E. 1000mb should become 1gb. Ideally with 1024 bit division support. Very sorry if this is already a feature and I missed it, seems hard to believe nobody needed this before

**Describe the solution you'd like**
An overload of tostring or perhaps another function that finds the best unit before conversion. For mb it would require a concept of bits and bytes as it some cases I wish to show 1 or the other.
Another option is to take a unit and to provide a function to set the optimal base unit

**Describe alternatives you've considered**
We currently use a custom conversion system which is messy, I am hoping UnitsNet will do much better however without this vital feature it's, unfortunately, effectively useless to us. So sticking with that is the current alternative.

**Additional context**
An example:
```
var a = new UnitsNet.Information(2000, InformationUnit.Megabit);
var b = a.ToString(); // Gives 2000mb
var c = a.ToOptimalUnit(InformationUnit.Bit); // Should give 2gb in bits
var c = a.ToOptimalUnit(InformationUnit.Bytes, 1); // Should give ~2gb in bytes. 1 is optional decimal points
```

angularsen:
Hi and thanks for the suggestion.
This is not implemented today, but if you are interested in doing a pull request then I think this would be a welcome feature.

We already have the following convention `x.ToUnit(whatUnitToConvertTo).ToString(paramsOnHowToPresentText)` so building on your example, how about something like this:

Output:
```
2,100 Mb
2,100 MB
2,148 Mib
2,148 MiB

ToUnit:
2.1 Gb
2.1 GB
2.1 Gib
2.1 GiB

ToOptimalUnit (inferred binary multiple):
2.1 Gb
2.1 GB
2.1 Gib
2.1 GiB

ToOptimalUnit (explicit binary multiple):
2.1 Gb
2.1 GB
2.1 Gib
2.1 GiB
```

LINQPad example:
```c#
void Main()
{
	var a = Information.FromMegabits(2100).Dump();
	var b = Information.FromMegabytes(2100).Dump();
	var c = Information.FromMebibits(2148).Dump();
	var d = Information.FromMebibytes(2148).Dump();
	
	
	// From this
	Console.WriteLine("\nToUnit:");
	a.ToUnit(InformationUnit.Gigabit).ToString().Dump();
[...]

PR #968, opened 2021-09-18
Overload * operator for BitRate and Duration

breckognize:
The change to overload the * operator for BitRate and Duration is simple, but doing so broke unit tests. To fix them, I needed to make Information a BaseDimension, which is the bulk of the diff. I believe I've done that successfully, but I confess I don't understand why "BaseDimensions" exists. (Does it enable a feature? Does it offer redundancy for testing?)

### DIFF
--- a/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
@@ -36,7 +36,8 @@ namespace CodeGen.Generators.UnitsNetGen
                               baseDimensions.I == 0 &&
                               baseDimensions.Θ == 0 &&
                               baseDimensions.N == 0 &&
-                              baseDimensions.J == 0;
+                              baseDimensions.J == 0 &&
+                              baseDimensions.b == 0;
 
         }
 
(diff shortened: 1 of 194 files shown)


angularsen:
Hi and thanks for contributing. It seems you hit an edge case here, sorry about that.
Some quantities, like BitRate and RatioChangeRate, were incorrectly configured as dimensionless quantities.
I pushed a fix now: 7b60facded2fa2e8ca2477726933c2ade781c0e0

If you could please try this PR again, you should really only have to add your BitRate.extra.cs file, no other changes should be necessary. Let me know if you still meet a wall. 

As for the intent behind BaseDimensions, it was an experiment to see if we could support dynamic conversions like `IQuantity * IQuantity` where the intermediate calculations may involve `(Length * Length * Length * Length) / Length` that have no strongly typed quantity for the representation inside the parantheses, but the end result is a `Volume`.

BaseDimensions is only intended for SI derived quantities, and as mentioned, it was an experiment we haven't fully explored or reaped the benefits from yet.

Main discussions
#651 
#709

Other related topics from a quick search
#630 
#524 
#831 
#588 
#587 
#180 

Best, Andreas.

Issue #780, opened 2020-04-25
Recommended way to create 3 dimensional magnetic field?


krwq:
We're currently considering to use this library in https://github.com/dotnet/iot/ . We have initially started developing very similar APIs and then noticed this library so we decided to give it a try since it already solves many problems we were about to hit.

We have converted our Temperature/Pressure structs to this library without any issues (see https://github.com/dotnet/iot/pull/1052) but now we're trying to use 3-dimensional magnetic field which is the output of IMU (i.e. BNO055 sensor). We haven't ourselves developed such unit yet and used Vector3.

What's the recommended way to achieve 3-dimensional magnetic field?

pgrawehr:
Do I understand correctly that you would like to have a `Vector3<MagneticField>` or `Vector3<Speed>` instance? I guess that will end up with the same problems that were also discussed in #695. 

Consider this code:
```
public class Vector3<T>
{
    private T _value;
    public Vector3(T value)
    {
        _value = value;
    }
    
    public static Vector3<T> operator+(Vector3<T> a, Vector3<T> b)
    {
        return new Vector3(a._value + b._value); // ERROR: T does not have an operator +
    }
    // ....
}
```

Unfortunatelly the above doesn't work, because the generic type "T" (aka any type) does not have a defined + operator. There's no way to add a type constraint on "something that is a number". And declaring interfaces (`IImplementsAddition`) for operator methods is also not possible, because in C# they need to be static by definition 😢 ...

This works:
```
    public static Vector3<T> operator+(Vector3<T> a, Vector3<T> b)
    {
        dynamic a1 = a;
        dynamic b1 = b;
        return new Vector3(a1._value + b1._value); // Works!
    }
```
But of course, this comes with a performance penalty (I don't know how big it is) and the downside that problems will pop up only at runtime and not at compile time.

angularsen:
Hi,

UnitsNet does not currently support XYZ-dimensional quantities in a _generic_ way.
Some discussion in #695 on why this was not straight forward to add, but it is absolutely possible if someone wants to champion it and a performant proposal is outlined in #666.

You can, however, easily add your own wrapper types for specific quantities.
Something like this:
```c#
public struct MagneticField3
{
    public MagneticField X { get; }
    public MagneticField Y { get; }
    public MagneticField Z { get; }
    
    public MagneticField3(MagneticField x, MagneticField y, MagneticField z) { ... }

    // Operator overloads for arithmetic
[...]

PR #977, opened 2021-10-14
Added implicit conversion from Scalar to double

cedricmoers:
- Added implicit conversion from Scalar to double
- Made exception for ScalarTestBase to ignore the Convert_ToBool_ThrowsInvalidCastException test for Scalar

### DIFF
--- a/CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
@@ -545,6 +545,10 @@ namespace UnitsNet.Tests
             Assert.Equal(quantity.ToString(CultureInfo.CurrentUICulture, ""g""), quantity.ToString(null, ""g""));
         }}
 
+");
+            if (!_quantity.Name.Equals("Scalar"))
+            {
+                Writer.WL($@"
 
         [Fact]
         public void Convert_ToBool_ThrowsInvalidCastException()
@@ -552,6 +556,9 @@ namespace UnitsNet.Tests
             var quantity = {_quantity.Name}.From{_baseUnit.PluralName}(1.0);
             Assert.Throws<InvalidCastException>(() => Convert.ToBoolean(quantity));
         }}
+");
+            }
+            Writer.WL($@"
 
         [Fact]
         public void Convert_ToByte_EqualsValueAsSameType()
--- a/UnitsNet.Tests/CustomCode/ScalarTests.cs
+++ b/UnitsNet.Tests/CustomCode/ScalarTests.cs
@@ -27,6 +27,18 @@ namespace UnitsNet.Tests.CustomCode
         // Override properties in base class here
         protected override double AmountInOneAmount => 1;
         protected override bool SupportsSIUnitSystem => false;
-       
+
+        [Theory]
+        [InlineData(0)]
+        [InlineData(1.0)]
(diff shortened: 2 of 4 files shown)


cedricmoers:
I also wonder if the implicit conversion from double to Scalar should also be implemented in a similar way:

`
        /// <summary>Defines an implicit conversion of a <see cref="Scalar"/> to a <see cref="double"/>.</summary>
        public static implicit operator double(Scalar scalar)
        {
            return scalar.Amount;
        }

        /// <summary>Defines an implicit conversion of a <see cref="double"/> to a <see cref="Scalar"/>.</summary>
        public static implicit operator Scalar(double value)
        {
            return Scalar.FromAmount(value);
        }

`

Also, should this mean that all arithmetic operations should be based on Scalar instead of Double? For example, if you divide a voltage by a voltage, would you then rather return a 'double', or a Scalar?

cedricmoers:
In my scenario I use (and love) UnitsNet for scientific measurements. I have some value processing blocks that are by nature 'unit independent'. One of these blocks is for example a value processor that returns an average value of multiple measurements:

```
    public interface IValueProcessor
    {
        IQuantity Get(Func<IQuantity> quantities);
    }

    public class Average
    {
        public int Count { get; set; }
        IQuantity Get(Func<IQuantity> quantityProvider)
        {
            int count = Count;
            IQuantity sum = null;
            for (int i = 0; i < count; i++)
            {
                var q = quantityProvider();
                if (sum is null)
                    sum = q;
                else
                    sum = (dynamic)sum + q;
            }

            IQuantity avg = (dynamic)sum / count;

            return avg;
        }
    } 
```

The scalar is a big improvement here, since before the Scalar, I had to create a separate class for processors taking quantities and processors taking int/double/decimal, ... Before this, I had to use 'object' as the common class and check if the object was an IQuantity or a double/int/decimal. This became a bit ugly and I created a 'wrapper': IQuantifiable and some helper Structs. I never was happy with that and Scalar is likely to solve that.

Currently I see one (edge-)usecase that would benefit of returning the Scalar from arithmetic operations. Imagine I have a processor that does a Division on a given quantity that I do not know the type of:

```
    public class Divider
    {
        public IQuantity DivideBy { get; set; }
        IQuantity Get(Func<IQuantity> quantityProvider)
        {
            var q = quantityProvider();
            return (dynamic)q / DivideBy;
        }
    }
```

If the measured quantity is 5V and I want it divided by 2V, then the result I expect is 2.5 (no unit). Unfortunately, because the division in the ElectricPotential struct returns a 'double', which is not a Quantity, and thus the above will throw an InvalidCastException. If the division would return a Scalar, the above (I think) would work fine. 

If the expected result is in fact a double, then the implicit cast will do the correct casting for you.

Thus I see an advantage in returning a Scalar from arithmetics when normally a 'double' is returned, but I cannot oversee what the impact will be and how that will break the current system. I am willing to try this out.

angularsen:
@cedricmoers Some more comments on Generic Math.
https://github.com/angularsen/UnitsNet/issues/780#issuecomment-958241530

angularsen:
I'm sorry, what is the status of this PR again? 

It seems my last suggestion was
> I guess, if we can change all arithmetic method overloads to return Scalar, and provide implicit casts to double, then hopefully 🤞 all existing tests and people's application code will compile and run exactly as before?

Should we try changing all arithmetic methods retuning `double` to return `Scalar` instead? Or should we merge this PR as-is?

PR #980, opened 2021-10-30
CHG : Remove explicit support for net40 and net47

ebfortin:
net47 is supported defacto through netstandard2.0. See reference : https://docs.microsoft.com/en-us/dotnet/standard/net-standard


tmilnthorp:
> net47 is supported defacto through netstandard2.0. See reference : https://docs.microsoft.com/en-us/dotnet/standard/net-standard

Yes but there are issues in the transient reference to the ValueTuple types depending on the runtime. See #960

tmilnthorp:
And #959

angularsen:
@tmilnthorp can you give latest 5.0 nuget a try and see if this is still a problem now that we removed all targets but netstandard2.0?

PR #981, opened 2021-10-30
CHG: Removed explicit support for net40 and net47.

ebfortin:
net47 support is done through netstandard2.0 support.


angularsen:
@tmilnthorp We're removing `net47` target, that you added in #960. I believe it should no longer be necessary, now that `net40` is removed. Also, a v5 branch is now rolling and will publish pre-release nugets moving forward that you can test with.

angularsen:
@ebfortin merged into v5 branch. Working on getting pre-release nugets out.

PR #984, opened 2021-11-02
✨ Experiment with Generic Math in .NET6

angularsen:
Experimenting with https://devblogs.microsoft.com/dotnet/preview-features-in-net-6-generic-math/

Example:
```cs
public interface IQuantity<TQuantity> :
        IAdditionOperators<TQuantity, TQuantity, TQuantity>,
        IParseable<TQuantity>,
        IComparisonOperators<TQuantity, TQuantity>,
        IAdditiveIdentity<TQuantity, TQuantity>
    where TQuantity : IQuantity<TQuantity>
{
    static abstract TQuantity Zero { get; }
}
```

Allows us to do:
```cs
void Foo<TQuantity>(TQuantity left, TQuantity right)
    where TQuantity : IQuantity<TQuantity>
{
    WriteLine(left.CompareTo(right));
    WriteLine(left + right);
    WriteLine(TQuantity.Parse("25 m", null);
}

TQuantity Sum<TQuantity>(IEnumerable<TQuantity> items)
    where TQuantity : IQuantity<TQuantity>
{
    return items.Aggregate(TQuantity.Zero, (acc, item) => acc + item);
}
```

Full sample output:
```
Foo(1 Meter, 10 Centimeter)
---

1                   <==  left.CompareTo(right)
1.1 Meter           <==  left + right
25 Meter            <==  TQuantity.Parse("25 m", null)
10 Meter            <==  Sum [1,2,3,4] array of meters
```

### DIFF
--- a/ConsoleNet6/ConsoleNet6.csproj
+++ b/ConsoleNet6/ConsoleNet6.csproj
@@ -1,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net6.0</TargetFramework>
+        <LangVersion>preview</LangVersion>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+  <!-- .NET6 Experimental! -->
+  <ItemGroup>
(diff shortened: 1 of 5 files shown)


tmilnthorp:
Love it. It doesn't seem like there is a way to make it backwards compatible to netstandard, etc. though :(

angularsen:
Merged by #1164

PR #986, opened 2021-11-13
Use CurrentCulture rather than CurrentUICulture

### LINKED ISSUE #787: Default number format should be CultureInfo.CurrentCulture, not CurrentUICulture

pgrawehr:
Fixes #787

Replaces #795 

This brings the changes agreed on in #795 to the v5 release branch. This might be a breaking change for whoever relies on the exiting, but unusual behavior.

FYI: @tmilnthorp

### DIFF
--- a/UnitsNet/CustomCode/UnitAbbreviationsCache.cs
+++ b/UnitsNet/CustomCode/UnitAbbreviationsCache.cs
@@ -70,7 +70,7 @@ namespace UnitsNet
         /// <typeparam name="TUnitType">The type of unit enum.</typeparam>
         public void MapUnitToAbbreviation<TUnitType>(TUnitType unit, params string[] abbreviations) where TUnitType : Enum
         {
-            MapUnitToAbbreviation(typeof(TUnitType), Convert.ToInt32(unit), CultureInfo.CurrentUICulture, abbreviations);
+            MapUnitToAbbreviation(typeof(TUnitType), Convert.ToInt32(unit), CultureInfo.CurrentCulture, abbreviations);
         }
 
         /// <summary>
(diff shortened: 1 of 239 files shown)


UnitsNet/CustomCode/UnitAbbreviationsCache.cs
```diff
@@ -300,7 +300,7 @@ internal bool TryGetUnitValueAbbreviationLookup(Type unitType, IFormatProvider?
         {
             unitToAbbreviations = null;
 
-            formatProvider ??= CultureInfo.CurrentUICulture;
+            formatProvider = formatProvider ?? CultureInfo.CurrentCulture;
```
angularsen, line 303: Is there a reason you reverted this syntax? Not significant, just curious.

angularsen (approved):

angularsen:
Great work! You even added a few improvements 🙌

angularsen:
Alpha 5.0 nuget should be on the way out shortly.

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
pgrawehr: Probably a merge conflict of some kind.

Issue #988, opened 2021-11-16
Temperature Gradient


dglozano:
Hi 👋🏻 I am using UnitsNet quite a lot and I really love it! So far it has covered most of my use cases, but I recently I got the need to use a `TemperatureGradient` quantity (https://en.wikipedia.org/wiki/Temperature_gradient). 

I know there is `LapseRate` quantity which is also `TemperatureDelta / Length`, but it seems to be "semantically" focused on the atmospheric temperature variation and it supports only the `DegreesPerKilometer` unit. 

I was wondering if adding a new `TemperatureGradient` is an option? If so, I would be willing to create a PR to add it.

Another option would be to make the `LapseRate` support more units I guess, but both the name and description of the quantity can be a little misleading.

Related (ish) to #324

angularsen:
Hi, without knowing this domain I would favor moving everything to TemperatureGradient. It might be more consistent with other quantities to name it TemperaturePerLength, but it seems TemperatureGradient is a well established quantity and maybe the better choice .

This quantity has been touched by in:
#316
#321 
#324 

@ferittuncer @jnyrup 
What do you guys think about merging it into TemperatureGradient (or TemperaturePerLength), and marking LapseRate as obsolete?

There is currently a v5 branch where we allow breaking changes, where we can remove anything made obsolete in v4.
#982

jnyrup:
I'm fine with obsoleting `LapseRate` as it is a quite meteorology specific unit and a special case of a temperature gradient. 
> It corresponds to the vertical component of the spatial gradient of temperature
https://en.wikipedia.org/wiki/Lapse_rate

Some observations regard naming.
UnitsNet has several `PerLength` units and only a single `Gradient` unit (`ElectricCurrentGradient`).
Googling a bit it seems TemperatureGradient is an established term.
UnitsNet has `ForcePerLength` for [Surface Tension](https://en.wikipedia.org/wiki/Surface_tension).

angularsen:
@dglozano It seems we agree then, please go ahead with the PR:
- Obsolete LapseRate (Add `"ObsoleteText": "Use TemperatureGradient instead."` on the quantity in LapseRate.json and make sure an obsolete attribute is added to the generated LapseRate struct afterwards.)
- Add TemperatureGradient with your units plus LapseRate's units
- Copy any usages of LapseRate: https://github.com/angularsen/UnitsNet/search?q=lapserate
[...]

PR #991, opened 2021-11-21
Add TemperatureGradient and deprecate LapseRate

dglozano:
I have added a new Quantity called TemperatureGradient and deprecated LapseRate, as agreed in #988 

In addition, I did a small change in the CodeGen project to place the `[Obsolete]` attribute below the XML documentation, to avoid getting some warnings. 

This is how it was generated before my change
image

I have a small issue though: after cloning the repo, and before doing any change at all, I run the CodeGen and I see tons of changes being made. It seems that much of the code was alphabetically ordered before, and the CodeGen by default in my machine overrides that (check https://github.com/angularsen/UnitsNet/commit/a9235843b197913f1b5a62b1476c36d03ce047e3)

Any suggestion on how to solve this ?


angularsen:
> Any suggestion on how to solve this ?

Huh, that is super weird. Seems you are on MacOS or Linux maybe, I am on Windows. Could it be anything there?

It should be alphabetically sorted. Could you debug the codegen console app and maybe see why it produces a different ordering? Maybe there is some subtle difference on LINQ not preserving the ordering or something. I would expect adding a OrderBy() at the right place should fix this.

angularsen:
> In addition, I did a small change in the CodeGen project to place the [Obsolete] attribute below the XML documentation, to avoid getting some warnings.

Sounds good to me 👍

dglozano:
That's right, I am using Intellij on a Mac. I will debug the CodeGen later today and try to sort it out 😀

dglozano:
It seems the culprit of the sorting issue was the `Directory.GetFiles` method  that's used to read the quantity json files.
It doesn't guarantee the order of the files. It seems that the result in Windows is always sorted alphabetically, but that wasn't the case in Mac. 

Ref.: https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.getfiles?view=net-6.0
image

Have added the fix in this [commit](https://github.com/angularsen/UnitsNet/pull/991/commits/3dfe9a4913d25466671355f09ed478cd5d2ae599) together with a small renaming.

angularsen:
Great that you figured it out!

UnitsNet.Tests/CustomCode/TemperatureGradientTests.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    public class TemperatureGradientTests : TemperatureGradientTestsBase
+    {
+        protected override bool SupportsSIUnitSystem => true;
+
+        protected override double DegreesCelciusPerKilometerInOneKelvinPerMeter => 1000;
+
+        protected override double DegreesCelciusPerMeterInOneKelvinPerMeter => 1;
+
+        protected override double DegreesFahrenheitPerFootInOneKelvinPerMeter => 0.54864;
```
angularsen, line 33: Verified

https://unitconverterapp.com/quantity/temperature%20interval%20per%20length
https://www.calculand.com/unit-converter/?gruppe=Temperature+gradient

UnitsNet/UnitsNet.csproj
```diff
@@ -60,6 +60,9 @@
     <Compile Remove="UnitDefinitions\**" />
     <EmbeddedResource Remove="UnitDefinitions\**" />
     <None Remove="UnitDefinitions\**" />
+    <None Include="..\Common\UnitDefinitions\TemperatureGradient.json">
+      <Link>Common\UnitDefinitions\TemperatureGradient.json</Link>
+    </None>
```
angularsen: This was probably accidental, but I can't push to your master branch so I will fix after merging

UnitsNet.Tests/CustomCode/TemperatureGradientTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        public void TemperatureDeltaDividedByTemperatureGradientEqualsLength()
+        {
+            Length length = TemperatureDelta.FromDegreesCelsius(50) / TemperatureGradient.FromDegreesCelciusPerKilometer(5);
+            Assert.Equal(length, Length.FromKilometers(10));
+        }
+
+        [Fact]
+        public void TemperatureDeltaDividedByLengthEqualsTemperatureGradient()
+        {
+            TemperatureGradient lapseRate = TemperatureDelta.FromDegreesCelsius(50) / Length.FromKilometers(10);
```
angularsen, line 47: lapse rate name remains, will fix after merge

angularsen (approved):
Thanks! Looks good to me, just a few minor things I'll fix after merging.

angularsen:
Nuget should be out shortly.

[Release UnitsNet/4.106.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.106.0)

PR #992, opened 2021-11-22
Add ForceChangeRate arithmetic operators

### LINKED ISSUE #989: ForceChangeRate calculation operators
Many units have additional calculation operators which do not require you to work with unitless values.
However, there are no additional operators for the ForceChangeRate unit.

I propose adding the following operators:
```cs
public static Force operator *(Duration left, ForceChangeRate right); //Duration.extra.cs
public static Force operator *(ForceChangeRate left, Duration right); //ForceChangeRate.extra.cs
public static Duration operator /(Force left, ForceChangeRate right); //Force.extra.cs
```

Chippychap:
Fixes #989

I added the additional calculation operators for `ForceChangeRate`, as agreed in #989.
Additionally, I added the appropriate unit tests.

I noticed some differences between creating the new unit in various other operators, where some use the `FromXxx(value)` static factory methods and others use the constructors directly.
I used the constructors directly to reduce method calls, but I am wondering what is preferred.

### DIFF
--- a/UnitsNet.Tests/CustomCode/DurationTests.cs
+++ b/UnitsNet.Tests/CustomCode/DurationTests.cs
@@ -199,5 +199,12 @@ namespace UnitsNet.Tests.CustomCode
             Speed speed = Duration.FromSeconds(10) * Acceleration.FromMetersPerSecondSquared(10);
             Assert.Equal(Speed.FromMetersPerSecond(100), speed);
         }
+
+        [Fact]
+        public void DurationTimesForceChangeRate()
+        {
+            Force force = Duration.FromSeconds(10) * ForceChangeRate.FromNewtonsPerSecond(100);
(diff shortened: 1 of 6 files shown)


angularsen:
> I used the constructors directly to reduce method calls, but I am wondering what is preferred.

Doesn't really matter, we have tons of tests on both ctors and factory methods so these tests should focus on the operator overloads.

angularsen:
Sorry, my bad, I read wrong and thought you meant in the tests.
Yes, using ctors in the implementations is preferred as it reduces the call chain 👍

angularsen:
Great work, looked perfect to me 🙌

angularsen:
Nuget should be out shortly 

[Release UnitsNet/4.105.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.105.0)

PR #999, opened 2021-12-04
Update nanoFramework dependencies

josesimoes:
- Fix bug in path when calling nuget CLI.
- Automatic update of all libraries with latest stable versions:
  + nanoFramework.CoreLibrary.1.11.7
  + nanoFramework.System.Math.1.4.3

### DIFF
--- a/CodeGen/Generators/NanoFrameworkGenerator.cs
+++ b/CodeGen/Generators/NanoFrameworkGenerator.cs
@@ -134,7 +134,7 @@ namespace CodeGen.Generators
             {
                 StartInfo = new ProcessStartInfo
                 {
-                    FileName = "Tools/nuget.exe",
+                    FileName = Path.Combine(rootDir, "Tools/nuget.exe"),
                     Arguments = $"update {path}\\UnitsNet.nanoFramework.sln -PreRelease",
                     UseShellExecute = false,
                     CreateNoWindow = true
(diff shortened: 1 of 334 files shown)


josesimoes:
@angularsen adding to this, is it possible to hide the "5.0-alpha" versions that are available on NuGet? Because that offers a higher version than the current one (4.x), in our automated update scripts, nuget is using those instead of the 4.x ones... completly screws this for our CI-CD... 😢

angularsen:
Sorry for the delay, I thought I already merged this.

angularsen:
@josesimoes Is version range an option for you?
https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges

I can hide the already published NF pre-release nugets if we need to, but we would also have to modify the scripts on v5 branch to not bump the NF nugets to avoid the problem on next deploy.

josesimoes:
> Sorry for the delay, I thought I already merged this.

No problem! 😄

josesimoes:
> @josesimoes Is version range an option for you? https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges
> 
> I can hide the already published NF pre-release nugets if we need to, but we would also have to modify the scripts on v5 branch to not bump the NF nugets to avoid the problem on next deploy.

Thanks for the pointer. Let me check if that works for us... not 100% sure because of the automation...

josesimoes:
@angularsen the above doesn't work as we are relying on nuget CLI to find what to update and grab the version from the output. Still... I've tweaked our scripts so they only allow stable versions for UnitsNet, which does the trick and filters out the 5.0- alpha.

angularsen:
Ok, great that you found a way to work around it 👏

josesimoes:
@angularsen sorry to bother you again with this. It seems that the version being used it's the same as the previous one, so there was no package published to NuGet... can you please bump the version and rebuild?

angularsen:
I just had the same thought 😅 

[Release UnitsNet/4.110.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.110.0)

PR #1008, opened 2022-01-10
Add gram-force torque units

Angerxzer:
Maybe there's a mistake somewhere.


UnitsNet.Tests/CustomCode/TorqueTests.cs
```diff
@@ -52,6 +52,12 @@ public class TorqueTests : TorqueTestsBase
 
         protected override double PoundalFeetInOneNewtonMeter => 23.730360404231938;
 
+        protected override double GramForceCentimetersInOneNewtonMeter => KilogramForceCentimetersInOneNewtonMeter / 1000;
```
angularsen, line 55: Test case values should be constants, preferably looked up in some reliable source like google, wiki or online converter app. The idea is to double check the functions in the JSON.

Common/UnitDefinitions/Torque.json
```diff
@@ -91,6 +91,42 @@
         }
       ]
     },
+    {
+      "SingularName": "GramForceMillimeter",
+      "PluralName": "GramForceMillimeters",
+      "FromUnitToBaseFunc": "x*0.0000098066502",
```
angularsen, line 97: These functions are preferably computations instead of constants, to obtain the highest possible precision.
I assume the gravitational constant multiplied by some power of 10 should be used here?

angularsen (changes requested):
Thank you! Almost good to go, just a few minor comments.

Common/UnitDefinitions/Torque.json
Angerxzer: This time I traveled to  
[unitconverters.net ](url)
and did the conversion🤣

Common/UnitDefinitions/Torque.json
```diff
@@ -118,8 +118,8 @@
     {
       "SingularName": "GramForceMeter",
       "PluralName": "GramForceMeters",
-      "FromUnitToBaseFunc": "x*0.009806650199607",
-      "FromBaseToUnitFunc": "x*0.000101971619222",
+      "FromUnitToBaseFunc": "x*0.00980665",
+      "FromBaseToUnitFunc": "x*101.9716213",
```
angularsen, line 122: Standard gravity is defined as `9.80665`.
I believe this can be changed for this and the other units to ensure the full precision is kept:

```
      // GramForceMeter
      "FromUnitToBaseFunc": "x*9.80665e-3",
      "FromBaseToUnitFunc": "x/9.80665e-3",

      // GramForceMillimeter
      "FromUnitToBaseFunc": "x*9.80665e-6",
      "FromBaseToUnitFunc": "x/9.80665e-6",
```

I noticed KilogramForceMeter is defined as `9.80665019960652`, but that doesn't seem right to me? Out of scope for this PR, but I think it's odd that they are slightly different from the standard definition of the constant.

angularsen (changes requested):

Common/UnitDefinitions/Torque.json
Angerxzer: Yes, I also found it strange that "FromBaseToUnit" should be the division method, but I saw that KilogramForceMeter is like this, so I wrote it as it is

angularsen:
Thanks!
Nuget should be out shortly.

[Release UnitsNet/4.113.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.113.0)

PR #1022, opened 2022-01-28
Improve update of .NET nanoFramework

josesimoes:
- Add extra step to UpdateNanoFrameworkDependencies to restore .NET nanoFramework solution before update.
(this is required on usage cases where one hasn't performed the restore previously, like automated updates)
- On run failure of nuget CLI, now outputs the error message.
- UpdateNanoFrameworkDependencies now returns a bool reporting running success.
- Update caller accordingly.
- Update nuget.exe to v6.0.0.


angularsen:
[Release UnitsNet/4.116.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.116.0)

Issue #865, opened 2020-12-05
Support extensibility use cases


generateui:
# Overview
As discussed in #858 and #864, I'd like to open a tracking issue for various extensibility features. I propose enumerating the different use cases and then prioritize per use case the feasibility and desirability of supporting and implementing them. I'll be available to assist in implementation too (limited).

# Use cases
Below follow use cases. Please mention your own in the discussion and describe them clearly. I will update this top post to include yours. For convenience and reference, I have included a technical name behind the title.
If some of these usecases are already supported, I will remove them.

## Derived quantities `DerivedQuantities`
While quantities can often be applied, it can often make more sense to have a *different* quantity of the *same* dimensions. Think of an application modelling geothermal pipes. Such application models the **depth** quantity of a pipe, not the **length**. Yet, both quantities share the same dimensions. Typically, the **depth** dimension only uses a subset of the **length** units. Lightyears and nanometers are simply not applicable, but meters and hectometers all the more. Also, as pipes come in prefabricated lengths, a unit to add here may be "pipe" equalling i.e. around 10m for one unit of measurement. Such unit should then be added to the specific **depth** quantity.

In short, **depth** is length with semantic difference and an inherited subset of units and added specific units.

## Extra unit `ExtraUnit`
As DervicedQuantity example, but then the application can simply use an existing quantity but misses a unit. The user of UnitsNet should be able to easily specify that unit, generate some code, and then use it.

## Extra quantity `ExtraQuantity`
As DervicedQuantity example, but then a completely new quantity with units. Think of an application that suddenly needs to use an esoteric unit of a quantity to do calculations.
- blocking issue done: #864
- PR to support this using codegen: #892

## Extra culture `ExtraCulture`
Assume for a moment, like Russian culture has implementations in UnitsNet already, that new i18n of unit formatting/parsing is desirable. There should be an option to simply add these i18n customizations and use them.

## Extra abbreviation within culture `ExtraAbbreviation`
Like ExtraCulture, but then a more specific abbreviation of a cultured unit. This happens often in specialized domains where the domain users have a different semantic pointer to an already existing unit. For example, I have seen instances of `m/year` in a domain where the more standard abbreviation is `m/y`.

## Extra prefix `ExtraPrefix`
Add a new prefix to an existing unit. 

## Have UnitsNet api methods work *with* extended quantities/units `ExtensibilityApi`
[...]

PR #1023, opened 2022-01-28
Use UnitConverter to do conversion functions and allow basic extensibility

tmilnthorp:
Here's a very-rough initial idea for supporting #865.

Rather than generating functions that are not extensible, we can reuse our existing UnitConverter class and be able to use it for conversions. Users could add/remove custom functions at-will.

Need to formalize this more as we should really use QuantityInfo for the parameters.


CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
```diff
@@ -264,9 +264,9 @@ public void ToUnit()
         }}
 
         [Fact]
-        public void ToBaseUnit_ReturnsQuantityWithBaseUnit()
+        public void ToUnit_WithBaseUnit_ReturnsQuantityWithBaseUnit()
         {{
-            var quantityInBaseUnit = {_quantity.Name}.From{_baseUnit.PluralName}(1).ToBaseUnit();
+            var quantityInBaseUnit = {_quantity.Name}.From{_baseUnit.PluralName}(1).ToUnit({_quantity.Name}.BaseUnit);
```
tmilnthorp, line 269: ToBaseUnit is internal. Not sure why we were testing it like this.

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
```diff
@@ -865,11 +872,31 @@ public double As(UnitSystem unitSystem)
...
+        /// <param name="unitConverter">The <see cref="UnitConverter"/> to use for the conversion.</param>
+        /// <returns>A Angle with the specified unit.</returns>
+        public Angle ToUnit(AngleUnit unit, UnitConverter unitConverter)
+        {
+            if(Unit == unit)
+                return this;
+
+            var inBaseUnits = ToUnit(BaseUnit);
+
+            if(!unitConverter.TryGetConversionFunction<Angle>(inBaseUnits.Unit, unit, out var conversionFunction))
```
angularsen, line 895: I like it. 

As a very minor performance improvement, can we call the innermost method overload? I think it has to go through 2 additional `TryGetConversionFunction` methods to get to the most specific one.

angularsen (approved):

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
angularsen: Besides the minor detail of failing tests, I think this looks good?

So in order to extend this, we can simply call `Acceleration.DefaultConversionFunctions.SetXXX()` on startup?

angularsen: > Need to formalize this more as we should really use QuantityInfo for the parameters.

What did you mean by this?

tmilnthorp: > I like it.
> 
> As a very minor performance improvement, can we call the innermost method overload? I think it has to go through 2 additional `TryGetConversionFunction` methods to get to the most specific one.

Done!

tmilnthorp: > Besides the minor detail of failing tests, I think this looks good?
> 
> So in order to extend this, we can simply call `Acceleration.DefaultConversionFunctions.SetXXX()` on startup?
> 
> Will there be a racing condition with the initialization of the global static `UnitConverter.Default`?

You certainly could. I do worry a bit though about that causing unintended consequences, especially since ```GetValueAs``` uses the ```DefaultConversionFunctions```. This would affect:
- operator +
- operator -
- operator <=
- operator >=
- operator <
- operator >
- CompareTo
- Equals
- As

That said, if you write a custom conversion function, arguably it **should** end up being used for * for example.

tmilnthorp:
Odd I am not able to reproduce the test failures locally. Can you @angularsen ?

tmilnthorp:
Hmmm let me add some tests. Kinda shocked we don't have any tests for to base conversion in ToUnit!

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
```diff
@@ -268,73 +309,41 @@ public void As_SIUnitSystem_ThrowsArgumentExceptionIfNotSupported()
             }
         }
 
-        [Fact]
-        public void ToUnit()
+        [Theory]
+        [MemberData(nameof(UnitTypes))]
```
tmilnthorp, line 313: I parameterized a few tests to make them nicer

tmilnthorp:
I think this is ready. I wonder why the new CodeCov report isn't coming through?

CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
```diff
@@ -45,37 +45,38 @@ internal class UnitTestBaseClassGenerator : GeneratorBase
         /// </summary>
         private readonly string _numberSuffix;
 
-        public UnitTestBaseClassGenerator(Quantity quantity)
+        public UnitTestBaseClassGenerator( Quantity quantity )
```
angularsen, line 48: Nitpick, but no whitespace inside ()'s is more consistent with the code base. It seems to be introduced several places in this file, probably auto formatted.

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
```diff
@@ -268,73 +309,41 @@ public void As_SIUnitSystem_ThrowsArgumentExceptionIfNotSupported()
...
+            // See if there is a unit available that is not the base unit.
+            var fromUnit = Acceleration.Units.FirstOrDefault(u => u != Acceleration.BaseUnit && u != AccelerationUnit.Undefined);
+
+            // If there is only one unit for the quantity, we must use the base unit.
+            if(fromUnit == AccelerationUnit.Undefined)
+                fromUnit = Acceleration.BaseUnit;
+
+            var quantity = Acceleration.From(3.0, fromUnit);
+            var converted = quantity.ToUnit(unit);
+            // TODO: Meaningful check possible?
```
angularsen, line 346: Can't we assert that the new quantity has `Unit == unit`?
If so, I would rename this method `ToUnit_ReturnsQuantityWithGivenUnit` unless there is already an equivalent test for that. No exception is an implicit assertion in tests.

UnitsNet/GeneratedCode/Quantities/ElectricChargeDensity.g.cs
```diff
@@ -642,24 +642,44 @@ public ElectricChargeDensity ToUnit(ElectricChargeDensityUnit unit)
...
+                // Already in requested units.
                 return this;
+            }
+            else if(unitConverter.TryGetConversionFunction((typeof(ElectricChargeDensity), Unit, typeof(ElectricChargeDensity), unit), out var conversionFunction))
+            {
+                // Direct conversion to requested unit found. Return the converted quantity.
+                var converted = conversionFunction(this);
+                return (ElectricChargeDensity)converted;
+            }
+            else if(Unit != BaseUnit)
```
angularsen: I know the code base is riddled with varying styles, and this is nitpick, but I notice `if()` without whitespace is consistently used instead of `if ()`. 

I believe the latter is the most common style in C#, and used in [Microsoft's C# coding style conventions](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#layout-conventions).

Can we use the latter form?
Then we can clean the old code to be consistent.

angularsen (changes requested):
Mostly nitpicks, which means the PR is good.

angularsen:
I have no idea what and how the codecov is triggered. It almost seems random.

angularsen:
It seems codecov failed to process the last uploaded coverage reports. Can't see any error codes or any details on what though.

https://codecov.io/github/angularsen/UnitsNet/commit/b20c987ade7301039ed68e63b16137101e966ac3

image

CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
tmilnthorp: Ah yes, stupid VS auto-formatting. Removed.

UnitsNet/GeneratedCode/Quantities/ElectricChargeDensity.g.cs
tmilnthorp: These blocks don't have a space between ```if (``` ?

angularsen: I'll fix it post merge in a solution wide fix, but what I meant is that I would like syntax `if (xxx)` instead of `if(xxx)`.

angularsen:
[Release UnitsNet/4.117.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.117.0)

lipchev:
Hey guys, I've just updated the release note- indicating that this is a breaking change for implementors of the IQuantity interface.

lipchev:
I think the impact of doing it now would be smaller (probably just me and you).

angularsen:
I agree to remove for now to avoid breaking change in v4.

I was thinking if we could maybe make use of [default interface methods](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods), but this seems limited to netstandard2.1 / netcore3 and I believe we should still support nestandard2.0.

tmilnthorp:
> I agree to remove for now to avoid breaking change in v4.
> 
> I was thinking if we could maybe make use of [default interface methods](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods), but this seems limited to netstandard2.1 / netcore3 and I believe we should still support nestandard2.0.

I had thought of that also, but I wouldn't be able to provide a default implementation anyways, since DefaultConverter lives on the individual quantity implementations, not interface definition.

PR #1028, opened 2022-02-02
Change replacement token in json files from `x` to `{x}`

tmilnthorp:
I was adding some better localization code (via resource files, at least for the prefixes at first), and noticed that if I wanted to use a const for SiPrefixes it would fail during text replacement:
```csharp
public class SiPrefixes
{
    public const double Milli = 1e-3;
}
```

x would match the x in ```SiPrefixes.Milli``` and the conversion functions (eg ```GetValueInBaseUnit```) would be incorrect:
```
case AccelerationUnit.MillistandardGravity: return (SiPrefi_valuees.Milli*9.80665) * SiPrefixes.Milli;
```

A more specific replacement token such as ```[x]``` or ```$x$``` would limit scope better since we're doing a string replacement.

Open to ideas for better replacement tokens 😃


angularsen:
I have no strong preference, but `[x]` looks like an array to me and `$x$` is unfamiliar.

`{0}` and `{myVariable}` are commonly used in templating. 
How about `{value}`? Or just `{x}`.

tmilnthorp:
> I have no strong preference, but `[x]` looks like an array to me and `$x$` is unfamiliar.
> 
> `{0}` and `{myVariable}` or `{{myVariable}}` are commonly used in templating. How about `{value}`? Or just `{x}`.

That looks good to me.

Is there any doc files I'm missing that need touched?

angularsen:
> Is there any doc files I'm missing that need touched?

Not that I know of, just search the repo for ... `x` 😆

angularsen:
[Release UnitsNet/4.117.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.117.0)

PR #1018, opened 2022-01-25
Move default unit converter registration to quantity types

tmilnthorp:
Move default unit converter registration to unit's static constructor. 

Attempt to resolve #1016

PR #1031, opened 2022-02-08
Move abbreviation mapping to individual quantity types

tmilnthorp:
In the spirit of #1018, I removed the enormous UnitAbbreviationsCache array and moved registration to a static method on the individual quantitys.


UnitsNet.WindowsRuntimeComponent/CustomCode/UnitAbbreviationsCache.cs
```diff
@@ -45,10 +46,16 @@ static UnitAbbreviationsCache()
 
         private void LoadGeneratedAbbreviations()
         {
-            foreach(var localization in GeneratedLocalizations)
+            var quantities = typeof(Length)
+                .Wrap()
+                .Assembly
+                .GetExportedTypes()
+                .Where(t => typeof(IQuantity).Wrap().IsAssignableFrom(t) && t != typeof(IQuantity));
```
angularsen, line 53: I'm thinking it might be better to just statically generate a `Type` array with the 100ish quantity types, since this will only pick up our own IQuantity types. Reflection is perhaps better suited for picking up types externally or not known at compile time. It is also not particularly performant, so I assume it will have some minor impact on startup time compared to a static list. 

What do you think?

angularsen: I see a couple of other places with the same reflection to look up our own quantity types:
https://github.com/angularsen/UnitsNet/search?q=GetExportedTypes

Maybe we can just have an internal static property to reuse the lookup of our quantity types.

angularsen (changes requested):
Nice improvement, just a suggestion about reflection.

UnitsNet.WindowsRuntimeComponent/CustomCode/UnitAbbreviationsCache.cs
tmilnthorp: I think that's a good suggestion. Can I do as a separate PR that hits all places?

angularsen: Absolutely, seem slike you already got the fix in though! That was quick. Let's merge this, and address any remaining places later.

tmilnthorp: I changed my mind, I did it now 😆 

Easy enough

UnitsNet/GeneratedCode/Quantity.g.cs
```diff
@@ -1233,5 +1233,121 @@ public static bool TryParse(IFormatProvider? formatProvider, Type quantityType,
                     return false;
             }
         }
+
+        internal static IEnumerable<Type> GetQuantityTypes()
+        {
```
angularsen, line 1238: Something tells me we may want to allocate an array for this if it's going to be called multiple times, but that is a minor optimization for later.

angularsen:
LGTM, just merge when you are done

tmilnthorp:
> LGTM, just merge when you are done

I'm done 👍🏻

angularsen:
[Release UnitsNet/4.118.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.118.0)

PR #1049, opened 2022-02-16
⚡ Prefer built-in conversions over extensibility for perf

### LINKED ISSUE #1069: Performance hit by commit 'Use UnitConverter to do conversion functions (e5aba5b)'
**Describe the bug**
Since release 4.117 we are experiencing some serious performance issues. I have been able to trace them back in the build logs of UnitsNet. This commit logs a warning about performance with a ratio of 40x: https://github.com/angularsen/UnitsNet/commit/e5aba5b75aaf0cc85ced1425dce0670f861b8eee

I have also noticed that the build of 4.121 still has this issue: https://github.com/angularsen/UnitsNet/runs/5207495227?check_suite_focus=true#step:7:2963

I am really noticing this slowdown in my project and I was wondering what the cause could be and how we could resolve it.

tmilnthorp:
Fixes #1069 

Per @lipchev's comments in #1023, I decided to add the local conversion method back (as private ```TryToUnit```). This will allow better performance for auto-generated conversions before falling back to the extensibility points.


CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -364,36 +364,37 @@ private void GenerateStaticMethods()
         /// <param name=""unitConverter"">The <see cref=""UnitConverter""/> to register the default conversion functions in.</param>
         internal static void RegisterDefaultConversions(UnitConverter unitConverter)
         {{
-            // Register in unit converter: BaseUnit -> {_quantity.Name}Unit");
+            // Register in unit converter: {_quantity.Name}Unit -> BaseUnit");
```
tmilnthorp, line 367: I reversed these to better match our "mental model" in json of going from a unit to base units first.

tmilnthorp:
@angularsen I will fill in some missing test coverage in this PR. 

Also, is there a way to run the branch marking suite on a PR? It would be nice to see proactively rather than after something is already merged.

tmilnthorp:
Alright. Looks good now. 

Can we add the benchmarking as an optional check that has to be manually run?

angularsen:
> Alright. Looks good now.
> 
> Can we add the benchmarking as an optional check that has to be manually run?

I don't know how to do that? :D 
The checks are optional, the PR can be merged even if they are red.
I believe you can re-trigger the checks via the Checks tab.

I just reduced the codecov precision from 1 to 0, and changed rounding from up to down (default).
Hoping this reduces the pesky failed checks on -0.0 or -0.1% diff.
a2b92aba34922bb5eb17dad474986b6650d923d0

angularsen:
Oh, I didn't read your question right.
**Benchmarking**, yes I guess we can make it optional. To be honest, I haven't paid much attention to them because we found that the results varied alot from run to run due to running on different VMs I guess.

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -898,36 +898,88 @@ public Acceleration ToUnit(AccelerationUnit unit)
...
-                return (Acceleration)converted;
+                // See if the unit converter has an extensibility conversion registered.
+                return (Acceleration)conversionFunction(this);
             }
             else if (Unit != BaseUnit)
             {
-                // Direct conversion to requested unit NOT found. Convert to BaseUnit, and then from BaseUnit to requested unit.
+                // Conversion to requested unit NOT found. Try to convert to BaseUnit, and then from BaseUnit to requested unit.
                 var inBaseUnits = ToUnit(BaseUnit);
                 return inBaseUnits.ToUnit(unit);
```
angularsen, line 922: Nitpick, but this could be a single line and read well: 
`return ToUnit(BaseUnit).ToUnit(unit)`

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -898,36 +898,88 @@ public Acceleration ToUnit(AccelerationUnit unit)
...
         {
-            if (Unit == unit)
+            if (TryToUnit(unit, out var converted))
             {
-                // Already in requested units.
-                return this;
+                // Try to convert using the auto-generated conversion methods.
+                return converted!.Value;
             }
             else if (unitConverter.TryGetConversionFunction((typeof(Acceleration), Unit, typeof(Acceleration), unit), out var conversionFunction))
```
angularsen, line 913: Shouldn't extensibility conversions be attempted first?

So it is possible to override the built-in functions. I'm not sure what would be a good reason to do so though, other than duct-tape fixing bugs or precision issues. More of a gut-feeling, that extensibility should have precedence. I don't know if the extra checks would have any meaningful impact on performance though.

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -898,36 +898,88 @@ public Acceleration ToUnit(AccelerationUnit unit)
...
             {
-                // Direct conversion to requested unit found. Return the converted quantity.
-                var converted = conversionFunction(this);
-                return (Acceleration)converted;
+                // See if the unit converter has an extensibility conversion registered.
+                return (Acceleration)conversionFunction(this);
```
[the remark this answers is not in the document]
tmilnthorp: I think it also depends if we want to allow conversion to<->from **any** quantity, not just a unit within a quantity. In that case, we need to return to IQuantity. To keep it the same type, a cast (and boxing) is required.

We could also create a generic unit converter (```UnitConverter<Acceleration>```) to allow for in-quantity conversions only which would avoid the boxing/unboxing. For out-of-quantity conversions you'd then fall back to the current IQuantity with the boxing problem.

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
tmilnthorp: Generally, I'd say yes, extensibility should have precedence. But if you went that route, you may as well abandon this PR and use as-is. 

This is purely a performance optimization as I saw the benchmark was taking 20x longer for As conversions 👎🏻 

However, this optimization would mean that you could not override the conversion functions that we define in the json files at runtime due to the if/else ordering.

I guess the question is
- Do we need performance? If so, merge this PR
- Is extensibility more important than performance? Abandon this!

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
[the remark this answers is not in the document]
angularsen: Ah, good point about converting to other quantities.

Yes, maybe in-quantity and out-of-quantity is an option to consider. But it sounds like a bit of complexity and I'm not so sure how much we gain really, other than for scenarios of performing unit conversions on loads of quantities.

angularsen:
@tmilnthorp I'm reopening this in light of #1069 .

I think maybe a fair compromise is to merge this PR, and not allow extensions to override built-in functions to mitigate performance issue. I'm not sure I see a good usecase to override built-in functions except to duct-tape fix bugs, but I would rather merge those fixes into the library.

What do you think?

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
angularsen: As per https://github.com/angularsen/UnitsNet/issues/1069, it seems the consensus is that the performance is now so bad that this is the better compromise.

angularsen:
[Release UnitsNet/4.142.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.142.0)

PR #1051, opened 2022-02-17
Add conversion properties that return a quantity

tmilnthorp:
- Adding conversion properties that return a quantity
- Obsolete conversion properties that return only a double

This allows you to always get back a quantity, then reuse that quantity for other operations.

For example to calculate ```Mass``` and you'd use the following method:
```csharp
public static Mass operator *(Density density, Volume volume)
```

This allows for nicer conversion
```csharp
Density density = Density.FromKilogramsPerCubicMeter( 3.0 );
Volume volume = Volume.FromLiters( 3.0 );

// Current method
Mass mass = density.ToUnit(DensityUnit.CentigramPerLiter) * volume.ToUnit(VolumeUnit.CubicFoot);

// With this PR
Mass mass = density.ToCentigramsPerLiter * volume.ToCubicFeet;
```

It also lets you care less about if ```Value``` is a ```double``` or ```decimal```. If you never need to get the ```Value```, it keeps you "in" a UnitsNet type.

I figured I'd throw this up here now, since it would be nice to remove the current conversion properties (that assume double, even for a decimal value), as it will bloat the binary size temporarily with 2x conversion properties.


tmilnthorp:
```
Downloading Tools/7-zip/7za.exe (536 KB)
Error downloading object: Tools/7-zip/7za.exe (fa252e5): Smudge error: Error downloading Tools/7-zip/7za.exe (fa252e501332b7[4](https://ci.appveyor.com/project/angularsen/unitsnet/builds/42619991#L4)86a972e7e471cf6915daa681af35c6aa102213921093eb2a3): batch response: This repository is over its data quota. Account responsible for LFS bandwidth should purchase more data packs to restore access.
```
Uh oh.

Muximize:
I don't really understand the benefit of these conversion properties. The whole point of UnitsNet is to not care too much about the underlying values, right?

So instead of your current method (which works, but is a bit verbose) why not just do:

```C#
Mass mass = density * volume;
```
And to get a raw value out at some later point:
```C#
double masskg = mass.Kilograms;
```
That is my preferred way of using UnitsNet, it makes the unit explicit when extracting a raw value. But it would be deprecated by this PR! :anguished:


 I try to avoid using `.Value`, I'd rather see _**that**_ deprecated. It enables the conversion and value extraction to be separate, which can be confusing:
```C#
Mass othermass = mass.ToKilograms;
...
...
...
double massgrams = othermass.Value; // Surely it was grams, right?
```

angularsen:
At first glance, I share the gut feeling of @Muximize .

Adding `.ToGrams()` provides a minor syntax improvement over `ToUnit()`.
Removing `.Grams` loses out on a lot of the static type safety I think UnitsNet brings.

Due to 1000+ units and binary size issues, I don't want to have both `.ToGrams()` and `.Grams`, so I think we would have to pick one.


The problem; if you take a `Mass` parameter into a method, you should not need to know what unit it was constructed with. You might as well have passed `double`. Or am I misunderstanding this PR?

tmilnthorp:
> The whole point of UnitsNet is to not care too much about the underlying values, right?

I think you're agreeing with me actually 😄 Encapsulation is important. These properties allow you to do conversion to a different unit and still keep encapsulation in a UnitsNet quantity as the returned type. Otherwise it's just a double (even if the quantity was not a double originally such as ```BitRate```). It __forces__ you to not care about the underlying value when asking for a quantity in a different unit.

Currently:
```csharp
Volume volume = Volume.FromLiters( 3.0 );

// Surely it was cubic feet, right?   ;)
double volumeInCubicFeet = volume.CubicFeet;
```
Is now:
```csharp
Volume volume = Volume.FromLiters( 3.0 );

// Definitely in cubic feet   ;)
Volume volumeInCubicFeet = volume.ToCubicFeet;
```

tmilnthorp:
> Due to 1000+ units and binary size issues, I don't want to have both .ToGrams() and .Grams, so I think we would have to pick one.

This is to propose a move to this style and remove the double versions at v5. But if you just change it, it's a major breaking change with no obsolete message.

> The problem; if you take a `Mass` parameter into a method, you should not need to know what unit it was constructed with. You might as well have passed `double`. Or am I misunderstanding this PR?

Perhaps that was a bad example. Let's go super simple: I have a length that I want to display in meters.
```csharp
private static string GetStringInMeters(Length length)
{
	return inches.Meters; // Can't do currently: just a double      
	return inches.ToMeters.ToString(); // Now it's easy! It's in a UnitNet type already
}
```

Currently your only option is:
```csharp
private static string GetStringInMeters(Length length)
{
	return inches.ToUnit(LengthUnit.Meter).ToString();
}
```
And getting the value is easy if you so choose:
```csharp
private static double GetLengthInMeters(Length length)
{
	return inches.ToMeters.Value;
}
```

What's the point of the handy conversion properties if using them loses the power of UnitsNet?

angularsen:
I think it comes to what the primary use case is.
Both solutions work, I believe they just optimize syntax for one way or the other.

1. `inches.ToMeters` syntax is optimized to get a new quantity in the given unit
1. `inches.Meters` syntax is optimized to get the `double` in the given unit.

Both are more cumbersome to get the other representation.

I honestly don't know what I would consider the primary use case, and I am probably colored by many years of seeing `.Kilograms` properties and expecting double values out. It does more closely resemble `TimeSpan` and feels familiar in that regard.

angularsen:
I agree with your sentiment @Muximize, that in my use of UnitsNet I generally put values in, maybe do some arithmetic, and later get values out in a given unit. For me, that makes the most intuitive sense.

However, I believe this comes down to syntax and discoverability. Feature-wise, this PR is identical to current.

If I understand right, we are looking at these before/after examples, so I think it is primarily a matter of syntax:
```cs
// Who knows what unit, but that is fine
var parsedMass = Mass.Parse(input);

// Some arithmetic maybe
var twiceMass = parsedMass * 2;

// Before/after
var grams1 = twiceMass.Grams;
var grams2 = twiceMass.AsGrams.Value;

// Before/after
var gramsString1 = twiceMass.As(MassUnit.Gram).ToString();
var gramsString2 = twiceMass.AsGrams.ToString();
```

Muximize:
You're right, it's mostly a syntax thing. But I don't think it is an improvement because it makes an (in my opinion, as explained above) unfortunate feature of UnitsNet **_more_** discoverable and easier to shoot yourself in the foot with. There should be some resistance and explicitness.

```C#
var parsedMass = Mass.Parse(input);

// Before/after
var someMass1 = parsedMass.ToUnit(MassUnit.Gram); // Quite explicitly changing the unit, good
var someMass2 = parsedMass.AsGrams;               // Would a new user understand what they are doing?

var someMass3 = parsedMass.As(MassUnit.Gram);     // Almost the same syntax, but returns a double?
```

tmilnthorp:
Sorry guys has been a busy week, I haven't had time to check back in.

Fine by me, you won't offend :)

angularsen:
I'm glad we can part as friends 😆

Just kidding, I appreciate your ideas 🙌

PR #1053, opened 2022-02-18
Creating build and test action

tmilnthorp:
Not working yet


tmilnthorp:
Surprisingly it checks out without LFS error :)

angularsen:
> Surprisingly it checks out without LFS error :)

Just bought an LFS data pack, so we are good for 30 days.

PR #1070, opened 2022-03-28
Fix Duration.ToTimeSpan() inaccuracy rounding to millisecond

AndreasLeeb:
When using .NET Framework (v4.8 in my case), the `TimeSpan.FromXXX` methods (like `FromSeconds` used in `Duration.ToTimeSpan()` up until now) round to the nearest millisecond, even though `TimeSpan`s are accurate to the tick (100 ns).

New unit tests before my fix:
image

Unit tests after my fix:
image

### DIFF
--- a/UnitsNet/CustomCode/Quantities/Duration.extra.cs
+++ b/UnitsNet/CustomCode/Quantities/Duration.extra.cs
@@ -18,7 +18,7 @@ namespace UnitsNet
             if ( Seconds > TimeSpan.MaxValue.TotalSeconds ||
                 Seconds < TimeSpan.MinValue.TotalSeconds )
                 throw new ArgumentOutOfRangeException( nameof( Duration ), "The duration is too large or small to fit in a TimeSpan" );
-            return TimeSpan.FromSeconds( Seconds );
+            return TimeSpan.FromTicks((long)(Seconds * TimeSpan.TicksPerSecond));
         }
 
         /// <summary>Get <see cref="DateTime"/> from <see cref="DateTime"/> plus <see cref="Duration"/>.</summary>
--- a/UnitsNet.Tests/CustomCode/DurationTests.cs
+++ b/UnitsNet.Tests/CustomCode/DurationTests.cs
@@ -63,31 +63,15 @@ namespace UnitsNet.Tests.CustomCode
             AssertEx.EqualTolerance(duration.Seconds, timeSpan.TotalSeconds, 1e-3);
         }
 
-        [Fact]
-        public static void ToTimeSpanShouldNotDiscardOneTickDuration()
+        [Theory]
+        [InlineData(100, Units.DurationUnit.Nanosecond)]
+        [InlineData(1, Units.DurationUnit.Microsecond)]
+        [InlineData(1.234, Units.DurationUnit.Millisecond)]
+        public static void ToTimeSpanShouldNotRoundToMillisecond(double value, Units.DurationUnit unit)
         {
-            const long ticksPerMicrosecond = TimeSpan.TicksPerMillisecond / 1000;
-            Duration duration = Duration.FromNanoseconds(100);
+            Duration duration = Duration.From(value, unit);
             TimeSpan timeSpan = duration.ToTimeSpan();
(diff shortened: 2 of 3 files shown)


UnitsNet/CustomCode/Quantities/Duration.extra.cs
```diff
@@ -18,7 +18,7 @@ public TimeSpan ToTimeSpan()
             if ( Seconds > TimeSpan.MaxValue.TotalSeconds ||
                 Seconds < TimeSpan.MinValue.TotalSeconds )
                 throw new ArgumentOutOfRangeException( nameof( Duration ), "The duration is too large or small to fit in a TimeSpan" );
-            return TimeSpan.FromSeconds( Seconds );
+            return TimeSpan.FromTicks((long)(Seconds * TimeSpan.TicksPerSecond));
```
angularsen, line 21: Oh wow, I never realized FromSeconds [rounded to milliseconds](https://docs.microsoft.com/en-us/dotnet/api/system.timespan.fromseconds?view=net-6.0). Today I learned!

angularsen (approved):
Thanks! This looks good to me, just requeued the build that failed on some weird timeout. When it's green, this should be ready to go.

UnitsNet/CustomCode/Quantities/Duration.extra.cs
AndreasLeeb: Yeah, it really depends on the framework you're using.

.NET 5/6 does not round, it uses exact ticks ([see](https://github.com/dotnet/runtime/blob/6019cdace12ec9bc5117e515b9a6fccfd5708bde/src/libraries/System.Private.CoreLib/src/System/TimeSpan.cs#L215)):
```
private static TimeSpan Interval(double value, double scale)
{
	if (double.IsNaN(value))
		ThrowHelper.ThrowArgumentException_Arg_CannotBeNaN();
	return IntervalFromDoubleTicks(value * scale);
}

private static TimeSpan IntervalFromDoubleTicks(double ticks)
{
	if ((ticks > long.MaxValue) || (ticks < long.MinValue) || double.IsNaN(ticks))
		ThrowHelper.ThrowOverflowException_TimeSpanTooLong();
	if (ticks == long.MaxValue)
		return TimeSpan.MaxValue;
	return new TimeSpan((long)ticks);
}
```

In .NET Framework 4.8, the values are rounded to milliseconds ([see](https://referencesource.microsoft.com/#mscorlib/system/timespan.cs,228)):
```
private static TimeSpan Interval(double value, int scale) {
	if (Double.IsNaN(value))
		throw new ArgumentException(Environment.GetResourceString("Arg_CannotBeNaN"));
	Contract.EndContractBlock();
	double tmp = value * scale;
	double millis = tmp + (value >= 0? 0.5: -0.5);
	if ((millis > Int64.MaxValue / TicksPerMillisecond) || (millis < Int64.MinValue / TicksPerMillisecond))
		throw new OverflowException(Environment.GetResourceString("Overflow_TimeSpanTooLong"));
	return new TimeSpan((long)millis * TicksPerMillisecond);
}
```

Mono also rounds ([see](https://github.com/mono/mono/blob/4b5e2591eb299eb34d60b5217b14c1ad9f11e238/mcs/class/referencesource/mscorlib/system/timespan.cs#L228)), its implementation is the same as in the .NET Framework.

So you can't rely on its accuracy, that's why I've resorted to custom TimeSpanExact.FromXXX methods which use FromTicks internally.

tmilnthorp:
Since the original ```Duration``` could have been specified in any unit, doing a conversion to ```Seconds``` may result in floating point error anyways, no? 

Just wondering if we should take the unit of the original definition into account.

angularsen:
Nuget should be out shortly
[Release UnitsNet/4.127.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.127.0)

angularsen:
> Since the original `Duration` could have been specified in any unit, doing a conversion to `Seconds` may result in floating point error anyways, no?
> 
> Just wondering if we should take the unit of the original definition into account.

I guess we could as an improvement, if you want to have a go at it.

PR #1073, opened 2022-04-05
Update install of nanoFramework build components

josesimoes:
- Dropping VSIX gallery as source of msbuild components. Replacing with GitHub version which is a much more reliable source.


angularsen:
[Release UnitsNet/4.129.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.129.0)

PR #1083, opened 2022-05-13
Fix QuantityTypeConverter ignoring given CurrentCulture without DisplayAsAttribute

AndreasLeeb:
Hi,

I've noticed a bug in the `QuantityTypeConverter` while using it in a PropertyGrid.

When `CultureInfo.CurrentCulture` and `CultureInfo.CurrentUICulture` differ, like in my case where CurrentCulture is de-AT (German uses a comma as decimal separator) and CurrentUICulture is en-US (dot as decimal separator), it happens that if the `QuantityTypeConverter` is given `CultureInfo.CurrentCulture` as culture when no `DisplayAsAttribute` is present in the context, `QuantityTypeConverter` calls the base implementation `TypeConverter.ConvertTo` which only passes the given culture to `ToString` if it's not equal to the CurrentCulture ([see](https://github.com/dotnet/runtime/blob/731b2c0b457ae7bc433dcca2e4fd57d3cd56d3cb/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/TypeConverter.cs#L124)). This results in the parameterless `ToString` method of the quantity being called, which uses `CultureInfo.CurrentUICulture`.
```
public override string ToString()
{
    return ToString("g");
}
public string ToString(string format)
{
    return ToString(format, CultureInfo.CurrentUICulture);
}
```

I have changed `QuantityTypeConverter` to always pass the given culture now.

New unit test before bugfix:
image

After bugfix:
image

### DIFF
--- a/UnitsNet.Tests/QuantityTypeConverterTest.cs
+++ b/UnitsNet.Tests/QuantityTypeConverterTest.cs
@@ -251,6 +251,24 @@ namespace UnitsNet.Tests
             Assert.Equal("10 dm", convertedQuantitySpecificCulture);
         }
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void ConvertTo_GivenCurrentCulture_ReturnValueFormattedAccordingToGivenCulture(bool useDisplayAsAttribute)
+        {
+            QuantityTypeConverter<Length> converter = new();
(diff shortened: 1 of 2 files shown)


angularsen:
I can't say I understand 100% all the details here, but your reasoning seems sound and I appreciate the test to help cover it. I can't spot any regressions in the logic, so looks good to me 👍

angularsen:
Thanks! Nuget is on the way out.

[Release UnitsNet/4.131.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.131.0)

PR #1084, opened 2022-05-13
QuantityValue: 16 bytes instead of 40 bytes

AndreasLeeb:
Fixes #875 

@angularsen 
Regarding https://github.com/angularsen/UnitsNet/discussions/875#discussioncomment-2690100 a PR draft for https://github.com/angularsen/UnitsNet/discussions/875#discussioncomment-2689636

Size comparison
```
Before: Marshal.SizeOf<QuantityValue>(): 40 B
After:  Marshal.SizeOf<QuantityValue>(): 16 B
```

`UnderlyingDataType` could support up to 256 different enum values, I guess we won't ever need more different underlying types that can fit into this struct 😄

### DIFF
--- a/UnitsNet/QuantityValue.cs
+++ b/UnitsNet/QuantityValue.cs
@@ -1,8 +1,10 @@
 ﻿// Licensed under MIT No Attribution, see LICENSE file at the root.
 // Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
 
-
+using System;
+using System.Diagnostics;
 using System.Runtime.InteropServices;
+using System.Text;
 using UnitsNet.InternalHelpers;
 
 namespace UnitsNet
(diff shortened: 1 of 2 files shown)


UnitsNet/QuantityValue.cs
```diff
@@ -20,31 +21,41 @@ namespace UnitsNet
     ///     From 8 (int, long, double, decimal + each nullable) down to 2 (QuantityValue and QuantityValue?).
     ///     This also adds more numeric types with no extra overhead, such as float, short and byte.
     /// </remarks>
-    public struct QuantityValue
+    [StructLayout(LayoutKind.Explicit)]
+    public readonly struct QuantityValue
```
angularsen, line 25: Can we get a short description in summary/remarks of xmldoc to help explain the new multiple numeric types thing? This will not be obvious to most.

UnitsNet/QuantityValue.cs
```diff
@@ -72,28 +83,84 @@ private QuantityValue(decimal val)
...
+            /// <inheritdoc cref="decimal"/>
+            /// <remarks>Has to be 0 due to the bit structure of <see cref="decimal"/>.</remarks>
+            Decimal = 0,
+            /// <inheritdoc cref="double"/>
+            Double = 1
         }
+
+        #region just for debugging
+#if DEBUG
+        [FieldOffset(0)]
```
angularsen, line 131: You might be able to use this instead
 https://docs.microsoft.com/en-us/dotnet/framework/debug-trace-profile/enhancing-debugging-with-the-debugger-display-attributes

UnitsNet/QuantityValue.cs
```diff
@@ -72,28 +83,84 @@ private QuantityValue(decimal val)
...
+            };
+
+        /// <summary>
+        ///     Describes the underlying type of a <see cref="QuantityValue"/>.
+        /// </summary>
+        public enum UnderlyingDataType : byte
         {
-            return _value.HasValue ? _value.ToString() : _valueDecimal.ToString();
+            /// <inheritdoc cref="decimal"/>
+            /// <remarks>Has to be 0 due to the bit structure of <see cref="decimal"/>.</remarks>
```
angularsen, line 123: `Decimal must have the value 0 due to ...`

angularsen (approved):
I think this looks great? As long as tests pass, I think this should be ready to go. Just a few minor comments.

angularsen:
Thanks a lot!
v5 branch currently has trouble building, #1094 .

When that is resolved, we will publish a new prerelease nuget for v5 branch.

angularsen:
Build is now fixed and nuget UnitsNet/5.0.0-alpha006 should be out shortly.

PR #1100, opened 2022-06-14
✨ Add back IEquatable with strict equality

### LINKED ISSUE #1017: GetHashCode returns different hash for Equal units
**Describe the bug**
`.GetHashCode()` returns different hash for 2 units for which `.Equals()` return true

**To Reproduce**
```csharp
Length inCm = Length.FromCentimeters(100);
Length inM = Length.FromMeters(1);

Assert.AreEqual(inCm, inM);
Assert.AreNotEqual(inCm.GetHashCode(), inM.GetHashCode());
```

**Expected behavior**
The `GetHashCode()` should return the same value because the .Equals() returns true

**Additional context**
From the [remarks](https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-6.0#remarks):

> Two objects that are equal return hash codes that are equal. However, the reverse is not true: equal hash codes do not imply object equality, because different (unequal) objects can have identical hash codes.

I've peeked in the sources and I think the issue is, that the value from which the unit is constructed is used when calculating the hashcode:

https://github.com/angularsen/UnitsNet/blob/7f67ffba914705eb27000b7d2962a90e0aebf70a/UnitsNet/GeneratedCode/Quantities/Length.g.cs#L1025-L1032

Shouldn't this be normalized to the unit's base unit or something along those lines?

Reason I'm asking is because I'm doing `IEquatable` implementations on my objects and in the `.Equals(...)` method, I'm comparing Units to a max of 5 decimals while in the `GetHashCode()` method I'm simply using the `GetHashCode()` method from the unit.

During the PR someone pointed out that for the `GetHashCode()`, the unit should also be rounded using the same logic as in the `.Equals(...)` method:

```csharp
        /// <inheritdoc />
        public bool Equals(ResultAnalysis1DInternalForces other)
        {
            if (ReferenceEquals(null, other))
            {
                return false;
            }

            if (ReferenceEquals(this, other))
            {
                return true;
            }

            return base.Equals(other)
                && Equals(Member, other.Member)
                && Equals(MemberRib, other.MemberRib)
                && Index == other.Index
                && Section.UnitsNetEquals(other.Section)
                && N.UnitsNetEquals(other.N)
                && Vy.UnitsNetEquals(other.Vy)
                && Vz.UnitsNetEquals(other.Vz)
                && Mx.UnitsNetEquals(other.Mx)
                && My.UnitsNetEquals(other.My)
                && Mz.UnitsNetEquals(other.Mz);
        }

        /// <inheritdoc />
        public override bool Equals(object obj)
        {
            return ReferenceEquals(this, obj) || obj is ResultAnalysis1DInternalForces other && Equals(other);
        }

        /// <inheritdoc />
        public override int GetHashCode()
        {
            unchecked
            {
                int hashCode = base.GetHashCode();
                hashCode = (hashCode * 397) ^ (Member != null ? Member.GetHashCode() : 0);
                hashCode = (hashCode * 397) ^ (MemberRib != null ? MemberRib.GetHashCode() : 0);
                hashCode = (hashCode * 397) ^ Index;
                hashCode = (hashCode * 397) ^ Section.GetHashCode();
                hashCode = (hashCode * 397) ^ N.GetHashCode();
                hashCode = (hashCode * 397) ^ Vy.GetHashCode();
                hashCode = (hashCode * 397) ^ Vz.GetHashCode();
                hashCode = (hashCode * 397) ^ Mx.GetHashCode();
                hashCode = (hashCode * 397) ^ Mz.GetHashCode();
                return hashCode;
            }
        }
```

```csharp
        public static bool UnitsNetEquals<TUnit>(this TUnit value, TUnit other)
        {
            if (value == null && other == null)
            {
                return true;
            }

            if (value != null && other != null && value is IQuantity thisUnit && other is IQuantity otherUnit)
            {
                try
                {
                    double thisValue = thisUnit.Value;
                    double otherValueInThisUnits = otherUnit.As(thisUnit.Unit);

                    return Comparison.Equals(thisValue, otherValueInThisUnits, ComparingConstants.DoubleComparisionDelta, ComparisonType.Absolute);
                }
                catch (ArgumentException e)
                {
                    return false;
                }
            }

            return EqualityComparer<TUnit>.Default.Equals(value, other);
        }
```

angularsen:
Fixes #1017

- Reverted removing `IEquatable`.
- Changed the implementation to strict equality on both `Value` and `Unit`.
- Improved tests.
- Improved xmldocs for equality members.
- `GetHashCode()` left unchanged, which includes quantity name in addition to value and unit, so that `LengthUnit.Meter = 1` and `MassUnit.Gram = 1` are still considered different in collections of `IQuantity`.

Reverts commit f3c7e2559319b88219413348c0a27864ad0d306a.
"🔥 Remove IEquatable<T> and equality operators/methods"


CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -760,12 +786,40 @@ public int CompareTo(object obj)
...
+        public override bool Equals(object obj)
+        {{
+            if(obj is null || !(obj is {_quantity.Name} obj{_quantity.Name}))
+                return false;
+
+            return Equals(obj{_quantity.Name});
+        }}
+
+        /// <inheritdoc />
+        /// <summary>Returns true if either <see cref=""Value"" /> or <see cref=""Unit"" /> are not exactly equal for both quantities.</summary>
```
dschuermans, line 816: Is it just me, or is this confusing wording?

Shouldn't it be:

Returns true if either Value or Unit are exactly equal for both quantities?

angularsen: This is for `!=` not equals, but yes I agree it was clumsy wording.

I propose these:
`==`
> Returns true if both <see cref=""Value"" /> and <see cref=""Unit"" /> are exactly equal for both quantities.

`!=`
> Returns true if either <see cref=""Value"" /> or <see cref=""Unit"" /> are not exactly equal for both quantities.

angularsen:
> > @dschuermans I implemented strict equality, if you want to look over it.
> > Based on decision in [#1017 (comment)](https://github.com/angularsen/UnitsNet/issues/1017#issuecomment-1028402841).
> 
> Nice,
> 
> But if I understand correctly, the case I made in the mentioned issue still won't work right? _(1m equals 100cm and thus should yield the same hashcode)_

No, `1m` and `100cm` are no longer equal as of this PR. Their units differ. So GetHashCode does not have to be equal for these either.

Or did I misunderstand?

dschuermans:
> > > @dschuermans I implemented strict equality, if you want to look over it.
> > > Based on decision in [#1017 (comment)](https://github.com/angularsen/UnitsNet/issues/1017#issuecomment-1028402841).
> > 
> > 
> > Nice,
> > But if I understand correctly, the case I made in the mentioned issue still won't work right? _(1m equals 100cm and thus should yield the same hashcode)_
> 
> No, `1m` and `100cm` are no longer equal as of this PR. Their units differ. So GetHashCode does not have to be equal for these either.
> 
> Or did I misunderstand?

Erm, for me personally that sounds wrong if UnitsNet will now report that `1m` is not the same as `100cm`.

Imagine that you're filling out forms to get a building permit and you provide your plan which has values in cm whereas the city's system works in m and your request for a permit gets denied because the clerk tells you "well, we only allow this to be a max of `4m` wide and your application states `400cm`"

But in any case, if i recall my initial question was that if it was possible to globally configure the default comparison's number of decimals, so that I wouldn't have to use the overload everywhere in code.

We "fixed" this by using an extension method which allows us to configure the number of decimals in a single spot
I then bumped into the issue with the hash codes not being the same (`1m` has different hash then `100cm`) but in our case, those values **are** the same.

So for me and my project, things could've stayed the way they were, as I created work-arounds for it.
We're still on 4.72 and as long as everything works like it's supposed to we're good.
I just hope we don't need any new units, as I do not wish to upgrade to something that'll say that `1m`.Equals(`100cm`) is false 😅

angularsen:
I hear your pain @dschuermans 😅 This discussion has literally gone for years, on and off.

I think the reasoning is best outlined in this comment https://github.com/angularsen/UnitsNet/issues/1017#issuecomment-1024605778

I simply don't see a trivial solution here that makes everyone happy.
Strict equality seems to be the lesser evil. People will run into subtle bugs with option 2, trying to apply some rounding on the equality for various sizes of values and units - plus `GetHashCode()` seems tricky to get right with rounding.
Strict equality is at least explainable, but I agree not intuitive.

As summarized in the comment:
> If the value + unit doesn't match exactly, they are not equal.
> Simple, arguably correct, will 100% piss off consumers trying to equate 100 cm to 1 meter.

### Static config of equality
I guess we could allow for a static configuration of a default `IEqualityComparer`, where you can configure the equivalent of calling one of these:
```cs
Length.FromMeters(1).Equals(Length.FromCentimeters(100), 1e-5, ComparisonType.Absolute) // 1m +/- 0.00001
Length.FromMeters(1).Equals(Length.FromCentimeters(100), 1e-3, ComparisonType.Relative) // 1m +/- 0.1% 
```

At least then you are making an informed choice on the rounding.
However, `GetHashCode()` would still produce values for strict equality in that case, and differ as you described in your issue.

If you see a better solution or have convincing arguments for option 2 or 3, then I'm all ears.

angularsen:
Thank you for the comments and suggestions.

> 2 separate packages of UnitsNet

I don't like this option. It will confuse people what package to choose, and it complicates transient dependencies on UnitsNet, where libraries A and B take in two different UnitsNet nugets. It also feels overkill to solve a rather small problem.

> Uses sane equality and produces the same hash code for units that are equal in a sane way.

This is the challenge. How do we produce the same hash code for units that are approximately equal, for very small or very large values and when comparing very small units to very large units?


> Maybe not even static, but it could be provided when constructing the unit?

Not sure I favor this approach. Consumers may run into runtime exceptions if quantities happen to be constructed with different equality settings, such as getting quantities computed by a library out of your control. 

I am not a fan of static config either, but if we had to choose between ctor parameter and static config, I would in this case prefer the latter. I'm still not convinced we should offer the static config though, since we would still have to solve the challenge about the hashcode mentioned above.

angularsen:
Some other options not mentioned:
- For comparison methods on collections, you can pass `IEqualityComparer` to provide your own equality comparers. UnitsNet could maybe provide some default instances based on typical relative and absolute comparison.
- The consumer can write an extension method to make it syntactically easy to compare two quantities with their own equality comparer. UnitsNet could possibly also provide overloads or extension methods for typical relative/absolute comparison.

However, none of these really address the problem that many will expect 100 cm to equal 1m out of the box.

UnitsNet.Tests/CustomCode/AreaTests.cs
```diff
@@ -49,14 +49,14 @@ public void AreaDividedByLengthEqualsLength()
         public void AreaTimesMassFluxEqualsMassFlow()
         {
             MassFlow massFlow = Area.FromSquareMeters(20) * MassFlux.FromKilogramsPerSecondPerSquareMeter(2);
-            Assert.Equal(massFlow, MassFlow.FromKilogramsPerSecond(40));
+            Assert.Equal(40, massFlow.KilogramsPerSecond);
```
lipchev, line 52: I think this would be good place to hammer down the fact the actual value might not be exactly 40. However I'm not sure as to what precision should be used: 5, 9, 15? We could hide the fact that we don't know behind the AssertEx.Equals - but we'd have to reference the other test for the tolerance.

lipchev (commented):
I've suggested fixes for the comments on the Equals overload.
As for the question of the precision on the custom tests- I have no suggestions :)

angularsen:
Sorry for the late reply, I have been missing out on email notifications for some time and also been very busy.

Will review soon.

CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
```diff
@@ -373,7 +390,7 @@ public void ToUnit_WithSameUnits_AreEqual({_unitEnumName} unit)
         public void ToUnit_FromNonBaseUnit_ReturnsQuantityWithGivenUnit({_unitEnumName} unit)
```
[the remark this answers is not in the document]
angularsen: I like the Skip one 👏

CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
```diff
@@ -373,7 +390,7 @@ public void ToUnit_WithSameUnits_AreEqual({_unitEnumName} unit)
...
         {{
             // See if there is a unit available that is not the base unit, fallback to base unit if it has only a single unit.
-            var fromUnit = {_quantity.Name}.Units.Where(u => u != {_quantity.Name}.BaseUnit).DefaultIfEmpty({_quantity.Name}.BaseUnit).FirstOrDefault();
+            var fromUnit = {_quantity.Name}.Units.Where(u => u != {_quantity.Name}.BaseUnit).DefaultIfEmpty({_quantity.Name}.BaseUnit).First();
```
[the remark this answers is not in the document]
angularsen: Done

UnitsNet.Tests/CustomCode/AreaTests.cs
angularsen: We have separate tests for roundtrip tolerance, this test simply verifies the arithmetic.

PR #1104, opened 2022-06-30
Add Mach speed unit

tjrtfarnham:
Hello @angularsen.  Thanks for the help...github is a bit different.  Note, these changes generate the testing error I described in the discussions about the two lists of enumerations not being identical since the order of the entries are different from one another.  I'm also curious why the generated enum is 42 when it seems like it should have chosen 33?

### DIFF
--- a/Common/UnitDefinitions/Speed.json
+++ b/Common/UnitDefinitions/Speed.json
@@ -313,6 +313,22 @@
           "Abbreviations": [ "миль/ч" ]
         }
       ]
+    },
+    {
+      "SingularName": "Mach",
+      "PluralName": "Mach",
+      "FromUnitToBaseFunc": "{x} * 340.29",
+      "FromBaseToUnitFunc": "{x} / 340.29",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "Ma" ]
+        },
+		    {
+          "Culture": "ru-RU",
+          "Abbreviations": [ "мах" ]
+        }
+      ]
     }
   ]
 }
--- a/Common/UnitDefinitions/Speed.json
+++ b/Common/UnitDefinitions/Speed.json
@@ -322,7 +322,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "Ma" ]
+          "Abbreviations": [ "M", "Ma", "MN", "MACH" ]
         },
 		    {
           "Culture": "ru-RU",
--- a/Common/UnitEnumValues.g.json
+++ b/Common/UnitEnumValues.g.json
@@ -1343,7 +1343,8 @@
     "UsSurveyFootPerSecond": 29,
     "YardPerHour": 30,
     "YardPerMinute": 31,
-    "YardPerSecond": 32
+    "YardPerSecond": 32,
+    "Mach": 42
   },
   "StandardVolumeFlow": {
     "StandardCubicCentimeterPerMinute": 1,
--- a/UnitsNet.Tests/CustomCode/SpeedTests.cs
+++ b/UnitsNet.Tests/CustomCode/SpeedTests.cs
@@ -73,7 +73,7 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double UsSurveyFeetPerHourInOneMeterPerSecond => 1.1811E4;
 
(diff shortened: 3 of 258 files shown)


tjrtfarnham:
@angularsen Hoping you might find some time to have a look at this to see what we can do to remedy the test failure.

angularsen:
Hi yes I hope to get to this in the next few days, just a bit wrapped up in work before the summer holidays.

angularsen:
Just an update, am on a vaction this week and will probably not have time to look before the end of the weekend.

Common/UnitDefinitions/Speed.json
```diff
@@ -313,6 +313,22 @@
...
+    },
+    {
+      "SingularName": "Mach",
+      "PluralName": "Mach",
+      "FromUnitToBaseFunc": "{x} * 340.29",
+      "FromBaseToUnitFunc": "{x} / 340.29",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "Ma" ]
```
angularsen, line 325: It seems both "M" and "Ma" are used, I don't know which one is more widely used, but we should add both here and the most widely used one should be the first in the array, since that will be used by `ToString()`.

Common/UnitEnumValues.g.json
```diff
@@ -1343,7 +1343,8 @@
     "UsSurveyFootPerSecond": 29,
     "YardPerHour": 30,
     "YardPerMinute": 31,
-    "YardPerSecond": 32
+    "YardPerSecond": 32,
+    "Mach": 42
```
angularsen, line 1347: You asked about why the number is 42 and not 33. This is intentional to reduce merge conflicts, when multiple pull requests adds units to the same quantity. It picks a random value for the next 10 available values.

UnitsNet.Tests/CustomCode/SpeedTests.cs
```diff
@@ -73,6 +73,8 @@ public class SpeedTests : SpeedTestsBase
 
         protected override double UsSurveyFeetPerHourInOneMeterPerSecond => 1.1811E4;
 
+        protected override double MachInOneMeterPerSecond => (1.0 / 340.29);
```
angularsen, line 76: This should be a calculated number, not a calculation.

The idea is to look up this value from a reliable source, such as another unit converter, wikipedia, wolfram, google and similar, in order to verify we got the conversion functions right - and in particular, got the _direction_ of the from/to conversion functions right.

angularsen (changes requested):
Sorry for the delay, work and vacation 🌞

I fixed the test case that you hit, in #1108 , then rebased this PR to bring that fix in.

This is almost ready to merge, just a few minor changes required.

UnitsNet.Tests/CustomCode/SpeedTests.cs
tjrtfarnham: Thanks @angularsen.  I have pushed what I hope are the requested changes to tjrtfarnham:add-mach-speed-unit.  Do I need to create another pull request.  Sorry, I am a total newbie when it comes to using github.

UnitsNet.Tests/CustomCode/SpeedTests.cs
```diff
@@ -73,7 +73,7 @@ public class SpeedTests : SpeedTestsBase
 
         protected override double UsSurveyFeetPerHourInOneMeterPerSecond => 1.1811E4;
 
-        protected override double MachInOneMeterPerSecond => (1.0 / 340.29);
+        protected override double MachInOneMeterPerSecond =>2.9386699579770E-3;
```
angularsen, line 76: Verified at https://www.metric-conversions.org/speed/mach-to-meters-per-second.htm

UnitsNet.Tests/CustomCode/SpeedTests.cs
angularsen: Perfect, that is the way to go. The new commits arrive in the pull request for me to review 👍

angularsen (approved):

angularsen:
Thanks! The nuget should be out in an hour or so. The builds are a bit slow these days.

[Release UnitsNet/4.139.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.139.0)

PR #1106, opened 2022-07-12
FeetInches to architectural string

### LINKED ISSUE #1075: FeetInches.ToString as architectural format
Not really

It would be nice if FeetInches had a method to output a string in architectural format (including fractional inches). Using the example from below, it would output a string like `5' - 4 1/2"`.  

https://github.com/angularsen/UnitsNet/blob/ebd3275b84991f5d90113a03fb569cddd655c86e/UnitsNet/CustomCode/Quantities/Length.extra.cs#L235

**Describe alternatives you've considered**
An extension method works fine, and I'm happy to do this if it's not desired in this codebase but figured there may be others wishing for similar functionality.

**Additional context**
This is a pretty common notation in the US in certain contexts. I'm not sure if (the few) other cultures that use feet and inches have a different architectural format, so not sure how CultureInfo might play a role if any.

Related issue (never implemented AFIAK) #676. This would have to be implemented at the same time, so another thing to consider is how to deal with the fractional precision for rounding. Maybe via an argument, like an integer for the denominator?

I can contribute if this is wanted. I'm assuming we wouldn't want to override `ToString()` or any of it's overloads since that's already established. Maybe it'd be like `ToArchitecturalString(int fractionDenom)` ?

lwestfall:
Fixes #1075. 

Also fixes broken image URL in readme from a long time ago :)

My coverage gutters weren't working locally so not 100% sure the new method has full coverage. Will wait for codecov to do its thing and will add test cases as needed.

### DIFF
--- a/UnitsNet.Tests/CustomCode/LengthTests.cs
+++ b/UnitsNet.Tests/CustomCode/LengthTests.cs
@@ -54,7 +54,7 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double ShacklesInOneMeter => 0.0364538;
 
-        protected override double NauticalMilesInOneMeter => 1.0 / 1852.0;
+        protected override double NauticalMilesInOneMeter => 1.0/1852.0;
 
         protected override double HandsInOneMeter => 9.8425196850393701;
 
@@ -82,38 +82,38 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double DataMilesInOneMeter => 0.000546807;
 
-        [Fact]
+        [ Fact]
         public void AreaTimesLengthEqualsVolume()
         {
-            Volume volume = Area.FromSquareMeters(10) * Length.FromMeters(3);
+            Volume volume = Area.FromSquareMeters(10)*Length.FromMeters(3);
             Assert.Equal(volume, Volume.FromCubicMeters(30));
         }
 
         [Fact]
         public void ForceTimesLengthEqualsTorque()
         {
-            Torque torque = Force.FromNewtons(1) * Length.FromMeters(3);
+            Torque torque = Force.FromNewtons(1)*Length.FromMeters(3);
             Assert.Equal(torque, Torque.FromNewtonMeters(3));
         }
 
         [Fact]
         public void LengthTimesAreaEqualsVolume()
         {
-            Volume volume = Length.FromMeters(3) * Area.FromSquareMeters(9);
+            Volume volume = Length.FromMeters(3)*Area.FromSquareMeters(9);
             Assert.Equal(volume, Volume.FromCubicMeters(27));
         }
 
         [Fact]
         public void LengthTimesForceEqualsTorque()
         {
-            Torque torque = Length.FromMeters(3) * Force.FromNewtons(1);
+            Torque torque = Length.FromMeters(3)*Force.FromNewtons(1);
             Assert.Equal(torque, Torque.FromNewtonMeters(3));
         }
 
         [Fact]
         public void LengthTimesLengthEqualsArea()
         {
-            Area area = Length.FromMeters(10) * Length.FromMeters(2);
+            Area area = Length.FromMeters(10)*Length.FromMeters(2);
             Assert.Equal(area, Area.FromSquareMeters(20));
         }
 
@@ -218,6 +218,20 @@ namespace UnitsNet.Tests.CustomCode
 
             Assert.Equal(0.0508, inSI.Value);
             Assert.Equal(LengthUnit.Meter, inSI.Unit);
+        }
+
+        [Theory]
+        [InlineData(-1.0, -1.0)]
+        [InlineData(-2.0, -0.5)]
+        [InlineData(0.0, 0.0)]
+        [InlineData(1.0, 1.0)]
+        [InlineData(2.0, 0.5)]
+        public static void InverseReturnsReciprocalLength(double value, double expected)
+        {
+            var length = new Length(value, Units.LengthUnit.Meter);
+            var inverseLength = length.Inverse();
+
+            Assert.Equal(expected, inverseLength.InverseMeters);
         }
 
         [Theory]
--- a/UnitsNet.Tests/CustomCode/LengthTests.cs
+++ b/UnitsNet.Tests/CustomCode/LengthTests.cs
@@ -54,7 +54,7 @@ namespace UnitsNet.Tests.CustomCode
 
         protected override double ShacklesInOneMeter => 0.0364538;
 
-        protected override double NauticalMilesInOneMeter => 1.0/1852.0;
+        protected override double NauticalMilesInOneMeter => 1.0 / 1852.0;
 
         protected override double HandsInOneMeter => 9.8425196850393701;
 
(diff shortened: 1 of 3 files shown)


UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -236,5 +236,70 @@ public string ToString(IFormatProvider? cultureInfo)
             // So inches are rounded when converting from base units to feet/inches.
             return string.Format(cultureInfo, "{0:n0} {1} {2:n0} {3}", Feet, footUnit, Math.Round(Inches), inchUnit);
         }
+
+        /// <summary>
+        ///     Outputs feet and inches on the format: {feetValue}' - {inchesValueIntegral}[ {inchesValueFractional}]"
+        ///     The inches are rounded to the nearest fraction of the fractionDenominator argument and reduced over the greatest common divisor.
+        /// </summary>
+        /// <param name="fractionDenominator">The maximum precision to express the rounded inch fraction part</param>
+        /// <example>Length.FromFeetInches(3,2).ToArchitecturalString(32).ToString() outputs: "3 ft 2 in"</example>
```
angularsen, line 245: Shouldn't the example output string be `3' - 2"` ?

UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -236,5 +236,70 @@ public string ToString(IFormatProvider? cultureInfo)
             // So inches are rounded when converting from base units to feet/inches.
             return string.Format(cultureInfo, "{0:n0} {1} {2:n0} {3}", Feet, footUnit, Math.Round(Inches), inchUnit);
         }
+
+        /// <summary>
+        ///     Outputs feet and inches on the format: {feetValue}' - {inchesValueIntegral}[ {inchesValueFractional}]"
+        ///     The inches are rounded to the nearest fraction of the fractionDenominator argument and reduced over the greatest common divisor.
+        /// </summary>
```
angularsen, line 243: Missing `/` before `{inchesValueIntegral}`.
Should probably also give a short description that the fractional will be omitted if it is `1`.

UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -236,5 +236,70 @@ public string ToString(IFormatProvider? cultureInfo)
...
+        /// <param name="fractionDenominator">The maximum precision to express the rounded inch fraction part</param>
+        /// <example>Length.FromFeetInches(3,2).ToArchitecturalString(32).ToString() outputs: "3 ft 2 in"</example>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown when fractionDenominator is &lt; 1 (use 1 to round to nearest integer, with no fraction)</exception>
```
angularsen, line 246: I would move the `use 1 to round to nearest integer, with no fraction` description to `fractionDenominator` parameter xmldoc, so it helps explain what values you can give.

UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -236,5 +236,70 @@ public string ToString(IFormatProvider? cultureInfo)
...
+        public string ToArchitecturalString(int fractionDenominator)
+        {
+            if (fractionDenominator < 1)
+            {
+                throw new ArgumentOutOfRangeException(nameof(fractionDenominator), "Denominator for fractional inch cannot be zero");
```
angularsen, line 251: To be more precise, maybe change it to `"Denominator for fractional inch must be greater than zero."`

angularsen: You could also put this text in the `ArgumentOutOfRangeException` xmldoc.

UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -236,5 +236,70 @@ public string ToString(IFormatProvider? cultureInfo)
...
+                        if (a > b)
+                            a %= b;
+                        else
+                            b %= a;
+                    }
+
+                    return a | b;
+                }
+
+                var gcd = greatestCommonDivisor((int)Math.Abs(numerator), fractionDenominator);
```
angularsen, line 285: Nitpick, but `int gcd = ` would be slightly easier to read here.

UnitsNet/CustomCode/Quantities/Length.extra.cs
angularsen: I would love a few more examples to help illustrate how different `fractionDenominator` values determines the output for the same input.

For example:
```xmldoc
<example>
<code>
var length = Length.FromFeetInches(3, 2.6);
length.ToArchitecturalString(1)   => 3' - 3"
length.ToArchitecturalString(2)   => 3' - 2 1/2"
length.ToArchitecturalString(4)   => 3' - 2 1/2"
length.ToArchitecturalString(8)   => 3' - 2 5/8"
length.ToArchitecturalString(16)  => 3' - 2 5/8"
length.ToArchitecturalString(32)  => 3' - 2 19/32"
length.ToArchitecturalString(128) => 3' - 2 77/128"
</code>
</example>
```

UnitsNet.Tests/CustomCode/LengthTests.cs
```diff
@@ -233,5 +233,24 @@ public static void InverseReturnsReciprocalLength(double value, double expected)
 
             Assert.Equal(expected, inverseLength.InverseMeters);
         }
+
+        [Theory]
+        [InlineData(3, 2.563, 16, "3' - 2 9/16\"")]
+        [InlineData(3, 2.563, 32, "3' - 2 9/16\"")]
+        [InlineData(3, 2, 16, "3' - 2\"")]
+        [InlineData(3, 2.5, 1, "3' - 2\"")]
+        [InlineData(0, 2, 32, "2\"")]
```
angularsen, line 242: Nice tests.

Can we also add these test values from my example in another PR comment? They helped me understand how it works better. 

Adapt the example to [InlineData].

```
var length = Length.FromFeetInches(3, 2.6);
length.ToArchitecturalString(1)   => 3' - 3"
length.ToArchitecturalString(2)   => 3' - 2 1/2"
length.ToArchitecturalString(4)   => 3' - 2 1/2"
length.ToArchitecturalString(8)   => 3' - 2 5/8"
length.ToArchitecturalString(16)  => 3' - 2 5/8"
length.ToArchitecturalString(32)  => 3' - 2 19/32"
length.ToArchitecturalString(128) => 3' - 2 77/128"
```

angularsen (changes requested):
Thanks! This looks great, just a few minor suggestions.

lwestfall:
Thanks for the feedback! I'll address these and try to push by the end of the week, work is a little busy for me the next couple of days.

lwestfall:
Fixes pushed! I think I addressed all comments, and I left them open so you can re-review. Let me know if you find anything else

angularsen:
Awesome, great work! The nuget should be out in a few minutes.

[Release UnitsNet/4.138.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.138.0)

angularsen:
Next hour, build times are very slow lately.

PR #1109, opened 2022-07-20
luminance quantity definition is added.

umutsoysalansys:


umutsoysalansys:
@tmilnthorp could you also take a look at this one please?

PR #1111, opened 2022-07-20
Add Luminance units

umutsoysalansys:


Common/UnitDefinitions/Luminance.json
```diff
@@ -0,0 +1,67 @@
...
+    },
+	{
+      "SingularName": "CandelaPerSquareInch",
+      "PluralName": "CandelasPerSquareInch",
+      "FromUnitToBaseFunc": "{x}* 1.5500031e3",
+      "FromBaseToUnitFunc": "{x}/ 1.5500031e3",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "Cd/s·in²" ]
```
angularsen, line 49: Should this not be `Cd/in²`?

angularsen (changes requested):
This looks almost ready to go, just a minor issue found.

Common/UnitDefinitions/Luminance.json
umutsoysalansys: Thanks for cathing, yes it was a typo by going from sq-in to in². 
Additionally I realized that UnitEnumValues.g.json file is not overwritten and keeping some extra enums that is initially generated, I tried to delete this file totally and regenerate it by generate code but it seems to have a larger footprint so I avoided. What would you suggest to clear these extra fields?

angularsen: Ah, I did not spot these myself.

Try deleting everything under `"Luminance": { ... }` in the JSON file, then re-run codegen.
It should then allocate unit enum values only for the units that actually exist.

As you noticed, by design it does not remove any values if units are removed, to preserve backwards compatibility.

umutsoysalansys: I see, yes it makes sense. Cleaned that part, also thanks to @tmilnthorp for pointing out autogenerated parts as well.

angularsen:
Great, thanks!
The nuget should be out in an hour or so.

[Release UnitsNet/4.140.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.140.0)

PR #1112, opened 2022-07-30
Add ElectricConductivity.SiemensPerCentimeter

QuantumPhoton1:
Common unit for electrical conductivity in water quality sensors is the microsiemen per centimeter.

Added SiemenPerCentiMeter as well as the milli and micro prefixes.


angularsen:
Thanks, looks good to me!

Nuget should be out in an hour or so.
[Release UnitsNet/4.141.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.141.0)

PR #1113, opened 2022-08-03
Add Energy divided by Power equals TimeSpan operator

dglozano:
This is useful in the context of battery charging. If you have an empty battery with
a capacity of 100 kwH and you connect it to a 20kw power supply, you might want
to know how long it will take to fully charge (5 hours) by doing energy / power.

Example
```
            var electricCurrent = ElectricCurrent.FromAmperes(32);
            var electricPotential = ElectricPotential.FromVolts(230);
            var chargingPower = electricCurrent * electricPotential;
            var batteryCapacity = Energy.FromKilowattHours(1000);

            TimeSpan chargingTime = batteryCapacity / chargingPower;
```

### DIFF
--- a/UnitsNet/CustomCode/Quantities/Energy.extra.cs
+++ b/UnitsNet/CustomCode/Quantities/Energy.extra.cs
@@ -19,6 +19,12 @@ namespace UnitsNet
             return Power.FromWatts(energy.Joules / duration.Seconds);
         }
 
+        /// <summary>Get <see cref="Duration"/> from <see cref="Energy"/> divided by <see cref="Power"/>.</summary>
+        public static Duration operator /(Energy energy, Power power)
+        {
+            return Duration.FromSeconds(energy.Joules / power.Watts);
+        }
+
         /// <summary>Get <see cref="Power"/> from <see cref="Energy"/> times <see cref="Frequency"/>.</summary>
         public static Power operator *(Energy energy, Frequency frequency)
         {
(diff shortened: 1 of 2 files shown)


UnitsNet/CustomCode/Quantities/Energy.extra.cs
```diff
@@ -19,6 +19,12 @@ public partial struct Energy
             return Power.FromWatts(energy.Joules / duration.Seconds);
         }
 
+        /// <summary>Get <see cref="TimeSpan"/> from <see cref="Energy"/> divided by <see cref="Power"/>.</summary>
+        public static TimeSpan operator /(Energy energy, Power power)
```
angularsen, line 23: Can we change TimeSpan to Duration?

This way it is an `IQuantity`. We usually support both Duration and TimeSpan in overloads that take duration as a parameter, but when it is the result we prefer Duration.

angularsen (changes requested):
Just a minor suggestion

UnitsNet/CustomCode/Quantities/Energy.extra.cs
dglozano: Perfect, was wondering which one was preferred. Have updated 🙌

angularsen:
Great, thanks! Nuget should be out in the next hour or so.

[Release UnitsNet/4.143.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.143.0)

PR #1124, opened 2022-08-23
IQuantity with QuantityValue as Value type (PoC)

lipchev:
As per our discussion over at #875 here's the proof of concept, based on the explicit struct layout proposed by @AndreasLeeb and @pgrawehr (_could you please check my modification - I had to make the fields nullable in order to avoid having the DataContractSerializers spew the trash value produced by the overlapping bits_). 
- `QuantityValue` implements most of the `INumber` interface members / operators (operations  preserving decimal when possible, failing back to double, clamped to the real values- infinities excluded) (see the [QuantityValueTests](https://github.com/lipchev/UnitsNet/blob/96882e631c6c786c5f09eee283d572451784f60b/UnitsNet.Tests/QuantityValueTests.cs) and the [QuantityComparisonTests](https://github.com/lipchev/UnitsNet/blob/96882e631c6c786c5f09eee283d572451784f60b/UnitsNet.Tests/QuantityComparisonTests.cs))
- changed the type of `IQuantity.Value/As` to `QuantityValue` **(explicit cast required)**
- added back the `IEquatable` interface: behavior depending on the runtime type of the quantity values (strict equality comparison for double/double)
- added back the fix for the non-reflexive `CompareTo` issue (and the corresponding tests)
- added a check preventing the possible stack overflow exception when calling `ToUnit()` with `default(TUnit)` (or any other invalid Enum value)
- DataContracts changed to reflect the possible use of either double or decimal as ValueType (the Json converters were not changed- but probably should be)
- removed the obsolete Json converter tests

Obviously this is quite a breaking change: 

1. `Mass.FromKilograms(1)` - the implicit conversion is to decimal instead of double, but I think is the correct behavior: [here](https://github.com/lipchev/UnitsNet/blob/96882e631c6c786c5f09eee283d572451784f60b/UnitsNet.Tests/GeneratedCode/TestsBase/MassTestsBase.g.cs#L365)'s an example where the resulting type is decimal, failing back to double as we get into the astronomical units (in contrast to [this other example](https://github.com/lipchev/UnitsNet/blob/96882e631c6c786c5f09eee283d572451784f60b/UnitsNet.Tests/GeneratedCode/TestsBase/MassTestsBase.g.cs#L2022) - where all values are doubles). I've intentionally left out [a margin](https://github.com/lipchev/UnitsNet/blob/96882e631c6c786c5f09eee283d572451784f60b/UnitsNet/QuantityValue.cs#L36) for the range check, as to avoid having such units loose all of their precision. 
2. The bigger issue is that an explicit cast (or some `ToDouble`/`ToDecimal` method) is required- but there is no way around that.
3. The default DataContractSerializer (both json and xml) format had to change - for a WCF service that shouldn't be much of a problem, if we're breaking the interface it's guaranteed that both client and service would have to be updated (unless using a converter/data surrogate etc.). I haven't touched the json converters - the tests still pass but only because I haven't touched the `IDecimalQuantity` interface (which currently stands for "a quantity using decimal-based conversion functions"- not sure what we want to do about that..).
4. The `IEquatable` implementation seems to be correct although possibly confusing- `Mass.FromKilograms(1m) == Mass.FromGrams(1000d)`,  `Mass.FromGrams(1000d) == Mass.FromGrams(1000d)`. `Mass.FromKilograms(1d) != Mass.FromGrams(1000d)`. 
5. Once the [GenericMath](https://docs.microsoft.com/en-us/dotnet/standard/generics/math) becomes generally available we could think about using an INumber for the Value type


pgrawehr:
I haven't looked at it very closely yet, but isn't this now conflicting with #1074 ?

lipchev:
@pgrawehr Yes, quite so I'm afraid- I actually started off with the implementation from your branch, but things evolved quickly from there - up to the point where It's now a case of _either_ or _neither_.  
We both bring the same breaking changes to the interface- however I went one further and replaced the internal representation of the `_value` field as well. This prompted the need to update the DataContract which in turn made me have to change the struct layout of the `QuantityValue` by making the fields Nullable..

lipchev:
Ah, damn it- I just re-read what I wrote last night, realizing that the equality contract is still broken- due to the lack of transitivity (again).. If we wanted to be strict in that regard we'd have to return `false` for `Mass.FromKilograms(1m) == Mass.FromGrams(1000d)`..

pgrawehr:
> @pgrawehr Yes, quite so I'm afraid- I actually started off with the implementation from your branch, but things evolved quickly from there - up to the point where It's now a case of _either_ or _neither_.  
> We both bring the same breaking changes to the interface- however I went one further and replaced the internal representation of the `_value` field as well. This prompted the need to update the DataContract which in turn made me have to change the struct layout of the `QuantityValue` by making the fields Nullable.. 

Then, to make it easier to compare and to comply with the license, please commit your changes separately and do not squash contributions from different users to one commit.

angularsen:
@lipchev 
> Ah, damn it- I just re-read what I wrote last night, realizing that the equality contract is still broken- due to the lack of transitivity (again).. If we wanted to be strict in that regard we'd have to return `false` for `Mass.FromKilograms(1m) == Mass.FromGrams(1000d)`..

I lost a bit momentum on this one after getting backlash on the change, but I personally think strict equality is the better compromise after ages of discussions. If you agree, please help me review the PR so we can get it merged.

[✨ Add back IEquatable with strict equality PR 1100](https://github.com/angularsen/UnitsNet/pull/1100)

Original discussion and decision for strict equality:
https://github.com/angularsen/UnitsNet/issues/1017#issuecomment-1028402841

PR #1125, opened 2022-08-29
Fix wrong SolarMass abbreviation

Brunni:
The wrong UTF-8 symbol was used.

Correct: https://en.wiktionary.org/wiki/%E2%98%89

Wrong:
https://en.wikipedia.org/wiki/Circled_dot

### DIFF
--- a/Common/UnitDefinitions/Mass.json
+++ b/Common/UnitDefinitions/Mass.json
@@ -248,7 +248,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "M⊙" ]
+          "Abbreviations": [ "M☉" ]
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


tmilnthorp:
Interesting! This would be a breaking parse change. We should leave the old incorrect string as an abbreviation (non default), so that parsing the incorrect symbol will work until a major version. Can we mark symbols as obsolete @angularsen?

angularsen:
> Interesting! This would be a breaking parse change. We should leave the old incorrect string as an abbreviation (non default), so that parsing the incorrect symbol will work until a major version. Can we mark abbreviation strings as obsolete @angularsen?

I don't think so, we will just have to mark it as a breaking change in #982 and document it, I guess.

angularsen:
@Brunni If you can add the correct one first in the array, and keep the old one as second item in the array, then parsing won't break, while ToString() will prefer the first one. Then we can merge this.

angularsen:
Thank you! Nuget should be out in the next hour.

[Release UnitsNet/4.145.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.145.0)

PR #1133, opened 2022-09-02
💚 Fix CodeGen on Linux/WSL (#1133)

Brunni:
The codegen did not start using `dotnet run --project CodeGen` because there were still back-slashes used in NanoFrameworkGenerator. 

Tested in WSL


angularsen:
[Release UnitsNet/4.145.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.145.0)

PR #1136, opened 2022-09-10
Upgrade .NET runtime to 6.0 for tests, CodeGen and Benchmark

Brunni:
.NET 5 is out of active support.

So I don't have it installed on my computer but I'm using .NET6 as it is an LTS version.

Therefore, I would suggest to upgrade things which are executed on the developer machine / CI (e.g. tests, benchmark, codegen) to .NET6 target framework.

https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core


angularsen:
See #982 

I believe it is already upgraded there, and if not, please target that branch instead.

PR #1140, opened 2022-09-27
Integration with the JS package

haimkastner:
- Trigger JS package re-generate once a new release created
- Add link in docs to the other platforms' packages

### DIFF
--- a/.github/workflows/trigger-js-release.yml
+++ b/.github/workflows/trigger-js-release.yml
@@ -10,11 +10,11 @@ jobs:
     steps:
       - name: Call re-run generate & publish job 
         env:
-          UNITISNET_JS_WORKFLOW_TOKEN: ${{ secrets.UNITISNET_JS_WORKFLOW_TOKEN }}
+          UNITSNET_JS_WORKFLOW_TOKEN: ${{ secrets.UNITSNET_JS_WORKFLOW_TOKEN }}
         run: |
           # Trigger dispatches event to rebuild unitsnet-js package 
           curl --location --request POST 'https://api.github.com/repos/haimkastner/unitsnet-js/actions/workflows/nodejs.yml/dispatches' \
           --header 'Accept: application/vnd.github.everest-preview+json' \
           --header 'Content-Type: application/json' \
-          --header "Authorization: Basic $UNITISNET_JS_WORKFLOW_TOKEN" \
+          --header "Authorization: Basic $UNITSNET_JS_WORKFLOW_TOKEN" \
           --data-raw '{"ref": "master" }'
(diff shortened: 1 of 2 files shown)


angularsen:
Sorry for the late reply, I have been missing out on email notifications for some time.

First off, kudos to your project. I love that a TypeScript version is being maintained.

I'm not very familiar with Github actions, so I don't know what is the best way to achieve this. It seems maybe it is not possible for repo B to subscribe to changes to repo A, it must be repo A that upon change notifies repo B.

Your proposed solution might work, but have you looked into "dispatch" events?

https://github.com/orgs/community/discussions/26323#discussioncomment-3703865
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

haimkastner:
Thank you :)

And right, I have looked for a way to trigger a build by "subscribing" to other repo events, but looks like there is no option like that...
Seems like the only way is to trigger it by the other repo.

I have changed the call to use GitHub's dispatches API instead of getting the latest job and re-running it, this way is more elegant to achieve the rebuild, thank you for the tip.

angularsen:
Great, let's give it a try.
I fixed a typo in the env name, it is now `UNITSNET_JS_WORKFLOW_TOKEN`.

Will you pass me the secret somehow? And I also need to figure out how to configure it.

angularsen:
You can email me at andreas.larsen84@gmail.com, or set up some other more secure channel if you wish. Let me know what works for you.

haimkastner:
Great! I have sent you the key and how to extract it to your email.

Once you have the key you only need to add a new key to the reop's actions keys.
Go to the repo's `settings` tab-> `Secrets` -> `Actions` -> `New repository secret` add it, and that's it.

angularsen:
@haimkastner Seems the token is missing admin privileges. Let me know if you want to try a different token or if you can upgrade the existing token.

https://github.com/angularsen/UnitsNet/actions/runs/3373552228/jobs/5598237504

```json
{
  "message": "Must have admin rights to Repository.",
  "documentation_url": "https://docs.github.com/rest/reference/actions#create-a-workflow-dispatch-event"
}
```

haimkastner:
Yes agree, I will send a commit fix for it.

Regarding the token, I have just tested it now using curl and it worked for me. 
Maybe you triggered the release before you set up the environment?  
Can you try to run it directly using curl/postman? 
```bash
curl --location --request POST 'https://api.github.com/repos/haimkastner/unitsnet-js/actions/workflows/nodejs.yml/dispatches' \
--header 'Accept: application/vnd.github.everest-preview+json' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic <base64 string key>' \
--data-raw '{"ref": "master" }'
```

haimkastner:
The fix for a request failure non-zero exit code #1147

PR #1150, opened 2022-11-03
Fix installing .NET nanoFramework msbuild in AZDO pipeline

josesimoes:


josesimoes:
I can see that this will fail because I've removed the install part that is required in AppVeyor... Not sure if you want me to add that back to keep the pipeline building there and you'll remove it later...

josesimoes:
Reverted that change in the install PS1 and left a TODO comment for future reference.

angularsen:
Neat, thanks!

angularsen:
@josesimoes I can't find the task on the marketplace or google?

image

josesimoes:
That's odd...  😯 
Here it is: https://marketplace.visualstudio.com/items?itemName=nanoframework.install-nanoframework-msbuild-components

PR #1151, opened 2022-11-03
Fix build of nanoFramework with msbuild in VS2022

josesimoes:
- Add path to msbuild x64 and update PS1.


josesimoes:
The issue is now with the build components. #1152 need to be merged first and then this will build fine.

PR #1157, opened 2022-11-23
Add UnitMath.Clamp(value, min, max)

Muximize:
This adds `UnitMath.Clamp` similar to `Math.Clamp`.

### DIFF
--- a/UnitsNet.Tests/UnitMathTests.cs
+++ b/UnitsNet.Tests/UnitMathTests.cs
@@ -318,5 +318,40 @@ namespace UnitsNet.Tests
             Assert.Equal(150, sum.Value);
             Assert.Equal(LengthUnit.Centimeter, sum.Unit);
         }
+
+        [Fact]
+        public void ClampCalculatesCorrectly()
+        {
+            var min = Length.FromMeters(-1);
+            var max = Length.FromCentimeters(150);
+
+            var value1 = Length.FromMillimeters(33);
+
+            Length clampedValue = UnitMath.Clamp(value1, min, max);
+            Assert.Equal(33, clampedValue.Value);
+            Assert.Equal(LengthUnit.Millimeter, clampedValue.Unit);
+
+            var value2 = Length.FromMillimeters(-1500);
+
+            Length clampedMin = UnitMath.Clamp(value2, min, max);
+            Assert.Equal(-1, clampedMin.Value);
+            Assert.Equal(LengthUnit.Meter, clampedMin.Unit);
+
+            var value3 = Length.FromMillimeters(2000);
+
+            Length clampedMax = UnitMath.Clamp(value3, min, max);
+            Assert.Equal(150, clampedMax.Value);
+            Assert.Equal(LengthUnit.Centimeter, clampedMax.Unit);
+        }
+
+        [Fact]
+        public void ClampMinGreaterThanMaxThrowsException()
+        {
+            var min = Length.FromMeters(2);
+            var max = Length.FromCentimeters(150);
+            var value = Length.FromMillimeters(33);
+
+            Assert.Throws<ArgumentException>(() => UnitMath.Clamp(value, min, max));
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


UnitsNet.Tests/UnitMathTests.cs
```diff
@@ -318,5 +318,40 @@ public void SumOfLengthsWithSelectorCalculatesCorrectly()
...
+
+            Length clampedValue = UnitMath.Clamp(value1, min, max);
+            Assert.Equal(33, clampedValue.Value);
+            Assert.Equal(LengthUnit.Millimeter, clampedValue.Unit);
+
+            var value2 = Length.FromMillimeters(-1500);
+
+            Length clampedMin = UnitMath.Clamp(value2, min, max);
+            Assert.Equal(-1, clampedMin.Value);
+            Assert.Equal(LengthUnit.Meter, clampedMin.Unit);
```
angularsen, line 338: I'm not sure it's intuitive that the unit is changed when clamped.
Should it perhaps instead convert to the min/max in the value's unit?

```
min.ToUnit(value.Unit)
max.ToUnit(value.Unit)

**Review by angularsen — CHANGES_REQUESTED** [review]

File: UnitsNet.Tests/UnitMathTests.cs (hunk 1 above)
Comment (Muximize) [reply UnitsNet.Tests/UnitMathTests.cs:338]: Hmm, that sounds about right. But there is a consideration:

I took this from the tests of `UnitMath.Max` and `UnitMath.Min`, but there you have no obvious leading input value, `val1` and `val2` are equal and thus can be returned as-is.

I modeled  clamp as `UnitMath.Min(max, UnitMath.Max(value, min)`, so the semantics and documentation matches `Math.Clamp`, "returns min if value < min".

But you're right, that might not be what a user is expecting in this case. It's leaky abstractions biting us, if the internal unit and value where inaccessible this wouldn't be an issue. [I have an opinion about that 😄](https://github.com/angularsen/UnitsNet/pull/1051#issuecomment-1051428016).

Given this, if you still think we should convert min/max (and change the docs) I'll gladly update this PR.

Comment (Muximize) [reply UnitsNet.Tests/UnitMathTests.cs:338]: I just had a realization. The way we currently solve this for other methods where the expected return unit is not clear (min/max on collections, `Sum`) is to explicitly specify the unit. We could do that here as well:
```
UnitMath.Clamp(value, min, max, unitType)
```

angularsen: Min/max/sum need to specify the unit, but for Clamp I really think there is only one natural unit we should use, the input unit.

If you do feel there is a real use case for choosing the min/max unit (only when clamped), then we could:
- Add parameter `bool toBoundaryUnit = false`
- Use min/max unit if clamped and `toBoundaryUnit == true`

Muximize: No, that would be overkill, I agree we should follow the principle of least surprise here. Changes pushed!

angularsen: Great!

angularsen:

[Release UnitsNet/5.0.0-rc004 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.0.0-rc004)

Muximize:
@angularsen What's your opinion on backporting this to v4?

angularsen:
@Muximize Sure, seems straight forward.
Merged: #1213

angularsen:
[Release UnitsNet/4.151.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.151.0)

Muximize:
Thanks! :heart: Can you do a Nuget release?

angularsen:
I tried, but the CI build failed on net5.0 no longer being included out of the box.
Attempted a fix now, we'll see 🤞

https://dev.azure.com/unitsnet/Units.NET/_build/results?buildId=356

Muximize:
Your fix worked! 🚀

PR #1161, opened 2022-12-06
update benchmarks to net6.0

kasperk81:
.net 5 has reached its end of life and [framework name](https://learn.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-frameworks) for .net 5 and above is `netX.Y` rather than `netcoreappX.Y`


angularsen:
Thank you!

Issue #1129, opened 2022-08-30
Fully embrace INumber and generic math of .NET 7


bitbonk:
.NET 7 will introduce the [INumber&lt;TSelf&gt;](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.inumber-1?view=net-7.0) interface (along with accompanying with C# 11 features) that make a feature called [generic math](https://devblogs.microsoft.com/dotnet/preview-features-in-net-6-generic-math/) possible.

I think this would be a perfect fit for the quantities of UnitsNet and could potentially make the usage way more intuitive, flexible and overall enjoyable.

Please consider to fully embrace `INumber<TSelf>`  and generic math in a future version of UnitsNet.

Here is a short introduction to this feature set by @MadsTorgersen (first 25 minutes): [Where's C# headed? - Mads Torgersen - NDC Copenhagen 2022](https://youtu.be/v8bqAm4aUFM)
Official documentation: https://docs.microsoft.com/en-us/dotnet/standard/generics/math

tmilnthorp:
The biggest question is the supported deployments. Do we abandon netstandard/netframework support just for generic math? Neither are EOL yet.

tmilnthorp:
We already do, but without too many #ifs. I fear that the API support for generic math vs precompiled lambdas would be difficult to maintain with equivalent functionality. But definitely worth a try.

bitbonk:
My guess would be that [IQuantity&lt;TUnitType&gt;>](https://github.com/angularsen/UnitsNet/blob/77b1ef2e86fa8961650c281ec3aad6521dcd3963/UnitsNet/IQuantity.cs#L112) would need to derive from [INumber&lt;TSelf&gt;](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.inumber-1?view=net-7.0) somehow, right?

angularsen:
@bitbonk An early prototype here to explore how it can be used.
#984

I haven't yet looked into what the changes were compared to the experimental interfaces at the time, but I assume it is in the ballpark similar.

angularsen:
Some support for generic math was added in #1164 , it can be extended further.

bitbonk:
> Some support for generic math was added in #1164 , it can be extended further.

Awesome! What version of UnitsNet has generic math included?

angularsen:
[...]

PR #1164, opened 2022-12-08
Generic math for UnitsNet in .NET 7

AndreasLeeb:
Hi,

since it's already been quite some time since #984 and as #1129 shows that the topic is of interest, I also gave it a shot.
So far, it's been a surprisingly minimally invasive undertaking.

Changes so far:
- build UnitsNet in .NET 7 as well with very limited conditional compilation for the generic math interfaces
- extended IQuantity interface hierarchy to support generic math (as the TSelf parameter for CRT pattern, and the underlying value type are needed)
- automatically generated operator overloads and parsing methods are covered by the respective generic math interfaces

Still to do:
- also cover the manually added operator overloads in the *.extra.cs files with generic math interfaces
- also implement ISpanParsable, which will need some work, especially in the QuantityParser
- maybe for symmetry reasons also overload the unary + operator and implement IUnaryPlusOperators. However, I'm unsure because it's a pretty useless operator
- resolve nullable reference types chaos: since multi-targeting with net7.0 even when disabling the nullability in the csproj I kept getting tons of errors for both netstandard2.0 and net7.0 builds, so for now I've disabled the respective warnings. We should fully use the nullable reference types feature in the whole code base also for .netstandard2.0 to solve this tangle

Pain points:
- multiplication with one operand of Quantity type and one operand of double/decimal type only supported with Quantity being the first parameter (when adding the commutative case, I get errors 
grafik
- I suspect we'll run into the same issues with some custom overloads involving commutative multiplication with different quantity types


angularsen:
Awesome! I will have to take a closer look later.

angularsen:
I fixed nullable annotations in #1175 and added some initial extension methods `Sum` and `Average` to test out the new generic math interfaces in practice.

Some changes made to the PR:
- Merge in latest master, with nullable fix
- Re-enable nullable, remove mute of warnings
- Tweak `IQuantity` constraints for generic math, specifically the `IDivisionOperators` had to be changed to support dividing by a number for `Average()`. Details in ed27b33fa626296a577a65805e273249851c0919.
- The extension methods were written so we don't have to specify the generic type arguments, in particular supporting `Average` for both `double` and `decimal` quantities required two separate extension method classes to keep the same method name. I tried adding both value types to the generic math interface list, but that resulted in ambiguous method errors.


Please review my commits and take a look whether we should expand on the extension methods or make other tweaks, but I think this is more or less ready to go. 

There are definitely some paper cuts with regards to having units in the mix, such as addition picking the left hand side unit so my initial attempt at Average() resulted in the base unit instead of the unit in the collection. We'll just have to test this out more and get a better feel for it, but it does seem to kind of work.

AndreasLeeb:
Looks good, I think 👍 

Regarding the operators in the *.extra.cs files, that could be tackled easily by describing the dependencies (operations) between different quantities in the quantity JSON files, and then the operator overloads and the generic math interfaces for the quantity structs could also be automatically generated. But that's a topic for another time 😄

angularsen:
Thanks for the great work! The nuget should be out to play with shortly.

[Release UnitsNet/5.0.0-rc008 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.0.0-rc008)

Issue #1180, opened 2023-01-07
Split into subpackages


hughesjs:
**Is your feature request related to a problem? Please describe.**
It's unfortunate to have to ship all units if my application is only using a small subsection.
For instance, if I'm making a package concerned with weights, mass, and distance, there's not much point including `Candela` in my application.

**Describe the solution you'd like**
Split the solution into domain packages along the lines of `UnitsNet.Newtonian`, `UnitsNet.Electricity` etc.

**Describe alternatives you've considered**
A package for each unit rather than each domain, but that could get cumbersome

angularsen:
Hi, yes I have been doing some investigation into this.
Work in progress branch here: #1181

Please review the design proposal and the problems discovered so far, if you have any ideas or comments about this approach. If you want to help move that project forward, that would also be helpful. I don't have much time to spend on the project so it moves slowly.

PR #1181, opened 2023-01-07
XP One nuget per quantity

angularsen:
Related to #372, #1180

Experiment with splitting up UnitsNet nuget into:
- UnitsNet.Core (common types and abstractions)
- UnitsNet.Length
- UnitsNet.Mass
- etc..

This is already done for UnitsNet.nanoFramework nugets:
https://www.nuget.org/packages?q=unitsnet.nanoframework

We might also provide meta nuget packages that bring in multiple nugets in a single package:
- UnitsNet (all nugets, like today)
- UnitsNet.Meta.SI (all SI and SI derived nugets, such as Length, Force, Mass, Area, Speed etc)
- UnitsNet.Meta.Main (the most widely used nugets, striving for 80/20 rule)

## Benefits
- Smaller size, only bring the quantities you need. The `UnitsNet.dll` is ~2MB as of 2023-01-07.
- More extensible, it forces a design where plugging in 3rd party quantities is simpler.

## Problems
- `QuantityType` enum must be replaced by strings and using `QuantityInfo` instead
- `UnitSystem` and `BaseUnits` rely on the 7 SI quantities Length, Mass, etc.
- Arithmetic across quantities may not be feasible for the library to provide, such as `Speed = Length / Duration` and `Mass.FromGravitationalForce(Force f)`. We might be able to provide this with meta nuget packages that bring in the most common quantities and arithmetic/conversions between them.

## Design proposal
- Use reflection to look up all quantities in loaded assemblies at startup instead of hard coding it, to allow plugging in 3rd party quantities. Need to provide some configuration for assemblies to search or quantities to load.
- Make use of `QuantityInfo` to fully describe a quantity, instead of relying on generated code for known quantities spread over various types.
  - Unit abbreviations for different cultures, currently loaded by `UnitAbbreviationsCache.LoadGeneratedAbbreviations()` based on generated code.
  - Unit conversion functions, currently defined by generated code in `Acceleration.RegisterDefaultConversions` and similar for other quantities.


angularsen:
test comment

angularsen:
test 2

angularsen:
test 3

angularsen:
test 4

angularsen:
test 5

PR #1184, opened 2023-01-20
Add unit definition "MegaJoulePerTonne" to SpecificEnergy

kuhnm:
Used in electric power plants, where they fire some hundreds kilo of coal per minute.
Examples: https://www.enbw.com/unternehmen/konzern/energieerzeugung/fossile-energie/standorte.html


angularsen:
Thanks a lot, it was perfect!

Nuget should be out shortly. It is still in prerelease, but 5.0 is stable and will leave prerelease soon.

UnitsNet/5.0.0-rc012
https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.0.0-rc012

PR #1187, opened 2023-01-28
1 mile is 1609.344 metres

charleskclo:
A mile is 1609.344 metres instead of 1609.34 metres, according to this Wikipedia article:
https://en.wikipedia.org/wiki/Mile

"The statute mile was standardised between the British Commonwealth and the United States by an international agreement in 1959, when it was formally redefined with respect to SI units as exactly 1,609.344 metres (1.609344 km)."

### DIFF
--- a/Common/UnitDefinitions/Length.json
+++ b/Common/UnitDefinitions/Length.json
@@ -36,8 +36,8 @@
       "BaseUnits": {
         "L": "Mile"
       },
-      "FromUnitToBaseFunc": "{x} * 1609.34",
-      "FromBaseToUnitFunc": "{x} / 1609.34",
+      "FromUnitToBaseFunc": "{x} * 1609.344",
+      "FromBaseToUnitFunc": "{x} / 1609.344",
       "Localization": [
         {
           "Culture": "en-US",


Common/UnitDefinitions/Length.json
```diff
@@ -36,8 +36,8 @@
       "BaseUnits": {
         "L": "Mile"
       },
-      "FromUnitToBaseFunc": "{x} * 1609.34",
-      "FromBaseToUnitFunc": "{x} / 1609.34",
+      "FromUnitToBaseFunc": "{x} * 1609.344",
+      "FromBaseToUnitFunc": "{x} / 1609.344",
```
angularsen, line 40: Verified 
https://www.convertworld.com/en/length/mile/miles-to-meters.html

It seems the test case is accurate enough already so we can perhaps leave it for now.

https://github.com/angularsen/UnitsNet/blob/master/UnitsNet.Tests/CustomCode/LengthTests.cs#L42
```
protected override double MilesInOneMeter => 0.000621371;
```

The same converter offers `0.0006213712`, but the computed value on my calculator is `0.00062137119223733396961743418436332`.

angularsen (approved):
Nice catch, thanks!

angularsen:
Nuget should be out shortly. It is still in pre-release, but considered stable and ready to use. Will leave pre-release in not too long.

[Release UnitsNet/5.0.0-rc013 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.0.0-rc013)

angularsen:
@heathbm That error was awfully small, are you sure this is not due to a double precision error more than incorrect constant used?

However, if the constant is indeed wrong, would you like to attempt a pull request? I'm happy to assist. 

Fix the conversion functions in JSON, fix the test case and regenerate the code. Detailed steps in the wiki.

PR #1195, opened 2023-02-08
Convert Power to double

### LINKED ISSUE #1194: Power is using the decimal type
I'm upgrading to 5.1.0 in my projects and noticed that Power is using the decimal type.

I don't think this is a bug so much as questioning if this is expected? It has been this way since [2014](https://github.com/angularsen/UnitsNet/pull/44), I only just noticed since the value property in v5 now returns the underlying type rather than always double.

I would expect this type to be consistent with much of the other types: double.

Per the PR:

> I was thinking about this previously and I think decimal might be a better backing datatype because it can exactly represent powers of 10 because it's a decimal floating point type (double is binary floating point). It would make sense to use it since we're mainly concerned with SI units which are powers of 10, however it's a major breaking change and decimal is less efficient for calculations...

I don't think this is applicable anymore given the test checking within a tolerance. Let me make a PR and we can discuss,.

tmilnthorp:
Fixes #1194


tmilnthorp:
Fails in json tests. Unsure what to do here.

angularsen:
3 tests failing in CI it seems, all related to JSON converters.

image
https://dev.azure.com/unitsnet/Units.NET/_build/results?buildId=311&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=9d4476d9-f6e2-5f97-1036-4124d48c41e1

tmilnthorp:
> 3 tests failing in CI it seems, all related to JSON converters. We probably handpicked decimal quantities for these tests, but we would have to create a new dummy quantity for testing purposes I guess if we want to keep the decimal JSON functionality around.
> 
> It would also be important to verify that old serialized JSON for Power/BitRate/Information can still be deserialized after switching to double.
> 
> image https://dev.azure.com/unitsnet/Units.NET/_build/results?buildId=311&view=logs&j=3dc8fd7e-4368-5a92-293e-d53cefc8c4b3&t=9d4476d9-f6e2-5f97-1036-4124d48c41e1

Yea the tests were just using Power to test decimal types. I just convered them to another existing decimal type.

Def should add a test to test old output.

UnitsNet.Tests/DecimalOverloadTests.cs
```diff
@@ -18,7 +18,7 @@ public static void CreatingQuantityWithDoubleBackingFieldFromDecimalReturnsCorre
         public static void CreatingQuantityWithDecimalBackingFieldFromDecimalReturnsCorrectValue()
         {
             Power power = Power.FromWatts(1m);
-            Assert.Equal(1.0m, power.Watts);
+            Assert.Equal(1.0, power.Watts);
```
angularsen, line 21: Should change this to Information to keep test's intent of decimal-backed quantity

UnitsNet.Tests/IntOverloadTests.cs
```diff
@@ -18,7 +18,7 @@ public static void CreatingQuantityWithDoubleBackingFieldFromIntReturnsCorrectVa
         public static void CreatingQuantityWithIntBackingFieldFromIntReturnsCorrectValue()
         {
             Power power = Power.FromWatts(1);
-            Assert.Equal(1.0m, power.Watts);
+            Assert.Equal(1.0, power.Watts);
```
angularsen, line 21: Can be a separate PR, but these `Int/Long/DecimalOverloadTests` are weirdly named and probably stem from an older design of having multiple overloads. Probably better moved to `QuantityTests`, to test that quantities are correctly constructed for different numeric types as input.

UnitsNet.Tests/DecimalOverloadTests.cs
angularsen: Never mind, this is out of scope for this PR, see below comment on fixing up these old tests.

UnitsNet.Serialization.JsonNet.Tests/UnitsNetBaseJsonConverterTest.cs
```diff
@@ -54,8 +54,7 @@ public void UnitsNetBaseJsonConverter_ConvertValueUnit_works_as_expected()
 
             Assert.NotNull(result);
             Assert.IsType<Power>(result);
-            Assert.True(Power.FromWatts(10.2365m).Equals((Power)result, 1E-5m, ComparisonType.Absolute));
-
+            Assert.True(Power.FromWatts(10.2365).Equals((Power)result, 1e-5, ComparisonType.Absolute));
```
angularsen, line 57: Duplicate the test case, so we test both double and decimal quantities.

Nitpick, but I would rather test `result.Value` and `result.Unit` instead of using `Equals()` in this test, to not mix too many implementations.

angularsen (changes requested):
Looks good, just a few minor suggestions.

Question; what do you think of also changing Information/BitRate from decimal to double? 

On one hand, it is sort of weird to risk getting rounding errors like 8.00000001 bits. On the other hand, for all other units (kilobyte, etc) that are probably way more used, floating point is perhaps the expected value representation.

If we change all 3 quantities to double, we have the potential to clean up a LOT of QuantityValue complexity.

tmilnthorp:
> Question; what do you think of also changing Information/BitRate from decimal to double?

As much as it pains me, I think these actually do make sense as decimal by default. If we had generic support, it could be whatever the client wishes.

angularsen:
I'm still not sure about Information, there's a lot of code just to avoid rounding errors on the unit Bit. For all other units Kilobyte etc, a non-integer value is probably expected anyway. 

Although, with #1206 it would probably read 5.299999999 kB instead of 5.3 kB in some cases.
But, with #1206 we kind of force everyone to use `ToString("s2")` anyway, and that would return 5.30 kB.

ebfortin:
> Looks good, just a few minor suggestions.
> 
> Question; what do you think of also changing Information/BitRate from decimal to double?
> 
> On one hand, it is sort of weird to risk getting rounding errors like 8.00000001 bits. On the other hand, for all other units (kilobyte, etc) that are probably way more used, floating point is perhaps the expected value representation.
> 

I agree. Floating point error is a fact of life anyway. Some number can't even be represented and we still use them for computation everywhere. I don't think it's worth the complexity just for one unit.

UnitsNet.Tests/IntOverloadTests.cs
angularsen: removed them

UnitsNet.Tests/DecimalOverloadTests.cs
angularsen: I removed these

UnitsNet.Serialization.JsonNet.Tests/UnitsNetBaseJsonConverterTest.cs
angularsen: Keeping it to have some tests still deserializing decimal values

Issue #1183, opened 2023-01-17
QuantityFormatter inconsequential format strings


AndreasLeeb:
Hi,

as we were having issues with how the QuantityTypeConverter displayed small values, I've dug through the code of the QuantityFormatter and found a pecularity.

If you pass the general format string "G" (or none), FormatUntrimmed always uses 2 significant digits instead of the the number specified after G (e.g. "G5"). And there is the format string "S" that allows to specify the amount of significant digits. And, unfortunately, there is no way to always display the number in normal fixed-point notation, as values below 1e-3 are displayed in scientific notation.

So I propose to move the feature that you can specify the number of significant digits and use scientific notation below 1e-3 to the format string "G" (so "G" without a number still means "G2", and for "Gn" n significant digits are shown). And for never using scientific notation we could either use "S" (as that would become obsolete when "Gn" works) or introduce a separate letter for it. However, I favor the former option.

If you agree I would change this, so that it will be part of the v5 release.

angularsen:
Hm, the wiki docs seem outdated: https://github.com/angularsen/UnitsNet/wiki/String-Formatting

This is the latest PR that changed this:
[Add support for standard numeric format strings](https://github.com/angularsen/UnitsNet/pull/788)

In particular, this xmldoc describes the current capabilities and the wiki should be updated to reflect this. 
https://github.com/angularsen/UnitsNet/pull/788/files#diff-66fe4faace86846b656488e7f6cb47d9f5c74459b9803237656684483956efcf

Back to your proposal:
- Generally, yes, let's try to mimic [.NET's standard numeric format strings](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings)
- G2, G4 etc sounds useful, f.ex. `Length.FromMeters(-1.234567890e-25).ToString("G4")` should be `-1,235E-25 m` to match `(-1.234567890e-25).ToString("G4")`.
- G and s are not the same, so we probably should keep both. G is significant digits (G4 => 1.234), while S is significant digits AFTER radix (S4 => 1.2345).
- G equals S2 so `Length.FromMeters(1.23456).ToString()` or `ToString("G")` or `ToString("S2")`, all produce `1.23 m`. This differs from .NET standard, but we wanted the default ToString() to be nice and readable.

Or so I recall, I may be off on some details.

AndreasLeeb:
[...]

PR #1206, opened 2023-02-17
Don't override .NET format string G

tmilnthorp:
Addresses #1183

### DIFF
--- a/UnitsNet/QuantityFormatter.cs
+++ b/UnitsNet/QuantityFormatter.cs
@@ -16,7 +16,7 @@ namespace UnitsNet
         /// <summary>
         /// The available UnitsNet custom format specifiers.
         /// </summary>
-        private static readonly char[] UnitsNetFormatSpecifiers = { 'A', 'a', 'G', 'g', 'Q', 'q', 'S', 's', 'U', 'u', 'V', 'v' };
+        private static readonly char[] UnitsNetFormatSpecifiers = { 'A', 'a', 'Q', 'q', 'S', 's', 'U', 'u', 'V', 'v' };
 
         /// <summary>
         /// Formats a quantity using the given format string and format provider.
@@ -142,9 +142,6 @@ namespace UnitsNet
 
                 switch(formatSpecifier)
                 {
-                    case 'G':
-                    case 'g':
-                        return ToStringWithSignificantDigitsAfterRadix(quantity, formatProvider, 2);
                     case 'A':
                     case 'a':
                         var abbreviations = UnitAbbreviationsCache.Default.GetUnitAbbreviations(quantity.Unit, formatProvider);
--- a/UnitsNet.Tests/QuantityTests.ToString.cs
+++ b/UnitsNet.Tests/QuantityTests.ToString.cs
@@ -18,9 +18,9 @@ namespace UnitsNet.Tests
             {
                 var culture = CultureInfo.InvariantCulture;
                 Assert.Equal("5 kg", Mass.FromKilograms(5).ToString(culture));
-                Assert.Equal("5,000 g", Mass.FromGrams(5000).ToString(culture));
-                Assert.Equal("1e-04 long tn", Mass.FromLongTons(1e-4).ToString(culture));
-                Assert.Equal("3.46e-04 dN/m", ForcePerLength.FromDecinewtonsPerMeter(0.00034567).ToString(culture));
+                Assert.Equal("5000 g", Mass.FromGrams(5000).ToString(culture));
+                Assert.Equal("0.0001 long tn", Mass.FromLongTons(1e-4).ToString(culture));
+                Assert.Equal("0.00034567 dN/m", ForcePerLength.FromDecinewtonsPerMeter(0.00034567).ToString(culture));
                 Assert.Equal("0.0069 dB", Level.FromDecibels(0.0069).ToString(culture));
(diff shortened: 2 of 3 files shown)


tmilnthorp:
Fun times. Standard .NET formatting for ```double.MinValue``` is different between net48 and net7

angularsen:
Awesome!

tmilnthorp:
@angularsen alright. I made everything match the runtime default (which is "G" by the way, not "g").

Obviously not an API breaking change, but it is a behavior change. Perhaps for the better. Merge now or is this a v6 thing?

UnitsNet/QuantityFormatter.cs
```diff
@@ -142,9 +156,6 @@ private static string FormatUntrimmed<TUnitType>(IQuantity<TUnitType> quantity,
 
                 switch(formatSpecifier)
                 {
-                    case 'G':
-                    case 'g':
-                        return ToStringWithSignificantDigitsAfterRadix(quantity, formatProvider, 2);
```
angularsen, line 147: I think this change is for the better. It reduces the wtf-factor a bit by being more consistent with .NET, such as TimeSpan and numbers.  It also fixes the issue with `"g1"` not rounding to 1 decimals.

Precision errors will become more evident too when doing arithmetic, for people running into equality issues beyond units.

`ToString("s")` would equal today's behavior, and is hopefully fairly straight forward to setup in most UI frameworks.

However, I suspect a lot of people are using `ToString()` directly or indirectly in a lot of UI places so I think I would move it to v6.

Create a `release/v6` branch and target this PR there.

angularsen (changes requested):

tmilnthorp:
@angularsen updated per your suggestions.

I think we should also remove
- P/p
- U/u
- Q/q

These don't really add anything you can't do via the ```IQuantity``` interface.

Also maybe ```An```/```An``` should format the quantity using abbreviation n rather than just returning abbreviation n (e.g. ```123.4 deg``` instead of the current ```deg```).

angularsen:
Yes I'm thinking the same for Q/U. Remove it. You don't obtain "Seconds" from TimeSpan.ToString() either. It's an odd thing to do and it is available elsewhere.

Regarding P (percentage, numeric standard format), I think we should not support neither C (currency) nor P (percentage). They don't make sense. So I think we should choose the subset of standard formats we support, and document them.

tmilnthorp:
> Yes I'm thinking the same for Q/U. Remove it. You don't obtain "Seconds" from TimeSpan.ToString() either. It's an odd thing to do and unit/quantity name is available elsewhere.
> 
> Regarding P (percentage, numeric standard format), I think we should not support neither C (currency) nor P (percentage). They don't make sense. So I think we should choose the subset of standard formats we support, and document them.

Sorry, I meant I'll remove Q/U/V. 

Regarding P, or any other default format string that's invalid _for us_. Should we just blindly allow strings of out say ```15.3% m```, or throw a ```FormatException```? I'm leaning towards the exception.

angularsen:
> Regarding P, or any other default format string that's invalid _for us_. Should we just blindly allow strings of out say `15.3% m`, or throw a `FormatException`? I'm leaning towards the exception.

FormatException 👍

> Also maybe An/An should format the quantity using abbreviation n rather than just returning abbreviation n (e.g. 123.4 deg instead of the current deg).

It would be more consistent, but then you can't control the number formatting, for example to 2 digits after radix. I think abbreviations are more useful as a custom format for string interpolation:
```cs
var x = $"{myLength.Value:f2} {myLength:a0}"; // 5 ft
var y = $"{myLength.Value:f2} {myLength:a1}"; // 5 '
```

I think this is how it would look like:

```cs
// Standard formats with .NET standard/custom numeric format + default unit abbreviation
Print("e");         // 1.000235e+003 ft
Print("e2");        // 3.14e+000 ft
Print("f");         // 1000.235 ft
Print("f1");        // 1000.2 ft
Print("n");         // 1,000.235 ft
Print("n2");        // 1,000.23 ft
Print("r");         // 1000.23456789 ft
Print("#.0");       // 1000.2 ft
Print("00000.0");   // 01000.2 ft

// Custom formats -- For string interpolation.
Print("a");         // ft 
Print("a1");        // '
```

```cs
// FormatException thrown by dotnet, only for integer types:
// Print("x"); 
// Print("d");

// FormatException thrown by us, no longer supported
Print("c");         // ¤1,000.23 ft    
Print("p");         // 100,023.457% ft 
Print("q");         // Length          
Print("u");         // Foot            
Print("v");         // 1000.23456789   
```

angularsen:
@lipchev It's been ages, but isn't this more or less ready to go? Still need to remove QUV?
Can you update the PR description with a summary of the changes?

lipchev:
@angularsen do we still want this? can I include it in my v6 proposal?

angularsen:
Yeah I belive this is still useful. 

What do you mean by including it in your v6 proposal? 
If you mean taking it into your fractions pull request, then I'd rather merge this as a separate PR and avoid a godzilla PR 🐲😄

lipchev:
> Yeah I belive this is still useful.
> 
> What do you mean by including it in your v6 proposal? If you mean taking it into your fractions pull request, then I'd rather merge this as a separate PR and avoid a godzilla PR 🐲😄

Don't worry, I'll split it and it's _still_ going to be..._godzilla_ 🐲 

PS Lol, Is that a dragon emoji? I don't think you should be allowed to go around, claiming you're old.. 🤣

angularsen:
Lol 😆

angularsen:
Merged in #1450

PR #1208, opened 2023-02-20
Added Megameter (1e+6 meters)

Mando501:
Megameter is a useful intermediate unit between kilometer and light second

Use cases tend towards astronomical such as satellite orbits


tmilnthorp:
@Mando501 this can be added to the existing Prefixes for Meter

Mando501:
> 

I must have missed Prefixes. I'll take a look, thanks.

Mando501:
Went back and removed the entry in Length.json and added Mega to the Prefixes for Meter.

I can't believe I missed that prefixes were a thing - that makes so much more sense 😂

angularsen:
Thanks! Nuget is on the way.

[Release UnitsNet/5.4.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.4.0)

Issue #1115, opened 2022-08-08
Different abbreviations for volume unit Liters used


geseru:
Not the same abbreviation for Liters is used in the whole library.
Sometimes it's an uppercase letter "L", sometimes a lowercase letter "l" is used.

Some examples:
Volume                      Liter                    => abbreviation "l"
Volume flow              LiterPerMinute   => abbreviation  "L/min"
Volume flow              LiterPerHour      => abbreviation  "L/h" 
Volume flow              LiterPerDay       => abbreviations "l/day", "L/d", "LPD"
... 

The same abbreviation should be used in the whole library.

Brunni:
According to SI Symbols both l and L are correct: https://en.wikipedia.org/wiki/International_System_of_Units

According to CEFACT rec20 UN standard the lower case variant is used. 

According to a US National Institute website: https://physics.nist.gov/cuu/Units/outside.html#:~:text=Thus%2C%20although%20both%20l%20and,approved%20symbols%20for%20the%20liter.

> The liter in Table 6 deserves comment. This unit and its symbol l were adopted by the CIPM in 1879. The alternative symbol for the liter, L, was adopted by the CGPM in 1979 in order to avoid the risk of confusion between the letter l and the number 1. Thus, although both l and L are internationally accepted symbols for the liter, to avoid this risk the preferred symbol for use in the United States is L. Neither a lowercase script letter l nor an uppercase script letter L are approved symbols for the liter.

angularsen:
I too favor uppercase "L", so I propose we change them all in #982 ?

hughesjs:
The standard for units, to my understanding, is that unless it's named after someone, it's lower case. Sure `L` might be used about as widely as `l` but that's a side-effect of how common the usage is amongst laypeople. I feel that when there's the potential to use two "equally correct" abbreviations, we should follow convention.

angularsen:
I'd like to land on a convention here, but I don't want this to come down to personal preferences, mine the least. I want to follow industry standards, however it seems to be split on American vs European/international.

@Brunni listed 3 standards:

- [SI](https://en.wikipedia.org/wiki/International_System_of_Units) (European/international), uses "litre" and lowercase "l".
- [UN/CEFACT](https://unece.org/DAM/cefact/recommendations/rec20/rec20_rev3_Annex3e.pdf) (European) uses "litre" and lowercase "l".
[...]

PR #1209, opened 2023-02-20
Show example of en vs en-US liters preference (l/L)

tmilnthorp:
Example for addressing #1115


angularsen:
I think it is a neat idea, but I am not convinced there is a strong enough correlation between the culture you are running and the user's preference for L vs l.

As a counter-example, in Norwegian we use both "l" and "L".
I had no specific intuition on when to use what, but apparently our national language council recommends "L" for technical domains.

```
l | liter
L | liter (i teknisk språk)
```

https://www.sprakradet.no/sprakhjelp/Skriveregler/Forkortinger/

SI standard accepts both, so there is bound to be a mix internationally.
I'm thinking we should maybe instead just make it configurable and maybe that configuration's default could be based on the culture?

Not exactly sure how to make it configurable, beyond calling some static method on startup that re-configures unit abbreviations for liter units. Maybe we can have something enumerate all units with "liter" in the unit name and replace "L" with "l" and vice versa?

angularsen:
Fixed by #1404

PR #1210, opened 2023-02-21
Move unit abbreviation strings into culture-specific resource files

### LINKED ISSUE #1126: Bad Performance on first Power.From() call in .netCore
**Description**
The very first call on Power.From takes up to several seconds. 
All the following calls are fine (regarding the performance).

**Environment**
- .net Core 3.1
- 32bit ARM Platform
- Embedded Linux

**Repro/Code**
`var result = Power.From(5, PowerUnit.Watt);`

**Expected behavior**
The call should return within milliseconds, but it takes up to more than 10 seconds.

**Additional context**
I Experienced this issue after updating from Version 4.72.0 to Version 4.144.0.
It did not happen in the older version.

tmilnthorp:
Fixes #1126 

A quick draft to move the unit abbreviations into resource files. I also want to move methods for getting culture-specific abbreviations to the individual quantities.

This should:
- Reduce initialization time by removing the MapGeneratedLocalizations methods (TODO)
- Reduce in-memory footprint by only loading a requested culture
- Allow formatting to use methods on the quantities themselves
  - Which should allow for better extensibility as well
- Remove the unit abbreviations cache (deprecate)

### TODOs
- [ ] Test fallback to invariant culture `export DOTNET_SYSTEM_GLOBALIZATION_INVARIANT='1'` #1238


tmilnthorp:
Check #1126 after merge

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
```diff
@@ -1211,6 +1211,10 @@ public void ToString_ReturnsValueAndUnitAbbreviationInCurrentCulture()
                 Assert.Equal("1 nm/s²", new Acceleration(1, AccelerationUnit.NanometerPerSecondSquared).ToString());
                 Assert.Equal("1 g", new Acceleration(1, AccelerationUnit.StandardGravity).ToString());
             }
+            catch
+            {
+                throw;
```
tmilnthorp, line 1216: Although there should never be an exception in a ToString class, I want it to show up in the tests if there is!

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -256,6 +261,12 @@ private void GenerateStaticProperties()
             }
 
             Writer.WL($@"
+
+        /// <summary>
+        /// The per-culture abbreviations. To add a custom default abbreviation, add to the beginning of the list.
+        /// </summary>
+        public static IDictionary<(CultureInfo Culture, {_unitEnumName} Unit), List<string>> Abbreviations {{ get; }}
```
angularsen, line 268: Before even reading the rest, shouldn't this rather sit in `QuantityInfo`?
- This is where units and other data for a quantity is stored
- Reusable/extensible when working with `IQuantity`
- Reduce member count per quantity (binary size)

angularsen: Reading more, yes, `GetAbbreviations()` etc seems to me would sit nicely in `QuantityInfo`.

angularsen (commented):
I didn't get time to read much, the kids are demanding my presence in the couch for TV and candy.

Some initial thoughts here.

UnitsNet.Tests/GeneratedCode/TestsBase/TorqueTestsBase.g.cs
```diff
@@ -1343,6 +1343,10 @@ public void ToString_ReturnsValueAndUnitAbbreviationInCurrentCulture()
                 Assert.Equal("1 tf·m", new Torque(1, TorqueUnit.TonneForceMeter).ToString());
                 Assert.Equal("1 tf·mm", new Torque(1, TorqueUnit.TonneForceMillimeter).ToString());
             }
+            catch
+            {
+                throw;
+            }
```
angularsen, line 1349: This doesn't seem too useful, maybe just omit the catch in this case?

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
angularsen: The syntax would only be slightly longer `Length.Info.GetAbbreviation()`.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -396,7 +385,34 @@ public static string GetAbbreviation({_unitEnumName} unit)
...
+        public static IReadOnlyList<string> GetAbbreviations({_unitEnumName} unit, CultureInfo? culture = null)
+        {{
+            culture ??= CultureInfo.CurrentCulture;
+
+            if(!Abbreviations.TryGetValue((culture, unit), out var abbreviations))
+            {{
+                abbreviations = new List<string>();
+
+                const string resourceName = $""UnitsNet.GeneratedCode.Resources.{_quantity.Name}"";
+                var resourceManager = new ResourceManager(resourceName, typeof({_quantity.Name}).Assembly);
```
angularsen, line 406: I'm not familiar with manually instantiating `ResourceManager`. Should we cache and reuse these instances?

angularsen: Generally though, I think you are on to something good about reducing the generated code and moving things like localization to files that are read at runtime. 

The JIT clearly has issues with our codebase. Not sure if it is the large number of types/members or if it is just the large amount of code, or what exactly.

angularsen: Moving things into `QuantityInfo` is another way to avoid duplicating methods N times for N quantities.

The biggest hit is M members for M units though, which is roughly 10x the number of quantities.

angularsen (commented):
Another partial review, generally I think this is a good idea and would love to see some benchmarks on improved startup time due to JIT.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
tmilnthorp: @angularsen I like that. Let me give it a shot here

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
tmilnthorp: @angularsen I am doing it lazily to load into a List. However there is some caching done internally, if my read of the docs is correct.

tmilnthorp:
@angularsen moved to ```UnitInfo```. Just have to do the custom mapping now.

I have no clue why auto-gen for ```ElectricCharge``` gets moved around.

```
// * Summary *

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.19042.2604 (20H2/October2020Update)
AMD Ryzen 9 3900X, 1 CPU, 24 logical and 12 physical cores
.NET SDK=7.0.102
  [Host]     : .NET 6.0.14 (6.0.1423.7309), X64 RyuJIT
  DefaultJob : .NET 6.0.14 (6.0.1423.7309), X64 RyuJIT


|                 Method |          Mean |         Error |      StdDev |  Gen 0 |  Gen 1 | Allocated |
|----------------------- |--------------:|--------------:|------------:|-------:|-------:|----------:|
|            Constructor |      6.669 ns |     0.0742 ns |   0.0657 ns |      - |      - |         - |
|         Constructor_SI |    333.957 ns |     2.4437 ns |   2.2859 ns | 0.0229 |      - |     192 B |
|             FromMethod |     15.249 ns |     0.0954 ns |   0.0796 ns |      - |      - |         - |
|             ToProperty |     37.741 ns |     0.6000 ns |   0.5612 ns |      - |      - |         - |
|                     As |     38.237 ns |     0.4616 ns |   0.4318 ns |      - |      - |         - |
|                  As_SI |    333.398 ns |     6.2248 ns |   6.1136 ns | 0.0229 |      - |     192 B |
|                 ToUnit |     35.854 ns |     0.2862 ns |   0.2537 ns |      - |      - |         - |
|              ToUnit_SI |    340.429 ns |     6.7290 ns |   8.5101 ns | 0.0229 |      - |     192 B |
|           ToStringTest |  1,383.815 ns |    26.9084 ns |  28.7917 ns | 0.0801 |      - |     680 B |
|                  Parse | 90,754.795 ns |   911.6428 ns | 852.7512 ns | 6.3477 | 0.1221 |  53,538 B |
|          TryParseValid | 89,569.940 ns |   951.8402 ns | 794.8296 ns | 6.3477 | 0.2441 |  53,530 B |
|        TryParseInvalid | 64,205.361 ns | 1,022.9905 ns | 906.8544 ns | 5.0049 | 0.2441 |  42,766 B |
|           QuantityFrom |     66.576 ns |     1.3068 ns |   1.2835 ns | 0.0067 |      - |      56 B |
|           IQuantity_As |     41.443 ns |     0.6139 ns |   0.6029 ns | 0.0029 |      - |      24 B |
|        IQuantity_As_SI |    333.429 ns |     6.6227 ns |   6.1949 ns | 0.0229 |      - |     192 B |
|       IQuantity_ToUnit |     42.520 ns |     0.6126 ns |   0.5430 ns | 0.0067 |      - |      56 B |
| IQuantity_ToStringTest |  1,426.165 ns |    27.2386 ns |  27.9721 ns | 0.0801 |      - |     680 B |

// * Hints *
Outliers
  UnitsNetBenchmarks.Constructor: Default      -> 1 outlier  was  removed (8.33 ns)
  UnitsNetBenchmarks.FromMethod: Default       -> 2 outliers were removed, 4 outliers were detected (16.51 ns, 16.62 ns, 16.96 ns, 17.03 ns)
  UnitsNetBenchmarks.As_SI: Default            -> 1 outlier  was  detected (320.38 ns)
  UnitsNetBenchmarks.ToUnit: Default           -> 1 outlier  was  removed (38.08 ns)
  UnitsNetBenchmarks.TryParseValid: Default    -> 2 outliers were removed (94.28 us, 95.10 us)
  UnitsNetBenchmarks.TryParseInvalid: Default  -> 1 outlier  was  removed, 3 outliers were detected (62.01 us, 62.88 us, 66.01 us)
  UnitsNetBenchmarks.IQuantity_As: Default     -> 2 outliers were removed (45.16 ns, 45.31 ns)
  UnitsNetBenchmarks.IQuantity_As_SI: Default  -> 3 outliers were removed, 5 outliers were detected (321.38 ns, 324.70 ns, 347.70 ns..351.21 ns)
  UnitsNetBenchmarks.IQuantity_ToUnit: Default -> 1 outlier  was  removed (47.58 ns)

// * Legends *
  Mean      : Arithmetic mean of all measurements
  Error     : Half of 99.9% confidence interval
  StdDev    : Standard deviation of all measurements
  Gen 0     : GC Generation 0 collects per 1000 operations
  Gen 1     : GC Generation 1 collects per 1000 operations
  Allocated : Allocated memory per single operation (managed only, inclusive, 1KB = 1024B)
  1 ns      : 1 Nanosecond (0.000000001 sec)

// * Diagnostic Output - MemoryDiagnoser *


// ***** BenchmarkRunner: End *****
// ** Remained 0 benchmark(s) to run **
Run time: 00:05:34 (334.18 sec), executed benchmarks: 17

Global total time: 00:05:56 (356.86 sec), executed benchmarks: 17
```

tmilnthorp:
@angularsen now at the point where everything works - however routing everything via ```UnitInfo``` means instances of ```UnitAbbreviationsCache``` have data races on the static ```UnitInfo``` instances. So tests randomly fail 👎🏻

tmilnthorp:
> @angularsen now at the point where everything works - however routing everything via `UnitInfo` means instances of `UnitAbbreviationsCache` have data races on the static `UnitInfo` instances. So tests randomly fail 👎🏻

FYI tests run locally on re-run since the scope is narrower. Not sure what to do here if you want the lists in ```UnitInfo```.

angularsen:
@tmilnthorp I pushed a possible fix for resetting global static state in tests. All tests are green locally now. If we are happy with the solution, we can make things internal, cleanup names, document better etc.

https://github.com/tmilnthorp/UnitsNet/pull/1

angularsen:
@tmilnthorp  I like where this is going, but a few observations:

- `Quantity.Default` is assigned one instance for its static property
- `UnitAbbreviationsCache` is assigned an instance for each UACache instance
- Parse/TryParse instance methods copied from `Quantity` don't do much yet, but this can be extended later to support custom units.

I built on this and added `UnitsNetSetup` to gather global state in a single place as a singleton, with the possibility of passing instances of it to static methods later to override the defaults.
This way, state is no longer scattered among multiple classes that depend on each other in non-obvious ways and where initialization order has caused issues up until now.

Pull request with some changes on top of this branch:
https://github.com/tmilnthorp/UnitsNet/pull/2

angularsen:
@tmilnthorp I understand you are busy these days, I'll merge this and work some more on it.

tmilnthorp:
@angularsen thanks! I’ve had this on my mind, just have not had time to come back to it recently. Hope to get back into some things in the coming months.

angularsen:
@tmilnthorp No worries! 🙌

angularsen:
[Release UnitsNet/5.22.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.22.0)

Issue #1193, opened 2023-02-07
After update to v5, Equals() does not work as expected when using FluentAssersions' .BeEquevalentTo()


dmytro-gokun:
After updating to v5 when using FluentAssertions's .BeEquiavalentTo() on classes having UnitNet's types as members, we get false negatives, like:

```
Expected property actual.Mileage to be 0 m, but found 0 km
```

This worked properly in v4.

As far as I can understand, the reason for this is that .BeEquiavalentTo() uses .Equals() method which is implemented as:

```
        public bool Equals(Length other)
        {
            return new { Value, Unit }.Equals(new { other.Value, other.Unit });
        }
```

IMO, this is incorrect behavior. 0 km == 0m, 1km == 1000 m etc. Is not that a valid assumption when writing unit tests?

angularsen:
This is unfortunately by design and it has been discussed at length over years.

It seems trivial on the surface, but different quantities have units of different scale and you very quickly run into rounding errors that affect equality. We simply couldn't find a one size fits all for equality, strict equality is at least easy to reason about although not always intuitive.

I believe it is best summarized here:
https://github.com/angularsen/UnitsNet/pull/1100#issuecomment-1156012162

wenndemann:
Having the same problem after updating to v5. UnitsNet become useless for me since I have to deal with different units all around my project. 
I totally agree to @dmytro-gokun that 0km == 0m. Your argument that 0 Celsius != 0 Fahrenheit is true but 0 Celsius == 32 Fahrenheit should be true. 
Is the a possibility to change the default behavior of Equal? If not it should be considered to add support for this.

dmytro-gokun:
@angularsen While i agree that '0 Fahrenheit != 0 Celsius', it also seems apparent to me that for cases where it makes sense (e.g. 0km and 0m) the comparison should yield true. Think of it this way: Length.Zero produces '0 m' and we may have some parsing code which yields results in kilometers. When we write test code, we are forced to do Length.FromKilometers(0) to make the library happy, which looks kind of cumbersome and flaky.

One could argue, that we should somehow instruct FluentAssertions to use proper Equals overload with proper ComparisonType and tolerance. I'd would argue though, that having some sensible default tolerance to be used by Equals() w/o parameter would make much more sense. It will keep happy all the folks who were happy with old behavior, while leaving an escape hatch for those who want to specify tolerance/ComparisonType manually.

angularsen:
[...]

PR #1215, opened 2023-02-26
✨ Add Equals() to IQuantity interfaces

angularsen:
Ref #1193

In v5, the default equality implementation changed to strict equality and the existing methods to compare across units with a tolerance, but this was not available in `IQuantity` interfaces.

### Changes
- Add `Equals(IQuantity? other, IQuantity tolerance)` to `IQuantity`
- Add `Equals(TSelf? other, TSelf tolerance)` to `IQuantity<TSelf, TUnitType, TValueType>` for strongly typed comparisons
- Obsolete `Equals(TQuantity other, double tolerance, ComparisonType comparisonType)` method in quantity types

### DIFF
--- a/UnitsNet.Tests/QuantityTests.cs
+++ b/UnitsNet.Tests/QuantityTests.cs
@@ -1,6 +1,8 @@
 ﻿// Licensed under MIT No Attribution, see LICENSE file at the root.
 // Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
 
+using System;
+using System.Globalization;
 using UnitsNet.Units;
 using Xunit;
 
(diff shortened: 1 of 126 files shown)


UnitsNet.Tests/QuantityTests.cs
```diff
@@ -16,5 +17,42 @@ public void GetHashCodeForDifferentQuantitiesWithSameValuesAreNotEqual()
...
+        [InlineData("10 m", "8.9 m"  , ComparisonType.Absolute, 0.1, false)] // +/- 1m relative tolerance (10%) and some additional margin tolerate rounding errors in test case.
+        [InlineData("10 m", "9.1 m"  , ComparisonType.Relative, 0.1, true)]
+        [InlineData("10 m", "10.9 m" , ComparisonType.Relative, 0.1, true)]
+        [InlineData("10 m", "11.1 m" , ComparisonType.Relative, 0.1, false)]
+        public void Equals_IEquatableQuantity(string q1String, string q2String, ComparisonType comparisonType, double tolerance, bool expectedEqual)
+        {
+            IQuantity<Length, LengthUnit, double> q1 = Length.Parse(q1String, CultureInfo.InvariantCulture);
+            IQuantity<Length, LengthUnit, double> q2 = Length.Parse(q2String, CultureInfo.InvariantCulture);
+
+            Assert.Equal(expectedEqual, q1.Equals(q2, tolerance, comparisonType));
```
angularsen, line 36: Assert.NotEqual would not work with these interfaces, complaining about undefined generic members.

angularsen:
@tmilnthorp Would like your eyes on this whenever you have time.

tmilnthorp:
> @tmilnthorp Would like your eyes on this whenever you have time.

I have the same comment as before

angularsen:
> > @tmilnthorp Would like your eyes on this whenever you have time.
> 
> I have the same comment as before

I must have missed it, what comment?

tmilnthorp:
@angularsen:

Perhaps instead of using the ```Comparison``` parameter, we could have a single method:
```csharp
bool Equals(TQuantity other, TQuantity tolerance);
```

Or if you want asymmetric bounds:
```csharp
bool IsBetween(TQuantity other, TQuantity min, TQuantity max);
```

This gives a lot of flexibility to use the same method for handling both situations (as unit test frameworks do):
- Use as-is for absolute tolerance
- Multiply the expected result by a scaling factor for relative tolerance

This also lets us remove the ```Comparison``` enum entirely for the future.

dmytro-gokun:
@angularsen What did you decide to do about this PR? Is it going to be merged at some moment or?

angularsen:
This slipped off my radar honestly. 

@tmilnthorp Hm, yes, I like this contract better. Providing the error margin as a quantity. 

`IsBetween` could be an extension method, but it is just a minor convenience to an if-statement. 

So, I propose:
- Add `bool Equals(TQuantity other, TQuantity tolerance)` to `IQuantity` and implementations
- Mark existing `Equals(IQuantity, double, ComparisonType)` methods as obsolete
- Optional: Add extension method `IsBetween`.

@dmytro-gokun How does this work for you?

tmilnthorp:
@angularsen sounds good to me. Users can also use ```min < value < max```.

angularsen:
[Release UnitsNet/5.16.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.16.0)

PR #1221, opened 2023-03-06
Fix quantity name in obsolete attributes

jnyrup:
Noticed that all quantities suggested an overload taking an `Angle` 🙃


angularsen:
Thanks!

Nuget should be out shortly.

[Release UnitsNet/5.7.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.7.0)

PR #1224, opened 2023-03-09
ToUnit for IValueQuantity<T> should return IValueQuantity<T>

tmilnthorp:


UnitsNet.Tests/IValueQuantityTests.cs
```diff
@@ -50,5 +50,33 @@ public void IValueQuantityTDecimal_AsUnitSystem_ReturnsDecimal()
...
+        public void IValueQuantityTDouble_ToUnitUnitSystem_ReturnsIValueQuantityTDouble()
+        {
+            IValueQuantity<double> doubleQuantity = Temperature.FromDegreesCelsius(1234.5);
+            Assert.IsAssignableFrom<IValueQuantity<double>>(doubleQuantity.ToUnit(UnitSystem.SI));
+        }
+
+        [Fact]
+        public void IValueQuantityTDecimal_ToUnitEnum_ReturnsIValueQuantityTDecimal()
+        {
+            IValueQuantity<decimal> decimalQuantity = Power.FromWatts(1234.5);
```
angularsen, line 71: Use Information instead, since Power will be changed to double in v6.

UnitsNet/IValueQuantity.cs
```diff
@@ -48,5 +48,24 @@ public interface IValueQuantity<out TValueType> : IQuantity
...
         new TValueType As(UnitSystem unitSystem);
+
+        /// <summary>
+        ///     Converts this <see cref="IValueQuantity{TValueType}"/> to an <see cref="IValueQuantity{TValueType}"/> in the given <paramref name="unit"/>.
+        /// </summary>
+        /// <param name="unit">
+        ///     The unit <see cref="Enum"/> value. The <see cref="Enum"/> must be compatible with the units of the <see cref="IValueQuantity{TValueType}"/>.
+        ///     For example, if the <see cref="IValueQuantity{TValueType}"/> is a <see cref="Length"/>, you should provide a <see cref="LengthUnit"/> value.
+        /// </param>
+        /// <exception cref="NotImplementedException">Conversion was not possible from this <see cref="IValueQuantity{TValueType}"/> to <paramref name="unit"/>.</exception>
```
angularsen, line 59: I'm thinking NotSupportedException or InvalidOperationException instead, since units can be configured at runtime so it is generally not due to a missing implementation, but rather a mistake by the caller (wrong enum) or missing configuration for custom units.

This would have to change in other places too, could be a separate PR though.

angularsen (changes requested):
Looks good, just a few minor comments.

UnitsNet.Tests/IValueQuantityTests.cs
tmilnthorp: @angularsen that will work for the Enum overload, but Information doesn't support SI units.

UnitsNet/IValueQuantity.cs
tmilnthorp: @angularsen sounds good. I'll keep as-is for this PR then (for consistency) and we can take a pass for another PR.

I assume this would be a v6 issue since it would be a breaking behavior change.

UnitsNet.Tests/IValueQuantityTests.cs
tmilnthorp: Just noticed we do the same for ```IValueQuantityTDecimal_AsUnitSystem_ReturnsDecimal```.

tmilnthorp:
@angularsen updated per your review (minus the UnitSystem overload which I had to keep as Power like the other tests).

angularsen:
[Release UnitsNet/5.9.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.9.0)

PR #1229, opened 2023-03-17
Publish test results to ADO pipeline

tmilnthorp:


Build/build-functions.psm1
```diff
@@ -74,7 +74,6 @@ function Start-Tests {
   write-host -foreground blue "Run tests...`n---"
   foreach ($projectPath in $projectPaths) {
     $projectFileNameNoEx = [System.IO.Path]::GetFileNameWithoutExtension($projectPath)
-    $reportFile = "$testReportDir\${projectFileNameNoEx}.xunit.xml"
```
tmilnthorp, line 77: Was not used

angularsen:
Nice! I was thinking the same thing the other day, but you actually did something about it 😅

You are on fire these days, I'll try better to keep up on the reviewing side of things. 🤞 Work has been busier than usual lately.

PR #1233, opened 2023-03-28
Determine BaseUnits of Prefix-Unit

bdbonazz:
I found a way to determine come BaseUnits of Units added dinamically by prefix.
The main change is to file "CodeGen\Generators\QuantityJsonFilesParser.cs"
New to Open-Source, Open to criticism.

### DIFF
--- a/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
+++ b/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
@@ -213,29 +213,20 @@ namespace UnitsNet.Tests
         [Fact]
         public void GetDefaultAbbreviationFallsBackToUsEnglishCulture()
         {
-            var oldCurrentCulture = CultureInfo.CurrentCulture;
+            // CurrentCulture affects number formatting, such as comma or dot as decimal separator.
+            // CurrentCulture also affects localization of unit abbreviations.
+            // Zulu (South Africa)
+            var zuluCulture = CultureInfo.GetCultureInfo("zu-ZA");
+            // CultureInfo.CurrentCulture = zuluCulture;
 
-            try
-            {
-                // CurrentCulture affects number formatting, such as comma or dot as decimal separator.
-                // CurrentCulture affects localization, in this case the abbreviation.
-                // Zulu (South Africa)
-                var zuluCulture = CultureInfo.GetCultureInfo("zu-ZA");
-                CultureInfo.CurrentCulture = zuluCulture;
+            var abbreviationsCache = new UnitAbbreviationsCache();
+            abbreviationsCache.MapUnitToAbbreviation(CustomUnit.Unit1, AmericanCulture, "US english abbreviation for Unit1");
 
-                var abbreviationsCache = new UnitAbbreviationsCache();
-                abbreviationsCache.MapUnitToAbbreviation(CustomUnit.Unit1, AmericanCulture, "US english abbreviation for Unit1");
+            // Act
+            string abbreviation = abbreviationsCache.GetDefaultAbbreviation(CustomUnit.Unit1, zuluCulture);
 
-                // Act
-                string abbreviation = abbreviationsCache.GetDefaultAbbreviation(CustomUnit.Unit1, zuluCulture);
-
-                // Assert
-                Assert.Equal("US english abbreviation for Unit1", abbreviation);
-            }
-            finally
-            {
-                CultureInfo.CurrentCulture = oldCurrentCulture;
-            }
+            // Assert
+            Assert.Equal("US english abbreviation for Unit1", abbreviation);
         }
 
         [Fact]
--- a/UnitsNet/CustomCode/UnitAbbreviationsCache.cs
+++ b/UnitsNet/CustomCode/UnitAbbreviationsCache.cs
@@ -25,7 +25,7 @@ namespace UnitsNet
         ///     culture, but no translation is defined, so we return the US English definition as a last resort. If it's not
         ///     defined there either, an exception is thrown.
         /// </example>
-        internal static readonly CultureInfo FallbackCulture = CultureInfo.GetCultureInfo("en-US");
+        internal static readonly CultureInfo FallbackCulture = CultureInfo.InvariantCulture;
 
         /// <summary>
         ///     The static instance used internally for ToString() and Parse() of quantities and units.
--- a/UnitsNet/InternalHelpers/CultureHelper.cs
+++ b/UnitsNet/InternalHelpers/CultureHelper.cs
@@ -1,0 +1,45 @@
+// Licensed under MIT No Attribution, see LICENSE file at the root.
+// Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
+
+using System;
+using System.Collections.Concurrent;
+using System.Globalization;
+
+namespace UnitsNet.InternalHelpers;
+
+/// <summary>
+///     Helper class for <see cref="CultureInfo"/> and related operations.
+/// </summary>
+internal static class CultureHelper
+{
+    private static readonly ConcurrentDictionary<string, CultureInfo> CultureCache = new();
+
+    /// <summary>
+    ///     Attempts to get the culture by name, with fallback to invariant culture if not found.<br/>
+    ///     <br/>
+    ///     This is particularly useful for Linux and Raspberry PI environments, where cultures may not always be installed.
+    ///     To simulate the behavior, set environment variable DOTNET_SYSTEM_GLOBALIZATION_INVARIANT='1' when running the application.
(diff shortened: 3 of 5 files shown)


CodeGen/Generators/QuantityJsonFilesParser.cs
```diff
@@ -109,6 +111,72 @@ private static void AddPrefixUnits(Quantity quantity)
...
+            else
+                return null;
+        }
+
+        public static string[] ToWords(string str)
+        {
+            return Regex.Replace(str, "[a-z][A-Z]", m => $"{m.Value[0]} {m.Value[1]}")
+                .Split(' ');
+        }
+
```
angularsen, line 179: It is an interesting approach, but it only solves 82 of 216 prefix units.
I pushed a commit with some debug console output.

We might be able improve the name matching, but there is risk of matching incorrectly and I don't think we can get to 100% coverage with this approach for units like Farad and StandardGravity anyway.


```
Assigned base units to 134 prefix units, failed to assign for 82 prefix units:
---
StandardGravity           L=Meter T=Second                            Prefixes: [Milli]
Farad                     I=Ampere L=Meter M=Kilogram T=Second        Prefixes: [Pico, Nano, Micro, Milli, Kilo, Mega]
Volt                      I=Ampere L=Meter M=Kilogram T=Second        Prefixes: [Nano, Micro, Milli, Kilo, Mega]
VoltPerSecond             I=Ampere L=Meter M=Kilogram T=Second        Prefixes: [Micro, Milli, Kilo, Mega]
VoltPerMicrosecond        I=Ampere L=Meter M=Kilogram T=Microsecond   Prefixes: [Micro, Milli, Kilo, Mega]
VoltPerMinute             I=Ampere L=Meter M=Kilogram T=Minute        Prefixes: [Micro, Milli, Kilo, Mega]
VoltPerHour               I=Ampere L=Meter M=Kilogram T=Hour          Prefixes: [Micro, Milli, Kilo, Mega]
Joule                     L=Meter M=Kilogram T=Second                 Prefixes: [Milli, Kilo, Mega, Giga, Tera, Peta]
JoulePerCubicMeter        L=Meter M=Kilogram T=Second                 Prefixes: [Kilo, Mega, Giga, Tera, Peta]
Newton                    L=Meter M=Kilogram T=Second                 Prefixes: [Micro, Milli, Deca, Kilo, Mega]
WattPerSquareMeter        M=Kilogram T=Second                         Prefixes: [Nano, Micro, Milli, Centi, Deci, Kilo]
StandardGravitiesPerSecond L=Meter T=Second                           Prefixes: [Milli]
Watt                      L=Meter M=Kilogram T=Second                 Prefixes: [Femto, Pico, Nano, Micro, Milli, Deci, Deca, Kilo, Mega, Giga, Tera, Peta]
WattPerCubicMeter         L=Meter M=Kilogram T=Second                 Prefixes: [Pico, Nano, Micro, Milli, Deci, Deca, Kilo, Mega, Giga, Tera]
Pascal                    L=Meter M=Kilogram T=Second                 Prefixes: [Micro, Milli, Deca, Hecto, Kilo, Mega, Giga]
CubicMeter                L=Meter                                     Prefixes: [Hecto, Kilo]
---
```


I guess it is better than nothing though? We could go with this.

One obvious solution is to extend the JSON to explicitly map base units for each individual prefix unit in the JSON, but I think it will be messy and more cumbersome to add new units for contributors. I don't see enough utility value or upside to justify this effort.

angularsen: I couldn't push changes to your master branch, so I created a pull request to update your pull request: https://github.com/bdbonazz/UnitsNet/pull/1

angularsen (changes requested):
Hi, this looks interesting. Some comments below.

CodeGen/Generators/QuantityJsonFilesParser.cs
angularsen: @tmilnthorp what do you think of this approach? Do you see any better ways to achieve this?

bdbonazz: > I don't think we can get to 100% coverage

I don't think that too, but I cannot see at the moment any case of possible incorrect match

> One obvious solution is to extend the JSON to explicitly map base units for each individual prefix unit in the JSON, but I think it will be messy and more cumbersome to add new units for contributors. I don't see enough utility value or upside to justify this effort.

I totally agree with that. I'm either non totally sure this is possible anyway
Ex. MicroPascal could etiher be "g/m*s^2" or "kg/km*s^2"

angularsen: @tmilnthorp Any input on this? I'm inclined to merge it, it seems like a solid improvement although not 100%. Can't spot any mistakes in the automatically determined base units.

angularsen: @tmilnthorp Just a nudge on your opinion here before I merge.

tmilnthorp: > @tmilnthorp Just a nudge on your opinion here before I merge.

Sorry. Work and personal life got busy. Let me take some time to look in the next day or two. If you don’t hear by then go for it.

tmilnthorp:
So just thinking about this - wouldn't the base units just be the same as the ones without the prefixes? Per your example:
> MicroPascal could etiher be "g/ms^2" or "kg/kms^2"

CodeGen/Generators/QuantityJsonFilesParser.cs
bdbonazz: @tmilnthorp 
I hope I will be clear enough, English is not my first language
I don't know why I cannot reply to your other comment
"So just thinking about this - wouldn't the base units just be the same as the ones without the prefixes? Per your example:"
No, this is exactly what this pull request is about
I'm taking MetersPerSecondSquared [L=Meter T=Second] and saying that the prefix Nano "NanoMetersPerSecondSquared" changes that to [L=NanoMeter T=Second]
So I was saying I don't know how (or even if it's possible) to uniquely link MicroPascal to [M=Kilogram L=KiloMeter T=Second] or [M=Gram L=Meter T=Second]
And Pascal is even an easy example, Farad could be even more impossible in my opinion
Farad in SI is (s^4*A^2)/(m^2*kg). The only "correct" way to write MilliFarad is (s^4*A^2)/(m^2*g) but more often that not this is wrong. Often (s^4*A^2)/(m^2*kg) are the correct units and people use MilliFarad cause the results are small

angularsen:
> So just thinking about this - wouldn't the base units just be the same as the ones without the prefixes? Per your example:

I have never used base units in UnitsNet for anything myself, but my understanding of the use case for defining base units is that it makes it easier for an application developer to let the user choose what SI base units to use, then all quantities will automatically be shown in compatible units where possible.

For example, if user selects Centimeter as the SI base unit for Length, then all quantities would be represented using centimeters, such as Area cm², Volume cm³ and so on. 

It may not be able to find any unit for quantities not based on length, such as Angle.
It may not be able to find a unique unit for all quantities were multiple units share the same SI base units.

Since I haven't used this in any practical application myself, it all is a bit theoretical for me at this point and not so easy to reason about. 

Personally, if I were to offer unit preferences in an application, I would probably create a custom mapping between SI base units and units in the quantities I care about.

However, since we do have base units and maybe it does add some value, then I still think this PR seems worthwhile to merge.

angularsen (approved):

angularsen:
@bdbonazz Sorry this took a long time. 
I'd like to move ahead with merging this, but I noticed a few additional changes that I don't think should be included here.

### Please revert/undo these changes

AmountOfSubstance.json
LuminousIntensity.json
Mass.json

They add new units, and this should be a separate pull request from determining base units of prefix units.

### Tip for next time
Create a new branch on your fork instead of committing on your `master` branch, this way you can more easily have multiple features on different branches, and maintainers trying to merge are allowed to make changes to the branch that can help speed up the review process by fixing minor things immediately.

angularsen:
@bdbonazz Seems you closed this without creating a new pull request. Do you still intend to complete this?

bdbonazz:
> @bdbonazz Seems you closed this without creating a new pull request. Do you still intend to complete this?

I am sorry
I changed jobs for a month so I don't have much time left for github
I tried to do what you told me, but in the branch where i reverted the changes and only added some Units, some tests stopped working and I didn't have time to figure out why
I didn't close this PR on purpose, I must have done it by mistake when I reverted the changes

angularsen:
@bdbonazz I fully understand time is hard to come by, no pressure from my end.

The tests that stopped working was probably a broken change in `master` branch that I recently fixed in #1267 .

If you could update your feature branch to take in the latest changes, I believe you should be good to go. If you push your branch and create a pull request for it, then you have some UI buttons to help update the branch by either rebase or merge.

I'm happy to help if you need assistance.

PR #1235, opened 2023-04-01
codegen: 🐛 Fix consistent ordering of units

angularsen:
The ordering seemed to vary on computers resulting in annoying diffs for PRs.
Fixed by specifying the string comparer, assuming the default was affected by the system's culture.

Verified fix in a recent PR where running codegen locally resulted in a diff. With this fix, there was no diff.


angularsen:
FYI @tmilnthorp , I believe you mentioned this

PR #1254, opened 2023-06-02
Add ImperialQuart to Volume

Eternali:


UnitsNet.Tests/CustomCode/VolumeTests.cs
```diff
@@ -106,6 +106,11 @@ public class VolumeTests : VolumeTestsBase
 
         protected override double DecalitersInOneCubicMeter => 1e2;
 
+        /// <summary>
+        /// https://www.legislation.gov.uk/uksi/1995/1804/made
+        /// </summary>
+        protected override double ImperialQuartsInOneCubicMeter => 879.876993196;
```
angularsen, line 112: Verified.

Weird, wolfram normally is very precise, but seems to round here.
https://www.wolframalpha.com/input?i=1+cubic+meter+in+imperial+quarts

Common/UnitDefinitions/Volume.json
```diff
@@ -461,6 +461,18 @@
         }
       ]
     },
+    {
+      "SingularName": "ImperialQuart",
+      "PluralName": "ImperialQuarts",
+      "FromUnitToBaseFunc": "{x} * 1.1365225e-3",
+      "FromBaseToUnitFunc": "{x} / 1.1365225e-3",
```
angularsen, line 468: Verified 
> The imperial quart, which is used for both liquid and dry capacity, is equal to one quarter of an imperial gallon, or exactly 1.1365225 liters. 

https://en.wikipedia.org/wiki/Quart

angularsen (approved):

angularsen:
Looks perfect, thanks. 
We could also add US quarts, if you are up for it? Merging this as-is anyway.

Eternali:
Thank you for the quick response!
Looks like U.S. quarts is already part of the Volume class which is why I only needed to add the Imperial quart.

angularsen:
Lol, it was literally right next to it and I still missed it 😅

Nuget should be out shortly.
[Release UnitsNet/5.17.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.17.0)

Eternali:
No problem lol, glad I wasn't missing anything, thanks for the release!

PR #1257, opened 2023-06-10
Fixed tests without an explicit culture specification

Jagailo:
I fixed all the tests that didn't pass due to a non-English operating system.
I didn't change the test logic.

I didn't make any changes to the core code of the library, so this PR doesn't need a separate version.

Also, I can't guarantee that I've fixed all cases of missing culture.
For example, the test Assert.False("5 m" != "6 м") could return a false result not because 5 != 6, but because _meter_ is not a _метр_.

### DIFF
--- a/UnitsNet.Tests/CustomCode/StonePoundsTests.cs
+++ b/UnitsNet.Tests/CustomCode/StonePoundsTests.cs
@@ -31,13 +31,12 @@ namespace UnitsNet.Tests
         }
 
         [Fact]
-        public void StonePoundsToString_FormatsNumberInDefaultCulture()
+        public void StonePoundsToString_FormatsNumberInCurrentCulture()
         {
-            Mass m = Mass.FromStonePounds(3500, 1);
-            StonePounds stonePounds = m.StonePounds;
-            string numberInCurrentCulture = 3500.ToString("n0", CultureInfo.InvariantCulture);
+            CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo("en-US");
+            StonePounds stonePounds = Mass.FromStonePounds(3500, 1).StonePounds;
 
-            Assert.Equal($"{numberInCurrentCulture} st 1 lb", stonePounds.ToString(CultureInfo.InvariantCulture));
+            Assert.Equal("3,500 st 1 lb", stonePounds.ToString());
         }
 
         // These cultures use a thin space in digit grouping
--- a/UnitsNet.Tests/CustomCode/StonePoundsTests.cs
+++ b/UnitsNet.Tests/CustomCode/StonePoundsTests.cs
@@ -35,9 +35,9 @@ namespace UnitsNet.Tests
         {
             Mass m = Mass.FromStonePounds(3500, 1);
             StonePounds stonePounds = m.StonePounds;
-            string numberInCurrentCulture =  3500.ToString("n0", CultureInfo.CurrentCulture); // Varies between machines, can't hard code it
+            string numberInCurrentCulture = 3500.ToString("n0", CultureInfo.InvariantCulture);
 
-            Assert.Equal($"{numberInCurrentCulture} st 1 lb", stonePounds.ToString());
+            Assert.Equal($"{numberInCurrentCulture} st 1 lb", stonePounds.ToString(CultureInfo.InvariantCulture));
         }
 
         // These cultures use a thin space in digit grouping
--- a/UnitsNet.Tests/QuantityTest.cs
+++ b/UnitsNet.Tests/QuantityTest.cs
@@ -135,30 +135,30 @@ namespace UnitsNet.Tests
         [Fact]
         public void TryParse_GivenInvalidQuantityType_ReturnsFalseAndNullQuantity()
         {
-            Assert.False(Quantity.TryParse(typeof(DummyIQuantity), "3.0 cm", out IQuantity? parsedLength));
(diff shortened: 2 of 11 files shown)


UnitsNet.Tests/CustomCode/StonePoundsTests.cs
```diff
@@ -35,9 +35,9 @@ public void StonePoundsToString_FormatsNumberInDefaultCulture()
         {
             Mass m = Mass.FromStonePounds(3500, 1);
             StonePounds stonePounds = m.StonePounds;
-            string numberInCurrentCulture =  3500.ToString("n0", CultureInfo.CurrentCulture); // Varies between machines, can't hard code it
+            string numberInCurrentCulture = 3500.ToString("n0", CultureInfo.InvariantCulture);
```
angularsen, line 38: The test name indicates it tests the default culture, meaning ToString() without explicitly passing a culture.

For these kinds of tests, we could just set `CultureInfo.CurrentCulture` instead. 
As far as I know, manipulating `CurrentCulture` in one test case does not affect other test cases due to how the test runner works. I might be wrong.

I'll comment similar instances in other places with `TODO Set CurrentCulture instead of passing culture argument to ToString()`.

UnitsNet.Tests/QuantityTest.cs
```diff
@@ -135,30 +135,30 @@ public void TryFrom_GivenValueAndUnit_ReturnsQuantity()
...
-            Assert.False(Quantity.TryParse(typeof(Length), "x cm", out IQuantity? parsedLength));
+            Assert.False(Quantity.TryParse(CultureInfo.InvariantCulture, typeof(Length), "x cm", out IQuantity? parsedLength));
             Assert.Null(parsedLength);
 
-            Assert.False(Quantity.TryParse(typeof(Mass), "xt", out IQuantity? parsedMass));
+            Assert.False(Quantity.TryParse(CultureInfo.InvariantCulture, typeof(Mass), "xt", out IQuantity? parsedMass));
             Assert.Null(parsedMass);
 
-            Assert.False(Quantity.TryParse(typeof(Pressure), "foo", out IQuantity? parsedPressure));
+            Assert.False(Quantity.TryParse(CultureInfo.InvariantCulture, typeof(Pressure), "foo", out IQuantity? parsedPressure));
```
angularsen, line 151: You can use static imports to shorten this to just `InvariantCulture`

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-directive#static-modifier

UnitsNet.Tests/QuantityTypeConverterTest.cs
```diff
@@ -245,7 +245,7 @@ public void ConvertTo_GivenSomeQuantityAndContextWithDisplayAsUnitAttributes_Ret
             string convertedQuantityDefaultCulture = (string)converter.ConvertTo(length, typeof(string))!;
             string convertedQuantitySpecificCulture = (string)converter.ConvertTo(context, Culture, length, typeof(string))!;
 
-            Assert.Equal("1 m", convertedQuantityDefaultCulture);
+            Assert.Equal(Length.FromMeters(1).ToString(), convertedQuantityDefaultCulture);
             Assert.Equal("10 dm", convertedQuantitySpecificCulture);
```
angularsen, line 249: TODO Set CurrentCulture instead of passing culture argument to ToString()

angularsen (changes requested):
Very nice, just a few suggestions for you to consider.

UnitsNet.Tests/CustomCode/StonePoundsTests.cs
Jagailo: Disagree with the remark, except that the test title is incorrect.

If this test is checking for the Default culture, which is the Current one, then why does Assert.Equal use the English language?

The result will always be false for those cultures that have unit localizations.

angularsen: This is what I propose:

```cs
        [Fact]
        public void StonePoundsToString_FormatsNumberInCurrentCulture()
        {
            CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo("en-US");
            StonePounds stonePounds = Mass.FromStonePounds(3500, 1).StonePounds;

            Assert.Equal("3,500 st 1 lb", stonePounds.ToString());
        }
```

Rename it to `InCurrentCulture`, set the current culture at the top and simply call `ToString()` without any culture argument. Other tests in this test class test the behavior of passing an explicit culture, so it makes sense to let this test the implicit culture.

Does that make sense?

Jagailo: I understand the difference, but the check will be the same.
Either we are explicitly specifying InvariantCulture, either we are explicitly specifying the English culture.
Anyway, I did as you said.

angularsen:
Thanks! 

I see your point that providing English in a test case, either via CurrentCulture or as method argument, does not help distinguish from the "default" culture it would fall back to if the machine is running say Norwegian culture that has no unit abbreviation translations. Only the number formatting would differ, which does get tested here, but it is not as obvious as seeing the unit abbreviations in the correct language.

Russian is more or less the only other language with translations, so it would be an improvement to use that instead. This probably goes for many of our tests, so I'd make that a separate PR since this was primarily about fixing tests on machines with. If you are interested in taking a look at that, let me know.

Issue #1256, opened 2023-06-08
Swagger schema reports all unit properties and not UnitValue ones


danielecapursi:
**Describe the bug**
When using **UnitsNetIQuantityJsonConverter** (and hence **Newtonsoft.JSon** serializer), **Swagger** generates the schema of all units (e.g. **Length**) as they are as C# classes, reporting **all their properties** (value, unit, feetInches, quantityInfo, dimensions and all related subobjects).
Bug **UnitsNetIQuantityJsonConverter** convert values such that they are (de)serialized as **ValueUnit**, i.e. having **value and unit properties only**.

**To Reproduce**
Use this solution: [WebTest.zip](https://github.com/angularsen/UnitsNet/files/11688994/WebTest.zip)

It exposes a single method as a Rest API (/Api), returning a MyDto object, having Location = 12.345 Km.
**Try calling /Api.** 
The result will correctly be:
```
{
  "location": {
    "unit": "LengthUnit.Kilometer",
    "value": 12.345
  }
}
```
But MyDto will have Example Values:
```
{
  "location": {
    "feetInches": {
      "feet": 0,
      "inches": 0
    },
    "value": 0,
    "unit": 1,
    "quantityInfo": {
      "name": "string",
      "unitInfos": [
        {
          "value": 1,
          "pluralName": "string",
          "baseUnits": {
            "length": 1,
            "mass": 1,
            "time": 1,
            "current": 1,
            "temperature": 1,
...
}
```
I truncated it a lot, since it is veeeeeery long (more than 360 lines).
**Please note that this is not coherent with the JSON actually returned by the API. This is the bug!**

And MyDto will have schema:
```
MyDto{
location
	Length{
		feetInches	FeetInches{...}
		value	[...]
		unit	LengthUnit[...]
		quantityInfo	LengthUnitQuantityInfo{...}
		dimensions	BaseDimensions{...}
		}
}
```
**Expected behavior**
Example Values should be:
```
{
  "location": {
    "value": 0,
    "unit": "string"
  }
}
```
Schema should be:
```
MyDto{
	location	
		Length
		{
			value	number($double)
			unit	string
		}
}
```
**Additional context**
The problem is very easy to spot and seems easy to fix: we are using a converter, but Swagger has no idea of it. We just need to inform Swagger about the substitution of all IQuantityValue with ValueUnit (ExtendedValueUnit, in some cases).

danielecapursi:
[...]

PR #1258, opened 2023-06-10
✨ Add Quantity.From/TryFrom quantity name and unit name

angularsen:
Make it easier to construct quantities from strings describing the quantity name and unit name, if the unit enum value is not readily available.

This is useful for custom serialization of quantities, such as mapping to simple custom quantity DTO types instead of relying on JSON serialization converters.

### Changes
- Add `Quantity.From` and `TryFrom`
- Refactor `UnitConverter` to reuse code for constructing quantity from strings
- Add tests for `From` and `TryFrom`
- Add test `CustomSerializationTests.CanMapToJsonAndBackViaCustomDto` to showcase serialization

### Example
```cs
public record QuantityDto(double Value, string QuantityName, string UnitName); 

// The original quantity.
IQuantity q = Length.FromCentimeters(5);

// Map to DTO.
QuantityDto dto = new(
    Value: (double)q.Value,
    QuantityName: q.QuantityInfo.Name,
    UnitName: q.Unit.ToString());

/* Serialize to JSON:
{
    "Value": 5,
    "QuantityName": "Length",
    "UnitName": "Centimeter"
}
*/
var json = System.Text.Json.JsonSerializer.Serialize(dto);

// Deserialize from JSON.
QuantityDto deserialized = System.Text.Json.JsonSerializer.Deserialize<QuantityDto>(json)!;

// Map to IQuantity.
Quantity.TryFrom(deserialized.Value, deserialized.QuantityName, deserialized.UnitName, out IQuantity? deserializedQuantity);

```


angularsen:
@tmilnthorp I'm merging this, but I'd like your eyes on this when you have the chance.

It was inspired by https://github.com/angularsen/UnitsNet/issues/1256. 

We have already started migrating from enum-based to string-based methods in the codebase, since this works better with custom quantities, and this change follows that thinking.

Offering custom serializers is still convenient for many use cases, but in APIs and other areas where a breaking change would be disastrous, I would never trust a third party library to do the serialization. In my mind, custom mapping is the way to go here and this PR helps facilitate that. 
I also updated the wiki with an example on how to do this, and took the liberty of declaring it the "recommended" approach. Would love your perspective on this.

PR #1267, opened 2023-06-18
✨Add UnitsNetSetup to hold global static state

angularsen:
Added `UnitsNetSetup` to gather global state in a single place as a singleton, with the possibility of passing instances of it to static methods later to override the defaults.

This way, state is no longer scattered among multiple classes that depend on each other in non-obvious ways and where initialization order has caused issues up until now.

### Changes
- Add `UnitsNetSetup` to hold global state as a singleton property `Default` that controls default state initialization.
  - Add properties for all "services" that depend on global state: `UnitConverter, UnitAbbreviationsCache, UnitParser, QuantityParser`
- Forward all other `Default` singleton properties from these services to `UnitsNetSetup.Default` and mark obsolete
- Add some TODOs where it seems functionality is missing

### Testing
❌ Still running into a handful of flaky tests due to racing conditions. 
This was a regression in #1210, but still not fixed.
Will investigate and fix in separate PR.

```
UnitAbbreviationsCacheTests.AllUnitsImplementToStringForInvariantCulture
UnitAbbreviationsCacheTests.MapUnitToAbbreviation_AddCustomUnit_DoesNotOverrideDefaultAbbreviationForAlreadyMappedUnits
```

### DIFF
--- a/UnitsNet.Serialization.JsonNet/AbbreviatedUnitsConverter.cs
+++ b/UnitsNet.Serialization.JsonNet/AbbreviatedUnitsConverter.cs
@@ -45,7 +45,7 @@ namespace UnitsNet.Serialization.JsonNet
         /// </summary>
         /// <param name="comparer">The comparer used to compare the property/quantity names (e.g. StringComparer.OrdinalIgnoreCase) </param>
         public AbbreviatedUnitsConverter(IEqualityComparer<string?> comparer)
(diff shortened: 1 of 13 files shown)


angularsen:
[Release UnitsNet/5.22.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.22.0)

Issue #1260, opened 2023-06-13
ReciprocalVolume?


chucker:
**Is your feature request related to a problem? Please describe.**
I have a particulate matter sensor that gives measured values in µg/m^3 (so I can use `Density`), but then also measured _counts_ in #/m^3.

Wikipedia seems to call this https://en.wikipedia.org/wiki/Number_density.

I've also found the `ReciprocalArea` type, which is pretty close, except I need `ReciprocalVolume`.

**Describe the solution you'd like**
`NumberDensity` and/or `ReciprocalVolume` types would be nice, with factory methods such as `FromInverseCubicMeter`.

chucker:
(`NumberConcentration` might also be the correct name here. I'm not sure.)

ricardoboss:
See also: https://en.wikipedia.org/wiki/Concentration#Number_concentration

angularsen:
This feels a tiny bit fringe/exotic, but I'll accept it. 

From googling a bit, it seems number density is maybe the more commonly used term.
I have no hard preference, but reciprocal volume seems the least common of them.

I'd suggest NumberDensity, alternatively NumberConcentration.
If you want to go ahead with a pull request, steps are outlined here: https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit

I'm happy to assist.

chucker:
@ricardoboss made a PR (#1409); reopen

PR #1280, opened 2023-07-10
Add Cell Density for cell counting

eugene-zhary:
Numerous procedures in biology and medicine require the counting of cells
The most commonly used units of measure for this purpose are cells/cm² and cells/mL


angularsen:
Sorry, this is too domain specific.
There are extremely many variants of X units per volume, per area or per length. We can't add them all I'm afraid.

If it helps, there is ReciprocalArea, ReciprocalLength.

An issue describes adding `ReciprocalVolume` or `NumberDensity` for volume, would that help for you?
https://github.com/angularsen/UnitsNet/issues/1260

eugene-zhary:
no problem 🥲

eugene-zhary:
anyway, thank you for your suggestion!

angularsen:
Thanks for taking it like a champ 🙌 

If you are looking to add more units later, it can be helpful to create an issue first and make sure it fits the library before you put in too much effort.

PR #1281, opened 2023-07-10
🛡️Restrict number extension methods to INumber<T>

angularsen:
Fixes #1269 

Quantities implement `IConvertible.ToDouble()` by converting the `Value` property to `double`, and similar for other number types.

Number extension methods expects a number argument, but take any argument type and call `Convert.ToDouble()` on it. This means you can run into subtle bugs:

```cs
Length.FromCentimeters(100).Millimeters(); // Length of 100 mm
Length.FromCentimeters(100).Millimeters; // Value of 1000
```

The quantity implements conversions like this:
```cs
        double IConvertible.ToDouble(IFormatProvider? provider)
        {
            return Convert.ToDouble(_value);
        }
```

The extension method looks like this:
```cs
public static Length Millimeters<T>(this T value)
{
	return Length.FromMillimeters(Convert.ToDouble(value));
}
```

So `Convert.ToDouble()` happily converts `Length.FromCentimeters(100)` to `100`.

### Changes
- Restrict number extension methods to `INumber<T>` for net7.0 and later


angularsen:
[Release UnitsNet/5.25.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.25.0)

PR #1287, opened 2023-07-13
💥Add CoefficientOfThermalExpansion units, change base unit

### LINKED ISSUE #1259: Missing CoefficientOfThermalExpansion operators and units.
A typical use case for a coefficient of thermal expansion is to calculate a change in length due to a change in temperature.
`dL = (cte * dT) * L;`

`dL` is a change in length.
`cte` is a coefficient of thermal expansion.
`dT` is a temperature delta, the change in temperature from some reference temperature.
`L` is a length at the reference temperature.

It would be useful if there were operators of the form:
`public static double operator * (CoefficientOfThermalExpansion cte, TemperatureDelta dt)`
And:

`public static double operator * (TemperatureDelta dt, CoefficientOfThermalExpansion cte)`

tm-ren:
Fixes #1259 

### Add units
PerKelvin
PerDegreeCelsius
PerDegreeFahrenheit
PpmPerKelvin
PpmPerDegreeCelsius
PpmPerDegreeFahrenheit

### 💥Change base unit
From `InverseKelvin` to `PerKelvin`.

### Obsolete units (renamed)
InverseKelvin
InverseDegreeCelsius
InverseDegreeFahrenheit

### DIFF
--- a/Common/UnitDefinitions/CoefficientOfThermalExpansion.json
+++ b/Common/UnitDefinitions/CoefficientOfThermalExpansion.json
@@ -18,7 +18,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": []
+          "Abbreviations": [ "1/K" ]
         }
       ]
     },
@@ -34,7 +34,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": []
+          "Abbreviations": [ "1/°C" ]
         }
       ]
     },
@@ -50,7 +50,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": []
+          "Abbreviations": [ "1/°F" ]
         }
       ]
     },
@@ -65,7 +65,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "K⁻¹", "1/K" ]
+          "Abbreviations": [ "K⁻¹" ]
         }
       ]
     },
@@ -80,7 +80,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "°C⁻¹", "1/°C" ]
+          "Abbreviations": [ "°C⁻¹" ]
         }
       ]
     },
@@ -95,7 +95,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "°F⁻¹", "1/°F" ]
+          "Abbreviations": [ "°F⁻¹" ]
         }
       ]
     },
--- a/Common/UnitDefinitions/CoefficientOfThermalExpansion.json
+++ b/Common/UnitDefinitions/CoefficientOfThermalExpansion.json
@@ -1,6 +1,6 @@
 {
   "Name": "CoefficientOfThermalExpansion",
-  "BaseUnit": "InverseKelvin",
+  "BaseUnit": "PerKelvin",
   "XmlDocSummary": "A unit that represents a fractional change in size in response to a change in temperature.",
   "BaseDimensions": {
     "Θ": -1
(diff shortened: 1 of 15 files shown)


Common/UnitDefinitions/CoefficientOfThermalExpansion.json
```diff
@@ -50,6 +50,51 @@
           "Abbreviations": [ "°F⁻¹", "1/°F" ]
         }
       ]
+    },
+    {
+      "SingularName": "PartsPerMillionPerKelvin",
```
angularsen, line 55: `PartPer` for singular. Other places too.

UnitsNet.Tests/CustomCode/CoefficientOfThermalExpansionTests.cs
```diff
@@ -35,6 +35,12 @@ public class CoefficientOfThermalExpansionTests : CoefficientOfThermalExpansionT
 
         protected override double InverseKelvinInOneInverseKelvin => 1.0;
 
+        protected override double PartsPerMillionPerDegreeCelsiusInOneInverseKelvin => 1e6;
```
angularsen, line 38: I don't know what naming conventions are more widely used in this domain, but it seems to me from googling that "inverse" is maybe non-standard? 

Instead, I see PerKelvin, PerDegreeCelsius and PartsPerMillionDegreeCelsius in places like this source:

https://www.sensorsone.com/temperature-coefficient-converter/
image

Should we rename the Inverse units or are they fine as-is?
It would require adding new units and obsoleting the old ones, until the next major version bump.

angularsen: It also seems like we could maybe abbreviate to `PpmPerDegreeCelsius` to keep it a bit shorter, while still fairly clear what it means. Thoughts?

angularsen (changes requested):

UnitsNet.Tests/CustomCode/CoefficientOfThermalExpansionTests.cs
tm-ren: Yes agree on a ppm prefix, do you prefer that capitalisation rather than say `PPMPerDegreeCelsius`?

Also prefer the `Per...` over `Inverse...`, as far as I can tell, changing this would involve:
- Adding the three new `Per...` units, with same definitions as existing units. Adding unit tests definitions as required etc.
- Adding "ObsoleteText: " to the `Inverse...` unit definitions in json.
- Change the base units from `InverseKelvin` to `PerKelvin`, or should this be left until the next major change?

angularsen: `Ppm` would follow .NET naming conventions for abbreviations in my experiencee.

Yes, your approach looks exactly right.

I'm trying to think of scenarios for when changing base units breaks anything, but I can't really think of any. It is primarily used in conversion functions to convert via, but since they are equal that requires no change. 

All constructors and factory methods specify a unit.
Serialization specifies the unit.

One minor change would be that `Zero` property would have a different unit, but again, I don't think this will be a breaking change for most.

I say let's just change it 👍

Common/UnitDefinitions/CoefficientOfThermalExpansion.json
```diff
@@ -32,13 +34,59 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "°C⁻¹", "1/°C" ]
+          "Abbreviations": []
```
angularsen, line 37: We can't remove the existing abbreviations. 

Did you run into trouble with tests failing on ambiguous units when parsing these abbreviations?

I guess an improvement to avoid ambiguity when renaming units would be to add support for specifying in JSON that this unit cannot be parsed, so the abbreviations are only used for ToString().

Maybe something like `"SkipParse": true` or declaring a separate `"ParseAbbreviations": []` to override abbreviations for parse. The latter one is more flexible.

What do you think?

angularsen (changes requested):

Common/UnitDefinitions/CoefficientOfThermalExpansion.json
angularsen: I believe what we have done in the past is to keep one of the least used abbreviations in the old unit, and all other abbreviations in the new unit. In this case we could keep `1/°C` here, and move  `°C⁻¹` to the new unit, but it's not a very good solution.

tm-ren: In all honesty I haven't really paid much attention to the abbreviation functionality in what I have been using the library for. Your workaround seems reasonable to me for now, but I guess your proposed solution seems preferable.

The reason you would want to skip it during parsing is because the unit is obsolete, so perhaps an `"IsObsolete" : true`?

For now shall I implement the workaround pending a better solution via a different issue?

angularsen: I'm fine with the workaround

angularsen:

[Release UnitsNet/5.27.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.27.0)

PR #1289, opened 2023-07-15
Allow NaN, Infinity values for quantities

ebfortin:
Add possibility to have double.NaN, double.NegativeInifinity and double.PositiveInifnity special values.


angularsen:
Changed this to target v6 branch.

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
```diff
@@ -118,25 +118,6 @@ public void DefaultCtor_ReturnsQuantityWithZeroValueAndBaseUnit()
             Assert.Equal(AccelerationUnit.MeterPerSecondSquared, quantity.Unit);
         }
 
-        [Fact]
-        public void Ctor_WithInfinityValue_ThrowsArgumentException()
```
angularsen, line 122: Shouldn't we change the tests to reflect that Nan/Inf are allowed, instead of removing the tests?

angularsen (changes requested):
Moved this to v6 branch, primarily to give it more time to mature and have some time for people to discuss the change before we make it official.

Would also like to keep the tests, but change them to test that nan/inf is allowed.

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
ebfortin: Make sense. Will do that.

ebfortin: Finally, finally! made the changes.

angularsen:
Lol, better late than never :D Nice!

PR #1293, opened 2023-07-21
json: 🐛Fix stack overflow deserializing empty JSON

### LINKED ISSUE #1292: Deserialization of an empty unit cause a stackoverflow
Deserialization of an empty unit cause a stackoverflow.

```
            var jsonSerializerSettings = new JsonSerializerSettings()
            {
                Converters = new List<JsonConverter>() { new UnitsNetIQuantityJsonConverter(), new UnitsNetIComparableJsonConverter() },
            };

            JsonConvert.DeserializeObject<Length>("{}", jsonSerializerSettings);
```

**To Reproduce**
Steps to reproduce the behavior (just an example):
1. Add nuget UnitsNet 5.26.0 and UnitsNet.Serialization.JsonNet 5.5.0 to a dotnet 7.0 project.
2. Build using Visual Studio 2022 
3. Run the code above

**Expected behavior**
No stackoverflow

shaeussler:
Fixes #1292 

The call to "token.ToObject<IQuantity>(serializer)" with an empty or invalid json calls `JsonConvert.DeserializeObject` again and results in infinite recursion and stack overflow.

### Changes
- Throw `JsonSerializationException` on invalid JSON in `UnitsNetIQuantityJsonConverter`
	- Include helpful info in exception, link to wiki, truncated JSON token in message and full JSON token in `Data["JsonToken"]` property.
- Add tests for new behavior

### DIFF
--- a/UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonDeserializationTests.cs
+++ b/UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonDeserializationTests.cs
@@ -99,10 +99,9 @@ namespace UnitsNet.Serialization.JsonNet.Tests
         }
 
         [Fact]
-        public void EmptyJsonValue_NullableType_ExpectNullReturned()
+        public void EmptyJsonValue_NullableType_ExpectJsonSerializationException()
         {
-            Mass? deserializedNullMass = DeserializeObject<Mass?>("{}");
-            Assert.Null(deserializedNullMass);
+            Assert.Throws<JsonSerializationException>(() => DeserializeObject<Mass?>("{}"));
         }
 
         [Fact]
--- a/UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonDeserializationTests.cs
+++ b/UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonDeserializationTests.cs
@@ -99,12 +99,16 @@ namespace UnitsNet.Serialization.JsonNet.Tests
         }
 
         [Fact]
-        public void EmptyJsonValue_ExpectNullReturned()
+        public void EmptyJsonValue_NullableType_ExpectNullReturned()
         {
-            var json = "{}";
-            var deserializedNullMass = DeserializeObject<Mass?>(json);
-
+            Mass? deserializedNullMass = DeserializeObject<Mass?>("{}");
             Assert.Null(deserializedNullMass);
+        }
+
+        [Fact]
+        public void EmptyJsonValue_NonNullableType_ExpectJsonSerializationException()
+        {
+            Assert.Throws<JsonSerializationException>(() => DeserializeObject<Mass>("{}"));
         }
 
         [Fact]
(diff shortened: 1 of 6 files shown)


UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonDeserializationTests.cs
```diff
@@ -98,6 +98,15 @@ public void NullValue_ExpectNullReturned()
             Assert.Null(deserializedNullMass);
         }
 
+        [Fact]
+        public void EmptyJsonValue_ExpectNullReturned()
```
angularsen, line 102: What happens when deserializing to `Mass`, not nullable? Can this be a separate test?

angularsen (changes requested):
Looks good, a minor suggestion

UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonDeserializationTests.cs
shaeussler: It throws a Newtonsoft.Json.JsonSerializationException: 'No JSON content found and type 'UnitsNet.Mass' is not nullable. Path '', line 0, position 0.'

shaeussler:
Why do you call "token.ToObject<IQuantity>(serializer);" inside ReadJson? 
Here is another fix:
```
            if (valueUnit == null)
            {
                return existingValue; // token.ToObject<IQuantity>(serializer);
            }
```

UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonDeserializationTests.cs
angularsen: Alright, let's add a test case asserting this exception type is thrown in that case.

angularsen:
> Why do you call "token.ToObject(serializer);" inside ReadJson? Here is another fix:

I think this was bug. `UnitsNetIComparableJsonConverter` uses `UnitsNetBaseJsonConverter<T>.CreateLocalSerializer` to create a new serializer internally, excluding itself.

I'm guessing a similar approach was intended here, but we should not need to deserialize a second time here and can instead just return `existingValue` as you proposed.

I pushed some changes in that direction, take a look

UnitsNet.Serialization.JsonNet.Tests/UnitsNetJsonDeserializationTests.cs
angularsen: I added it

angularsen:
It's hard to find any official documentation on how to handle errors in deserialization, but I changed it to explicitly throw an exception if it fails to deserialize the JSON. 

It seems an exception is the convention when trying to deserialize "{}" to DateTime and other built-in supported types in json.net, rather than returning null or the existing value.

shaeussler:
Throwing a JsonSerializationException in case of an empty json is ok in my eyes. I added the token to the error message.

angularsen:
Great! JSON tokens can potentially be huge, maybe we should truncate it to a max length of say 100?

This example is 50 long: 
`{ "Value": 5.9999999, "Unit": "LengthUnit.Meter" }`

We can also include the full token in the `Data` dictionary property instead of in the message, which allows debuggers to view it as well as log frameworks to log it if configured so.

angularsen:
Thanks a lot!

Nuget should be out shortly.

[Release UnitsNet/5.28.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.28.0)

[Release JsonNet/5.28.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/JsonNet%2F5.28.0)

Issue #1303, opened 2023-08-11
Concurrency checks in UnitValueAbbreviationLookup no longer working correctly


oysteinkrog:
**Describe the bug**
The concurrency checks in UnitAbbreviationsCache are no longer working correctly.
Specificially, we sometimes see this error in our tests:
```	
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
	   at System.Collections.Generic.Dictionary`2.FindValue(TKey key)
	   at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
	   at UnitsNet.UnitAbbreviationsCache.GetAbbreviations(UnitInfo unitInfo, IFormatProvider formatProvider)
	   at UnitsNet.UnitAbbreviationsCache.TryGetUnitAbbreviations(Type unitType, Int32 unitValue, IFormatProvider formatProvider, String[]& abbreviations)
	   at UnitsNet.UnitAbbreviationsCache.GetUnitAbbreviations(Type unitType, Int32 unitValue, IFormatProvider formatProvider)
	   at UnitsNet.UnitAbbreviationsCache.GetDefaultAbbreviation(Type unitType, Int32 unitValue, IFormatProvider formatProvider)
	   at UnitsNet.UnitFormatter.GetFormatArgs[TUnitType](TUnitType unit, Double value, IFormatProvider culture, IEnumerable`1 args)
	   at UnitsNet.QuantityFormatter.ToStringWithSignificantDigitsAfterRadix[TUnitType](IQuantity`1 quantity, IFormatProvider formatProvider, Int32 number)
	   at UnitsNet.QuantityFormatter.FormatUntrimmed[TUnitType](IQuantity`1 quantity, String format, IFormatProvider formatProvider)
	   at UnitsNet.QuantityFormatter.Format[TUnitType](IQuantity`1 quantity, String format, IFormatProvider formatProvider)
	   at UnitsNet.Mass.ToString(String format, IFormatProvider provider)
	   at UnitsNet.Mass.ToString(String format)
	   at UnitsNet.Mass.ToString()
```

**To Reproduce**
Set up a concurrency situation (multiple threads) with e.g. ```UnitsNetSetup.Default.UnitAbbreviations.GetDefaultAbbreviation(unit, culture)``` 

**Expected behavior**
Should never throw Exception.

**Additional context**
[...]

PR #1304, opened 2023-08-11
🐛Fix concurrency in UnitAbbreviationsCache

oysteinkrog:
Naive/quick fix, maybe a better way?

Fixes https://github.com/angularsen/UnitsNet/issues/1303


angularsen:
I decided to bring back ConcurrentDictionary instead, take a look at #1305 .

PR #1308, opened 2023-08-21
Fix force conversion rates

### LINKED ISSUE #1306: Incorrect standard gravity/unit conversion constant
Working in the force calibration industry, and needing to occasionally perform mass conversion, I don't agree on the values that are used for force, ~~and mass~~, unit conversions. While not _technically_ the standard acceleration due to standard gravity (standard gravity), these unit conversion rates are derived from that value and are accepted as the standard unit conversion rate.

I'm not sure where the constant originally came from, but the furthest back I could find it referenced in this repository was at the below linked file. This value has carried through significant overhauls to the library for the last decade without any documentation about where it came from. I propose that the value `9.80665002864` be changed to `9.80665` as per the standard definition of acceleration due to standard gravity.

https://en.wikipedia.org/wiki/Gravitational_acceleration

> A conventional standard value is defined **exactly** as `9.80665 m/s2` (32.1740 ft/s2)

https://en.wikipedia.org/wiki/Standard_gravity

> It is defined by standard as `9.80665 m/s2` (about 32.17405 ft/s2). This value was established by the 3rd General Conference on Weights and Measures (1901, CR 70) and used to define the standard weight of an object as the product of its mass and this nominal acceleration.

> Although the actual acceleration of free fall on Earth varies according to location, the above standard figure is always used for metrological purposes.

### Original

https://github.com/angularsen/UnitsNet/blob/986d10b9d49cdfc97555e543dcf8e5843ef688e3/Src/UnitsNet/Constants.cs#L25

### Current

https://github.com/angularsen/UnitsNet/blob/bf080938604ee6d59a6c8e27e25887861c1e9199/Common/UnitDefinitions/Force.json#L35

It may be worth pointing out that the conversion rate for newtons to kilograms-force is also derived from standard gravity and would require the value here to be changed as well. Other conversion rates, and quantities, are impacted by this "constant" as well.

For reference, we use `4.4482216152605 N/lbf` as the conversion rate for pounds-force to newtons and `0.45359237 kgf/lbf` as the conversion rate for pounds-force to kilograms-force. And to calculate the value of standard gravity, we take the conversion rate of `lbf -> N` divided by the conversion rate of `lbf -> kgf`. In the context of Units.NET, it should be the conversion rate of `N -> N` divided by the conversion rate of `N -> kgf`. This can be used as a unit test to verify the values aren't changed in the future.

https://en.wikipedia.org/wiki/Kilogram-force

> Therefore, one kilogram-force is by definition equal to 9.80665 N.

---

**Edit:** I originally included the mass equations because I thought I remembered them being wrong as well. After double checking today, it appears the mass conversions are correct, but the force conversions are not. The values look like they are the result of some rounding errors but I can't quite put my finger on what it might be.

MH-ZShearer:
Fixes #1306

Corrects the conversion rates of the force quantity to be the precisely defined values, relative to standard gravity.

Adds a unit test that verifies that the conversion rate from kilograms divided by the conversion rate from newtons equals standard gravity (`9.80665` m/s<sup>2</sup>).


Common/UnitDefinitions/Force.json
```diff
@@ -142,8 +156,10 @@
     {
       "SingularName": "OunceForce",
       "PluralName": "OunceForce",
-      "FromUnitToBaseFunc": "{x} * 2.780138509537812e-1",
-      "FromBaseToUnitFunc": "{x} / 2.780138509537812e-1",
+      "FromUnitToBaseFunc": "{x} * 0.27801385095378125",
+      "FromBaseToUnitFunc": "{x} / 0.27801385095378125",
+      "XmlDocSummary": "An ounce-force is 1⁄16 of a pound-force, or about 0.2780139 newtons.",
```
angularsen, line 161: We can divide the function in pound-force by 16 here to preserve as much precision as possible:
`"FromUnitToBaseFunc": "{x} * 4.4482216152605 / 16"`

UnitsNet.Tests/CustomCode/ForceTests.cs
```diff
@@ -7,33 +7,36 @@ namespace UnitsNet.Tests
...
 
         protected override double MillinewtonsInOneNewton => 1.0e3;
 
         protected override double MicronewtonsInOneNewton => 1.0e6;
 
-        protected override double OunceForceInOneNewton => 3.596943089595368;
+        protected override double OunceForceInOneNewton => 1.0 / (PoundsConversionRate / 16.0);
 
-        protected override double ShortTonsForceInOneNewton => 1.12404471549816e-4;
+        protected override double ShortTonsForceInOneNewton => 1.0 / (PoundsConversionRate * 2_000.0);
```
angularsen, line 40: I understand your thinking to make the values more readable, but the tests are intentionally using constant values looked up in reference sources like a conversion table, wolfram alpha, online converter or similar. 

This way we sanity check any calculation described in the JSON files instead of doing the same calculation twice.

angularsen (changes requested):
Very nice, just a few minor things to look at.

UnitsNet.Tests/CustomCode/ForceTests.cs
MH-ZShearer: As with the other suggestion, I was expecting this process to be the other way around, sorry.

MH-ZShearer: After discussing this further internally, I can't say we agree with this request. However, it is your call but would like you to consider #1309 prior to making a final decision.

angularsen: Until we conclude otherwise in #1309, let's merge this PR with improvements to the conversion constants. It is an improvement regardless of what changes to the test setup we end up with, and this PR is almost ready to go.

We can spend more time discussing the unit testing setup in general in #1309 .

The library guideline favors calculations in JSON conversion functions to preserve the highest precision possible, while favoring finite constants (less precise) in the test assertions to double check the conversion function based on a secondary source of documentation, such as wikipedia, wolfram alpha, etc. 

I think this is a sound approach, but I'm open to learning otherwise.
Let's have that discussion in #1309.

UnitsNet.Tests/CustomCode/ForceTests.cs
```diff
@@ -7,33 +7,36 @@ namespace UnitsNet.Tests
...
```
angularsen, line 39: Please also change these test values to constants looked up in some reference source, so we don't repeat the same calculation as is performed in JSON and follow the current testing convention.

angularsen (changes requested):
Almost there 😊

UnitsNet.Tests/CustomCode/ForceTests.cs
MH-ZShearer: I'm not sure I understand the request. Is there a similar solution you would like me to follow?

angularsen: The original value was `1.12404471549816e-4`. You can either revert back to this, or find a more accurate value in wikipedia, wolfram alpha, or some online converter. Similar for the other properties.

The purpose of these test values is to double check the conversion function in JSON and to be easily verifiable in some online converter.

MH-ZShearer: Sorry about that. I thought per your other requested change you wanted to hold off on making this change until #1309 was resolved.

angularsen:
Thanks! A lot of great improvements here.
Nuget should be out shortly. 

[Release UnitsNet/5.33.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.33.0)

Now that this is merged, I can put my attention to #1309 the next time I find some free time.

Issue #1296, opened 2023-07-28
Ohm's Law Overloads for ElectricPotentialAc & ElectricPotentialDc


McNeight:
**Describe the bug**
Per Ohm's Law, multiplying an ElectricPotential object with an ElectricCurrent object results in a Power object. Attempting to use an ElectricPotentialAc or ElectricPotentialDc object results in error CS0019.

**To Reproduce**
1. var x = new ElectricPotentialDc(5, UnitsNet.Units.ElectricPotentialDcUnit.VoltDc) * new ElectricCurrent(2, UnitsNet.Units.ElectricCurrentUnit.Ampere);
2. See error: CS0019

**Expected behavior**
x is a Power object of 10 Watts.

**Additional context**
Perhaps I'm not understanding the purpose of the Ac and Dc objects and their difference from the "basic" ElectricPotential unit? I'm not sure if adding more operators to cover the Ac and Dc classes, or possibly having them inherit the operator from a base class is the better approach.

lamarch:
I think that you should go with `ElectricPotential` instead, because it properly implements Ohm's law.

As we can see, theses special AC/DC units were added in commit f82e49e .
However, looking in `UnitsNet/CustomCode/Quantities`, no multiplication behavior have been implemented, for both `ElectricalPotentialAc` and `ElectricalPotentialDc`. 

Frst, I think that `ElectricalPotentialDc` is redundant with `ElectricalPotential`, which implements the correct multiplication operations with current.

Moreover, we can see that the same commit added `ReactivePower` and `ApparentPower`, and if I can correctly remember my electrical engineering course, reactive power and apparent power are calculated using complex current and complex potential (that uses the frequency of the quantity as well as its phase shift).
I honestly think that theses very special quantities are beyond the scope of this API.
Another thing catches my curiosity here : `ElectricalCurrentAc` and `ElectricalCurrentDc` are both missing, which is inconsistency.

The direct way to solve this misunderstanding would be to remove theses 2 quantities : `ElectricalPotentialDc` & `ElectricalPotentialAc`. I don't think that `ReactivePower` & `ApparentPower` should be removed, that being said.
Considering that it would be a breaking change, I propose to mark them as obsolete, and to redirect on `ElectricalPotential`.

We could also implement complex quantities, using complex numbers, for AC current and potential, which would partially solve this issue, however, we would still need to include the frequency parameter with theses quantities. Outrageous complexity...

angularsen:
[...]

Issue #1200, opened 2023-02-12
v6 Wishlist


angularsen:
- Remove BtuPerSquareFootDegreeFahrenheit, replaced by BtuPerHourSquareFootDegreeFahrenheit

angularsen:
Decide on a naming and abbreviation convention for imperial/UK units and US units; like gallon, ounces, cup etc.

It is a bit of a mess right now:
![image](https://user-images.githubusercontent.com/787816/229309571-7f89b59a-3bbc-4a32-ad3d-80879477f8ec.png) ![image](https://user-images.githubusercontent.com/787816/229309566-f036a771-9cef-4909-8206-fd5819001636.png)

![image](https://user-images.githubusercontent.com/787816/229309580-70a06f69-b309-4b55-a08c-7b568b01ab8c.png)

<details>
<summary>LINQPad sample</summary>

```cs
void ListUnits(string withPartialName)
{
    Quantity.Infos
        .SelectMany(i => i.UnitInfos
            .Where(n => n.Name.Contains(withPartialName))
            .Select(u => new 
            { 
                u.Name, 
                Abbrev = UnitAbbreviationsCache.Default.GetDefaultAbbreviation(u.Value.GetType(), Convert.ToInt32(u.Value)) 
            }))
        .Dump(withPartialName);
}

ListUnits("Uk");
ListUnits("Imperial");
ListUnits("Us");
```
</details>



I personally like "UK", since it is shorter and better matches "US". 
[Wolfram](https://www.wolframalpha.com/input?i=1.0000000000000000+cubic+meter+per+meter+to+imperial+gallon+per+mile) also uses US gallon and UK gallon.

Some conventions for imperial gallons:
https://en.wikipedia.org/wiki/Gallon#:~:text=the%20imperial%20gallon%20(imp%20gal,American%20and%20Caribbean%20countries%3B%20and

Today the abbreviation convention is `gal (U.S.)` and `gal (imp.)`.
Depending on landing on UK or imperial naming, we could go with one of these pairs:
- `US gal` and `UK gal`
- `US gal` and `imp gal`
- `gal (US)` and `gal (UK)`
 
We can support multiple of these plus the old ones, but the primary form should be standardized across the different units.

ebfortin:
Do not throw ArgumentException on NaN, Inf+ or Inf- or give a way to configure the behavior globally.

angularsen:
> I think UnitsNet should be more extensible. 

Conceptually, I agree. 

[...]

PR #1312, opened 2023-09-04
WIP: ElectricAcDc

McNeight:
This is a rough cut at the reference type suggested in #1296. Primarily for review and to see if I'm on the right track or not.


Common/UnitDefinitions/ElectricPotentialAc.json
```diff
@@ -1,20 +0,0 @@
-{
-  "Name": "ElectricPotentialAc",
```
angularsen, line 2: Just a heads up that we can't delete these, we have to mark them obsolete in JSON with `ObsoleteText` on the quantity for example just below `Name` on line 3, then add to #1200 , then remove them in next major version.

The obsolete text should describe what they should use instead for AC and DC operations.
This can wait until the design is more complete of course.

angularsen (changes requested):
It would helpful if you could summarize the changes and what the changes aim to accomplish in the PR description.

On the surface, this looks useful enough.
However, I don't know this domain well enough to give any meaningful comments on the design of the wrapper types. It's just one those things that someone who actually needs it will have to start using and massage until it solves several use cases.

It is important for me to only add something to UnitsNet that will actually be used and solves a need that many UnitsNet consumers will benefit from.

So to be very direct 😄 
Do you plan to use this yourself? 
Will it solve a real for you? 
Are there any known unknowns, things to still explore or figure out in the design or use cases to test the design against?
Do you know other developers working in this domain that we could involve in this discussion? 

It doesn't have to be perfect on the first go, but it is helpful to think along these lines from the start.

angularsen:
If I understand correctly, we concluded to not continue with this PR?

https://github.com/angularsen/UnitsNet/issues/1296#issuecomment-2213785220

PR #1316, opened 2023-09-18
Add micro and nano prefixes to joule

pliebe91:
Hello,

For the correct display of the values of our laser energy meters we are currently missing the prefixes micro and nano for the unit joule.

It would be nice if this could be added.

Many thanks


angularsen:

[Release UnitsNet/5.34.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.34.0)

PR #1318, opened 2023-09-19
Added Gigameter (1e+9 meters)

yigitlevent:
Gigameter is really useful for my case, even though it is a bit obscure, when rendering astronomical distances it seems to hit the perfect spot. While I use AU and meter during the calculations, these intermediate units really help when it comes to rendering the simulation itself. It seems to be a good fit for making sure that both short and long stellar distances can be rendered without too much trouble.

I am completely fine with this getting rejected as the number of meter units this library is probably getting out of hand, but I wanted to create a PR to see if it would be a good fit for the library.

I followed the contribution page and also checked out the PR for Megameters (#1208) but please let me know if I missed something.

### DIFF
--- a/Common/UnitDefinitions/Length.json
+++ b/Common/UnitDefinitions/Length.json
@@ -14,7 +14,7 @@
       },
       "FromUnitToBaseFunc": "{x}",
       "FromBaseToUnitFunc": "{x}",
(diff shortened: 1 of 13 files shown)


angularsen:
Google tells me it is not frequently used, neither in terrestrial nor in astronomy, but there are references to this unit so I'll accept it, mostly for completeness and since you have at least encountered one real use case for it.

angularsen:

[Release UnitsNet/5.34.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.34.0)

PR #1323, opened 2023-10-18
Added some nullability annotations to v4

ptasev:
In my project I treat warnings as errors, and when using the source generator from `System.Text.Json` for a type that has `UnitsNet` properties, it was generating code that had nullability warnings on `BaseDimensions` and `BaseUnits` for `operator ==`.

Since there's no good way to suppress this warning only for source generated code, the only way to fix this, short of disabling the warnings-as-errors across my whole project, is to add these annotations.

### DIFF
--- a/CodeGen/Generators/UnitsNetWrcGen/QuantityGenerator.cs
+++ b/CodeGen/Generators/UnitsNetWrcGen/QuantityGenerator.cs
@@ -531,7 +531,7 @@ namespace UnitsNet
             Writer.WL($@"
         #region Equality / IComparable
 
-        public int CompareTo(object obj)
+        public int CompareTo(object? obj)
         {{
             if (obj is null) throw new ArgumentNullException(nameof(obj));
             if (!(obj is {_quantity.Name} obj{_quantity.Name})) throw new ArgumentException(""Expected type {_quantity.Name}."", nameof(obj));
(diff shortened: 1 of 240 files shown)


CodeGen/Generators/UnitsNetWrcGen/QuantityGenerator.cs
```diff
@@ -531,7 +531,7 @@ private void GenerateEqualityAndComparison()
             Writer.WL($@"
         #region Equality / IComparable
 
-        public int CompareTo(object obj)
+        public int CompareTo(object? obj)
```
tmilnthorp, line 534: Since we throw ```ArgumentNullException```, this should not be marked as accepting a nullable type.

tmilnthorp (approved):

tmilnthorp:
@angularsen looks like a NuGet failure in ADO.

We should merge this into main also.

angularsen:
Trying to get v4 branch to build on ADO.

angularsen:
@ptasev Merged, sorry this took so long. It fell off my radar.

The v4 branch has not been built in a long time and required some tweaks in the pipeline, but hopefully a nuget should be out shortly.

[Release UnitsNet/4.152.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F4.152.0)

@tmilnthorp If you want to, see if we can merge this into master. I assume there will be some conflicts to sort out and I don't have the time to attempt that right now.

angularsen:
Had to fix one more thing, but the nuget 4.152.0 is finally out.

PR #1329, opened 2023-11-13
Generate operators from unit relations defined in JSON

Muximize:
Related #1200 

In the PR adding generic math (#1164) @AndreasLeeb states:
> Regarding the operators in the *.extra.cs files, that could be tackled easily by describing the dependencies (operations) between different quantities in the quantity JSON files, and then the operator overloads and the generic math interfaces for the quantity structs could also be automatically generated. But that's a topic for another time 😄

I decided to give this a shot.

`UnitRelations.json` contains relations extracted from the existing *.extra.cs files. I decided on a new file because multiplication is commutative and I didn't want to duplicate these in the individual quantity JSON files, or risk missing one or the other, so it's best to define them once in one place. The generator handles this by generating two operators for a single multiplication relation.

The relations format uses the quantities method names. This is a bit unfortunate, but it's the best I could come up with without making the CodeGen project depend on UnitsNet, which would create a bit of a chicken/egg problem. This is not unheard of (self-hosted compilers) but I wanted to keep it simple for now.

The generated code enables the removal of 44 *.extra.cs files, and the 17 remaining contain much less code.


Muximize:
@angularsen What's your opinion about this? Is this a worthwhile endeavor in the right direction?

angularsen:
Apologies, I thought this was still being worked on so I haven't looked at it yet.
The PR description sounds very useful, I'm just curious if there are any challenges to this approach or not.

Let me read more through the code and get back to you 👍 Will try to find time the next couple of days.

Muximize:
No worries, I should have been more clear that this is a proposal in need of feedback. God jul! :smile:

Common/UnitRelations.json
```diff
@@ -0,0 +1,165 @@
...
+  { "Operator": "*", "Result": "Energy.FromJoules", "Left": "EnergyDensity.JoulesPerCubicMeter", "Right": "Volume.CubicMeters" },
+  { "Operator": "*", "Result": "Energy.FromJoules", "Left": "Power.Watts", "Right": "Duration.Seconds" },
+  { "Operator": "*", "Result": "Energy.FromJoules", "Left": "SpecificEnergy.JoulesPerKilogram", "Right": "Mass.Kilograms" },
+  { "Operator": "*", "Result": "Energy.FromJoules", "Left": "TemperatureDelta.Kelvins", "Right": "Entropy.JoulesPerKelvin" },
+  { "Operator": "*", "Result": "Entropy.FromJoulesPerKelvin", "Left": "SpecificEntropy.JoulesPerKilogramKelvin", "Right": "Mass.Kilograms" },
+  { "Operator": "*", "Result": "Force.FromNewtons", "Left": "ForceChangeRate.NewtonsPerSecond", "Right": "Duration.Seconds" },
+  { "Operator": "*", "Result": "Force.FromNewtons", "Left": "ForcePerLength.NewtonsPerMeter", "Right": "Length.Meters" },
+  { "Operator": "*", "Result": "Force.FromNewtons", "Left": "Mass.Kilograms", "Right": "Acceleration.MetersPerSecondSquared" },
+  { "Operator": "*", "Result": "Force.FromNewtons", "Left": "Pressure.Pascals", "Right": "Area.SquareMeters" },
```
angularsen, line 19: I think this is a sound design.

### 1. How about making it more compact
It could make it easier to read and maintain. It also resembles code using these operators.

This structure should be possible to parse, and CodeGen has access to the the plural form of units to substitute in the generated code.

```json
[ 
"ElectricCurrent.Ampere * Duration.Hour = ElectricCharge.AmpereHour",
"ElectricCurrent.Ampere * ElectricResistance.Ohm = ElectricPotential.Volt",

"Power.Watt / ElectricPotential.Volt = ElectricCurrent.Ampere",

"1 / Length.Meter = ReciprocalLength.InverseMeter"
]
```

### 2. Document sort order
One weakness with this approach, and the old approach too, is that it is not easy to see if a particular relation is defined. In particular for multiplication, where only one of two permutations of left/right is defined.

I propose a sorting the the document alphabetically on: 
- Left side argument first, then by right side argument

This reads better when following the string-based schema above, or changing the JSON to move "Result" last.

It's going to be a pain to enforce this for contributors, but I think we should at least define a standard layout/order for the document that we can maybe even have CodeGen auto-sort for us.


Let me know what you think.

UnitsNet/CustomCode/Quantities/Force.extra.cs
```diff
@@ -19,59 +19,5 @@ public static Force FromMassByAcceleration(Mass mass, Acceleration acceleration)
...
-        /// <summary>Get <see cref="ForcePerLength"/> from <see cref="Force"/> divided by <see cref="Length"/>.</summary>
-        public static ForcePerLength operator /(Force force, Length length)
-        {
-            return ForcePerLength.FromNewtonsPerMeter(force.Newtons / length.Meters);
-        }
-
-        /// <summary>Get <see cref="Duration"/> from <see cref="Force"/> divided by <see cref="ForceChangeRate"/>.</summary>
-        public static Duration operator /(Force force, ForceChangeRate forceChangeRate)
-        {
-            return new Duration(force.Newtons / forceChangeRate.NewtonsPerSecond, DurationUnit.Second);
```
angularsen, line 74: Does this PR replace the existing operators exactly, or are some new operators also added?

UnitsNet/GeneratedCode/Quantities/MassConcentration.g.cs
```diff
@@ -41,6 +44,11 @@ namespace UnitsNet
     [DataContract]
     public readonly partial struct MassConcentration :
         IArithmeticQuantity<MassConcentration, MassConcentrationUnit, double>,
+#if NET7_0_OR_GREATER
+        IMultiplyOperators<MassConcentration, Volume, Mass>,
+        IDivisionOperators<MassConcentration, MolarMass, Molarity>,
+        IDivisionOperators<MassConcentration, Density, VolumeConcentration>,
+#endif
```
angularsen, line 51: Nice touch on adding this 👍

angularsen (changes requested):
I think this looks very promising and a big improvement on the existing approach. 

A few ideas below.

UnitsNet/CustomCode/Quantities/Force.extra.cs
Muximize: Some new multiplication operators are added where they where not defined in a commutative way. For example, `HeatFlux * Area = Power` was defined in `HeatFlux.extra.cs` but the other way round did not exist.
This PR generates both, in `HeatFlux.g.cs` and `Area.g.cs`, based on a single definition in the json file. 

Also, some operators where defined only for `Duration` or `TimeSpan`, some for both. In the definitions file I only use `Duration` as an operand and then generate the `TimeSpan` equivalent too.

This makes the API more consistent, but also has an effect on code size (net +469 lines in e7ba65004) and test coverage. I explicitly chose not to touch any tests for now but to make sure all existing tests pass. Do you have any guidance on this? Should we generate tests for these operators and then remove the manual tests?

UnitsNet/GeneratedCode/Quantities/MassConcentration.g.cs
Muximize: That was the main goal of this exercise, the generated operators where just the way to get there 😄

UnitsNet/CustomCode/Quantities/Force.extra.cs
angularsen: Ok, those additions sounds good. Keeping the change as small as possible compared to existing code.

Code size is fine I think, I favor consistency.
Yes, let's leave the tests for now to make sure what existed before will still work, then we can generate tests in a second PR after merging this.

Common/UnitRelations.json
angularsen: Just realized, since multiplication and possibly addition only defines one of two permutations, `sorting on left side argument first` is ambiguous on what quantity to choose as the left side.

### Option 2.1
Pick the left/right side by alphabetical order also, so `ElectricCurrent.Ampere * Duration.Hour = xxx` would instead be `Duration.Hour * ElectricCurrent.Ampere = xxx`.

### Option 2.2
Or we could forego this problem altogether and move the `result` first and sort on that, similar to your schema. 

Either like this:
- Result type (alphabetically)
- Then operation order (`* / + -`)


```json
[
"ElectricCharge.AmpereHour = ElectricCurrent.Ampere * Duration.Hour",
"ElectricCurrent.Ampere = Power.Watt / ElectricPotential.Volt",
"ElectricPotential.Volt = ElectricCurrent.Ampere * ElectricResistance.Ohm",
"ReciprocalLength.InverseMeter = 1 / Length.Meter"
]
```

Or like this:
- Operation order (`* / + -`)
- Then result type (alphabetically)

```json
[ 
// Multiplication
"ElectricCharge.AmpereHour = ElectricCurrent.Ampere * Duration.Hour",
"ElectricPotential.Volt = ElectricCurrent.Ampere * ElectricResistance.Ohm",

// Division, including inverse
"ReciprocalLength.InverseMeter = 1 / Length.Meter",
"ElectricCurrent.Ampere = Power.Watt / ElectricPotential.Volt",
]
```

### Option 2.3
Duplicate all two-way methods, and possibly use CodeGen to enforce this by first scanning and correcting the document before parsing it.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -696,6 +732,81 @@ private void GenerateLogarithmicArithmeticOperators()
 " );
         }
 
+        private void GenerateRelationalOperators()
```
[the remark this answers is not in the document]
angularsen: I know, but it would still be an improvement. I would rather add more xmldoc to the remaining code, but not part of this PR.

Private vs public does not matter much in the context of CodeGen, since contributors are the only ones who are going to read this code anyway.

CodeGen/JsonTypes/Relation.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Licensed under MIT No Attribution, see LICENSE file at the root.
+// Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
+
+namespace CodeGen.JsonTypes
+{
+    internal record Relation
```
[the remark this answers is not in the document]
Muximize: Agreed on the name change. Docs would be repetitive, I've added them to `QuantityRelationsParser`

Common/UnitRelations.json
Muximize: Thanks, awesome feedback. The schema was the thing I was least happy with and I like your options better.

I've reworked this like option 2.2, but initially just sorted alphabetically on the whole string. Also added:

- Auto-sort and write to disk on codegen
- Validation of relations based on unit definitions

angularsen: > Another idea: generate division operators based on multiplication

Interesting idea, my gut tells me this may be a notch too much magic, since it is not clear whether a division operator implementation comes from a multiplication definition or an explicit division definition. Also, there are some dragons with some of the quantities, and maybe some division operations either don't make sense or are not desirable.

Then again, we can just try and see how it works out. I'm all for experimenting a bit on this 👍

angularsen: > alphabetically on the whole string

That works too, simpler even.

Muximize: I did some experimenting with generating division operators based on multiplication. I do agree it might be a bit magic, but I managed to avoid some confusion by removing all explicit division definitions 😄 
There are indeed some dragons so I had to pull a few tricks, but in the end it seems to work. See #1354

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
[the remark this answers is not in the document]
Muximize: Done 👍

Common/UnitRelations.json
angularsen: Cool! Let's continue the division experiment separately after merging this first, I think the two changes can be isolated.

Is there anything else that remains on this PR now, or is it ready for the final review?

Muximize: While it stays backwards compatible, the API surface grows quite a bit, and the units get generic math interfaces for the first time. It's also a conceptual change for contributors. Wouldn't we want to target v6?

angularsen: No I think it's fine. 
We're just adding stuff and I don't see anything too experimental here that needs to mature its design before release.

Muximize: Allright, I pushed some changes to clear up some code, add some comments and sort the operators after inferring new ones. Ready for final review 👍

CodeGen/Generators/QuantityRelationsParser.cs
```diff
@@ -0,0 +1,174 @@
...
+        {
+            var relationsFileName = Path.Combine(rootDir, "Common/UnitRelations.json");
+
+            try
+            {
+                var text = File.ReadAllText(relationsFileName);
+                var relationStrings = JsonConvert.DeserializeObject<List<string>>(text) ?? [];
+                relationStrings.Sort();
+
+                var parsedRelations = relationStrings.Select(relationString => ParseRelation(relationString, quantities)).ToList();
```
angularsen, line 98: Should we throw on duplicates here, with a helpful explanation? 

It probably gives compile errors in the generated code anyway, but it would be helpful if codegen failed early on invalid input to make it easier for contributors to find out what they did wrong.

If we eventually do #1354 , then duplicates could also occur implicitly by one explicit definition conflicting with the implicit definition for division operators.

Another option is to remove duplicates with a HashSet or similar, then fix the document when writing it back. But it may be complicated to know what to remove, in particular with the magic division operators, so it is probably better to just throw and have the author fix their mistake.

CodeGen/Generators/QuantityRelationsParser.cs
```diff
@@ -0,0 +1,174 @@
...
+    {
+        /// <summary>
+        ///     Parse and apply relations to quantities.
+        ///     Each defined relation can be applied multiple times to one or two quantities depending on the operator and the operands.
+        ///
+        ///     The format of a relation definition is "Quantity.Unit operator Quantity.Unit = Quantity.Unit".
+        ///     "double" can be used as a unitless operand.
+        ///     "1" can be used as the left operand to define inverse relations.
+        /// </summary>
+        /// <param name="rootDir">Repository root directory.</param>
```
angularsen, line 27: A few examples with some actual quantities and units in a `<example></example>` tag could be helpful for the reader

CodeGen/Generators/QuantityRelationsParser.cs
```diff
@@ -0,0 +1,174 @@
...
+
+            if (segments is not [_, "=", _, "*" or "/", _])
+            {
+                throw new Exception($"Invalid relation string: {relationString}");
+            }
+
+            var @operator = segments[3];
+            var left = segments[2].Split('.');
+            var right = segments[4].Split('.');
+            var result = segments[0].Split('.');
```
angularsen, line 123: This is fine, I was just thinking maybe regex was a good fit for this parsing?
Not important at all, just a tip to consider.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -17,6 +17,8 @@ internal class QuantityGenerator : GeneratorBase
         private readonly string _valueType;
         private readonly Unit _baseUnit;
 
+        private readonly string[] _decimalTypes = { "BitRate", "Information", "Power" };
```
angularsen, line 20: If it is not too much hassle, I would rather take a list of all quantities in the ctor to determine the value type from their `Quantity` type.

We are not likely to get more `decimal` types anytime soon, rather changing them to `double` I think, but it may take some time and this is one less place to maintain when that changes.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -696,6 +732,94 @@ private void GenerateLogarithmicArithmeticOperators()
...
+");
+                }
+                else
+                {
+                    var leftParameter = relation.LeftQuantity.Name.ToCamelCase();
+                    var leftConversionProperty = relation.LeftUnit.PluralName;
+                    var rightParameter = relation.RightQuantity.Name.ToCamelCase();
+                    var rightConversionProperty = relation.RightUnit.PluralName;
+
+                    if (relation.LeftQuantity.Name == "TimeSpan")
```
angularsen, line 766: Maybe reuse some constants for all these repeating strings

UnitsNet/CustomCode/Quantities/Area.extra.cs
```diff
@@ -23,47 +23,5 @@ public static Area FromCircleRadius(Length radius)
...
-        /// Calculates the inverse of this quantity.
-        /// </summary>
-        /// <returns>The corresponding inverse quantity, <see cref="ReciprocalArea"/>.</returns>
-        public ReciprocalArea Inverse()
-        {
-            if (SquareMeters == 0.0)
-                return new ReciprocalArea(0.0, UnitsNet.Units.ReciprocalAreaUnit.InverseSquareMeter);
-
-            return new ReciprocalArea(1 / SquareMeters, UnitsNet.Units.ReciprocalAreaUnit.InverseSquareMeter);
-        }
```
angularsen, line 37: Are you confident that all existing operators still exist so we don't break anyone? 

Not so easy to verify in review, but if there is doubt we could maybe do some reflection at runtime before/after PR to list all quantities and their public methods, properties, operators and do a text diff on any difference in the public API.

angularsen (changes requested):
I think this looks solid! This is some impressive work 👏 

Some minor suggestions for you to consider.

The biggest one for me is to get some peace of mind that we are not breaking anyone as it is hard to verify by review.

I added a suggestion to maybe use reflection to compare before/after of all quantities and their public members, by outputting to text and diffing that.

CodeGen/Generators/QuantityRelationsParser.cs
```diff
@@ -119,7 +112,7 @@ private static QuantityRelation ParseRelation(string relationString, IReadOnlyDi
         {
             var segments = relationString.Split(' ');
 
-            if (segments.Length != 5 || segments[1] != "=")
+            if (segments is not [_, "=", _, "*" or "/", _])
             {
                 throw new Exception($"Invalid relation string: {relationString}");
```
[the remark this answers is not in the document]
Muximize: It could, but isn't `UnitRelations.json` full of valid examples?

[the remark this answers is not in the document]
angularsen: Absolutely, but I still think it improves the developer experience a bit.

CodeGen/Generators/QuantityRelationsParser.cs
Muximize: I tried that but it wasn't a big improvement, the regex is quite ugly and less comprehensible than the list pattern:
```csharp
^(\w+)\.?(\w*) = (\w+)\.?(\w*) (\*|\/) (\w+)\.?(\w*)$

[_, "=", _, "*" or "/", _]
```

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
Muximize: This wasn't needed anymore since I reworked the codegen, so I removed it 👍

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
Muximize: I tried to tidy up a bit, but code generation is always a bit messy and throwing around consts called `_double` and `_value` didn't make things better 😅

CodeGen/Generators/QuantityRelationsParser.cs
Muximize: Done 👍

CodeGen/Generators/QuantityRelationsParser.cs
Muximize: Good idea, I added a check after all relations are inferred.

Duplicate definitions in `UnitRelations.json` are automatically removed with a `SortedSet`.

UnitsNet/CustomCode/Quantities/Area.extra.cs
Muximize: Yes, this is an interesting one. I'll see what I can do, but it's one of the reasons I would feel more comfortable with targeting v6 with this.

I know there are some longstanding architectural issues you would like to tackle too, but maybe those could be deferred to a v7? Then v6 could offer, amongst other things:

- Generated relational operators
- NaN, Inf values for double quantities
- Fixed typos in unit names
- Power using `double`

Muximize: Alright, using reflection to get the exact API changes was quite doable.

This PR adds 26 new operators to quantities and doesn't remove any.
```
AmountOfSubstance op_Multiply(System.TimeSpan, MolarFlow)              in MolarFlow
AmountOfSubstance op_Multiply(Duration, MolarFlow)                     in Duration
ElectricCharge op_Multiply(System.TimeSpan, ElectricCurrent)           in ElectricCurrent
ElectricCharge op_Multiply(ElectricCurrent, System.TimeSpan)           in ElectricCurrent
ElectricCurrent op_Division(ElectricCharge, System.TimeSpan)           in ElectricCharge
ElectricCurrent op_Multiply(System.TimeSpan, ElectricCurrentGradient)  in ElectricCurrentGradient
ElectricCurrent op_Multiply(Duration, ElectricCurrentGradient)         in Duration
Energy op_Multiply(Volume, EnergyDensity)                              in Volume
Force op_Multiply(System.TimeSpan, ForceChangeRate)                    in ForceChangeRate
Force op_Multiply(ForceChangeRate, System.TimeSpan)                    in ForceChangeRate
Force op_Multiply(Length, ForcePerLength)                              in Length
Jerk op_Division(Acceleration, System.TimeSpan)                        in Acceleration
LuminousIntensity op_Multiply(Area, Luminance)                         in Area
Mass op_Multiply(Area, AreaDensity)                                    in Area
Mass op_Multiply(Length, LinearDensity)                                in Length
MassFlow op_Multiply(MolarMass, MolarFlow)                             in MolarMass
Power op_Multiply(Area, HeatFlux)                                      in Area
Pressure op_Multiply(System.TimeSpan, PressureChangeRate)              in PressureChangeRate
Pressure op_Multiply(Duration, PressureChangeRate)                     in Duration
RotationalStiffness op_Multiply(Length, RotationalStiffnessPerLength)  in Length
Speed op_Multiply(System.TimeSpan, Acceleration)                       in Acceleration
Speed op_Multiply(Acceleration, System.TimeSpan)                       in Acceleration
Torque op_Multiply(Angle, RotationalStiffness)                         in Angle
Torque op_Multiply(Area, ForcePerLength)                               in Area
Volume op_Multiply(System.TimeSpan, VolumeFlow)                        in VolumeFlow
Volume op_Multiply(Mass, SpecificVolume)                               in Mass
```

29 existing operators get moved to a different quantity because we generate them as part of the left parameter:
```
Angle op_Multiply(Duration, RotationalSpeed)                   from RotationalSpeed to Duration
Area op_Multiply(Duration, KinematicViscosity)                 from KinematicViscosity to Duration
Energy op_Multiply(Duration, Power)                            from Power to Duration
Energy op_Multiply(Entropy, TemperatureDelta)                  from TemperatureDelta to Entropy
Energy op_Multiply(Mass, SpecificEnergy)                       from SpecificEnergy to Mass
Entropy op_Multiply(Mass, SpecificEntropy)                     from SpecificEntropy to Mass
Force op_Multiply(Acceleration, Mass)                          from Mass to Acceleration
Force op_Multiply(Area, Pressure)                              from Pressure to Area
ForcePerLength op_Multiply(Area, SpecificWeight)               from SpecificWeight to Area
Length op_Division(Force, ForcePerLength)                      from ForcePerLength to Force
Length op_Division(TemperatureDelta, TemperatureGradient)      from TemperatureGradient to TemperatureDelta
Length op_Multiply(Duration, Speed)                            from Speed to Duration
Mass op_Division(Mass, MassFraction)                           from MassFraction to Mass
Mass op_Multiply(Duration, MassFlow)                           from MassFlow to Duration
Mass op_Multiply(Mass, MassFraction)                           from MassFraction to Mass
Mass op_Multiply(MolarMass, AmountOfSubstance)                 from AmountOfSubstance to MolarMass
Mass op_Multiply(Volume, Density)                              from Density to Volume
Mass op_Multiply(Volume, MassConcentration)                    from MassConcentration to Volume
MassConcentration op_Multiply(Density, VolumeConcentration)    from VolumeConcentration to Density
MassConcentration op_Multiply(MolarMass, Molarity)             from Molarity to MolarMass
MassFlow op_Multiply(Density, VolumeFlow)                      from VolumeFlow to Density
Molarity op_Multiply(VolumeConcentration, Molarity)            from Molarity to VolumeConcentration
Power op_Multiply(Frequency, Energy)                           from Energy to Frequency
Power op_Multiply(Speed, Force)                                from Force to Speed
SpecificEnergy op_Multiply(SpecificEntropy, TemperatureDelta)  from TemperatureDelta to SpecificEntropy
TemperatureDelta op_Multiply(Duration, TemperatureChangeRate)  from TemperatureChangeRate to Duration
TemperatureDelta op_Multiply(Length, TemperatureGradient)      from TemperatureGradient to Length
Torque op_Multiply(Force, Length)                              from Length to Force
Volume op_Multiply(Area, Length)                               from Length to Area
```

angularsen: I missed this update, but that sounds great!
Will look over a bit more closely then I believe we are ready to merge.

angularsen: > 29 existing operators get moved to a different quantity

I think this should be safe and not break anyone, just a bit unsure if it's binary compatible. Will it break an app depending on `Library X`, which depends on UnitsNet, when the app upgrades UnitsNet and library still expects the old version.

Will look into it.

angularsen: Sorry this is taking so long, I'm just having a busy time with work not finding much time or energy for other things. But I intend to help wrap this up soon.

angularsen: Just to comment on versioning

> I know there are some longstanding architectural issues you would like to tackle too, but maybe those could be deferred to a v7

Generally speaking, I have over the years started to appreciate libs that make an effort to not break anyone when upgrading. It doesn't matter if we slap a major version bump on the change, it will still annoy the hell out of people if stuff stops working on upgrade or they are blocked by intermediate dependencies that also need to upgrade first. So the change better be worth it, trivial to fix or affect a very small group of users.

angularsen: I did a quick test on binary compatibility, and it seems moving the operator to a different class is a breaking change.

image


To reproduce:
1. Create local nuget of this PR's change (e.g. 5.99.0)
2. Add library Lib1, which depends on current latest 5.43.0, and performs multiplication on some given quantities

You get a `MissingMethodException`, because the multiply operator method no longer exists on `Power`, it was moved to `Duration`.

> Energy op_Multiply(Duration, Power)                            from Power to Duration

```cs
using System.Diagnostics;
using System.Reflection;
using Lib1;
using UnitsNet;

LogVersion(typeof(Lib1Helpers).Assembly);
LogVersion(typeof(Power).Assembly);

Energy energy1 = Lib1Helpers.Multiply(Power.FromWatts(100), Duration.FromSeconds(10));
Energy energy2 = Lib1Helpers.Multiply(Duration.FromSeconds(10), Power.FromWatts(100));

Console.WriteLine("Energy1: " + energy1);
Console.WriteLine("Energy2: " + energy2);
return;

void LogVersion(Assembly assembly)
{
    Console.WriteLine($"Assembly: {assembly.FullName}");
    Console.WriteLine("Version: " + assembly.GetName().Version);
    Console.WriteLine("File version: " + FileVersionInfo.GetVersionInfo(assembly.Location).FileVersion);
    Console.WriteLine("Product version: " + FileVersionInfo.GetVersionInfo(assembly.Location).ProductVersion);
    Console.WriteLine("\n---\n");
}
```

I don't think this will affect a lot of people, but it is definitely a breaking change, so let's target v6 branch. 
To refer back to my previous comment on breaking changes, this is one change I think is worth it because I don't think it will affect many users.

### In summary
- I'll retarget this PR to v6 and merge it there

Muximize: Cool! 👍 Thanks for your guidance with this.

PR #1331, opened 2023-11-22
Add Mechanical Inertance

Eternali:
This is a fairly niche request but I wanted to see if it was able to be accepted as it is used fairly commonly in certain parts of motorsport.

An inertance constant is a way of defining a tuned mass-damper-inerter system, which is a component of the suspensions of several types of racecar including IndyCar (and formerly Formula 1) and is fairly well documented as a type of mechanical balancer.

Just as a brief introduction to the concept: just as a spring is defined by its resistance to displacement, and a damper is defined by its resistance to velocity, a mass-damper-inerter is defined by its resistance to acceleration.  Following the pattern of units we see a spring constant is expressed as N/m, a damping constant is expressed as N/(m/s), and an inertance constant is expressed as N/(m/s<sup>2</sup>), which just so happens to be mass.

[Here](https://www.annualreviews.org/doi/full/10.1146/annurev-control-053018-023917) is paper that explains the concept more completely. And [here](https://en.wikipedia.org/wiki/Mobility_analogy#Inertance) is a related Wikipedia section.

The interesting part, from a units perspective, comes from the fact that many tools these racing teams use to simulate and manage their vehicles define the inertance constant using those component units, instead of simplifying to a SI unit of mass. This is mainly due to convenience in computation, but also seems to just be the standard adopted in the industry and what engineers like to view the numbers as. 

The reason the units I included have been selected is that they have been explicitly requested by the teams I work with, there may be more I am unaware of.

I also opted to name the unit "MechanicalInertance" as I am aware there are more common definitions of the term "inertance" and while a formal unit hasn't been defined in this library as of yet, I did not want to cause potential future conflict if, for example, fluid-flow inertance was to be added.

If this is too much of a special case to be accepted I understand so feedback is welcome regardless.


angularsen:
Hi and thanks for sharing. On the surface, this seems too niche for Units.NET, sorry. In fact I have trouble finding much references inertance and units at all, probably because as you say, it is dimensionally equivalent to mass.

You can always fork UnitsNet and add it to your private version.
It is also possibly to get some minor reuse by adding it at runtime via [Custom units](https://github.com/angularsen/UnitsNet/wiki/Extending-with-Custom-Units), but this is still considered experimental.

Hope this helps!

PR #1333, opened 2023-12-05
Added SiDecitonne mass unit of measure.

IhorMukhortv:


angularsen:
At first I was sure this was a typo 😄 

1. It seems the unit centner is also called quintal. Which name should we use?

2. There are two different units with the same name. Do we need to add both? And should we name them so it is clear which unit we mean?

> In [British English](https://en.wikipedia.org/wiki/British_English), it referred to the [hundredweight](https://en.wikipedia.org/wiki/Hundredweight); in [American English](https://en.wikipedia.org/wiki/American_English), it formerly referred to an uncommon measurement of 100 kg (220 lb).

https://en.wikipedia.org/wiki/Quintal

We have previously used `gal (U.S.)` vs `gal (imp.)` to disambiguate US and imperial units. 
We could maybe use `centner (U.S.)` and `centner (imp.)` to match that, or "quintal" if we go with that name.

https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#abbreviation-naming-conventions

IhorMukhortv:
> At first I was sure this was a typo 😄
> 
> 1. It seems the unit centner is also called quintal. Which name should we use?
> 2. There are two different units with the same name. Do we need to add both? And should we name them so it is clear which unit we mean?
> 
> > In [British English](https://en.wikipedia.org/wiki/British_English), it referred to the [hundredweight](https://en.wikipedia.org/wiki/Hundredweight); in [American English](https://en.wikipedia.org/wiki/American_English), it formerly referred to an uncommon measurement of 100 kg (220 lb).
> 
> https://en.wikipedia.org/wiki/Quintal
> 
> We have previously used `gal (U.S.)` vs `gal (imp.)` to disambiguate US and imperial units. We could maybe use `centner (U.S.)` and `centner (imp.)` to match that, or similar for "quintal" if we go with that name.
> 
> https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#abbreviation-naming-conventions

I have found a suitable standardized name - 'decitonne.' I haven't delved into the details of the U.S. and Imperial measurement systems. In my opinion, there are two primary systems: Metric (or SI, as the modern variant) and Imperial. Considering that 'decitonne' could still cause confusion, I suggest adding 'decitonne (SI).' In the code, this might appear as 'SiDecitonne.'

If you have any further thoughts or concerns, please let me know, and we can proceed with the implementation. Thank you for your attention to detail!

It appears that 'Hundredweight' is not included in the PDF document provided in the Wiki under 'Adding a New Unit.

Best regards.

IhorMukhortv:
Hi, it's frustrating that my straightforward suggestion—just adding one unit of measurement to your structure—was overlooked. I'm disappointed in the lack of attention to such a simple improvement. I regret investing my time in this, but regardless, I wish you success in your development endeavors.

angularsen:
@IhorMukhortv I think you need to adjust your expectations. This is the nature of many open source projects that are maintained on evenings and nights outside busy work life and family life. This was also not a simple improvement, there were questions and stuff to figure out.

PR #1334, opened 2023-12-05
Add density unit PoundsPerCubicYard + Kilo prefix

aharres:
This commit just adds two new density units: PoundsPerCubicYard and KilopoundsPerCubicYard.

PR #1335, opened 2023-12-06
Improve precision of PR 1334

angularsen:
Fixup of 202eafa - Add density unit PoundsPerCubicYard + Kilo prefix (#1334)

Use precise calculation from definitions.


angularsen:
@aharres I realized the calculation could be improved after I merged, so I did it here.

PR #1339, opened 2023-12-19
Add Kilo prefix to yards

jmnauman:
I'm submitting a pull request to add a "kilo" prefix to yards (similar to kilofeet). 1kyd = 1000yd. It's not a common measurement, but is needed in a project where I use UnitsNet. See the bottom paragraph of [this Wikipedia article](https://en.wikipedia.org/wiki/Metrication_in_the_United_States#Hybrid_units) for an example of where this is practically used.

The [instructions](https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit) for adding a new unit mentioned adding a comment to the KiloyardsInOneMeter getter in LengthTests.cs about the number's source. I didn't see that on any other getters in this file, so I left it out to be consistent. I can add a source if needed.

### DIFF
--- a/Common/UnitDefinitions/Length.json
+++ b/Common/UnitDefinitions/Length.json
@@ -63,6 +63,7 @@
       },
       "FromUnitToBaseFunc": "{x} * 0.9144",
       "FromBaseToUnitFunc": "{x} / 0.9144",
+      "Prefixes":  [ "Kilo" ],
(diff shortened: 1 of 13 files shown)


angularsen:
Looks perfect to me!

Nuget should be out shortly.

[Release UnitsNet/5.39.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.39.0)

Issue #1337, opened 2023-12-10
New Unit - Grade (Slope)


mariuszhermansdorfer:
**Is your feature request related to a problem? Please describe.**
In architecture and landscape architecture, there are multiple ways of representing a [grade](https://en.wikipedia.org/wiki/Grade_%28slope%29):
![image](https://github.com/angularsen/UnitsNet/assets/49192999/bf02cbd4-d770-4f21-97f5-c363c257af68)

It can be defined as:
- an angle (26.6°)
- a percentage value (50%)
- a per mille value (500‰)
- a ratio written as a fraction (1/2)

![image](https://github.com/angularsen/UnitsNet/assets/49192999/49067df0-26e0-406b-9dd4-46f796b39da7)

**Describe the solution you'd like**
I'd like `UnitsNet` to group all the above representations within one `GradeUnit` to allow for seamless conversions. Currently, using the `RatioUnit`, it is possible to convert from a percent value to per mille, but angle is missing. Also, it'd be great to have support for fractions. similarly to how the `LengthUnit` supports feet and inches.

angularsen:
I think this sounds useful and has a wide range of applications.
Would you be interested in attempting a pull request?

Detailed steps here:
https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit

I'm happy to assist.

mariuszhermansdorfer:
Thanks @angularsen!

The description for a PR looks very manageable, I’ll gladly give it a shot. I could definitely use some help with the fraction part, though. 

Let me cook up a first sketch and submit it as a draft PR so that I can ask for more specific feedback.

mariuszhermansdorfer:
@angularsen,

I have a draft PR which passes all the tests except for this one:
![image](https://github.com/angularsen/UnitsNet/assets/49192999/2a5dff00-2c3e-4439-b952-d787dbad76d7)
Any ideas why this can be the case?

Also, I defined `Percent` as the main unit, do you agree with this approach, or would you prefer using `Fraction` as default?

```json
{
  "Name": "Grade",
  "BaseUnit": "Percent",
  "XmlDocSummary": "The grade, or slope, of a surface is the tangent of its angle compared to the horizontal, indicating its steepness. It's often expressed as a ratio or fraction of vertical rise to horizontal run, where a larger value means a steeper angle.",
  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Grade_%28slope%29",
  "BaseDimensions": {
[...]

PR #1342, opened 2023-12-22
Introduce conversion methods between Angle and Ratio

mariuszhermansdorfer:
@angularsen, following your suggestions from #1337, here is an updated PR adding the conversion methods between Angle and Slope.

How do you suggest we proceed with creating strings like "1 in 2, 1 in 5, 1:2, 1:5" etc. ?

### DIFF
--- a/UnitsNet/CustomCode/Quantities/Angle.extra.cs
+++ b/UnitsNet/CustomCode/Quantities/Angle.extra.cs
@@ -18,5 +18,15 @@ namespace UnitsNet
         {
             return RotationalSpeed.FromRadiansPerSecond(angle.Radians / duration.Seconds);
         }
+
+        /// <summary>
+        /// Convert an <see cref="Angle"/> to a slope <see cref="Ratio"/>.
+        /// </summary>
+        /// <returns>A <see cref="Ratio"/> representing the slope.</returns>
+        public Ratio ToSlopeRatio()
+        {
+            var slope = Math.Tan(this.Radians);
+            return Ratio.FromDecimalFractions(slope);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


mariuszhermansdorfer:
I could simply add the following method to Ratio.extra.cs:

```c#
    /// <summary>
    /// Converts the ratio to a string representing slope in fraction format.
    /// </summary>
    /// <param name="cultureInfo">The culture info to format the string. If null, the current culture is used.</param>
    /// <returns>A string representing the slope in the format "1 in X" or "1 : X".</returns>
    public string ToSlopeString(CultureInfo cultureInfo = null)
    {
        cultureInfo ??= CultureInfo.CurrentCulture;

        string slopeFormat = cultureInfo.Name == "en-US" ? "1 in {0}" : "1 : {0}";

        if (this.DecimalFractions == 0)
        {
            return string.Format(cultureInfo, slopeFormat, "∞");
        }

        var reciprocal = 1 / this.DecimalFractions;
        var roundedReciprocal = Math.Round(reciprocal);

        return string.Format(cultureInfo, slopeFormat, roundedReciprocal);
    }
```

angularsen:
I don't know what the correct term is, but I'd expect something like `ToRatioString()` to output on the format `1:3` for 1 to 3 ratio.

This could be a method just like you proposed above.
Then create some unit tests to verify the behavior for a range of values. 
For example, how does it work when the value is almost but not exactly `1:3`, like `1:2.999` ? I assume it would be represented as `1000:2999`?

angularsen:
There is some similar work here, for `Length.ToArchitecturalString()`:
https://github.com/angularsen/UnitsNet/blob/a95e33fafa466bf27706b54fd6ea9069a9dce97c/UnitsNet/CustomCode/Quantities/Length.extra.cs#L238-L314

PR #1343, opened 2023-12-24
Add the missing 4 SI prefixes.

adamradocz:
Sorurce: https://www.nist.gov/pml/owm/metric-si-prefixes

### DIFF
--- a/CodeGen/JsonTypes/Prefix.cs
+++ b/CodeGen/JsonTypes/Prefix.cs
@@ -1,8 +1,10 @@
-namespace CodeGen.JsonTypes
+﻿namespace CodeGen.JsonTypes
 {
     internal enum Prefix
     {
         // SI prefixes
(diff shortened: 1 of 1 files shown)


angularsen:
I mean, we can definitely add them, but I'd like to first investigate if we are at all able to represent round-trip conversions between units this large/small without too much precision loss due to using 64-bit `double` value type internally.

Are there any units in particular you are looking to add, that require these unit prefixes?

adamradocz:
Not really, just wanted to make it complete. But the double is not going to be enough, for those prefixes decimal type has to be used.

angularsen:
Let's hold off on this change then, we are for better or worse stuck with the limitations of `double` in the foreseeable future. 

I only want to add prefixes that will actually be used and that we have verified work with `double` for the units we define with those prefixes.

PR #915, opened 2021-04-08
Added Radioactivity

nforwood:
The Si unit for radioactivity is the Becquerel (Bq) which is one disintegration per second. The non-SI unit is the Curie (Ci) where 1 Curie = 3.7e10 Becquerel

PR #1345, opened 2023-12-28
Add radioactivity

piotrstenke:
Added Radioactivity (radioactive decay), along with three units: Bq - Becquerel (SI-derived), Ci - Curie and Rd - Rutherford (both non-SI). 

The scope of prefixes is fairly wide (Pico- to Exa- for Becquerel , Pico- to Tera- for Curie and Rutherford) in order to support most common scenarios, like measuring concentration of radioactive elements in a substance (e.g. 10 Bq per m3 of air is considered normal), rate of radioactive decay in electrical parts,  Geiger counter scales, cosmic ray interactions or even global inventory of specific radioactive elements (per [Wikipedia](https://en.wikipedia.org/wiki/Becquerel#Examples)). It does not, however, support prefixes for things like activity produced by [nuclear explosions](https://books.google.pl/books?id=CJqcq2C792UC&pg=PA74&redir_esc=y#v=onepage&q=hiroshima&f=false) (values beyond 10^24 Bq), since this seems too niche even for C# devs.

Similar PR was already created a quople years back (https://github.com/angularsen/UnitsNet/pull/915), but it was never actually merged.


piotrstenke:
If this PR gets accepted, I will continue with more radioactivity-related stuff. Here, feel warned.

angularsen:
Looks perfect to me, thanks! 

I think this is a good fit for the library, so any further work on radioactivity is probably also a good fit. Just keep in mind that the library generally tries to strike a balance between providing what most people need, while not adding too many esoteric or very specialized units that will only be useful to a handful of users.

angularsen:
Nuget should be out shortly

[Release UnitsNet/5.41.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.41.0)

PR #1347, opened 2024-01-02
Fix typo: Celcius -> Celsius

Muximize:
Came across this while working on #1329. I'm afraid it's an API-breaking change, so maybe it should wait for v6?


angularsen:
Nice catch.
Yes, this would be a breaking change on `FromXXX` factory methods among other things. Let's target v6 branch and merge it there. We don't have a good way of marking only the plural form as obsolete though, so people will find only be able to migrate when upgrading.

Muximize:
Sounds like a plan! I rebased this and targeted v6.

PR #1348, opened 2024-01-02
Add radiation exposure

piotrstenke:
Added radiation exposure along with two units - Coulomb per kilogram (SI-based) and Roentgen.

Coulombs per kilogram come with four prefixes (pico, nano, micro and milli), which is two less (with kilo and mega missing) than the definition of Coulomb itself in ElectricCharge.json. This is due to the fact that C/kg is quite large and there are not many practical usecases for anything beyond than that. 

Roentgens come with two prefixes (milli and micro). Most often than not, Roentgens are used 'as is', without prefixes, e.g. in case of the famous 20000R/h during Chernobyl disaster, but in some sources milli and micro are also used (example articles below).

https://www.epa.gov/radnet/about-exposure-and-dose-rates
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7155967/

Wikipedia pages:

https://en.wikipedia.org/wiki/Radiation_exposure
https://en.wikipedia.org/wiki/Roentgen_(unit)


UnitsNet.Tests/CustomCode/RadiationExposureTests.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public class RadiationExposureTests : RadiationExposureTestsBase
+    {
+        protected override bool SupportsSIUnitSystem => true;
+        protected override double CoulombsPerKilogramInOneCoulombPerKilogram => 1;
+        protected override double MillicoulombsPerKilogramInOneCoulombPerKilogram => 1e+3;
+        protected override double MicrocoulombsPerKilogramInOneCoulombPerKilogram => 1e+6;
+        protected override double NanocoulombsPerKilogramInOneCoulombPerKilogram => 1e+9;
+        protected override double PicocoulombsPerKilogramInOneCoulombPerKilogram => 1e+12;
+        protected override double RoentgensInOneCoulombPerKilogram => 3875.9689922;
```
angularsen, line 16: Verified: `3875.968992248`

https://www.translatorscafe.com/unit-converter/en-US/radiation-exposure/4-1/roentgen-coulomb/kilogram/

angularsen (approved):

angularsen:
Looks good to me!

[Release UnitsNet/5.42.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.42.0)

PR #1352, opened 2024-01-06
Add radiation equivalent dose

piotrstenke:
Added radiation equivalent dose and its two units - Sievert (SI-derived) and Roentgen equivalent man (or, simply, rem).

Sievert comes with three prefixes - milli, micro, nano. This will support 99% of usecases, ranging from radiation caused by Earth itself (~6  to 83nSv/h, per this [article](https://www.epa.gov/radnet/about-exposure-and-dose-rates)), [eating bananas](https://en.wikipedia.org/wiki/Banana_equivalent_dose) (yep, that's a real unit), X-rays or tomography scans, as well as measuring health risk of radiation-induced cancer and radiation sickness.

Roentgen equivalent man is mostly used in the US, and it can be easily converted to sieverts (1 sievert = 100 rems). It includes only one prefix - milli - as I did not find any sources with other prefixes being used.

Wikipedia pages:

https://en.wikipedia.org/wiki/Equivalent_dose
https://en.wikipedia.org/wiki/Sievert
https://en.wikipedia.org/wiki/Roentgen_equivalent_man
https://en.wikipedia.org/wiki/Orders_of_magnitude_(radiation)


piotrstenke:
The name of the quantity should also be discussed. UnitsNet already has another radiation dose-related quantity - AbsorbedDoseOfIonizingRadiation. If we are about to add the equivalent dose (and, potentialy, other dose quantities, like the [effective dose](https://en.wikipedia.org/wiki/Effective_dose_(radiation)), both names should follow the same convention. I suggest to go with Radiation{X}Dose, as it is simpler and much easier to read, as well as more in line with other already existing types (e.g. PressureChange, not ChangeOfPressure, ElectricSurfaceChargeDensity, not DensityOfElectricSurfaceCharge). Also, the word 'Ionizing' can be safely dropped here, as I don't believe there are any similar dose quantities for non-ionizing radiation that this quantity could be possibily confused with.

However, this approach requires the AbsorbedDoseOfIonizingRadiation to be marked as obsolete (and removed in later version) and re-implemented as a new type - RadiationAbsorbedDose, possibly with additional units (e.g. ergs per gram, as seen [here](https://www.ncbi.nlm.nih.gov/books/NBK230653/), roentgens equivalent physical (reps), [wikipedia](https://en.wikipedia.org/wiki/R%C3%B6ntgen_equivalent_physical)).

angularsen:
Sorry this fell off my radar.

> I suggest to go with Radiation{X}Dose, as it is simpler and much easier to read
> However, this approach requires the AbsorbedDoseOfIonizingRadiation to be marked as obsolete (and removed in later version) and re-implemented as a new type - RadiationAbsorbedDose, possibly with additional units

I like it, let's do it.

I added a TODO for vNext to rename/obsolete AbsorbedDoseOfIonizingRadiation.
#1200 

As far as I can tell, this PR is good to go. Good work!

angularsen:
Thanks, nuget should be out shortly.

[Release UnitsNet/5.46.0-pre · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.46.0-pre)

angularsen:
Note that this is a pre-release version, we are currently testing out some other bits before releasing as stable: #1363

lipchev:
> Sorry this fell off my radar.
> 
> > I suggest to go with Radiation{X}Dose, as it is simpler and much easier to read
> > However, this approach requires the AbsorbedDoseOfIonizingRadiation to be marked as obsolete (and removed in later version) and re-implemented as a new type - RadiationAbsorbedDose, possibly with additional units
> 
> I like it, let's do it.
> 
> I added a TODO for vNext to rename/obsolete AbsorbedDoseOfIonizingRadiation. #1200
> 
> As far as I can tell, this PR is good to go. Good work!

@angularsen What's the plan for v6 on this: clone + [Obsolete] or rename right away?

angularsen:
We can just rename it on v6 I think, there will be a breaking change guide based on #1200 that will help people migrate.

PR #1354, opened 2024-01-07
Inferred division from defined multiplication relations

Muximize:
See #1329


Muximize:
I rebased this on the latest changes in #1329 which makes it much cleaner.

Example on an interesting situation that occurs here, we previously defined:
```
MassFlow.GramPerSecond = Area.SquareMeter * MassFlux.GramPerSecondPerSquareMeter
MassFlux.KilogramPerSecondPerSquareMeter = MassFlow.KilogramPerSecond / Area.SquareMeter
```

If we infer division from multiplication, one of those has to add or lose a kilo 😅

angularsen:
Will review this next, hopefully in the next few days.

> If we infer division from multiplication, one of those has to add or lose a kilo 😅

Haha 😆 But it will still work, right?
The inferred division will just use grams instead?

Muximize:
Yes, it will work, except we're not sure why the original contributor chose a particular unit for an operator and if it matters, with precision for example.

Also, this PR is more of an experiment than a serious proposal. There's some questions about the "magic" of it, and I took some liberties with implicit/explicit conversion between `Density` and `MassConcentration` that I'm not quite sure about. Maybe @lipchev can chime in on this?

lipchev:
> Yes, it will work, except we're not sure why the original contributor chose a particular unit for an operator and if it matters, with precision for example.
> 
> Also, this PR is more of an experiment than a serious proposal. There's some questions about the "magic" of it, and I took some liberties with implicit/explicit conversion between `Density` and `MassConcentration` that I'm not quite sure about. Maybe @lipchev can chime in on this?

Although the Density and Mass Concentration share the same units, they are not interchangeable- they describe different types of ratios. Casting between them (either implicitly or explicitly) makes no sense..

Muximize:
Thanks for the explanation! That makes the concept of inferred division a bit harder, given that we currently have:
```
Mass.Kilogram = Density.KilogramPerCubicMeter * Volume.CubicMeter
Mass.Kilogram = MassConcentration.KilogramPerCubicMeter * Volume.CubicMeter
```

which leads to these inferred ambiguous operators, which is what I tried to solve with my casting shenanigans:
```
Density.KilogramPerCubicMeter           = Mass.Kilogram / Volume.CubicMeter
MassConcentration.KilogramPerCubicMeter = Mass.Kilogram / Volume.CubicMeter
```

The latter is currently not defined, so skipping it would not be a breaking change. I've reworked this PR to do that and removed the casting stuff.

CodeGen/Generators/QuantityRelationsParser.cs
```diff
@@ -59,6 +59,24 @@ public static void ParseAndApplyRelations(string rootDir, Quantity[] quantities)
...
+                    LeftUnit = r.ResultUnit,
+                    ResultQuantity = r.LeftQuantity,
+                    ResultUnit = r.LeftUnit,
+                })
+                // Skip division between equal quantities because the ratio is already generated as part of the Arithmetic Operators.
+                .Where(r => r.LeftQuantity != r.RightQuantity)
+                .ToList());
+            
+            // Remove inferred relation "MassConcentration = Mass / Volume" because it duplicates "Density = Mass / Volume"
+            relations.RemoveAll(r => r is { Operator: "/", ResultQuantity.Name: "MassConcentration", LeftQuantity.Name: "Mass", RightQuantity.Name: "Volume" });
```
angularsen, line 78: Maybe we can solve this in a way that makes it more discoverable by moving this into the JSON file, how about this:

### Option 1) Skip division if multiple competing definitions

If there more than one inferred division definition `QuantityA / QuantityB` with different resulting quantity type, then skip all of them. They will have to be manually defined in the JSON file.

### Option 2) Allow defining opt-out in the JSON file

We could move the array to a structure like this:
```json
{
  "skipInferredDivisionForDefinitions": [
    "Mass.Kilogram = MassConcentration.KilogramPerCubicMeter * Volume.CubicMeter"
  ],
  "definitions": [
    "1 = Area.SquareMeter * ReciprocalArea.InverseSquareMeter",
    "1 = ElectricResistivity.OhmMeter * ElectricConductivity.SiemensPerMeter",
    "1 = Length.Meter * ReciprocalLength.InverseMeter",
    "Acceleration.MeterPerSecondSquared = Jerk.MeterPerSecondCubed * Duration.Second",
    "AmountOfSubstance.Kilomole = MolarFlow.KilomolePerSecond * Duration.Second",
    "AmountOfSubstance.Mole = Molarity.MolePerCubicMeter * Volume.CubicMeter",
    "Mass.Kilogram = MassConcentration.KilogramPerCubicMeter * Volume.CubicMeter"
  ]
}
```

Or add a structure for each definition, allowing us to configure stuff for each entry:
 
```json
{
  "definitions": [
    { "d": "1 = Area.SquareMeter * ReciprocalArea.InverseSquareMeter" },
    { "d": "1 = ElectricResistivity.OhmMeter * ElectricConductivity.SiemensPerMeter" },
    { "d": "1 = Length.Meter * ReciprocalLength.InverseMeter" },
    { "d": "Acceleration.MeterPerSecondSquared = Jerk.MeterPerSecondCubed * Duration.Second" },
    { "d": "AmountOfSubstance.Kilomole = MolarFlow.KilomolePerSecond * Duration.Second" },
    { "d": "AmountOfSubstance.Mole = Molarity.MolePerCubicMeter * Volume.CubicMeter" },
    { "d": "Mass.Kilogram = MassConcentration.KilogramPerCubicMeter * Volume.CubicMeter", "skipInferredDivision": true }
  ]
}
```

UnitsNet/GeneratedCode/Quantities/Force.g.cs
```diff
@@ -705,12 +708,30 @@ public static bool TryParseUnit(string str, IFormatProvider? provider, out Force
...
+        public static ForceChangeRate operator /(Force force, Duration duration)
+        {
+            return ForceChangeRate.FromNewtonsPerSecond(force.Newtons / duration.Seconds);
+        }
+
+        /// <summary>Get <see cref="ForceChangeRate"/> from <see cref="Force"/> / <see cref="TimeSpan"/>.</summary>
+        public static ForceChangeRate operator /(Force force, TimeSpan timeSpan)
+        {
+            return ForceChangeRate.FromNewtonsPerSecond(force.Newtons / timeSpan.TotalSeconds);
+        }
```
angularsen, line 733: This may spawn a bigger discussion, but I'm thinking maybe we should clear up TimeSpan vs Duration a bit.

1. I like the idea of having `UnitsNet` using its own `Duration` as the primary type, and being consistent about it.
1. `Duration` already supports explicit cast to/from TimeSpan, let's change it to implicit cast.
2. Don't add duplicate method overloads in this PR just to support both types, implicit cast solves this.
3. I think we should also remove existing duplicate method overloads for TimeSpan/Duration, since we are doing a breaking change anyway.

I verified that with this change, then operator overloads also work with implicit cast of `TimeSpan`.

```cs
        [Fact]
        public void Foo()
        {
            Assert.Equal(ForceChangeRate.FromNewtonsPerSecond(100), Force.FromNewtons(100) / Duration.FromSeconds(1));
            Assert.Equal(ForceChangeRate.FromNewtonsPerSecond(100), Force.FromNewtons(100) / TimeSpan.FromSeconds(1));
        }
```

### Patch of `Duration` for implicit cast

```diff
diff --git forkSrcPrefix/UnitsNet/CustomCode/Quantities/Duration.extra.cs forkDstPrefix/UnitsNet/CustomCode/Quantities/Duration.extra.cs
index 4333d3c7450db4346bd224a8a0d152c075945e44..bac40311628a06a8700203e54f6557f166f8ee7e 100644
--- forkSrcPrefix/UnitsNet/CustomCode/Quantities/Duration.extra.cs
+++ forkDstPrefix/UnitsNet/CustomCode/Quantities/Duration.extra.cs
@@ -15,10 +15,7 @@ public partial struct Duration
         /// <returns>The TimeSpan with the same time as the duration</returns>
         public TimeSpan ToTimeSpan()
         {
-            if ( Seconds > TimeSpan.MaxValue.TotalSeconds ||
-                Seconds < TimeSpan.MinValue.TotalSeconds )
-                throw new ArgumentOutOfRangeException( nameof( Duration ), "The duration is too large or small to fit in a TimeSpan" );
-            return TimeSpan.FromTicks((long)(Seconds * TimeSpan.TicksPerSecond));
+            return TimeSpan.FromSeconds(Seconds);
         }
 
         /// <summary>Get <see cref="DateTime"/> from <see cref="DateTime"/> plus <see cref="Duration"/>.</summary>
@@ -33,14 +30,14 @@ public TimeSpan ToTimeSpan()
             return time.AddSeconds(-duration.Seconds);
         }
 
-        /// <summary>Explicitly cast <see cref="Duration"/> to <see cref="TimeSpan"/>.</summary>
-        public static explicit operator TimeSpan(Duration duration)
+        /// <summary>Implicitly cast <see cref="Duration"/> to <see cref="TimeSpan"/>.</summary>
+        public static implicit operator TimeSpan(Duration duration)
         {
             return duration.ToTimeSpan();
         }
 
-        /// <summary>Explicitly cast <see cref="TimeSpan"/> to <see cref="Duration"/>.</summary>
-        public static explicit operator Duration(TimeSpan duration)
+        /// <summary>Implicitly cast <see cref="TimeSpan"/> to <see cref="Duration"/>.</summary>
+        public static implicit operator Duration(TimeSpan duration)
         {
             return FromSeconds(duration.TotalSeconds);
         }
```

angularsen (changes requested):
I like where this is going and would like to get it merged.
Some comments so far.

UnitsNet/GeneratedCode/Quantities/ReciprocalLength.g.cs
```diff
@@ -630,11 +638,10 @@ public static bool TryParseUnit(string str, IFormatProvider? provider, out Recip
             return ReciprocalArea.FromInverseSquareMeters(left.InverseMeters * right.InverseMeters);
         }
 
-        /// <summary>Calculates the inverse of this quantity.</summary>
-        /// <returns>The corresponding inverse quantity, <see cref="Length"/>.</returns>
-        public Length Inverse()
```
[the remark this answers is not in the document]
angularsen: Maybe this was just this one file, you can disregard this.

UnitsNet.Tests/CustomCode/VolumeTests.cs
```diff
@@ -151,8 +151,8 @@ public void VolumeDividedByDurationEqualsVolumeFlow()
         [Fact]
         public void VolumeDividedByVolumeFlowEqualsTimeSpan()
         {
-            TimeSpan timeSpan = Volume.FromCubicMeters(20) / VolumeFlow.FromCubicMetersPerSecond(2);
-            Assert.Equal(TimeSpan.FromSeconds(10), timeSpan);
+            Duration duration = Volume.FromCubicMeters(20) / VolumeFlow.FromCubicMetersPerSecond(2);
+            Assert.Equal(Duration.FromSeconds(10), duration);
```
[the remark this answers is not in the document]
Muximize: Yes, that was my reasoning with this change 👍

UnitsNet/GeneratedCode/Quantities/ReciprocalLength.g.cs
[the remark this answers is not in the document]
Muximize: Not just this one, this happens in `Area`, `ReciprocalArea`, `Length` and `ReciprocalLength`.

But I think it's an improvement, `Inverse` is now always the first operator in the #region because it's now defined as
```
1 = Area.SquareMeter * ReciprocalArea.InverseSquareMeter
```
instead of
```
ReciprocalArea.InverseSquareMeter = 1 / Area.SquareMeter
```
and sorted on that string alphanumerically.

CodeGen/Generators/QuantityRelationsParser.cs
Muximize: I like option 2.2, but I also like the simple json structure we have now, so I propose a string-based solution of just appending `NoInferredDivision` to the relation, see last commits.

UnitsNet/GeneratedCode/Quantities/TemperatureDelta.g.cs
```diff
@@ -576,7 +590,7 @@ public static bool TryParseUnit(string str, IFormatProvider? provider, out Tempe
         /// <summary>Get <see cref="Length"/> from <see cref="TemperatureDelta"/> / <see cref="TemperatureGradient"/>.</summary>
         public static Length operator /(TemperatureDelta temperatureDelta, TemperatureGradient temperatureGradient)
         {
-            return Length.FromKilometers(temperatureDelta.Kelvins / temperatureGradient.DegreesCelsiusPerKilometer);
+            return Length.FromKilometers(temperatureDelta.DegreesCelsius / temperatureGradient.DegreesCelsiusPerKilometer);
         }
```
angularsen, line 594: This is from before, but isn't `PerKilometer` an odd choice here?
I'd expect `DegreesCelciusPerMeter`.
Same below.

Can fix in separate PR.

angularsen (approved):
Looks good to me, just one minor comment we can address later.

UnitsNet/GeneratedCode/Quantities/TemperatureDelta.g.cs
angularsen: @Muximize Just an extra ping to make sure you saw this:

This is from before, but isn't `PerKilometer` an odd choice here?
I'd expect `DegreesCelciusPerMeter`.
Same below.

Can fix in separate PR.

Muximize: I agree it seems odd, but it was added this way in #991. I guess this is used in meteorology, where kilometers make more sense?

The Kelvin part is weird tho. Previously we had:

```
TemperatureDelta.DegreeCelsius = TemperatureGradient.DegreeCelsiusPerKilometer * Length.Kilometer

TemperatureGradient.KelvinPerMeter = TemperatureDelta.Kelvin / Length.Meter
Length.Kilometer = TemperatureDelta.Kelvin / TemperatureGradient.DegreeCelsiusPerKilometer
```

We dropped the last two because they now get inferred, but with Celsius instead of Kelvin. I guess we need a domain expert to answer both there concerns?

angularsen:
Nuget on the way

[Release UnitsNet/6.0.0-pre003 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F6.0.0-pre003)

PR #1358, opened 2024-01-23
🐛Fix precision of Volume.CubicInch

### LINKED ISSUE #1357: Accuracy Issue when converting from CubicMeters to CubicInches
**Describe the bug**

When converting 10 cubicMeters into cubicInches, the result will be slightly different from the desired value as shown by the following unit test

image

this should be the correct value (it is possible to confirm it as well from other sources)

image

**To Reproduce**

image

angularsen:
Fixes #1357

An imprecise value was used, fixed by using the definition of inch as 2.54e-2 per meter.
Verified that the related unit CubicFoot is already precise.


angularsen:
[Release UnitsNet/5.43.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.43.0)

PR #1359, opened 2024-01-27
Remove decimal support

Muximize:
In #1195 @angularsen says:

This made me wonder how deep that complexity goes so I decided to experiment, and this is the result.

I must say some of these removals make me a bit sad. A lot of work and some very clever thinking went into supporting multiple numerical types, and I want to acknowledge that. 🙇 

Also, I took it as far as possible but that might not be the best outcome, for example we might want to keep deserialization support. This just demonstrates a possible direction we could go in.


angularsen:
I synced this with release/v6 branch after merging #1195 #1353 

Will do a review of this next.

Muximize:
I rebased this to clean up the commit history, which should aid in reviewing.

angularsen (approved):
Looks good to me! 

As you said, it's a bit sad to see a lot of thought and work being thrown out, but I still do think reducing this complexity is worth it until we find a better way to support multiple numeric types in a more holistic way.

I made a few minor improvements and pushed it.

PR #1360, opened 2024-02-04
💥🐛Change DataMember ordering to 1-indexed

### LINKED ISSUE #1356: Protobuf-Net support not working because of DataMember order 0
I tried transferring Quantities to a frontend via protobuf-net library. This library support DataContract and DataMember and therefore should easily work.

But:
The entry 0 is not a valid proto entry. So it does not work.

Any ideas on how to get it working?

See also:
https://stackoverflow.com/questions/1239295/stubborn-object-wont-serialize-with-protobuf-net

angularsen:
Related #1200 
Fixes #1356

Apparently, protobuf-net does not support 0-indexed order values.

Some investigation indicates that the actual order value is not important, only the relative ordering.
https://github.com/angularsen/UnitsNet/issues/1356#issuecomment-1900278438

### DIFF
--- a/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
@@ -99,7 +99,7 @@ namespace UnitsNet
                 Writer.WL(@$"
 #endif");
             }
-            
+
             if (_quantity.ValueType == "decimal") Writer.WL(@$"
         IDecimalQuantity,");
 
@@ -115,13 +115,13 @@ namespace UnitsNet
         /// <summary>
         ///     The numeric value this quantity was constructed with.
         /// </summary>
-        [DataMember(Name = ""Value"", Order = 0)]
+        [DataMember(Name = ""Value"", Order = 1)]
         private readonly {_quantity.ValueType} _value;
 
         /// <summary>
         ///     The unit this quantity was constructed with.
         /// </summary>
-        [DataMember(Name = ""Unit"", Order = 1)]
+        [DataMember(Name = ""Unit"", Order = 2)]
         private readonly {_unitEnumName}? _unit;
 ");
             GenerateStaticConstructor();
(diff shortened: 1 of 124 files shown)


angularsen:
@lipchev You originally added the data contract support. Do you have any insight on whether this change would be safe? The relative order is preserved, but it now starts with Order=1 instead of Order=0.

- WCF should tolerate this, according to its docs. https://github.com/angularsen/UnitsNet/issues/1356#issuecomment-1900278438
- Binary formatters hopefully only care about relative ordering and thus still compatible?
	- Protobuf-net, it never worked and should be OK
	- [BinaryFormatter](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter?view=net-8.0) is obsolete and generally not recommended for years, so hopefully no users are affected
	- A bunch of others exist too, haven't looked into how they handle this
- XML/JSON serializers should tolerate any order, can't imagine it breaking these?


@Brunni For your information

lipchev:
> @lipchev You originally added the data contract support. Do you have any insight on whether this change would be safe? The relative order is preserved, but it now starts with Order=1 instead of Order=0.
> 
>     * WCF should tolerate this, according to its docs. [Protobuf-Net support not working because of DataMember order 0 #1356 (comment)](https://github.com/angularsen/UnitsNet/issues/1356#issuecomment-1900278438)
> 
>     * Binary formatters hopefully only care about relative ordering and thus still compatible?
>       
>       * Protobuf-net, it never worked and should be OK
>       * [BinaryFormatter](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter?view=net-8.0) is obsolete and generally not recommended for years, so hopefully no users are affected
>       * A bunch of others exist too, haven't looked into how they handle this
> 
>     * XML/JSON serializers should tolerate any order, can't imagine it breaking these?
> 
> 
> @Brunni For your information

I'm not aware of any possible issues with this- personally I've only ever used the Order for display purposes (and have often had members start with an Order>0 without any issues).

angularsen:
@lipchev Thanks!

angularsen:
By all accounts and from what I've mustered to find in docs, it seems like this should be a safe change.
I'll release it as a pre-release on v5 to unblock you, but unless I hear any noise in the short term, it'll land in stable v5 soon.

PR #1362, opened 2024-02-13
Add VolumeFlow prefixes Deca, Hecto

### LINKED ISSUE #1253: VolumeFlowUnit: HectoliterPerTime as volume flow missing.
In the (beer) brewing sector the unit of hectoliter is used constantly. In the _VolumeUnit_ enum that unit is implemented, but is missing in the _VolumeFlowUnit_ enum.

It would be great if this unit could be implemented, since all machines we build use this as the base unit for volume flow.

If no one has the time to implement this feature, I would consider implementing it myself and therefore create a pull request once I implemented it. I'm just a total noob to this, since I have never contributed to any repo other then my own.

image

andrerodi:
Fixes #1253

Added the prefixes 'deca' and 'hecto', since, at least for the 'hecto'
prefix, it's a very common unit used in the brewing industry. Added the
'deca' prefix while I was at it.


angularsen:
Nuget should be out shortly. Note that it is a pre-release version, due to another recent change I can only publish pre-release versions until we have tested the other change a bit more.

[Release UnitsNet/5.45.0-pre · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.45.0-pre)

andrerodi:
@angularsen 
You're the best around
Nothing's gonna ever keep you down
- Joe Esposito

PR #1363, opened 2024-02-14
💥🐛Change DataMember ordering to 1-indexed (#1360)

### LINKED ISSUE #1356: Protobuf-Net support not working because of DataMember order 0

angularsen:
Backport of #1360 to v5 (master).

Related #1200
Fixes #1356

Apparently, protobuf-net does not support 0-indexed order values.

- Change `DataMember` explicit order from 0-indexed to 1-indexed

Some investigation indicates that the actual order value is not important, only the relative ordering: https://github.com/angularsen/UnitsNet/issues/1356#issuecomment-1900278438

- WCF should tolerate this, according to its docs. https://github.com/angularsen/UnitsNet/issues/1356#issuecomment-1900278438
- Binary formatters hopefully only care about relative ordering and thus still compatible?
	- Protobuf-net, it never worked and should be OK
	- [BinaryFormatter](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.serialization.formatters.binary.binaryformatter?view=net-8.0) is obsolete and generally not recommended for years, so hopefully no users are affected
	- A bunch of others exist too, haven't looked into how they handle this
- XML/JSON serializers should tolerate any order, can't imagine it breaking these?

### DIFF
--- a/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
@@ -82,13 +82,13 @@ namespace UnitsNet
         /// <summary>
         ///     The numeric value this quantity was constructed with.
         /// </summary>
-        [DataMember(Name = ""Value"", Order = 0)]
+        [DataMember(Name = ""Value"", Order = 1)]
         private readonly {_quantity.ValueType} _value;
 
         /// <summary>
         ///     The unit this quantity was constructed with.
         /// </summary>
-        [DataMember(Name = ""Unit"", Order = 1)]
+        [DataMember(Name = ""Unit"", Order = 2)]
         private readonly {_unitEnumName}? _unit;
 ");
             GenerateStaticConstructor();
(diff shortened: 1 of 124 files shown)


angularsen:
Nuget should be out shortly, note that it is a prerelease version so you need to enable that when updating your nuget version.

[Release UnitsNet/5.44.0-pre · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.44.0-pre)

angularsen:
@schalpat

Nuget should be out shortly, note that it is a prerelease version so you need to enable that when updating your nuget version.

[Release UnitsNet/5.44.0-pre · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.44.0-pre)

https://github.com/angularsen/UnitsNet/pull/1360#issuecomment-1940719687

PR #1365, opened 2024-02-19
Implicit cast between Duration and TimeSpan

Muximize:
See https://github.com/angularsen/UnitsNet/pull/1354#discussion_r1493344825

One issue is that the operator overloads only work when `TimeSpan` is the right operand.

I changed the code generation to take this into account, but another option would be to make a breaking change where we just don't support `TimeSpan` as the left operand at all.

Then users would have to cast explicitly, or for multiplication just reverse the operands.

This would affect 13 operators:

```
TimeSpan * Acceleration
TimeSpan * ElectricCurrent
TimeSpan * ElectricCurrentGradient
TimeSpan * ForceChangeRate
TimeSpan * KinematicViscosity
TimeSpan * MassFlow
TimeSpan * MolarFlow
TimeSpan * Power
TimeSpan * PressureChangeRate
TimeSpan * RotationalSpeed
TimeSpan * Speed
TimeSpan * TemperatureChangeRate
TimeSpan * VolumeFlow
```

Of which only 6 are used in tests so I assume are supported in v5:
```
TimeSpan * KinematicViscosity
TimeSpan * MassFlow
TimeSpan * Power
TimeSpan * RotationalSpeed
TimeSpan * TemperatureChangeRate
TimeSpan * Speed
```


Muximize:
Hmm, these failing CI tests pass on my machine. Any guidance?

angularsen:
@Muximize 
Beyond multiplication, are there any cases of `TimeSpan` as left hand argument?

I can't think of any, so I'd say let's just make a breaking change and not support it. 

For multiplication, they can switch order.
Otherwise, just do an explicit cast or maybe call an extension method `.ToDuration()` on it.

angularsen:
As for the failing test case, that was my bad in a previous merged PR.
Pushed a fix to this branch.

Muximize:
I completely removed `TimeSpan` from `Duration` inference, which is a nice cleanup.

Also removed some tests for `TimeSpan` as the left operand. They all still have a counterpart that tests the reverse.

UnitsNet.Tests/CustomCode/DurationTests.cs
```diff
@@ -33,20 +33,6 @@ public class DurationTests : DurationTestsBase
...
-            Assert.Throws<ArgumentOutOfRangeException>(() => duration.ToTimeSpan());
-        }
-
-        [Fact]
-        public static void ToTimeSpanShouldThrowExceptionOnValuesSmallerThanTimeSpanMin()
-        {
-            Duration duration = Duration.FromSeconds(TimeSpan.MinValue.TotalSeconds - 1);
-            Assert.Throws<ArgumentOutOfRangeException>(() => duration.ToTimeSpan());
-        }
-
```
angularsen, line 49: We can keep these tests, no? I would still expect an exception in these cases.

UnitsNet/CustomCode/Quantities/Duration.extra.cs
```diff
@@ -11,13 +11,9 @@ public partial struct Duration
         /// <summary>
         ///     Convert a Duration to a TimeSpan.
         /// </summary>
-        /// <exception cref="ArgumentOutOfRangeException">Throws if the TimeSpan can't represent the Duration exactly </exception>
         /// <returns>The TimeSpan with the same time as the duration</returns>
         public TimeSpan ToTimeSpan()
         {
-            if ( Seconds > TimeSpan.MaxValue.TotalSeconds ||
-                Seconds < TimeSpan.MinValue.TotalSeconds )
-                throw new ArgumentOutOfRangeException( nameof( Duration ), "The duration is too large or small to fit in a TimeSpan" );
```
angularsen, line 20: Oh, I guess this was here for a reason. It seems TimeSpan simply rolls over to negative if exceeding the max value. I don't like that, let's keep our checks in place and throw here.

UnitsNet.Tests/CustomCode/DurationTests.cs
angularsen: My bad, I removed some checks that seemed redundant but it appears TimeSpan simply rolls over to negative values if exceeding the max value.

I don't like that behavior, so I'd rather add these checks back.

angularsen (changes requested):
Another mistake on my part, I'll fix it.

UnitsNet/CustomCode/Quantities/Duration.extra.cs
angularsen: Added back

UnitsNet.Tests/CustomCode/DurationTests.cs
angularsen: Added back

PR #1374, opened 2024-03-02
Add type safety and improve type inference

UrielZyx:
Some of the extension methods in `UnitMath.cs` (e.g. Average) take an `Enum unitType` argument.
The compiler should be able to detect when the units type doesn't match the quantity type.

* This will break backwards compatibility, but only for:
  * People doing really weird things.
  * People using the explicit generic type instead of inference (e.g. `Average<Length>(LengthUnit.Inch)`), in which case they can fix the code by changing it to e.g. `Average<Length, LengthUnit>(LengthUnit.Inch)`
* This change might be warranted in `UnitConverter.cs` as well, but can't be implemented as a straight-forward refactor since it breaks compatibility in generated code (e.g. `unitConverter.SetConversionFunction<ElectricPotential>` in `ElectricPotential` should be `SetConversionFunction<ElectricPotential, ElectricPotentialUnit>`

In addition, had to remove a few unit tests that were asserting type safety.
All tests that were removed:
* Were only testing that an incorrect behavior throws an exception (`Assert.Throws`)
* Don't compile after the changes in `UnitMath.cs`

### DIFF
--- a/UnitsNet.Tests/UnitMathTests.cs
+++ b/UnitsNet.Tests/UnitMathTests.cs
@@ -46,27 +46,11 @@ namespace UnitsNet.Tests
         }
 
         [Fact]
-        public void AverageOfDifferentUnitsThrowsException()
-        {
-            var units = new IQuantity[] {Length.FromMeters(1), Volume.FromLiters(50)};
-
-            Assert.Throws<ArgumentException>(() => units.Average(LengthUnit.Centimeter));
-        }
-
-        [Fact]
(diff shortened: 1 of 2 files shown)


angularsen (approved):
Looks good to me!

angularsen:
Nuget on the way

[Release UnitsNet/6.0.0-pre003 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F6.0.0-pre003)

PR #1375, opened 2024-03-03
Upgrade to .NET 8

Muximize:
As we're doing breaking changes in v6, it might be worth upgrading to the current [Long Term Support](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) version of dotnet.

This enables us to use C# 12 features and remove a bunch of conditional compilation (net 5328 deletions)


CodeGen/Generators/UnitsNetGen/NumberExtensionsGenerator.cs
```diff
@@ -50,10 +47,7 @@ public static class NumberTo{_quantityName}Extensions
                 Writer.WLIfText(2, GetObsoleteAttributeOrNull(unit.ObsoleteText));
 
                 Writer.WL(2, $@"public static {_quantityName} {unit.PluralName}<T>(this T value)
-            where T : notnull
-#if NET7_0_OR_GREATER
-            , INumber<T>
-#endif
+            where T : notnull, INumber<T>
             => {_quantityName}.From{unit.PluralName}(Convert.ToDouble(value));
```
angularsen, line 51: We can upgrade CodeGen, tests and sample apps - but I think the generated code still needs to stay on `netstandard2.0`, which needs the pragma conditions for generic math interfaces.

I don't have any numbers, but I'm sure there are _many_ out there still stuck on older versions of .NET and I don't think it's wise to exclude all of them just for the sake of cleaning up a bit on our end. 😅

angularsen (changes requested):

CodeGen/Generators/UnitsNetGen/NumberExtensionsGenerator.cs
tmilnthorp: Agreed for public consumables. .NET framework is still shipped and supported as of Windows 11. NET6 is supported until later this fall.

Muximize: Yes, some folks are (or choose to be) stuck on .NET Framework, but if that is acceptable to them why would staying on UnitsNet 5.x be a problem? I guess most of these projects are not in active development but in maintenance mode anyway. We can choose to accommodate this according to the support timelines, but then we're stuck on .NET Standard 2.0 basically forever.

.NET 7 is supported until May 14, .NET 6 until November 12. Projects can (should) upgrade to .NET 8 at their leisure, after which they could upgrade to an eventually released UnitsNet 6 at their leisure.

angularsen: It is a fair point, but I just think that unless it is big burden for us to maintain netstandard2.0, then we can keep supporting it. We already support multi-targeting so those on net8 etc will still benefit from any extra features such as generic math, while everyone else gets `netstandard2.0` features as a minimum.

As far as I recall, it's just the generic math stuff that requires pragma statements.
The upside of dropping netstandard2 is very limited, the downside is potentially significant to a lot of users that will lose out on new units.

Muximize: Because new units are just data in json I was thinking for a few seconds we could maybe keep supporting v5 and v6 as different generated code from that, but then I realized that won't fly, like with the recent unit name fixes. And something like that would be way more complicated than the multi-targeting we do now 🙈 

I scaled back this PR, keeping `netstandard2.0` for public consumables and `net8.0` everywhere else. This drops `net6.0`, `net7.0` and `net48`.

I'm not completely sure about that last one, did that have a specific purpose in `UnitsNet.Tests` or `UnitsNet.Benchmark`?

angularsen: @Muximize It seems the benchmark project needs to have all the target frameworks we want to benchmark:

https://benchmarkdotnet.org/articles/configs/toolchains.html#multiple-frameworks-support
> If you want to test multiple frameworks, your project file MUST target all of them and you MUST install the corresponding SDKs:

angularsen: `Scripts\json-export-all-runtimes.bat` has `--runtimes net472 net48 netcoreapp2.1 netcoreapp3.1 net6.0 ^`, so I'm not sure if that was already broken or if net8 is backwards compatible?

It currently fails in the PR at least:
```
❯ ./Scripts/json-export-all-runtimes.bat
C:\Program Files\dotnet\sdk\8.0.102\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(266,5): error NETSDK1005:
Assets file 'X:\unitsnet\UnitsNet.Benchmark\obj\project.assets.json' doesn't have a target for 'net6.0'. Ensure that restore has run and th
at you have included 'net6.0' in the TargetFrameworks for your project. [X:\unitsnet\UnitsNet.Benchmark\UnitsNet.Benchmark.csproj]

The build failed. Fix the build errors and run again.
```

Muximize: Allright, and do we still want to benchmark all these obsolete targets, or just `net48` and `net8.0`?

angularsen:
Good point, I don't really know. The benchmarks are not actively used so we
can revisit later. Let's go with net48 and net8 for now. No need to fix all
the scripts, we can address that whenever benchmarking comes up again.

angularsen:
[Release UnitsNet/6.0.0-pre004 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F6.0.0-pre004)

PR #1381, opened 2024-03-27
👔Change nuget license from MIT to MIT-0

### LINKED ISSUE #1379: NuGets license: MIT instead of MIT-0
The declared license for the nugets is MIT, but the license in the sources is MIT-0.
I see in the issues history that there were some problems with releasing the nugets with the MIT-0 license. There was even a try to use the MIT-0 license again, but not successful. It was quite a long time ago :).
Could you check if the MIT-0 license can be used now in the nugets? If so, then could you set this license? It would be great to have the correct license defined in the nugets for legal purposes, i.e. the list of used open source components :).

angularsen:
Fixes #1379
Backport from v6: #1380

Change the license expression in the nugets to match the LICENSE file. Previously, nuget.org did not accept MIT-0 expression.


angularsen:
Giving v6 a bit of time before doing the same for v5.
Revisit in a week or so.

angularsen:
[Release UnitsNet/5.50.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.50.0)

PR #1383, opened 2024-04-04
Change license from MIT to MIT-0 in NuspecGenerator

Muximize:


angularsen:
Good catch, thanks!

Muximize:
I guess #1381 for v5 needs the same.

PR #1386, opened 2024-04-04
Enable Central Package Management

Muximize:
This PR removes some seemingly unused dependencies, and moves all dependency versions to a single file according to [Central Package Management](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management).

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -11,9 +11,6 @@
     <PackageVersion Include="Serilog.Sinks.Console" Version="5.0.1" />
     <PackageVersion Include="System.CommandLine.DragonFruit" Version="0.4.0-alpha.22272.1" />
     <PackageVersion Include="xunit" Version="2.7.0" />
-    <PackageVersion Include="xunit.runner.visualstudio" Version="2.5.7">
-      <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-    </PackageVersion>
+    <PackageVersion Include="xunit.runner.visualstudio" Version="2.5.7" />
   </ItemGroup>
 </Project>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,19 @@
+<Project>
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageVersion Include="BenchmarkDotNet" Version="0.13.12" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageVersion Include="NuGet.Protocol" Version="6.9.1" />
+    <PackageVersion Include="Serilog" Version="3.1.1" />
+    <PackageVersion Include="Serilog.Sinks.Console" Version="5.0.1" />
+    <PackageVersion Include="System.CommandLine.DragonFruit" Version="0.4.0-alpha.22272.1" />
+    <PackageVersion Include="xunit" Version="2.7.0" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="2.5.7">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageVersion>
+  </ItemGroup>
+</Project>
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -4,13 +4,13 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageVersion Include="BenchmarkDotNet" Version="0.13.12" />
-    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
+    <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.10.0" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
-    <PackageVersion Include="NuGet.Protocol" Version="6.9.1" />
-    <PackageVersion Include="Serilog" Version="3.1.1" />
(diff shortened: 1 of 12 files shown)


Directory.Packages.props
```diff
@@ -0,0 +1,19 @@
...
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
+    <PackageVersion Include="NuGet.Protocol" Version="6.9.1" />
+    <PackageVersion Include="Serilog" Version="3.1.1" />
+    <PackageVersion Include="Serilog.Sinks.Console" Version="5.0.1" />
+    <PackageVersion Include="System.CommandLine.DragonFruit" Version="0.4.0-alpha.22272.1" />
+    <PackageVersion Include="xunit" Version="2.7.0" />
+    <PackageVersion Include="xunit.runner.visualstudio" Version="2.5.7">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageVersion>
```
angularsen: I believe `PrivateAssets`, `IncludeAssets` etc belongs to `PackageReference` and not `PackageVersion`. I pushed a fix for this.

UnitsNet.Serialization.JsonNet/UnitsNet.Serialization.JsonNet.csproj
```diff
@@ -27,18 +27,6 @@
...
-    <PublishRepositoryUrl>true</PublishRepositoryUrl>
-
-    <!-- Optional: Embed source files that are not tracked by the source control manager in the PDB -->
-    <EmbedUntrackedSources>true</EmbedUntrackedSources>
-
-    <!-- Optional: Build symbol package (.snupkg) to distribute the PDB containing Source Link -->
-    <IncludeSymbols>true</IncludeSymbols>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-  </PropertyGroup>
-
```
angularsen, line 41: Aha, it seems .NET8 includes these by default now. Nice.

> Starting with .NET 8, Source Link for the following source control providers is included in the .NET SDK and enabled by default

https://github.com/dotnet/sourcelink


However, we target `netstandard2.0` in this project, so I guess we need to keep those? 

Including the packagereference `Microsoft.SourceLink.GitHub` below.

UnitsNet/UnitsNet.csproj
```diff
@@ -27,18 +27,6 @@
...
-  <PropertyGroup>
-    <PublishRepositoryUrl>true</PublishRepositoryUrl>
-
-    <!-- Optional: Embed source files that are not tracked by the source control manager in the PDB -->
-    <EmbedUntrackedSources>true</EmbedUntrackedSources>
-
-    <!-- Optional: Build symbol package (.snupkg) to distribute the PDB containing Source Link -->
-    <IncludeSymbols>true</IncludeSymbols>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-  </PropertyGroup>
```
angularsen, line 40: Same here, I think we need to keep this

angularsen (changes requested):
A few things to look over

UnitsNet.Serialization.JsonNet/UnitsNet.Serialization.JsonNet.csproj
Muximize: The way I understand this is that these settings and packagereference are only used at build/package time, not at runtime, so as long as UnitsNet is build with the .NET 8 SDK it should work with `netstandard2.0`.

I want to validate this but have some trouble. Even if I build the latest master unmodified with the `dotnet build` and `dotnet pack` commands from `Build/build-functions.psm1` the resulting `.nupkg` has missing symbols and compiler flags when checked with https://nuget.info

Muximize: @angularsen any thoughts on this? Am I doing something wrong? Does this only work in CI?

angularsen: @Muximize CI should not do anything different I can think of.

As far as I can tell, the symbols package `.snupkg` includes the `.pdb` file as expected and `.nupkg` includes source code and has valid SourceLink for the commit I had currently checked out: `91a85e2dac60b420fabc44ebde71fc85dab33d6c`

```
git checkout origin/master # Where origin is the original repository
cd UnitsNet
dotnet pack
```

>  UnitsNet -> X:\unitsnet\Artifacts\UnitsNet\net7.0\UnitsNet.dll
>  Successfully created package 'X:\unitsnet\Artifacts/UnitsNet\UnitsNet.5.50.0.nupkg'.
>  Successfully created package 'X:\unitsnet\Artifacts/UnitsNet\UnitsNet.5.50.0.snupkg'.


**Testing the `.nupkg` file on nuget.info**
image

**Testing the `.snupkg` symbols file on nuget.info** 
image

Muximize: Alright, I finally looked into this. Turns out `IncludeSymbols` and `SymbolPackageFormat` are indeed needed after al, so I restored that but as arguments to `dotnet pack` in `build-functions.psm1`. This also enables sourcelink for the NumberExtensions project without needing per-project settings.

The packagereference is not needed, the symbols and sourcelink are also generated for `netstandard2.0` if build with the .NET 8 SDK.

angularsen: Is it not better to keep it in csproj? Or for consistency, move it to `Directory.Build.props` at the repo root?

This way, it packs nugets the same way regardless of doing it with scripts or built-in tools in IDEs.

angularsen:
Thanks! I'll bump the version so we can get some new nugets produced via CI pipeline and verify it works.

angularsen:
https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F6.0.0-pre007

Muximize:
> Thanks! I'll bump the version so we can get some new nugets produced via CI pipeline and verify it works.

Seems to work fine 👍

angularsen:
Great! 👏

PR #1388, opened 2024-05-10
Add Mass Distance Quantity used in transportation industry

rowens-an:
MassDistance is a unit of measure in the transportation industry, particularly with freight transportation.  There are three primary Units that are used in the industry

- Kilogram - Kilometer
- Tonne - Kilometer
- Ton-Mile 

In general, this Quantity is for measuring the amount of freight transported across a distance.  We are using it for carbon emissions calculations, particularly t*km and t*mi. 

More information can be found on [Wikipedia](https://en.wikipedia.org/wiki/Units_of_measurement_in_transportation#Freight)


rowens-an:
As a note, I wasn't sure what to call this Quantity.  I originally thought it would be useful to call it `PayloadDistance` but settled on `MassDistance` as it fits the naming convention a bit more.

angularsen:
Hi, this feels too domain specific for UnitsNet, sorry. 

It is a balance, but we generally try to only bring in units that are widely used and preferably across multiple domains.

PR #1391, opened 2024-06-04
Add quantity level to carrier

AnOtherMuenchner:
Level To Carrier with unit dBc is widely used in telecommunications and signal processing. Although, the calucalation formula is simular to Level/dB, it is used and treated as an own quantity.

For more information see https://en.wikipedia.org/wiki/DBc


angularsen:
Hi, thanks for putting in the effort, but without knowing this domain it seems to me maybe this quantity falls short on our criteria list:

https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#great-but-before-you-start

In short, I'm not convinced it adds enough value for enough people to justify increasing the binary size for everyone. I think it's better to add this quantity per application for now, until UnitsNet maybe gets a way to offer quantities in individual nugets.

PR #1396, opened 2024-06-16
🐛ComparisonType ignored in obsolete Equals method

### LINKED ISSUE #1394: QuantityGenerator ignores ComparisonType in the obsolete Equals method
We are in the process of migrating UnitsNet to the latest and found that some of our obsolete equals methods were failing.  Inspecting the generated code shows that the comparison type is ignored.

https://github.com/angularsen/UnitsNet/blob/ce42aa89a7a0cac6f0f77ad28672f0769ac2376d/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs#L872

Seems like the comparison class still supports it, any reason to not keep it backwards compatible?

https://github.com/angularsen/UnitsNet/blob/ce42aa89a7a0cac6f0f77ad28672f0769ac2376d/UnitsNet/Comparison.cs#L60

lipchev:
fixes #1394 

note that the error is also present in the v6 branch

### DIFF
--- a/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
@@ -852,7 +852,7 @@ namespace UnitsNet
                 referenceValue: this.Value,
                 otherValue: other.As(this.Unit),
                 tolerance: tolerance,
-                comparisonType: ComparisonType.Absolute);
+                comparisonType: comparisonType);
         }}
 
         /// <inheritdoc />
--- a/CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
@@ -553,6 +553,8 @@ namespace UnitsNet.Tests
             var v = {_quantity.Name}.From{_baseUnit.PluralName}(1);
             Assert.True(v.Equals({_quantity.Name}.From{_baseUnit.PluralName}(1), {_baseUnit.PluralName}Tolerance, ComparisonType.Relative));
             Assert.False(v.Equals({_quantity.Name}.Zero, {_baseUnit.PluralName}Tolerance, ComparisonType.Relative));
+            Assert.True({_quantity.Name}.From{_baseUnit.PluralName}(100).Equals({_quantity.Name}.From{_baseUnit.PluralName}(120), ({_quantity.ValueType})0.3m, ComparisonType.Relative));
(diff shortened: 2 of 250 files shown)


angularsen:
> note that the error is also present in the v6 branch

I thought we removed these methods in v6? Aren't they marked obsolete?

lipchev:
Doesn't seems like it- I only looked at my local version (with the fractions) and see that I've just removed the [Obsolete] annotation (as the operation would be safe), but hadn't noticed the other issue..

angularsen:
[Release UnitsNet/5.52.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.52.0)

lipchev:
> Doesn't seems like it- I only looked at my local version (with the fractions) and see that I've just removed the [Obsolete] annotation (as the operation would be safe), but hadn't noticed the other issue..

Actually I must have been looking at one of the other overloads, cause the [Obsolete] annotation is apparently still there and as looked at it a little more closely, realized that there is in fact a good reason to remove it:
without knowing what unit the two quantities are in, there is no possible use of the absolute tolerance comparison 😄 

The two overloads (if we need to have them inside the quantity) should be something like `Equals(Density other, Density absoluteTolerance)` and `Equals(Density other, Ratio relativeTolerance)`

angularsen:
Will handle this in #1200 when reviewing all remaining `[Obsolete]` code. Added an action item for it.

lipchev:
I have a working version of #1377 with 0 warning and 0 tests skipped (14 actually but those are expected), with all of the above mentioned changes as well as all [Obsolete] cases handled and ready for review. Everything is looking very good, but I just wanted to test out the nuget with my own code-base first. However I'm going to be busy this week so, decide it to postpone it for the next (weekend hopefully). 

How do you want to handle it? All in one, or some part of it (like this PR) + the rest? I was hoping that we'd get at least some of these `json`s merged (perhaps skipping those that cause a test to fail)- so that it would be easier to review my changes later on.

angularsen:
OK I'll have to get into #1377 then, it's been looming over me for a long time. Started vacation now, so I may finally find some time for it.

lipchev:
Yeah, but don't look at #1377, it's obsolete.. I'll have another one soon.

What i did for this particular [Obsolete] Equals overload was to removed it- as introducing the overload with `Ratio` would have introduced an ambiguity within the `Ratio` quantity and I didn't want to deal with that. 

The user can either use the other overload (with the absolute tolerance) by multiplying the quantity by a ratio, or switch to directly using the `QuantityComparison` class.

angularsen:
OK, let me know when you have something I can look at

PR #1409, opened 2024-07-11
Add `NumberConcentration` unit

### LINKED ISSUE #1260: ReciprocalVolume?

ricardoboss:
Closes #1260


angularsen:
Hi, sure I think we can add this.

However we have a naming consistency issue. From before we have:
https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/ReciprocalLength.json
https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/ReciprocalArea.json

For consistency, we should maybe name this ReciprocalVolume.

However, https://en.wikipedia.org/?title=Number_concentration&redirect=no redirects to https://en.wikipedia.org/wiki/Number_density, which defines names for all 3:
- Volumetric number density
- Areal number density
- Linear number density

I think these names are better, because google does not really find much on reciprocal area or reciprocal volume.

### Proposal
1. Rename in this PR to `VolumetricNumberDensity`
2. Mark `ReciprocalLength` and `ReciprocalArea` as obsolete, to be replaced with `ArealNumberDensity` and `LinearNumberDensity` in #1200 

What do you think?

Pinging @bc913 that added ReciprocalLength and ReciprocalArea.

bc913:
Imho, it'd be better to keep existing `ReciprocalLength` and `ReciprocalArea` and implementing the new one as `ReciprocalVolume` to keep consistency. Some quantities might be named and frequently used in various forms within different fields of physics. In order to reduce confusion and achieve consistency, I think we should stick to mathematical definitions with easy to infer synonyms if there are multiple names for the same quantity and/or unit.

For instance, when you write code and make use of `ReciprocalLength` is more convenient and self explanatory for the developer/maintainer etc. Using `LinearNumberDensity` most likely would force the reader/developer to search for the meaning and so on.

Muximize:
Users who prefer different names in their project could use aliases:
```
global using LinearNumberDensity = UnitsNet.ReciprocalLength;
global using Voltage = UnitsNet.ElectricPotential;
```

PR #1416, opened 2024-08-21
Add `TonsOfRefrigeration` Unit

jasekiw:
This is the US standard used for rating refrigeration machinery.

### DIFF
--- a/Common/UnitDefinitions/Power.json
+++ b/Common/UnitDefinitions/Power.json
@@ -112,6 +112,19 @@
           "Abbreviations": [ "J/h" ]
         }
       ]
+    },
+    {
+      "SingularName": "TonOfRefrigeration",
+      "PluralName": "TonsOfRefrigeration",
+      "FromUnitToBaseFunc": "{x} * 3516.8528420667m",
+      "FromBaseToUnitFunc": "{x} / 3516.8528420667m",
+      "Prefixes": [ ],
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [ "TR" ]
+        }
+      ]
     }
   ]
 }
--- a/Common/UnitDefinitions/Power.json
+++ b/Common/UnitDefinitions/Power.json
@@ -116,8 +116,8 @@
     {
       "SingularName": "TonOfRefrigeration",
       "PluralName": "TonsOfRefrigeration",
-      "FromUnitToBaseFunc": "{x} * 3516.8528420667m",
-      "FromBaseToUnitFunc": "{x} / 3516.8528420667m",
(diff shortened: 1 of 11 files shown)


Common/UnitDefinitions/Power.json
```diff
@@ -112,6 +112,19 @@
           "Abbreviations": [ "J/h" ]
         }
       ]
+    },
+    {
+      "SingularName": "TonOfRefrigeration",
+      "PluralName": "TonsOfRefrigeration",
+      "FromUnitToBaseFunc": "{x} * 3516.8528420667m",
+      "FromBaseToUnitFunc": "{x} / 3516.8528420667m",
```
angularsen, line 120: According to wiki, it exactly equals `3.51685 kW`. Do you have a definition to base this on?
https://en.wikipedia.org/wiki/Ton_of_refrigeration

> The modern definition is exactly 12,000 [BtuIT/h](https://en.wikipedia.org/wiki/Power_(physics)) (3.516853 [kW](https://en.wikipedia.org/wiki/Kilowatt))

angularsen (changes requested):

Common/UnitDefinitions/Power.json
jasekiw: I found that value from this table at [unitconverters.net](https://www.unitconverters.net/power/ton-refrigeration-to-watt.htm)
image

Since it started with the same numbers I assumed it was a more precise version of the one listed on wikipedia and the number was irrational.

Digging further from the [source listed from wikipedia](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication811e2008.pdf) I can see it is indeed listed the same as the wikipedia so I will go ahead and change it to that.

image

jasekiw: I just pushed the adjustment. Let me know if there is anything else. Thanks!

angularsen: Thanks, yes I have found that online converters are not always very reliable on precise values. They probably do some calculations that introduce floating point errors.

Wiki is my goto source, since you can typically find the source they refer to.

angularsen:
Thanks! Nuget should be out in the next hour.
[Release UnitsNet/5.57.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.57.0)

angularsen:
Also
[Release UnitsNet/6.0.0-pre012 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F6.0.0-pre012)

PR #1419, opened 2024-08-29
add EnergyEfficiency unit

Chef904:
add unit for energy consumption from electric cars.

Wh/km


Common/.DS_Store
angularsen, line 1: Remove mac files, maybe add to the repo's `.gitignore` file to avoid this

Common/UnitDefinitions/EnergyEfficiency.json
```diff
@@ -0,0 +1,42 @@
+{
+  "Name": "EnergyEfficiency",
```
angularsen, line 2: What domain is this, is it related to transport?
If so, it feels a bit too domain specific for UnitsNet maybe.
Can you help me understand the domain and usecases better?

Here are some criteria we have to consider before adding new units:
https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#great--but-before-you-start-

angularsen (changes requested):

Common/UnitDefinitions/EnergyEfficiency.json
Chef904: Hello,
it's like the FuelEfficiency Quantity but related to electric powered vehicles. This domain is getting more and more important these days.

angularsen: Right, it is getting kind of mainstream so maybe it makes sense to add something here.
However, maybe we just need to name it something else. 

### EnergyEfficiency means?
Googling [energy efficiency units](https://www.google.com/search?q=energy+efficiency+units) returns a lot of different things, from energy generators, heating/cooling, and more. Not much about EV efficiency in the first few results.

### Rename EnergyEfficiency to...?
It seems there is no well defined name for this, but electric vehicle converters do pop up for these units.

"Electric car energy economy"
https://www.aqua-calc.com/one-to-all/electric-car-energy-economy

"Electric Vehicle Range / Efficiency Unit Conversions"
https://home.hasslers.net/ev/convert/

I propose `ElectricVehicleEfficiency`, since there does not seem to be a well defined, generic physical quantity with units like kWH/km. They are all domain specific; heating/cooling, energy generators, etc.

### Maybe also add kWh/100km unit?
> Most common in Europe is the specification kilowatt-hours per 100 kilometers (kWh/100km).
https://e-automobile.info/en/calculator.php

### Summary of proposed changes to PR
- Rename quantity to `ElectricVehicleEfficiency`
- Add unit `WattHourPer100Kilometer` with prefix Kilo

angularsen (changes requested):
Thanks, see below comments.

PR #1431, opened 2024-10-23
Fix a bug in BaseDimensions.ToString()

ygorshkov:
Current implementation of `AppendDimensionString` ignores value `-1`.
This PR fixes the bug.


angularsen:

[Release UnitsNet/5.61.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.61.0)

PR #1444, opened 2024-12-09
Add/rename electric quantities

McNeight:
Renamed quantities to have prefix "Electric", marking old name as obsolete and adding new duplicate quantities:
* `ApparentEnergy` -> `ElectricApparentEnergy`
* `ApparentPower` -> `ElectricApparentPower`
* `Capacitance` -> `ElectricCapacitance`
* `ReactiveEnergy` -> `ElectricReactiveEnergy`
* `ReactivePower` -> `ElectricReactivePower `
* `ElectricPotentialAc`
* `ElectricPotentialDc`
* `ElectricAdmittance` (obsolete due to being a complex number, not supported by UnitsNet, recommending `ElectricReactance` and `ElectricSusceptance` instead) 
 
Decided against adding `ElectricImpedance`, also being a complex number. 

New:
- Add `ElectricReactance`
- Add `ElectricSusceptance`
- Add `EletricCurrent = ElectricApparentPower / ElectricPotential`
- Add `ElectricPotential = ElectricApparentPower / ElectricCurrent`
- `ElectricAdmittance`: 
  - Add prefixes: Kilo, Mega, Giga, Tera
  - Add Mho unit, equivalent to Siemen
- `ElectricConductance`:
  - Add prefixes: Mega, Giga, Tera
  - Add Mho unit, equivalent to Siemen
- `EletricResistance`:
  - Add prefixes: Nano


Common/UnitDefinitions/Capacitance.json
```diff
@@ -3,6 +3,7 @@
   "BaseUnit": "Farad",
   "XmlDocSummary": "Capacitance is the ability of a body to store an electric charge.",
   "XmlDocRemarks": "https://en.wikipedia.org/wiki/Capacitance",
+  "ObsoleteText": "This class is obsolete. Please use ElectricCapacitance instead",
```
angularsen, line 6: Minor, but I think it would be helpful if the text said it was renamed from X to Y so it is clear that only the name changed and they are equivalent.

Common/UnitDefinitions/ElectricPotentialAc.json
```diff
@@ -1,7 +1,8 @@
 {
   "Name": "ElectricPotentialAc",
   "BaseUnit": "VoltAc",
-  "XmlDocSummary": "The Electric Potential of a system known to use Alternating Current.",
+  "XmlDocSummary": "If you want to map more parameters into the ElectricPotential class (volts RMS, phase angle, etc.), just use a named tuple.",
+  "ObsoleteText": "This class is redundant. Please use ElectricPotential instead.",
```
angularsen, line 5: I would keep the XmlDocSummary as before, and merge the new text into the ObsoleteText instead - since this is what's thrown in your face by intellisense and build warnings.

E.g. `ElectricPotentialAc was renamed to ElectricPotential, and will be removed in a later major version. If you want to map more parameters into the ElectricPotential class (volts RMS, phase angle, etc.), create your own wrapper type such as a record or named tuple.`

angularsen (changes requested):
Not sure if this was ready to review yet, but a few minor comments about the texts.

Common/UnitDefinitions/ElectricPotentialAc.json
angularsen: This text about named tuple would be helpful for the new types as well, not just the obsoleted ones.

Common/UnitDefinitions/Capacitance.json
McNeight: Does the ObsoleteText tag also parse XML comments? For example, would something like this work:
"<see cref="Capacitance" /> has been renamed to <see cref="ElectricCapacitance"> and marked as obsolete. Please change your code to use the new API before the next major revision of UnitsNet."

lipchev: > Does the ObsoleteText tag also parse XML comments? For example, would something like this work:
> 
> ```
> "<see cref="Capacitance" /> has been renamed to <see cref="ElectricCapacitance"> and marked as obsolete. Please change your code to use the new API before the next major revision of UnitsNet."
> ```

I don't think so- there is something about an [`UrlFormat`](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=net-9.0) property that could probably serve a similar purpose but I've never used it myself..

McNeight: > I don't think so- there is something about an [`UrlFormat`](https://learn.microsoft.com/en-us/dotnet/api/system.obsoleteattribute?view=net-9.0) property that could probably serve a similar purpose but I've never used it myself..

Yeah, that doesn't look like a good substitute. I'll confine the XML to the XmlDoc* tags for now, and will add it back if another method is found.

McNeight:
Should I make a second PR against default/v6, or would the code change here be merged into the next version as well?

angularsen:
@McNeight It will eventually be merged into v6 when syncing master into v6, but it is very helpful if you coulg do a separate PR for v6 simply to reduce the amount of merge conflicts 🙏 

Let's hold until this is merged though, is it ready to review now?

McNeight:
Not a problem!

I believe it's ready to review for v5, if you want to mark ElectricPotentialAc and Dc as obsolete in v5.

If you wanted to save the breaking change for v6, then I'll need to rework the commits a pull requests, but the content should be ready for review.

Common/UnitDefinitions/ElectricAdmittance.json
```diff
@@ -1,7 +1,9 @@
 {
   "Name": "ElectricAdmittance",
   "BaseUnit": "Siemens",
-  "XmlDocSummary": "Electric admittance is a measure of how easily a circuit or device will allow a current to flow. It is defined as the inverse of impedance. The SI unit of admittance is the siemens (symbol S).",
+  "XmlDocSummary": "Electric admittance is a measure of how easily a circuit or device will allow a current to flow by the combined effect of conductance and susceptance in a circuit. It is defined as the inverse of impedance. The SI unit of admittance is the siemens (symbol S).",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Electrical_admittance",
+  "ObsoleteText": "Admittance is a complex number, which is not currently supported by UnitsNet. Please use either ElectricConductance or ElectricSusceptance instead.",
```
angularsen, line 6: I see ElectricAdmittance is being obsoleted in favor of ElectricConductance and ElectricSusceptance. This is probably reasonable, but I don't know this domain well enough to say if this is a long term or short term solution. 

Is it likely we would add back EletricAdmittance at a later point?
If so, what would supporting imaginary numbers look like in UnitsNet?

Common/UnitDefinitions/ElectricImpedance.json
```diff
@@ -0,0 +1,28 @@
+{
+  "Name": "ElectricImpedance",
+  "BaseUnit": "Ohm",
+  "XmlDocSummary": "Electric impedance is the opposition to alternating current presented by the combined effect of resistance and reactance in a circuit. It is defined as the inverse of admittance. The SI unit of impedance is the ohm (symbol Ω).",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Electrical_impedance",
+  "ObsoleteText": "Impedance is a complex number, which is not currently supported by UnitsNet. Please use either ElectricResistance or ElectricReactance instead.",
```
angularsen, line 6: Should we still add this, if it's being obsoleted out of the box?

CodeGen/Generators/UnitsNetGen/NumberExtensionsGenerator.cs
```diff
@@ -47,7 +42,8 @@ public static class NumberTo{_quantityName}Extensions
                 Writer.WL(2, $@"
 /// <inheritdoc cref=""{_quantityName}.From{unit.PluralName}(UnitsNet.QuantityValue)"" />");
 
-                Writer.WLIfText(2, GetObsoleteAttributeOrNull(unit.ObsoleteText));
+                // Include obsolete text from the quantity per extension method, to make it visible when the class is not explicitly referenced in code.
+                Writer.WLIfText(2, GetObsoleteAttributeOrNull(unit.ObsoleteText ?? _quantity.ObsoleteText));
```
angularsen, line 46: Updated codegen to also mark number extension methods as obsolete.

angularsen (changes requested):
This is looking pretty good now, my only remaining comments are regarding `ElectricAdmittance` and `ElectricImpedance`.

Common/UnitDefinitions/ElectricReactiveEnergy.json
```diff
@@ -0,0 +1,25 @@
+{
+  "Name": "ElectricReactiveEnergy",
+  "BaseUnit": "VoltampereReactiveHour",
+  "XmlDocSummary": "The volt-ampere reactive hour (expressed as varh) is the reactive power of one Volt-ampere reactive produced in one hour.",
+  "BaseDimensions": {
+    "L": 2,
+    "M": 1,
+    "T": -1
```
lipchev, line 8: I maybe wrong here, but shouldn't this be `"T": -2`? I have it changed in one of my uncommitted stashes from months ago (back then it was still called `ReactiveEnergy` 😄 ).

Common/UnitDefinitions/ElectricReactiveEnergy.json
lipchev, line 1: Is the `Joule` an applicable / useful unit for this quantity? If yes- then it would allow us to have an SI unit for #1463

lipchev: Sorry, I actually meant to comment on #1467 but got my tabs confused..

lipchev: Alternatively `VoltampereReactiveSecond` would also do the trick, if you think it's a better fit (note, we don't have any rules about it- if you think neither units make sense, we could just [Skip] the tests and deal with it later).

Common/UnitDefinitions/ElectricAdmittance.json
McNeight: ElectricAdmittance and ElectricImpedance are two sides of the same coin. From the [Wikipedia page on admittance](https://en.wikipedia.org/wiki/Admittance): "admittance Y, measured in siemens, is defined as the inverse of impedance Z, measured in ohms."
Y≡1/Z

Beyond this is where my understanding of the math and theory gets fuzzy. I'm going to stick with impedance, because the math and units for that are directly usable in modelling radio antennas and predicting their performance. Impedance is a complex number, made up of a real part (the resistance, R, or the opposition to the flow of any current), and an [imaginary part](https://en.wikipedia.org/wiki/Imaginary_unit) (the reactance, X, or the opposition to the flow of alternating current), thus:
Z=R+jX, where j<sup>2</sup> = -1

(As an aside, why j and not i? In contexts in which use of the letter i is ambiguous or problematic, the letter j is sometimes used instead. For example, in electrical engineering and control systems engineering, the imaginary unit is normally denoted by j instead of i, because i is commonly used to denote electric current.)

So, while resistance is the only real number, and even though reactance is an imaginary number, it isn't too much of a mental leap to use an ElectricReactance class right now that stores its value in a double and just leaves out the imaginary symbol (j). The conversion of an ElectricReactance number back to an imaginary number, involving prepending a 'j' onto a ToString() call, is currently "left as an exercise for the reader".

Supporting complex numbers would either involve using the System.Numerics.Complex implementation in .NET, which would introduce a dependency on System.Numerics, or developing a UnitsNet implementation that would look like the reference source at https://github.com/microsoft/referencesource/blob/master/System.Numerics/System/Numerics/Complex.cs

In the particular case of ElectricImpedance, it would have to be a container class made up of one instance each of two other classes, an ElectricResistance class and an ElectricReactance class. For example, if the ElectricResistance class had a value of 31.8 Ω and the ElectricReactance class had a value of 2.92 Ω (or perhaps "j 2.92 Ω" at some point?), then ElectricImpedance.ToString() would create "31.8 + j 2.92 Ω".

This is why I obsoleted the existing ElectricAdmittance, and I "pre-obsoleted" the ElectricImpedance class as a kind of bookmark that says "Yes, we know this is a unit, but we don't have a technical solution at the moment." At this point, we are getting close to my suggestion of "use a named Tuple and do it yourself".

Common/UnitDefinitions/ElectricImpedance.json
McNeight: ElectricAdmittance and ElectricImpedance are two sides of the same coin. See my comments for ElectricAdmittance.json for more details.

Common/UnitDefinitions/ElectricReactiveEnergy.json
McNeight: Looks like you are correct. I'll make the change right now, in both ReactiveEnergy and ElectricReactiveEnergy, and regerate the code.

Thanks!

Common/UnitDefinitions/ElectricImpedance.json
angularsen: Alright, I guess it can help stave off future questions about why it isn't added.

Common/UnitDefinitions/ElectricReactiveEnergy.json
angularsen: I'll merge this now, we can still add this unit later if it makes sense.

Common/UnitDefinitions/ElectricAdmittance.json
angularsen: Thanks for a detailed explanation, makes sense to me.

angularsen:
Thanks @McNeight ! Nuget should be out shortly.

[Release UnitsNet/5.65.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.65.0)


Please see the [remaining question](https://github.com/angularsen/UnitsNet/pull/1444#discussion_r1898770232) above regarding SI unit. 

Other than that, we should be ready for the v6 PR next.

angularsen:
@McNeight Just a heads up regarding v6 PR:
- Had to fix compile of nanoFramework target after merging: https://github.com/angularsen/UnitsNet/pull/1478
- v6 uses https://github.com/angularsen/UnitsNet/blob/release/v6/Common/UnitRelations.json to define arithmetic operations between quantities, so `ElectricApparentPower.extra.cs` must be deleted and moved into the JSON file instead.

angularsen:
@McNeight I just merged in master into v6, so you no longer need to worry about PR for v6.

V6 nuget should be out shortly.

[Release UnitsNet/6.0.0-pre013 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F6.0.0-pre013)

Common/UnitDefinitions/ElectricAdmittance.json
lipchev: That's great- I think it may be possible in the future to come up with a solution which models this correctly (perhaps something like IComplexQuantity<TRealUnit, TImaginaryUnit, TValue> which could itself extend some `ICompositeQuantity<TFirstUnit, TSecondUnit, TValue>` or something of the sort..). 

Fun exercise for the _relatively far future..._

In the meantime, do you think the complex quantities have _any_ practical use, other than as a "documentation-section"? If not, I propose that we stash these two in PR and pin it as "Draft" or something.. It's kinda bugging me to have these "Is now obsolete.." warnings in the build output..

PR #1446, opened 2024-12-10
Fix nullability indicators for the Parse/TryParse overloads

### LINKED ISSUE #1445: UnitParser: fix the nullability of the Parse / TryParse overloads
Can we add this to #1200 

- [`public Enum Parse(string? unitAbbreviation, Type unitType, IFormatProvider? formatProvider = null)`](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/CustomCode/UnitParser.cs#L65):  `unitAbbreviation` should be non-nullable..
- [`public bool TryParse<TUnitType>(string unitAbbreviation, out TUnitType unit) where TUnitType : struct, Enum`](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/CustomCode/UnitParser.cs#L141) : `unitAbbreviation` should be nullable
- [`public bool TryParse(string unitAbbreviation, Type unitType, [NotNullWhen(true)] out Enum? unit)`](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/CustomCode/UnitParser.cs#L172):  `unitAbbreviation` should be nullable

lipchev:
(UnitParser, QuantityParser, Quantity)

fixes #1445


UnitsNet.Tests/UnitParserTests.cs
```diff
@@ -137,5 +137,20 @@ public void Parse_MappedCustomUnit()
...
+            Assert.Multiple(() =>
+            {
+                var success = unitParser.TryParse(null, out LengthUnit unit);
+                Assert.False(success);
+            }, () =>
+            {
+                var success = unitParser.TryParse(null, typeof(LengthUnit), out Enum? _);
+                Assert.False(success);
+            });
+        }
```
lipchev, line 154: Are you Ok with this type of testing? A lot of my v6 tests use the same convention (_merging only when **the method name is the same**_).

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -517,7 +517,7 @@ private void GenerateStaticParseMethods()
         /// <example>
         ///     Length.Parse(""5.5 m"", CultureInfo.GetCultureInfo(""en-US""));
         /// </example>
-        public static bool TryParse(string? str, out {_quantity.Name} result)
+        public static bool TryParse([NotNullWhen(true)]string? str, out {_quantity.Name} result)
```
lipchev, line 520: This is of course not necessary (and would probably increase our file size a little) but this is how the `IParsable` interface has it...
Haven't tried it, but I image we would get no warning in the block below:

``` csharp

if (Mass.TryParse(nullableStringAsInput, out _)){
  var firstChar = nullableStringAsInput[0]; // I guess this should be safe access..
}
```

UnitsNet.Tests/UnitParserTests.cs
angularsen: Absolutely, I have a very low bar for testing schemes 😄 as long as they do the job and the name of the test makes sense and kinda follows the other tests in the same area.

angularsen: Just to mention, we're using [FluentAssertions](https://github.com/fluentassertions/fluentassertions) at work and I do love that. 

The syntax is a breeze, assertion messages way more readable and in this case you would do `using var _ = new AssertionScope();` before all the assertions.

I wouldn't mind refactoring all tests over to it if I found surplus energy and time. _wink wink, nudge nudge_ 😆

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
angularsen: We should definitely do this, and seeing how widely used this is I don't imagine it increases any binary size in practice - meaning, the nuget is probably already loaded if not already bundled in the target framework.

UnitsNet.Tests/UnitParserTests.cs
lipchev: Don't know if you've got an AI agent- but given a little bit of preparing the right prompt- it should be able to reformat a whole class without much trouble...
If you want to start us up with an example class, I might try it out (_one day_).

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
angularsen: Okay, maybe a slight extra overhead in metadata of our methods, but still - worth it.

UnitsNet/GeneratedCode/Quantity.g.cs
```diff
@@ -451,7 +451,7 @@ public static bool TryFrom(double value, Enum? unit, [NotNullWhen(true)] out IQu
         /// <param name="quantityString">Quantity string representation, such as "1.5 kg". Must be compatible with given quantity type.</param>
         /// <param name="quantity">The resulting quantity if successful, otherwise <c>default</c>.</param>
         /// <returns>The parsed quantity.</returns>
-        public static bool TryParse(IFormatProvider? formatProvider, Type quantityType, string quantityString, [NotNullWhen(true)] out IQuantity? quantity)
+        public static bool TryParse(IFormatProvider? formatProvider, Type quantityType, string? quantityString, [NotNullWhen(true)] out IQuantity? quantity)
```
angularsen: I guess we could add `[NotNullWhen(true)]` to `string? quantityString` too.

angularsen: Seems you only edited the generated file by the way.
I'll make both changes in `StaticQuantityGenerator`.

angularsen (approved):
Looks perfect, I just added a small fix

UnitsNet/GeneratedCode/Quantity.g.cs
lipchev: This appear to have caused a compilation error in the `AbbreviatedUnitsConverter` - I've pushed a fix on my branch (removing the extra `?`), but I don't think the PR can be reopened..

PR #1448, opened 2024-12-10
Constrained the GenericMathExtensions to IQuantity

lipchev:
- replaced the aggregation with an iterator based version (slightly faster)
- the Average extension now throws an `InvalidOperationException` when the source is empty

### DIFF
--- a/UnitsNet.Tests/GenericMathExtensionsTests.cs
+++ b/UnitsNet.Tests/GenericMathExtensionsTests.cs
@@ -2,6 +2,8 @@
 // Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
 
 #if NET7_0_OR_GREATER
+using System;
+using System.Collections.Generic;
 using UnitsNet.GenericMath;
 using Xunit;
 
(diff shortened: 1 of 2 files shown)


UnitsNet.Tests/GenericMathExtensionsTests.cs
```diff
@@ -10,19 +12,59 @@ namespace UnitsNet.Tests;
...
+
+        Assert.Equal(Length.FromCentimeters(100), sumOfQuantities);
+    }
+
+    [Fact]
+    public void Sum_TwoQuantities_ReturnsTheExpectedSum()
+    {
+        IEnumerable<Length> values = [Length.FromCentimeters(100), Length.FromCentimeters(200)];
+
+        Length sumOfQuantities = GenericMathExtensions.Sum(values);
```
lipchev, line 37: The reason I'm not using `values.Sum()` is because when we add the `IVectorQuantity` interface, this is going to start testing the other extension method (which is a _better/more specific_ match for the list of `Length`).

UnitsNet/GenericMath/GenericMathExtensions.cs
```diff
@@ -2,67 +2,81 @@
...
-        // Put accumulator on right hand side of the addition operator to construct quantities with the same unit as the values.
-        // The addition operator implementation picks the unit from the left hand side, and the additive identity (e.g. Length.Zero) is always the base unit.
-        (T value, int count) result = source.Aggregate(
-            (value: T.AdditiveIdentity, count: 0),
-            (acc, item) => (value: item + acc.value, count: acc.count + 1));
+        using IEnumerator<TQuantity> e = source.GetEnumerator();
+        if (!e.MoveNext())
+        {
+            throw new InvalidOperationException("Sequence contains no elements");
+        }
```
lipchev, line 69: Notice this: in v5 this case would have caused an exception to be thrown because of the `NaN`, however as we disabled the guard in v6- this wouldn't have thrown an exception (which I believe is the [right behavior](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.average?view=net-9.0#system-linq-enumerable-average(system-collections-generic-ienumerable((system-double))))).

lipchev:
I think I already mentioned this in #1200 but these method are still applicable to the logarithmic quantities, which implement both `IAdditionOperators<TQuantity, TQuantity, TQuantity>` (rightly)  and `IAdditiveIdentity<TQuantity, TQuantity>` (incorrectly). 
``` csharp
        [Fact(Skip = "Fix in v6")]
        public void AdditiveIdentityIsIncorrect()
        {
            var expectedQuantity = PowerRatio.FromDecibelMilliwatts(24);
            AssertEx.EqualTolerance(expectedQuantity.DecibelMilliwatts, (expectedQuantity + PowerRatio.AdditiveIdentity).DecibelMilliwatts, DecibelMilliwattsTolerance);
        }
```
With the addition of the new interfaces, the logarithmic quantities are going to loose their `AdditiveIdentity` and become incompatible with the extensions from the `GenericMath` (they have their own `Geomatric/Aritmethic mean extensions).

lipchev:
Oh, and by the way- if you need to test the performance difference there are a few benchmarks available for the `Sum`:
- [UnitsNet.Benchmark/Operators/Additions/SumOfMassesWithRandomUnitsBenchmarks.cs](https://github.com/angularsen/UnitsNet/pull/1449/files#diff-6fdeebb2d07c82567b1237ba8090c74d4f4ff8889a802c20436e61ff9458e408)
- [UnitsNet.Benchmark/Operators/Additions/SumOfMassesWithSameUnitsBenchmarks.cs](https://github.com/angularsen/UnitsNet/pull/1449/files#diff-107fbfbde40b5298c1752414b28fa7880735ce0b644e8efbaeb4c89e28522265)

PR #1450, opened 2024-12-11
QuantityFormatter defaults to the "G" format

### LINKED ISSUE #1183: QuantityFormatter inconsequential format strings

lipchev:
- [x] `QuantityFormatter` defaults to the "G" format for null or empty string
- [x] `QuantityFormatter` no longer supports the `U`/`V`/`Q` formats (an exception is thrown)
- [x] `QuantityFormatter` explicitly throws for the `Cx`/`Px` formats
- [x] `QuantityFormatter` shouldn't throw for something like "P1: #.00" (instead it should output something like "P1: 12.34 mg")
- [x] `QuantityFormatter` unless explicitly given an incorrect `Ax` specifier, the `QuantityFormatter` uses the "default unit abbreviation" - which maybe `string.Empty` (if none are defined for a given unit)

Fixes #1183 
Duplicate of #1206

CC @tmilnthorp

### DIFF
--- a/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
+++ b/UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
@@ -20,26 +20,33 @@ namespace UnitsNet.Tests
         private static readonly IFormatProvider NorwegianCulture = CultureInfo.GetCultureInfo(NorwegianCultureName);
         private static readonly IFormatProvider RussianCulture = CultureInfo.GetCultureInfo(RussianCultureName);
 
-        // The default, parameterless ToString() method uses 2 sigifnificant digits after the radix point.
+        // The default, parameterless ToString() method represents the result with all significant digits, without a group separator.
         [Theory]
+        #if NET
+        [InlineData(double.MinValue, "-1.7976931348623157E+308 m")]
+        #else
+        [InlineData(double.MinValue, "-1.79769313486232E+308 m")]
+        #endif
+        [InlineData(-0.819999999999, "-0.819999999999 m")]
+        [InlineData(-0.111234, "-0.111234 m")]
+        [InlineData(-0.1, "-0.1 m")]
+        [InlineData(-0.0000012345, "-1.2345E-06 m")]
+        [InlineData(-0.000001, "-1E-06 m")]
         [InlineData(0, "0 m")]
+        [InlineData(0.000001, "1E-06 m")]
+        [InlineData(0.0000012345, "1.2345E-06 m")]
         [InlineData(0.1, "0.1 m")]
-        [InlineData(0.11, "0.11 m")]
-        [InlineData(0.111234, "0.11 m")]
-        [InlineData(0.115, "0.12 m")]
+        [InlineData(0.111234, "0.111234 m")]
+        [InlineData(0.819999999999, "0.819999999999 m")]
+        #if NET
+        [InlineData(double.MaxValue, "1.7976931348623157E+308 m")]
+        #else
+        [InlineData(double.MaxValue, "1.79769313486232E+308 m")]
+        #endif
         public void DefaultToStringFormatting(double value, string expected)
         {
-            string actual = Length.FromMeters(value).ToUnit(LengthUnit.Meter).ToString(AmericanCulture);
+            string actual = Length.FromMeters(value).ToString(AmericanCulture);
             Assert.Equal(expected, actual);
-        }
-
-        private enum CustomUnit
-        {
-            // ReSharper disable UnusedMember.Local
-            Undefined = 0,
-            Unit1,
-            Unit2
-            // ReSharper restore UnusedMember.Local
         }
 
         [Theory]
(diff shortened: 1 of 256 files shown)


UnitsNet.Tests/UnitAbbreviationsCacheTests.cs
lipchev, line 1: After you've finished reviewing these, let's move them to the `QuantityFormatterTests.cs` - none of the tests in this file should reference the `QuantityFormatter`.

UnitsNet/QuantityFormatter.cs
```diff
@@ -3,186 +3,222 @@
...
-            formatProvider ??= CultureInfo.CurrentCulture;
-            return FormatUntrimmed(quantity, format, formatProvider).TrimEnd();
+            switch (format[0])
+            {
+                case 'G' or 'g':
+                    return FormatWithValueAndAbbreviation(quantity, format, formatProvider);
+                case 'S' or 's':
+                    return ToStringWithSignificantDigitsAfterRadix(quantity, formatProvider, 0);
+                case 'A' or 'a':
+                    return UnitsNetSetup.Default.UnitAbbreviations.GetDefaultAbbreviation(quantity.Unit, formatProvider);
```
lipchev, line 149: Please note another "breaking" change (I _propose_): is that we would not longer from throw when given the 'a' format and the quantity's unit doesn't have any abbreviations(e.g. a `Volume` with the `AuTablespoon` unit)- instead we return the `default abbreviation` which is `string.Empty`.

UnitsNet/QuantityFormatter.cs
```diff
@@ -3,186 +3,222 @@
...
+                    return ToStringWithSignificantDigitsAfterRadix(quantity, formatProvider, precisionSpecifier);
+                case 'A' or 'a' when int.TryParse(format.AsSpan(1), out var abbreviationIndex):
+                {
+                    var abbreviations = UnitsNetSetup.Default.UnitAbbreviations.GetUnitAbbreviations(quantity.Unit, formatProvider);
 
-                int precisionSpecifier = 0;
+                    if (abbreviationIndex >= abbreviations.Length)
+                    {
+                        throw new FormatException($"The \"{format}\" format string is invalid because the index is out of range.");
+                    }
```
lipchev, line 176: We still throw when given an abbreviation index that's invalid (even for 'a0').

UnitsNet/QuantityFormatter.cs
```diff
@@ -3,186 +3,222 @@
...
-            string formatForSignificantDigits = UnitFormatter.GetFormat(quantity.Value, number);
-            object[] formatArgs = UnitFormatter.GetFormatArgs(quantity.Unit, quantity.Value, formatProvider, Enumerable.Empty<object>());
-            return string.Format(formatProvider, formatForSignificantDigits, formatArgs);
-        }
+    private static string FormatWithValueAndAbbreviation<TUnitType>(IQuantity<TUnitType> quantity, string format, IFormatProvider formatProvider)
+        where TUnitType : Enum
+    {
+        var abbreviation = UnitsNetSetup.Default.UnitAbbreviations.GetDefaultAbbreviation(quantity.Unit, formatProvider);
+        return string.Format(formatProvider, $"{{0:{format}}} {{1}}", quantity.Value, abbreviation);
+    }
```
lipchev, line 215: Another change from the existing behavior (which I think _is necessary_) is that the when in the case of the "standard numeric format" (which is the default for `Volume.FromAuTablespoon(1).ToString()`) we again rely on the default abbreviation (possibly `string.Empty`) instead of throwing, as would be the case if you attempt this on a previous version (e.g. v5):

`Volume.FromAuTablespoon(1).ToString("g"); // FormatException??`

lipchev:
@angularsen Eh, just wanted to note that I personally have no preference here regarding whether "G" or "S" is the default format- I just took the #1206 as a starting point.. Same goes for the exceptions. The removal of the extra formats- I'm very happy about...

angularsen:
> Eh, just wanted to note that I personally have no preference here regarding whether "G" or "S" is the default format

I think G is the better choice actually, if they want pretty formatting they should take control instead of us trying to guess if they prefer scientific notation or fixed point. I am no longer sure the S format is a good idea, but I guess we can keep it since it is now opt-in instead of default.

PR #1452, opened 2024-12-11
Minor performance improvements to the BaseUnits class (v5)

lipchev:
- avoid storing the `IsFullyDefined` value to a field (the property is only used when constructing a `UnitSystem`)
- add a reference check to the `Equals` method (improves the == Undefined)
- use the `HashCode` class for the `GetHashCode` method (note that the result would be different between Framework and NET projects)

### DIFF
--- a/UnitsNet/BaseUnits.cs
+++ b/UnitsNet/BaseUnits.cs
@@ -46,14 +46,6 @@ namespace UnitsNet
             Temperature = temperature;
             Amount = amount;
             LuminousIntensity = luminousIntensity;
-
-            IsFullyDefined = Length is not null &&
-                             Mass is not null &&
-                             Time is not null &&
-                             Current is not null &&
-                             Temperature is not null &&
-                             Amount is not null &&
-                             LuminousIntensity is not null;
         }
 
         /// <inheritdoc />
@@ -71,6 +63,9 @@ namespace UnitsNet
         {
             if (other is null)
                 return false;
+
+            if (ReferenceEquals(this, other))
+                return true;
 
             return Length == other.Length &&
                 Mass == other.Mass &&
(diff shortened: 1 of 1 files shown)


lipchev:
If you think the 3rd point is a "breaker", I could switch it over to v6 but I find it very unlikely that this would affect anyone (I'm frankly not even sure if the same code would produce the same result on both targets 🤷 ).
I have one more change I want to make- but that's more likely (although it shouldn't) to surprise somebody:

I'd like to change the `ToString` format:

```  csharp

        public override string ToString()
        {
            if (Equals(Undefined))
            {
                return "Undefined";
            }

            return string.Join(", ", GetUnitsDefined());

            IEnumerable<string> GetUnitsDefined()
            {
                if (Length is not null)
                {
                    yield return $"[Length]: {Length}";
                }
                if (Mass is not null)
                {
                    yield return $"[Mass]: {Mass}";
                }
                if (Time is not null)
                {
                    yield return $"[Time]: {Time}";
                }
                if (Current is not null)
                {
                    yield return $"[Current]: {Current}";
                }
                if (Temperature is not null)
                {
                    yield return $"[Temperature]: {Temperature}";
                }
                if (Amount is not null)
                {
                    yield return $"[Amount]: {Amount}";
                }
                if (LuminousIntensity is not null)
                {
                    yield return $"[LuminousIntensity]: {LuminousIntensity}";
                }
            }
        }
```

The difference is that we're not using the `UnitAbbreviations` - which removes the potential side effects that this could have when "configuring" the `UnitsNetSetup`.

angularsen:
Looks fine to me, I don't think this should break anyone or at least extremely few.

Regarding the ToString change, beware that `yield return` and enumerable are not really performant if that is the goal. No source right now, but if I recall correctly using `StringBuilder` is the way to go to construct strings performantly.

Of course, we're talking micro optimizations here, but also no real downside to that approach.

angularsen:
[Release UnitsNet/5.62.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.62.0)

PR #1453, opened 2024-12-12
Update the UnitRelations.json

lipchev:
As per my comments from #1200 here are the proposed modifications to the operators:
- [x] adding an `Inverse` relation between `Density` and `SpecificVolume` (replacing the existing multiplication)
- [x] adding standard multiplication for `MolarEnergy.JoulePerMole * AmountOfSubstance.Mole` (returning `Energy.Joule`)
- [x] adding standard multiplication for `Illuminance.Lux * Area.SquareMeter` (returning `LuminousFlux.Lumen`)
- [x] `TemperatureDelta.Kelvin * CoefficientOfThermalExpansion.PerKelvin` now returns a `Ratio` (instead of a `double`)
- [x] added the `-- NoInferredDivision` to the `KinematicViscosity.SquareMeterPerSecond * Duration.Second` operator (as the opposite of the relation was `Area / Duration` -> `KinematicViscosity` 😏 )
- [x] added the `-- NoInferredDivision` to the `Area.SquareMeter * ReciprocalArea.InverseSquareMeter` operator (as the opposite of the relation was `Ratio / Area` -> `InverseArea` does not appear intuitive )
- [x] change the multiplication units (using the `BaseUnit`) for the `Mass.Kilogram = AmountOfSubstance.Mole * MolarMass.KilogramPerMole`
- [x] change the multiplication units (using the `BaseUnit`) for the `AmountOfSubstance.Mole = MolarFlow.MolePerSecond * Duration.Second`

I've included tests for **_all_** operators from the affected classes (_most operators, introduced with the `UnitRelations.json` were never covered by tests_).

PS There are still many operators not covered by tests, but I'd rather finish them later (_hopefully once we no longer have to worry about rounding errors_).


lipchev:
You'll probably remark that there more than one styles for naming / testing the operators, I've been following the convention in the file (_or rather the AI was eager to suggest the same naming and I just went along_) : which one do you think we should prefer? There is a of course benefits from using the [Theory] variant everywhere, but frankly- writing in this style is quite tedious with the 6 parameters.. Donno, up to you..

angularsen:
@lipchev Holy cow, you have like 10 PRs active right now 😅 I'll try to get to them this weekend, any particular order or ones I should do first?

lipchev:
@angularsen I've selected them such that they shouldn't have any merge conflicts between each other.

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
```diff
@@ -122,5 +122,65 @@ public void VolumeSolutionFromComponentMassAndDesiredConcentration(
...
+            var amountOfSubstance = AmountOfSubstance.FromMass(mass, molarMass);
+
+            Assert.Equal(expectedAmountOfSubstance, amountOfSubstance);
+        }
+
+        [Theory]
+        [InlineData(5, AmountOfSubstanceUnit.Mole,
+            10, MolarEnergyUnit.JoulePerMole,
+            50, EnergyUnit.Joule)]
+        public void Multiplying_AmountOfSubstance_By_MolarEnergy_ReturnsEnergy(double amountOfSubstanceValue, AmountOfSubstanceUnit amountOfSubstanceUnit,
```
angularsen, line 142: If you only have one `InlineData`, I think a regular `[Fact]` is better and just inline the values in the code instead of passing as parameters. It was not particularly readable as it stands.

Theory is mostly useful when passing multiple variations of parameters.
Same for other places.

angularsen: Basically do what you did in AreaTests, it's short and sweet.

UnitsNet.Tests/CustomCode/CoefficientOfThermalExpansionTests.cs
```diff
@@ -44,21 +44,23 @@ public class CoefficientOfThermalExpansionTests : CoefficientOfThermalExpansionT
...
-            double ctePerDegC = 0.001;
-            CoefficientOfThermalExpansion cte = CoefficientOfThermalExpansion.FromPerDegreeCelsius(ctePerDegC);
-            TemperatureDelta dT = TemperatureDelta.FromDegreesCelsius(temperatureDeltaDegC);
-            AssertEx.EqualTolerance(dT * cte, temperatureDeltaDegC * ctePerDegC, 1e-10);
+            CoefficientOfThermalExpansion cte = CoefficientOfThermalExpansion.FromPerDegreeCelsius(2);
+            TemperatureDelta dT = TemperatureDelta.FromDegreesCelsius(0.001);
+
+            Ratio expansionRatio = dT * cte;
+
+            Assert.Equal(Ratio.FromDecimalFractions(0.002), expansionRatio);
```
angularsen, line 63: Nitpick, but while touching this, I'd shorten it to something like you did in AreaTests. Here and other places. It'll make it easier for the AI to mimic next time too 😆

UnitsNet.Tests/CustomCode/SpecificVolumeTests.cs
```diff
@@ -41,9 +41,9 @@ public static void SpecificVolumeTimesMassEqualsVolume()
         }
 
         [Fact]
-        public static void ConstantOverSpecificVolumeEqualsDensity()
+        public static void InverseSpecificVolumeEqualsDensity()
         {
-            Density density = 5 / SpecificVolume.FromCubicMetersPerKilogram(20);
+            Density density = SpecificVolume.FromCubicMetersPerKilogram(4).Inverse();
```
angularsen, line 46: Remind me, don't we want to support `double / QuantityX` when adding support for `Inverse()`? Shouldn't we offer both?

I see the division operator is removed for several quantities, e.g. `SpecificVolume`, `CoefficientOfThermalExpansion` and maybe others.

angularsen (changes requested):
A couple of questions and suggestions

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
lipchev: I was mostly lazy (_and partly afraid that picking another value would turn up a rounding error_)- but technically, for a "complete" set of values we need to include `NaN` / `Infinity` as well as both base and non-base units...

UnitsNet.Tests/CustomCode/SpecificVolumeTests.cs
lipchev: We don't implement `double / Quantity` for any of the other `Inverse` relationships - if we had to add an operation with `double` a more "natural looking" one would be the multiplication between inverse quantities - producing a `double` (or `Ratio`), however I don't personally think that makes a lot of sense (and even less does the `double` / `A` = `B`  version).

The two other operators that are removed are from the `Ratio` (I've simply added the `-NoInferredDivision` flag):
```
        IDivisionOperators<Ratio, ReciprocalArea, Area>,
        IDivisionOperators<Ratio, Area, ReciprocalArea>,
```
Again, this is subjective (and I'm definitely not an expert on the `InverseArea` domain 😆 )- but I don't see how anyone would expect to divide `20 %`  by `2 m^2` in order to obtain the `ReciprocalArea`..

Note that these were the only two operators defined on the `Ratio` type.

lipchev: Oh and by the way, the `Ratio` operators were only added in `v6` (I'm not 100% sure, but I think the only breaking change from v5 is the `Density`).

UnitsNet.Tests/CustomCode/AmountOfSubstanceTests.cs
lipchev: I've updated the tests (removing the [Theory])

lipchev:
I've added another small change (and updated the description in the PR)

> change the multiplication units (using the BaseUnit) for the AmountOfSubstance.Mole = MolarFlow.MolePerSecond * Duration.Second

I have a plan to change this in the future, but for the time being I think it's best to avoid using derived units in the operators (due to the additional conversions that are required).

UnitsNet.Tests/CustomCode/SpecificVolumeTests.cs
angularsen: Alright, I guess number divided by quantity is a bit weird. I was just concerned if we are removing something people are actual using, but they'll just have to multiply with the inverse quantity instead then.

We can try and see if any complaints come our way.

angularsen (approved):
Looks good to me now 👍

angularsen:
@lipchev that's what I got for now, I'll have to find time later this week for the remaining PRs. Awesome work though, thanks! ❤️

PR #1457, opened 2024-12-15
Fixing the nullability attribute of the AbbreviatedUnitsConverter (build failure)

lipchev:
fixing the nullability of the `unitAbbreviation` in the Parse method (following the build failure from #1446 )

BTW, note that this method isn't even used anywhere. I'm not sure if we've ever used it and then stopped or if it was meant to serve another purpose (having _extended_ the converter)..


angularsen:
I'm not sure if it's used either, we can clean up later.

PR #1459, opened 2024-12-16
Remove IParsable and IComparisonOperators from the IQuantity interface

### LINKED ISSUE #1454: Remove IComparisonOperators<TSelf, TSelf, bool> and IParsable<TSelf> from the IQuantity interface
Here is the current declaration:

``` csharp
    /// <summary>
    ///     An <see cref="IQuantity{TUnitType}"/> that (in .NET 7+) implements generic math interfaces for equality, comparison and parsing.
    /// </summary>
    /// <typeparam name="TSelf">The type itself, for the CRT pattern.</typeparam>
    /// <typeparam name="TUnitType">The underlying unit enum type.</typeparam>
#if NET7_0_OR_GREATER
    public interface IQuantity<TSelf, TUnitType>
        : IQuantity<TUnitType>
        , IComparisonOperators<TSelf, TSelf, bool>
        , IParsable<TSelf>
#else
    public interface IQuantity<in TSelf, TUnitType>
        : IQuantity<TUnitType>
#endif
        where TSelf : IQuantity<TSelf, TUnitType>
        where TUnitType : Enum
```

I think that the `IComparisonOperators<TSelf, TSelf, bool>` and `IParsable<TSelf>` declarations should be moved to the implementation section. 

The motivation: reduce the number of methods / operators required for the "custom quantities": consider how many methods are added by moving `HowMuch` from `: IQuantity` to `: IQuantity<HowMuch, HowMuchUnit>`.

A more _theoretical_ motivation is that by having it on the interface, it feels like we're expressing a _requirement_, as in "I need this method in order to be able to do something"- which is clearly not the case here. 
IMO anyone requiring the generic constraint can use a combination of constraints:

``` csharp
public List<TQuantity> ParseQuantities<TQuantity>(string text) where TQuantity: IParsable<TQuantity>;
public List<TUnit> ParseUnits<TQuantity, TUnit>(string text) where TQuantity: IQuantity<TUnit>, IParsable<TQuantity>;
```

Note that repeating the declaration 180 times would add few KB to the size, but if that's a priority, we could introduce some `IDefaultQuantity<TQuantity>` interface and move all interface declarations that are shared by all quantities:
``` csharp

#if NET7_0_OR_GREATER
        IComparisonOperators<TSelf, TSelf, bool>,
        IParsable<TSelf>,
#endif
        IComparable,
        IComparable<TSelf>,
        IConvertible,
        IEquatable<TSelf>,
        IFormattable

```

lipchev:
Fixes #1454


lipchev:
@angularsen This is not related to this PR but as of recently (I think) when the CodeGen generates the files (at least on my machine) I get a change in the `Quantity.g.cs` with regards to the encoding:

> These files have different encodings. Left file: Unicode (UTF-8) with signature. Right file: Cyrillic (Windows). You can resolve the difference by saving the right file with the encoding Unicode (UTF-8) with signature.  

If I covert the encoding (say using Notepad++) to `UTF-8-BOM` the file is removed from the change-list. Any ideas?

lipchev:
> @angularsen This is not related to this PR but as of recently (I think) when the CodeGen generates the files (at least on my machine) I get a change in the `Quantity.g.cs` with regards to the encoding:
> 
> > These files have different encodings. Left file: Unicode (UTF-8) with signature. Right file: Cyrillic (Windows). You can resolve the difference by saving the right file with the encoding Unicode (UTF-8) with signature.
> 
> If I covert the encoding (say using Notepad++) to `UTF-8-BOM` the file is removed from the change-list. Any ideas?

Hm, this isn't happening after merging the changes from my other PR- something must have changed in the encoding there already..

angularsen:
Aha, I do recall I found you had made some manual changes to this particular generated file, which I fixed by moving the code changes to the StaticQuantitySomething.cs file instead and running codegen. Maybe this fixed it.

lipchev:
> Aha, I do recall I found you had made some manual changes to this particular generated file, which I fixed by moving the code changes to the StaticQuantitySomething.cs file instead and running codegen. Maybe this fixed it.

My hypothesis for this case was that I must have accidentally reverted the change in the `StaticQuantityGenerator` (the change was still there in my other solution). 

In any case there isn't anything non-standard in terms of text in this file so I guess we'll see how it goes in the future...

PS I'm also a standard windows user...

Issue #1463, opened 2024-12-16
The `SupportsSIUnitSystem` property should no be needed


lipchev:
You know how many of the quantities would throw an exception from their `ToUnit`/`As(UnitSystem)` method when they don't have any unit with a proper SI `BaseUnit`.. well I'd like to change that (at least in regards to `UnitSystem.SI`):
- I think it's reasonable to say that no matter what unit system we use, "a dimensionless quantity should have the decimal fraction as it's base unit" - with the `%`,  `ppm`, `ppb` etc. being the derived units. In the future we could probably talk about extending our definition of the `BaseUnits` (e.g. consider `mg/mg` as `[M][M^-1]` or something) but as long as the _ratio_ is composed of the same base-units (i.e. no additional conversion such as `x * 100`) we could simplify the code generated for these methods to ignore the parameter (the unit system) and simply return the `As(BaseUnit)`. 
- from memory, there are only a couple of other quantities which were failing (mostly due to simply missing their `BaseUnits` in the json-definition): I've already got these ready (as I needed them for something else)
- from memory, the only breaking change w.r.t. the existing definitions was the `Angle` which used to have the `DegreeCelsius` as it's `BaseUnit` (which we've already fixed) and the `FuelEfficiency`, which I wanted to invert (_I don't remember where I mentioned this_), and for which I had to add the "SI unit" (_don't quote me on that_ ) `MeterPerCubicMeter` (m/m³) (we could go even further and say that this should probably be it's `BaseUnit` but I haven't gone _that far_..😆)

Overall, I know that this can be done (as I've already removed the `SupportsSIUnitSystem` from my version of the tests, a few months ago..), so if you don't mind I'd like to do it for the `v6`.

PS In the future (_or whenever you say_), I'd also like to move the `As(UnitSystem)` and the `ToUnit(UnitSystem)` methods out of the `IQuantity` interface (similarly to what I propose in #1461 w.r.t. the "Equals with tolerance", these could be transformed into extension methods, _without breaking the existing usages_).

PS2 There is of course a lot more work to be done regarding the support for the other unit systems but I think we can keep the _experimental_ flag on these for the `v6` release, and come back to it later (_I do have some ideas on the matter_)

PS3 Having merged most of the SI / BaseUnits related stuff here's a summary of what's left:

- [ ] `ElectricApparentEnergy` : see about adding an SI unit (VoltampereSecond, Joules?), if any of these make sense (CC @McNeight)
[...]

PR #1464, opened 2024-12-17
Add radiation equivalent dose rate

voronizer:
Add radiation equivalent [dose rate](https://en.wikipedia.org/wiki/Dose_rate).


angularsen:
Thanks! I ported this to v6 branch also.
https://github.com/angularsen/UnitsNet/pull/1468

Nuget should be out shortly.
[Release UnitsNet/5.62.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.62.0)

lipchev:
@angularsen 
Sorry, I'm perhaps a little late for the party, but I'd just like to point out that, IMO, the use of `Seconds` should be preferred to `Hours` for the `BaseUnit` definition, with the latter corresponding to the derived unit of time of SI. 

As it stands, the `RadiationEquivalentDose` would likely throw if asked for it's _official SI unit_ (i.e. using `ToUnit(UnitSystem.Si)`). We don't have any rules for _having a SI base unit_ (see #1463) but even if just as a performance consideration- given an operation (like the multiplication with `Duration`) that operates using the quantities `BaseUnit`s is significantly faster than one requiring a conversion to a non-base unit.

For example, multiplying some `RadiationEquivalentDose` by "30 minutes" would have the `Duration` perform a dual conversion from `Minutes` to `Seconds` and then from `Seconds` to `Hours`, which currently involves extra boxing of the `Enum`, in addition to asking the `ConcurrentDictionary`  of the `UnitConverter` for a custom conversion that _likely_ does not exist..

PS I love how the [build failed](https://app.codecov.io/gh/angularsen/UnitsNet/commit/36b783c6bec1113bfbe7628b60faf138b94a5511) due to the lack of coverage of the operator (_and the uncovered lines are finally showing up clearly_)- it's very useful in highlighting issues like the one I spotted just now.. 👍

angularsen:
Gotcha, I was thinking about it, but didn't know if seconds made sense in this domain. A quick google reveals that there are conversions for Sievert/second, so sure, we can add that and set it as base unit I guess.

angularsen:
https://github.com/angularsen/UnitsNet/pull/1469

voronizer:
Sorry, I'm perhaps a little late for the party :), but. I'm working for the company related to measurement of radiation an I could say that the amount of Sieverts per seconds value could be too small even for Double unit - its just no sense. I don't know internals of UnitsNet deeply. I added Sv(rem) per hour as base unit widely used in industry and

angularsen:
Not late at all, I'm just pushing too fast here, sorry my bad.

I did find references to Sievert/second, millisievert/second etc. so I assumed it would make sense to include it, mostly to have SI compatible units.
https://www.translatorscafe.com/unit-converter/en-US/radiation/21-27/sievert/second-milliroentgen/hour/
https://www.rp-alba.com/index.php?filename=radiationDoseRateConverter.php

I have no idea if these are reputable sources, but at least more than one source included it.

voronizer:
Ok. Lets see how it will work. Thank you!

angularsen:
Merged into V6 branch, nuget should be out shortly.

[Release UnitsNet/6.0.0-pre013 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F6.0.0-pre013)

PR #1472, opened 2024-12-26
💥Require unit enum types to also be struct

angularsen:
Change all `where TUnitType : Enum` generic type constraints to include `struct` constraint, to avoid usages like this.

```cs
UnitsNetSetup.Default.UnitParser.Parse<LengthUnit>("abc"); // Ok
UnitsNetSetup.Default.UnitParser.Parse<Enum>("abc"); // Compiles, but throws exception
```

This also fixes inconsistency, between `UnitAbbreviationsCache.GetUnitAbbreviations<TUnitType>` and `UnitAbbreviationsCache.GetDefaultAbbreviation<TUnitType>`.

### Changes
- Add `struct` constraint to all `Enum` constraints
- Remove two test cases `MapAndLookup_MapWithSpecificEnumType_LookupWithEnumType`, `MapAndLookup_WithEnumType`

### Noteworthy

There are some minor use cases for passing non-generic `Enum` value, like getting unit abbreviations for an `IQuantity` such as `UnitAbbreviations.GetDefaultAbbreviation(quantity.Unit)`. 
After this PR, you now have to do `GetDefaultAbbreviation(quantity.Unit.GetType(), Convert.ToInt32(quantity.Unit))` instead. 
`GetAbbreviations()` already had this requirement, so now it's more consistent at least.

### Sources
https://stackoverflow.com/a/74773273
https://devblogs.microsoft.com/premier-developer/dissecting-new-generics-constraints-in-c-7-3/


lipchev:
Awesome, now we just need to push this constraint up to the `IQuantity`:
```
public interface IQuantity<in TSelf, TUnitType> : IQuantity<TUnitType>
    where TSelf : IQuantity<TSelf, TUnitType>
    where TUnitType : Enum
```
(I think there are a few other usages two)

But most importantly, we should consider breaking the `UnitInfo`:
```
    public class UnitInfo<TUnit> : UnitInfo
        where TUnit : Enum
```

and make it so that the `HowMuch` is required to have a concrete Enum type (we could keep the `IQuantity.Unit` as it is, implemented explicitly).

I'm not sure how much of this can be done in the current `release-v6`, in my own branch I've actually made the `QuantityInfo` abstract, but I don't know if this would be easy to do here..

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
```diff
@@ -215,7 +215,7 @@ public string GetDefaultAbbreviation(Type unitType, int unitValue, IFormatProvid
         /// <param name="unit">Enum value for unit.</param>
         /// <param name="formatProvider">The format provider to use for lookup. Defaults to <see cref="CultureInfo.CurrentCulture" /> if null.</param>
         /// <returns>Unit abbreviations associated with unit.</returns>
-        public string[] GetUnitAbbreviations<TUnitType>(TUnitType unit, IFormatProvider? formatProvider = null) where TUnitType : Enum
+        public string[] GetUnitAbbreviations<TUnitType>(TUnitType unit, IFormatProvider? formatProvider = null) where TUnitType : struct, Enum
         {
             return GetUnitAbbreviations(typeof(TUnitType), Convert.ToInt32(unit), formatProvider);
         }
```
lipchev, line 221: This is great, introducing the `struct` constraint here opens up the possibility for using 
`Unsafe.As<TUnit, int>(ref unit)`
instead of 
`Convert.ToInt32(unit)` 
which is ~34x faster on `net8.0` and ~92x faster on `net48` 🚀

UnitsNet/QuantityDisplay.cs
lipchev, line 1: This one is really bugging me with it's 0% coverage - I'm definitely going to do something about it soon..

UnitsNet/QuantityInfo.cs
lipchev, line 1: If we want to be really strict about it, we should consider making the `QuantityInfo` class abstract- this way we know that the user won't be able to implement `IQuantity` without having actually implemented `IQuantity<TUnit>`.

lipchev (approved):
This all looks great. Just one question - do you have a preference with the way that the generic constraints are positioned (same line or new line)?

angularsen:
> This all looks great. Just one question - do you have a preference with the way that the generic constraints are positioned (same line or new line)?

Great! No, I'm too old to care about these things anymore 😅 Yolo! 
As long as it's follows most typical c# norms, I'm happy.

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
angularsen: Sounds good!

PR #1473, opened 2024-12-27
Adding the missing BaseUnits

### LINKED ISSUE #1043: Adding BaseUnits for Angstrom results in test failures due to UnitSystem logic
**Describe the bug**
Adding BaseUnits for Angstrom results in test failures due to UnitSystem logic. The BaseUnits for Armstrongs are coming as the first QuantityInfo in the UnitSystem constructor since A happens to be first in the alphabetic list.

**To Reproduce**
Steps to reproduce the behavior (just an example):
- Make sure Armstrongs are in the code (see #1042)
- Add BaseUnits of Meter

**Expected behavior**
Meters is still the expected LengthUnit for SI.

**Screenshots**
image

### LINKED ISSUE #1463: The `SupportsSIUnitSystem` property should no be needed

lipchev:
Fixes #1463 
Fixes #1043

- removed the `UnitSystem` constructor from the Dimensionless quantities (which was previously throwing)
- `As`/`ToUnit(UnitSystem)` for all dimensionless quantities now convert to their `BaseUnit` (i.e. the "DecimalFraction") *
- `As/ToUnit(UnitSystem)` for all other quantities refactored using the QuantityInfoExtensions
- added tests for the `UnitSystem` methods, skipping the tests for all quantities that fail with `UnitSystem.SI` (with a reason)

There are only two dimensionless quantities (IMO) that don't fit the definition: 
- `RelativeHumidity`:  currently has only the `Percent` unit, I think we should add the `DecimalFraction`, setting it to be the `BaseUnit`
- `FuelEfficiency`: I think this could be defined as `"L": -2` with the addition of the `MeterPerCubicMeter` unit (possibly setting it as its `BaseUnit`, if we want to satisfy the `BaseUnit_HasSIBase` test)

You can look for `As_UnitSystem_ReturnsValueInDimensionlessUnit` if you want to check the rest of the dimensionless quantities.

Regarding the removed `BaseUnits` (see `Force.json` or `Pressure.json`), those were detected by some earlier tests I had in place, regarding the multiplication/division operators where I used the following definition:
- A given operation between two quantities (either multiplication or division) such as `A / B = C` is only defined if `A.Dimensions / B.Dimensions = C.Dimensions` 
- When the intersection between `A.Dimensions` and `B.Dimensions` is the empty set, for every unit of `A` and `B` for which the `BaseUnits` is not `Unidefined`, and every unit of `C`, having `BaseUnits` = `A.BaseUnits union B.BaseUnits`, it must be true that `C.Value = A.Value / B.Value`.
- When the intersection between `A.Dimensions` and `B.Dimensions` is not empty, and the intersecting `BaseUnits` of `A` , `B` and `C` are all the same, then again we have the same condition, which I generally refer to as "the conversion coefficient is 1"
- The same logic can be used to infer the unit-conversion coefficient based off the `Dimensions` and a pair of `BaseUnits`, but special attention needs to be taken w.r.t. the exponents (e.g. `Area` is `L2` so the unit-conversion coefficients are squares of the ones from `Length`)
- If we want to extend this definition in the future, we should consider introducing a way to override the "default conversion coefficient" (`1`)..

Here are some links:
https://en.wikipedia.org/wiki/Dimensional_analysis
https://en.wikipedia.org/wiki/International_System_of_Units#Definition
https://en.wikipedia.org/wiki/Coherence_(units_of_measurement)


lipchev:
@angularsen I initially thought about not touching the unit definitions, and just [Skip] the tests - but after seeing that the tests failed for 61 quantities, decided that it would be simpler to just add the `BaseUnits`, rather than add to 61 test files (just to remove the addition in the next PR).

Hopefully you wont find the json files too difficult to review..

UnitsNet/CustomCode/QuantityInfo/QuantityInfoExtensions.cs
lipchev, line 1: I've got a bunch more extensions in here, for the time being they are internal- with one two good candidates for being public.
Anyway- tell me if you want to move them to another location / namepsace..

UnitsNet.Tests/CustomCode/TestsBase/QuantityTestsBase.cs
lipchev, line 1: The property is no longer used (other than the overrides) - but I didn't want to remove it just yet, as I wanted to only have the failing tests show up in this PR.
Once this is merged, I'll create another one  that just removes the property.

lipchev:
I took some other notes while preparing this PR:

The initial size for `UnitsNet.dll` was

> 2.23 MB (2 339 328 bytes)

1. After adding the extension and refactoring the constructor / `As/To(UnitSystem)` methods (with 61 quantities failing due to missing SI BaseUnit)

> 2.20 MB (2 311 680 bytes)

2. After adding all the missing `BaseUnits`

> 2.21 MB (2 321 920 bytes)

Still ahead, but if we include the "prefixed" base units- things will likely even out (possibly even setting us back some).

PS I'll update you with the actual size later..

lipchev:
Also noted the coverage 😄 :

> Information
Parser: 	MultiReport (3x DotCover)
Assemblies: 	3
Classes: 	293
Files: 	298
Line coverage
85%
Covered lines: 	30789
Uncovered lines: 	5179
Coverable lines: 	35968
Total lines: 	166435
Line coverage: 	85.6%

And now:

> Information
Parser: 	MultiReport (3x DotCover)
Assemblies: 	3
Classes: 	294
Files: 	299
Line coverage
88%
Covered lines: 	30364
Uncovered lines: 	3765
Coverable lines: 	34129
Total lines: 	163888
Line coverage: 	88.9%

lipchev:
Oh, forgot about the tests in `IQuantityTests` - the ones that are testing the `UnitSystem` methods are now redundant, but we can remove them as we complete the coverage of the As/ToUnit methods..

UnitsNet.NanoFramework/GeneratedCode/Quantities/ElectricCharge.g.cs
lipchev, line 1: This one wasn't re-generated from the last commit.

Common/UnitDefinitions/Pressure.json
```diff
@@ -310,11 +285,6 @@
     {
       "SingularName": "TonneForcePerSquareCentimeter",
       "PluralName": "TonnesForcePerSquareCentimeter",
-      "BaseUnits": {
-        "L": "Centimeter",
-        "M": "Tonne",
-        "T": "Second"
-      },
```
angularsen, line 317: Why are these removed? I see no duplicates.

Common/UnitDefinitions/SpecificFuelConsumption.json
```diff
@@ -3,6 +3,11 @@
   "BaseUnit": "GramPerKiloNewtonSecond",
```
angularsen, line 3: Out of scope of PR, but we should probably rename this to `Kilonewton` with lowercase `n`.

angularsen: Added to https://github.com/angularsen/UnitsNet/issues/1200

Common/UnitDefinitions/SpecificFuelConsumption.json
```diff
@@ -31,6 +36,10 @@
     {
       "SingularName": "GramPerKiloNewtonSecond",
       "PluralName": "GramsPerKiloNewtonSecond",
+      "BaseUnits": {
+        "L": "Meter",
+        "T": "Microsecond"
```
angularsen, line 41: Meter, Second?

Not entirely sure how to correctly resolve the SI base units for this.
I tried wolfram, but no clear cut answer.

https://www.wolframalpha.com/input?i=Gram+Per+KiloNewton+Second
image

UnitsNet.Tests/CustomCode/QuantityInfos/QuantityInfoExtensionsTest.cs
```diff
@@ -0,0 +1,61 @@
...
+            (ElectricCurrentUnit)(-1),
+            (TemperatureUnit)(-1),
+            (AmountOfSubstanceUnit)(-1),
+            (LuminousIntensityUnit)(-1)));
+
+        // Act
+        VolumeUnit defaultUnit = Volume.Info.GetDefaultUnit(unitSystem);
+
+        // Assert
+        Assert.Equal(VolumeUnit.CubicDecimeter, defaultUnit);
```
angularsen, line 59: I assume this isn't going to be random/flaky, but if it is prone to randomness or future changes we could make it slightly more robust by asserting it being one of all known possible units, e.g. CubicDecimeter and Liter. This does not help if adding new units in the future, of course.

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
```diff
@@ -292,20 +294,113 @@ public void As()
...
+        }
+
+        [Fact]
+        public void As_UnitSystem_ThrowsArgumentExceptionIfNotSupported()
+        {
+            var quantity = new Acceleration(value: 1, unit: Acceleration.BaseUnit);
+            // simulating a combination of units that won't match anything
+            var unsupportedUnitSystem = new UnitSystem(new BaseUnits(
+                (LengthUnit)(-1), (MassUnit)(-1), (DurationUnit)(-1), (ElectricCurrentUnit)(-1), (TemperatureUnit)(-1), (AmountOfSubstanceUnit)(-1), (LuminousIntensityUnit)(-1)));
+            Assert.Throws<ArgumentException>(() => quantity.As(unsupportedUnitSystem));
```
angularsen, line 329: It seems this particular test-only BaseUnits of all values `-1` is used alot for testing (yes yes, I realize we are generating tests), maybe we can move it to a static readonly named something like `TestConstants.UnsupportedBaseUnit` and reuse it?

Mostly for readability, and we can move comment to xmldoc of static field.

UnitsNet/GeneratedCode/Quantities/Acceleration.g.cs
```diff
@@ -886,25 +881,7 @@ public double As(AccelerationUnit unit)
...
-        }
-
-        /// <inheritdoc />
-        double IQuantity.As(Enum unit)
-        {
-            if (!(unit is AccelerationUnit typedUnit))
-                throw new ArgumentException($"The given unit is of type {unit.GetType()}. Only {typeof(AccelerationUnit)} is supported.", nameof(unit));
-
-            return As(typedUnit);
+            return As(Info.GetDefaultUnit(unitSystem));
```
angularsen, line 884: Nice cleanup this

UnitsNet.Tests/GeneratedCode/TestsBase/RatioTestsBase.g.cs
```diff
@@ -212,20 +191,70 @@ public void As()
...
+                Assert.Throws<ArgumentNullException>(() => quantity.ToUnit(nullUnitSystem));
+            }, () =>
+            {
+                IQuantity<RatioUnit> quantity = new Ratio(value: 1, unit: Ratio.BaseUnit);
+                Assert.Throws<ArgumentNullException>(() => quantity.ToUnit(nullUnitSystem));
+            }, () =>
+            {
+                IQuantity quantity = new Ratio(value: 1, unit: Ratio.BaseUnit);
+                Assert.Throws<ArgumentNullException>(() => quantity.ToUnit(nullUnitSystem));
+            });
```
angularsen, line 257: Good work on tests and handling dimensionless quantities 👏

UnitsNet/GeneratedCode/Quantities/Ratio.g.cs
```diff
@@ -97,25 +97,6 @@ public Ratio(double value, RatioUnit unit)
...
-        {
-            if (unitSystem is null) throw new ArgumentNullException(nameof(unitSystem));
-
-            var unitInfos = Info.GetUnitInfosFor(unitSystem.BaseUnits);
-            var firstUnitInfo = unitInfos.FirstOrDefault();
-
-            _value = value;
-            _unit = firstUnitInfo?.Value ?? throw new ArgumentException("No units were found for the given UnitSystem.", nameof(unitSystem));
-        }
-
```
angularsen, line 118: I'm fine removing this ctor as I hardly think anyone uses it, but would it not also be equally valid to let the ctor construct a ratio with unit `BaseUnit` (`DecimalFraction`)? 

It seems consistent with behavior that `As(UnitSystem)` also uses `BaseUnit`, and quantity implementations would be more consistent with each other.

UnitsNet/GeneratedCode/Quantities/Ratio.g.cs
```diff
@@ -822,28 +788,32 @@ private bool TryToUnit(RatioUnit unit, [NotNullWhen(true)] out Ratio? converted)
...
+            return As(typedUnit);
+        }
+
+        /// <inheritdoc />
+        IQuantity IQuantity.ToUnit(Enum unit)
+        {
+            if (!(unit is RatioUnit typedUnit))
+                throw new ArgumentException($"The given unit is of type {unit.GetType()}. Only {typeof(RatioUnit)} is supported.", nameof(unit));
+
+            return ToUnit(typedUnit, DefaultConversionFunctions);
```
angularsen, line 816: Out of scope of PR, but the call to `ToUnit()` looks a bit weird to me.

- `DefaultConversionFunctions` is an `UnitConverter` instance, should maybe be named `UnitConverter`
- Should this really be `public`? Should we not encourage `UnitsNetSetup.Default.UnitConverter`, or the shorthand `UnitConverter.Default` instead?
- I then realize, each quantity creates their own instance o` UnitConverter`, in addition to `UnitsNetSetup.Default.UnitConverter`, which creates another instance based on all quantities in `UnitConverter.CreateDefault`. This means, modifying `Angle.DefaultConversionFunctions`  does not modify `UnitsNetSetup.Default.UnitConverte` and they become inconsistent, and vice versa 👎 

Adding to #1200

angularsen (changes requested):
Great work! Just a few comments/questions.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -1123,29 +1113,44 @@ private bool TryToUnit({_unitEnumName} unit, [NotNullWhen(true)] out {_quantity.
...
+"
+                : $@"
         /// <inheritdoc cref=""IQuantity.ToUnit(UnitSystem)""/>
         public {_quantity.Name} ToUnit(UnitSystem unitSystem)
         {{
-            if (unitSystem is null)
-                throw new ArgumentNullException(nameof(unitSystem));
+            return ToUnit(Info.GetDefaultUnit(unitSystem));
+        }}
+");
```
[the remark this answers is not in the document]
lipchev: Yes, I think it might actually be better in terms of consistency - I hadn't thought about what a potential call to `Ratio.Info.GetDefaultUnit(UnitSystem.SI)` would yield..

Common/UnitDefinitions/CoefficientOfThermalExpansion.json
```diff
@@ -84,9 +78,6 @@
     {
       "SingularName": "PpmPerDegreeFahrenheit",
       "PluralName": "PpmPerDegreeFahrenheit",
-      "BaseUnits": {
-        "Θ": "DegreeFahrenheit"
-      },
```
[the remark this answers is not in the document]
angularsen: Ok so it seems that was not reason for removing these. So I guess duplicates are fine, but it also feels a bit icky that we don't control what units you get when requesting an SI unit, we just pick "the first one". 🤔 

If we had a policy of no duplicates, then at least it would be consistent what unit you get, but maybe there are good reasons for having multiple units specify the same base units?

Common/UnitDefinitions/BrakeSpecificFuelConsumption.json
```diff
@@ -4,6 +4,7 @@
   "XmlDocSummary": "Brake specific fuel consumption (BSFC) is a measure of the fuel efficiency of any prime mover that burns fuel and produces rotational, or shaft, power. It is typically used for comparing the efficiency of internal combustion engines with a shaft output.",
   "BaseDimensions": {
     "L": -2,
+    "M": 0,
     "T": 2
```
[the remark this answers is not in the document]
lipchev: I know, I've added it in a couple of places, where there was a hidden dimensions (I thought it could serve as a _tip_) - but I did also toyed with the idea of treating `null` and `0` differently..
I can remove them if you want (once you've reviewed everything)..

Common/UnitDefinitions/CoefficientOfThermalExpansion.json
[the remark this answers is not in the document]
lipchev: Yes, I've been thinking of different ways we could use the `BaseUnits` to test both the conversion coefficients and the provided `BaseUnits` (using the definitions from the summary comment) - but if we did that now, we'd have to account for the possible rounding errors (without a proper sense of scale etc..) - I think we're Ok for now so I'd say we do this later..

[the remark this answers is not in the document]
lipchev: No, we have to remove it, because the `PartsPerMillion` coefficient which multiplies the `BaseUnit`

Common/UnitDefinitions/Frequency.json
```diff
@@ -67,6 +76,9 @@
     {
       "SingularName": "BeatPerMinute",
       "PluralName": "BeatsPerMinute",
+      "BaseUnits": {
+        "T": "Minute"
+      },
```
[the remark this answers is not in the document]
lipchev: Only where the conversion coefficient of two duplicates isn't 1.

Common/UnitDefinitions/BrakeSpecificFuelConsumption.json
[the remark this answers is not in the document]
angularsen: I think it just adds confusion, at least for me I don't know it's there.

Common/UnitDefinitions/CoefficientOfThermalExpansion.json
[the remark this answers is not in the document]
angularsen: Ok, let's revisit later then

Common/UnitDefinitions/ElectricApparentEnergy.json
```diff
@@ -11,6 +11,11 @@
     {
       "SingularName": "VoltampereHour",
       "PluralName": "VoltampereHours",
+      "BaseUnits": {
+        "L": "Meter",
+        "M": "Kilogram",
+        "T": "Second"
```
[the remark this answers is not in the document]
angularsen: > Personally, I don't think that anyone should use the As(UnitSystem) as they are right now

I kind of agree. I believe @tmilnthorp has some insights here, but in my mind I don't yet see a compelling use case for the `UnitSystem` conversion methods and ctors - at least the simple ways I use this library, I have never needed it. 

But having definitions for SI base units seems useful and should allow us to infer stuff down the road.

UnitsNet/GeneratedCode/Quantities/Ratio.g.cs
lipchev: I think that both constructors should go, but since this would have been a _sudden_ breaking change I kept the one that could have actually be used by someone..

UnitsNet/GeneratedCode/Quantities/Ratio.g.cs
lipchev: I've actually dealt with this already- my upcoming release has this property marked as [Obsolete] and it now returns the `UnitConverter.Default` instance (which works for all conversions)... I've therefore completely removed, the need for `switching` and removed all of the reflection code..

Common/UnitDefinitions/ElectricApparentEnergy.json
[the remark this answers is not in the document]
lipchev: > I would imagine if I wrote an application, like Blender or Solidworks, I would have the user pick the unit of choice for known quantities like Length, Mass, and so on. This would affect how units were displayed, as well as inputs to set values.
> 
> I would then offer presets, something like:
> 
>     * `Metric [cm]`: Length = cm, Mass = kg, Time = s
> 
>     * `Metric [m]`: Length = m, Mass = kg, Time = s
> 
>     * `Imperial [ft]`: Length = ft, Mass = lbs, Time = s
> 
>     * `Imperial [in]`: Length = in, Mass = lbs, Time = s
> 
> 
> And this would all be configurable.
> 

Right, this was my original idea back in #709 where I was thinking about having the `QuantityInfo` / `UnitInfo` include a dictionary of default `UnitSystem` associations / configurations - which contain the `DefaultUnit` (resolving the possible duplicates) etc.. I was thinking of maybe re-opening this for `release-v9` 😄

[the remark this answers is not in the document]
lipchev: > > I would imagine if I wrote an application, like Blender or Solidworks, I would have the user pick the unit of choice for known quantities like Length, Mass, and so on. This would affect how units were displayed, as well as inputs to set values.
> > I would then offer presets, something like:
> > ```
> > * `Metric [cm]`: Length = cm, Mass = kg, Time = s
> > 
> > * `Metric [m]`: Length = m, Mass = kg, Time = s
> > 
> > * `Imperial [ft]`: Length = ft, Mass = lbs, Time = s
> > 
> > * `Imperial [in]`: Length = in, Mass = lbs, Time = s
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And this would all be configurable.
> 
> Right, this was my original idea back in #709 where I was thinking about having the `QuantityInfo` / `UnitInfo` include a dictionary of default `UnitSystem` associations / configurations - which contain the `DefaultUnit` (resolving the possible duplicates) etc.. I was thinking of maybe re-opening this for `release-v7` 😄

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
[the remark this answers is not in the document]
lipchev: Fixed in https://github.com/angularsen/UnitsNet/pull/1473/commits/4fc87b47ed74847fbb29eca5a936dab93107b4cc (I put it in the `QuantityTestsBase`, if you want a `TestConstants` class- pick the location)

Common/UnitDefinitions/BrakeSpecificFuelConsumption.json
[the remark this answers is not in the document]
lipchev: Fixed in https://github.com/angularsen/UnitsNet/pull/1473/commits/c326bbf51378dd2cb20c1b0417fab926e09d5a99

Common/UnitDefinitions/ElectricApparentEnergy.json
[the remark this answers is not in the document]
lipchev: I've removed the `BaseUnits` from the `ElectricApparentEnergy.VoltampereHour` and added "See about adding an SI unit (VoltampereSecond, Joules?)" as the [Skip] reason (similar to my previous comment/question in #1444 regarding the `ElectricReactiveEnergy`).
https://github.com/angularsen/UnitsNet/pull/1473/commits/93e940c89afb22349747bf4a05f375ef4ee802ec

UnitsNet.Tests/CustomCode/QuantityInfos/QuantityInfoExtensionsTest.cs
lipchev: Yes I know, initially this was returning the `VolumeUnit.ImperialPint` 😆 and I had left a comment about it, but then I went on and fixed the json files... 
Still, I don't care- let it break, if somebody adds a `VolumeUnit` starting with "A, B, or C", that has the same `BaseUnits` give me call, I'll fix the test 😎

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
lipchev: Fixed in https://github.com/angularsen/UnitsNet/pull/1473/commits/4fc87b47ed74847fbb29eca5a936dab93107b4cc (I put it in the `QuantityTestsBase`, if you want a `TestConstants` class- pick the location)

Common/UnitDefinitions/SpecificFuelConsumption.json
lipchev: I think the official SI unit here would be `GramPerNewtonSecond` - the `Mass` dimension is being simplified by the `Force`. Since we're left with a `1/Kilo` coefficient (and I didn't want to leave it undefined) - I've used the `Kilo` prefix of the `Time` dimension (I could have also picked the `LengthUnit` to scale on, but with an exponent of `"L": -1` this becomes a little harder to parse).

This is good example of the _choice_ I had to make in the (yet unpublished) fix for the generation of the prefix-bases - with more than one dimensions, there are often multiple possible combinations of prefixes to choose from, so I decided on the following order: "1, -1, 2, -2, ... n, -n" where the number represents a given dimension (e..g. if I have `"L" : 2, "T": -1` i would scale (if possible) using the `T` (as the scaling factor represents a _smaller deviation_).

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
angularsen: The locatino is fine.

Also, this is nitpicking, but I realize we could just also provide `new UnitSystem(UnsupportedBaseUnits)` in a similar `UnsupportedUnitSystem` singleton instance, since both UnitSystem and BaseUnits are basically immutable.

Common/UnitDefinitions/ElectricApparentEnergy.json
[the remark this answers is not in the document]
angularsen: Alright 👍

UnitsNet/GeneratedCode/Quantities/Ratio.g.cs
angularsen: Ok very nice

Common/UnitDefinitions/Pressure.json
lipchev: I believe the `PoundForce` has a _hidden_ coefficient representing the value of the earth's gravity, which makes it different from the `PoundPerInchSecondSquared` (or whatever the standard base) by a factor of `9.8`

UnitsNet.Tests/GeneratedCode/TestsBase/AccelerationTestsBase.g.cs
lipchev: Yeah, I know- but I wanted to have the type visible (otherwise it's not obvious what type we're giving as input).

Common/UnitDefinitions/SpecificFuelConsumption.json
angularsen: I still don't 100% follow here, but it kind of maybe makes sense. I'm lacking some basic understanding of how to think of this.

So, this is my first, uneducated attempt at breaking this down:

```
1 N = 1 kg⋅m⋅s −2

GramsPerKiloNewtonSecond =>
1g/(1000N * s) => 
1g/(1000kg⋅m⋅s−2 * s) => 
1g/(1000kg⋅m⋅s−1) =>
(1g * s) / (1000kg⋅m) =>
(1000 * 1g * s) / (kg⋅m) =>
(1kg * s) / (kg⋅m) =>
1s / m
```

So.. I'm probably way off here, but I arrive at 1 second per meter in SI base units.
Is this the way to go about it?

UnitsNet.Tests/CustomCode/QuantityInfos/QuantityInfoExtensionsTest.cs
angularsen: This comment will come back to haunt you 😆

UnitsNet/GeneratedCode/Quantities/Ratio.g.cs
angularsen: Alright

Common/UnitDefinitions/Pressure.json
angularsen: I'm not arguging with that

UnitsNet.Tests/CustomCode/QuantityInfos/QuantityInfoExtensionsTest.cs
lipchev: Right, don't think I'll be loosing much sleep over it..

angularsen:
@lipchev As far as I'm concerned, this is ready to go. If you want to make any last minute changes, just merge whenever you want 👍

angularsen (approved):
Looks good to me, merge whenever you are ready

Common/UnitDefinitions/SpecificFuelConsumption.json
lipchev: > I still don't 100% follow here, but it kind of maybe makes sense. I'm lacking some basic understanding of how to think of this.
> 
> So, this is my first, uneducated attempt at breaking this down:
> 
> ```
> 1 N = 1 kg⋅m⋅s −2
> 
> GramsPerKiloNewtonSecond =>
> 1g/(1000N * s) => 
> 1g/(1000kg⋅m⋅s−2 * s) => 
> 1g/(1000kg⋅m⋅s−1) =>
> (1g * s) / (1000kg⋅m) =>
> (1000 * 1g * s) / (kg⋅m) =>
> (1kg * s) / (kg⋅m) =>
> 1s / m
> ```
> 
> So.. I'm probably way off here, but I arrive at 1 second per meter in SI base units. Is this the way to go about it?

Yes, you're totally right- it's always the `Kilogram` that messes up my Si unit estimations...

lipchev: Fixed in https://github.com/angularsen/UnitsNet/pull/1473/commits/c32d2ffd787d5a039007aab8ee19af7f10a8bd89

angularsen: Ok nice, I learned something new today then 🎉

PR #1474, opened 2024-12-27
Added fluid resistance quantity

Jagailo:
https://en.wikipedia.org/wiki/Vascular_resistance

1. I know that negative powers are not welcome, but for dyn·s/cm⁵ and dyn·s·cm⁻⁵ units I added 2 options because both are equally common.
2. The unit MPa·s/m³ comes with the mega- prefix because that is the only combination that is used. I have not seen it as just Pa·s/m³ or in any other form.
3. WhodUnit actually has 3 different abbreviations, but they all mean the same thing. For mmHg·min/l, translating popular units into Russian was not difficult, but for WU and HRU I had to improvise. WU - Wood Units, which in turn is translated as "Единицы Вуда" (Units of Wood), so two words in the Russian version have changed places, and the capital letters form ЕВ (WU = ЕВ). For HRU I translated Hybrid Resistance Unit and used the capital letters of the Russian version. Translated as "Единицы Гибридного Сопротивления" (HRU = ЕГС)


Common/UnitDefinitions/VascularResistance.json
```diff
@@ -0,0 +1,76 @@
+﻿{
+  "Name": "VascularResistance",
+  "BaseUnit": "DyneSecondPerCentimeterToTheFifth",
+  "XmlDocSummary": "Vascular resistance is the resistance that must be overcome for blood to flow through the circulatory system.",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Vascular_resistance",
```
angularsen, line 5: This quantity on the surface feels very domain specific, and I am a little hesitant to add it as we strive to balance adding the most widely used units that many benefit from, without bloating the library with exotic units. There are extremely many domains in the world, with their own units. 

Could this be generalized into fluid resistance maybe? It seems to have the same dimensions of the units at least.

https://www.vcalc.com/wiki/Fluid-Resistance-Unit-Conversion

Jagailo: Sounds logical, I'll add it sometime after the holidays.

Jagailo:
I have no more ideas why base SI doesn't work, so I removed these tests.

image

UnitsNet.Tests/CustomCode/FluidResistanceTests.cs
```diff
@@ -0,0 +1,49 @@
...
+using System;
+
+namespace UnitsNet.Tests.CustomCode
+{
+    public class FluidResistanceTests : FluidResistanceTestsBase
+    {
+        // https://www.vcalc.com/wiki/Fluid-Resistance-Unit-Conversion
+
+        protected override bool SupportsSIUnitSystem => false;
+        protected override double DyneSecondsPerCentimeterToTheFifthInOnePascalSecondPerCubicMeter => 1e5; // dyn·s/cm⁵
```
angularsen, line 29: Shouldn't this be 1e-5?

```
1 Pa*s/m3 = 1e-6 MPa*s/m3

8 MPa*s/m3 = 80 dyn*s/cm5
1 MPa*s/m3 = 10 dyn*s/cm5

1e-6 MPa*s/m3 = 10 * 1e-6 dyn*s/cm5 = 1e-5 dyn*s/cm5
```

https://en.wikipedia.org/wiki/Vascular_resistance
image

UnitsNet.Tests/CustomCode/FluidResistanceTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        protected override double PascalMinutesPerCubicCentimeterInOnePascalSecondPerCubicMeter => 1.66666666666667e-8; // Pa·min/cm³
+        protected override double PascalMinutesPerCubicMeterInOnePascalSecondPerCubicMeter => 1.66666666666667e-2; // Pa·min/m³
+        protected override double PascalMinutesPerLiterInOnePascalSecondPerCubicMeter => 1.66666666666667e-5; // Pa·min/l
+        protected override double PascalMinutesPerMilliliterInOnePascalSecondPerCubicMeter => 1.66666666666667e-8; // Pa·min/ml
+        protected override double PascalSecondsPerCubicCentimeterInOnePascalSecondPerCubicMeter => 1e-6; // Pa·s/cm³
+        protected override double PascalSecondsPerCubicMeterInOnePascalSecondPerCubicMeter => 1; // Pa·s/m³
+        protected override double PascalSecondsPerLiterInOnePascalSecondPerCubicMeter => 1e-3; // Pa·s/l
+        protected override double PascalSecondsPerMilliliterInOnePascalSecondPerCubicMeter => 1e-6; // Pa·s/ml
+        protected override double WoodUnitsInOnePascalSecondPerCubicMeter => 1.25010625903202e-7; // WU
```
angularsen, line 47: All the others are verified at
https://www.vcalc.com/wiki/Fluid-Resistance-Unit-Conversion

angularsen (changes requested):
Looks good, just one question

UnitsNet.Tests/CustomCode/FluidResistanceTests.cs
Jagailo: true

Jagailo:
@angularsen Sorry for pinging you, but is everything fine with this PR? Do I need to change anything else?

angularsen:
No problem, it helps with a reminder if it takes more than a few days. Weekdays are kinda busy for me, but I usually find time in the weekends for small PRs like this.

Thanks! Nuget should be on the way.


[Release UnitsNet/5.68.0 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.68.0)

PR #1475, opened 2024-12-27
🔥Remove static Default getters, obsoleted by UnitsNetSetup

angularsen:
Ref #1200 

Remove static `Default` getter property for `UnitConverter`, `UnitAbbreviationsCache`, `QuantityParser`, `UnitParser`. These just pointed to the singleton `UnitsNetSetup.Default` and were made obsolete in v5.

### DIFF
--- a/UnitsNet/CustomCode/QuantityParser.cs
+++ b/UnitsNet/CustomCode/QuantityParser.cs
@@ -35,12 +35,6 @@ namespace UnitsNet
         private readonly UnitParser _unitParser;
 
         /// <summary>
-        ///     The default instance of <see cref="QuantityParser"/>, which uses <see cref="UnitsNetSetup"/>.<see cref="UnitsNetSetup.Default"/>.<see cref="UnitsNetSetup.UnitAbbreviations"/>.
-        /// </summary>
-        [Obsolete("Use UnitsNetSetup.Default.QuantityParser instead.")]
-        public static QuantityParser Default => UnitsNetSetup.Default.QuantityParser;
-
-        /// <summary>
         ///     Creates an instance of <see cref="QuantityParser"/>, optionally specifying an <see cref="UnitAbbreviationsCache"/>
         ///     with unit abbreviations to use when parsing.
         /// </summary>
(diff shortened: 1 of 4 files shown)


lipchev:
@angularsen Sorry, I've just gotten around to review these recent changes.. I know these had been made [Obsolete] for a while now, but are you sure there is sufficient reason to remove them? 
1. As long as the dependency between the types is correct (_it is_) then there shouldn't be any harm in having them (_do you see something I don't_?)
2. Even if we instantiate and use the concrete type, it's `Default` property isn't going to trigger the static constructor on the `UnitsNetSetup`, unless explicitly accessed
3. I think it's pretty clear that these are simply shortcuts to `UnitsNetSetup.Default.XXX` - I personally find `UnitParser.Default.Parse(..)` to be much more intuitive than `UnitsNetSetup.Default.UnitParser.Parse(..)`

angularsen:
I had the same thought, I don't have a strong opinion here. My thinking was that it may be better to have a single way to access these things, for docs/sample code and what not, but the syntax becomes more more clunky.

The main benefit of moving things into `UnitsNetSetup` was ensuring the same instance of `UnitAbbreviationCache` was used, but there is no harm in having these getter properties as convenience.

I'm fine reverting this, but we should then remove the obsolete flags in v5 too.

angularsen:
Sorry for not waiting for review and rushing a bit, but I have a little bit of extra time in the holidays and just want to get things done 😄 We can always revert, I'm glad to have a pair of extra eyes on these things 🙇

lipchev:
Well, it was on the wish-list- I hadn't gotten around to arguing against it :)

I wasn't 100% sure whether they are safe in `v5` but I was kinda hoping to remove the [Obsolete] tag in v6 (where I'm pretty sure it's all good).

Plus I'm sure there are many holdouts who ignored the [Obsolete] flag :)

lipchev:
> Plus I'm sure there are many holdouts who ignored the [Obsolete] flag :)

[Here](https://github.com/angularsen/UnitsNet/blob/master/Samples/UnitConverter.Wpf/UnitConverter.Wpf/UnitListItem.cs)'s one 🤣

PR #1484, opened 2024-12-30
🚸Describe why Parse fails due to case-insensitive ambiguity

### LINKED ISSUE #1423: Quantity.TryParse won't parse "10 MM" because "MM" is capitalized. Other abbreviations work fine.
**Describe the bug**
`Quantity.TryParse(type, string, out IQuantity value)` fails to parse Length strings with capital MM (i.e., "10 MM" should parse just as easily as "10 mm"). Other abbreviations appear to work, regardless of casing.

**To Reproduce**
See .NET Fiddle here: http://dotnetfiddle.net/3JmUsw

**Expected behavior**
The string should be converted to a Length based on a millimeter value

**Additional context**
I believe this _used_ to work correctly, but I'm not certain.

Thanks for this library, it's a massive time saver!

angularsen:
Fixes #1423

`UnitParser.Parse<LengthUnit>("MM")` fails due to matching both `Megameter` and `Millimeter` in case-insensitive matching, but matches neither of them in the case-sensitive fallback. It was confusing to get `UnitsNotFoundException` in this case, since case-insensitive usually works for most units.

### Changes
- Handle this case and throw `AmbiguousUnitParseException` instead of `UnitNotFoundException`
- Describe the case-insensitive units that matched
- Fix existing test `Parse_WithMultipleCaseInsensitiveMatchesButNoExactMatches_ThrowsUnitNotFoundException`
- Skip retrying with fallback culture if no specific `formatProvider` was given

### DIFF
--- a/UnitsNet.Serialization.JsonNet/AbbreviatedUnitsConverter.cs
+++ b/UnitsNet.Serialization.JsonNet/AbbreviatedUnitsConverter.cs
@@ -190,7 +190,7 @@ namespace UnitsNet.Serialization.JsonNet
         }
 
         /// <summary>
-        ///     Attempt to find an a unique (non-ambiguous) unit matching the provided abbreviation.
+        ///     Attempt to find a unique (non-ambiguous) unit matching the provided abbreviation.
         ///     <remarks>
         ///         An exhaustive search using all quantities is very likely to fail with an
         ///         <exception cref="AmbiguousUnitParseException" />, so make sure you're using the minimum set of supported quantities.
--- a/UnitsNet.Tests/QuantityParserTests.cs
+++ b/UnitsNet.Tests/QuantityParserTests.cs
@@ -1,4 +1,4 @@
-// Licensed under MIT No Attribution, see LICENSE file at the root.
+﻿// Licensed under MIT No Attribution, see LICENSE file at the root.
 // Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
 
 using UnitsNet.Tests.CustomQuantities;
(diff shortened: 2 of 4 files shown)


angularsen:
@lipchev relevant to your exception helpers PR, the exception message can be improved further to include information about the quantity and unit enum type, which I believe you already did there.

lipchev:
> @lipchev relevant to your exception helpers PR, the exception message can be improved further to include information about the quantity and unit enum type, which I believe you already did there.

Yes, I was also going to bring this is up (_eventually_) - if we want to go wild we could even include extra fields like in these examples:

``` csharp
               throw new UnitNotFoundException($"No unit was found for quantity '{quantityName}' with the name: '{unitName}'.")
               {
                   Data = { ["quantityName"] = quantityName, ["unitName"] = unitName }
               };
```

``` csharp
        throw new UnitNotFoundException($"No quantity was found with the specified unit type: '{unitType}'.") { Data = { ["unitType"] = unitType.Name } };
```

angularsen:
Fixed test case that broken in CI, due to different current culture: def3211200f240720e1d58b8eefec1ebfddd61b6

https://dev.azure.com/unitsnet/Units.NET/_build/results?buildId=1004&view=ms.vss-test-web.build-test-results-tab&runId=245&resultId=116706&paneView=debug

angularsen:
@lipchev Are you still reviewing this or can I merge?

lipchev:
> @lipchev Are you still reviewing this or can I merge?

Well, my earlier comment wasn't about the test that broke the build - it was a bit more subtle than that..
Nevertheless, I would say "merge it" as fixing _that very unlikely_ case would be an unpleasant endeavor (and testing it would probably require setting up custom abbreviation mappings etc.), so I think it would be easier if we just add it like it is and try to fix it _better_ in `v6`.

angularsen:
Gotcha, I'll merge this then.

PR #1485, opened 2024-12-30
BaseUnit generation for the prefixed units

lipchev:
- added a new  `UnitPrefixBuilder` (called from the `QuantityJonsFilesParser`)
- removed the skipped tests (with the "The BaseUnits are not yet supported by the prefix-generator" reason)
- added a `DebugDisplay` for the `Quantity` and the `Unit` (JsonTypes)  and 
- added a `ToString` implementation to the `BaseDimensions` (JsonTypes)

https://en.wikipedia.org/wiki/Metric_prefix


CodeGen/Generators/QuantityJsonFilesParser.cs
```diff
@@ -124,5 +126,246 @@ private static Localization[] GetLocalizationForPrefixUnit(IEnumerable<Localizat
...
+            yield return new KeyValuePair<string, Prefix[]>("Foot", [Prefix.Kilo]);
+            yield return new KeyValuePair<string, Prefix[]>("Parsec", [Prefix.Kilo, Prefix.Mega]);
+            yield return new KeyValuePair<string, Prefix[]>("LightYear", [Prefix.Kilo, Prefix.Mega]);
+        }
+
+        private static IEnumerable<KeyValuePair<string, Prefix[]>> GetDurationPrefixes()
+        {
+            yield return new KeyValuePair<string, Prefix[]>("Second", [Prefix.Nano, Prefix.Micro, Prefix.Milli]);
+        }
     }
```
lipchev, line 370: This is the most iffy part- I had to hard-code the prefixes that are applicable in the base quantities.. Those aren't likely to change, but if we want to make this _properly_ we either need to do pre-parse the 'Length.json', 'Mass.json' etc in order to get their prefixes before the start of the generation (or think of other ways of storing them).

CodeGen/Generators/QuantityJsonFilesParser.cs
```diff
@@ -124,5 +126,246 @@ private static Localization[] GetLocalizationForPrefixUnit(IEnumerable<Localizat
...
+        
+        /// <summary>
+        /// A dictionary that maps metric prefixes to their corresponding exponent values.
+        /// </summary>
+        /// <remarks>
+        /// This dictionary excludes binary prefixes such as Kibi, Mebi, Gibi, Tebi, Pebi, and Exbi.
+        /// </remarks>
+        private static readonly Dictionary<Prefix, int> PrefixFactors = PrefixInfo.Entries
+            .Where(x => x.Key is not (Prefix.Kibi or Prefix.Mebi or Prefix.Gibi or Prefix.Tebi or Prefix.Pebi or Prefix.Exbi)).ToDictionary(pair => pair.Key,
+                pair => (int)Math.Log10(double.Parse(pair.Value.Factor.TrimEnd('d'), NumberStyles.Any, CultureInfo.InvariantCulture)));
```
lipchev, line 295: I was a little worried when casting to `(int)` but I've checked the results, there are no rounding errors..
Still, we could simply expose these as properties on the `Prefix` type.

lipchev: > I was a little worried when casting to `(int)` but I've checked the results, there are no rounding errors.. Still, we could simply expose these as properties on the `Prefix` type.

Although these might have to be two properties (given that we have both binary and decimal prefixes)..

CodeGen/Generators/QuantityJsonFilesParser.cs
```diff
@@ -124,5 +126,246 @@ private static Localization[] GetLocalizationForPrefixUnit(IEnumerable<Localizat
...
+            foreach (var degree in dimensions.GetNonZeroDegrees().OrderBy(int.Abs).ThenByDescending(x => x))
+            {
+                if (TryPrefixWithDegree(dimensions, oldUnits, prefixInfo.Prefix, degree, out BaseUnits? prefixedUnits))
+                {
+                    return prefixedUnits;
+                }
+            }
+            
+            return null; 
+        }
```
lipchev, line 147: Note that when we have multiple dimensions present, there are often multiple combinations of base units that can fit a given prefix scale. This algorithm tries the smallest dimensions first, with a preference for the positive exponents.

Technically, if no exact scaling factor exists for a given quantity, having more than one dimensions- there could be another iteration after the `foreach` loop that tries to simultaneously change _two dimensions_ with a single prefix - but that would have really made my head hurt..

CodeGen/Generators/QuantityJsonFilesParser.cs
```diff
@@ -124,5 +126,254 @@ private static Localization[] GetLocalizationForPrefixUnit(IEnumerable<Localizat
...
+        /// <param name="prefix"></param>
+        /// <param name="degree"></param>
+        /// <param name="prefixedBaseUnits"></param>
+        /// <returns></returns>
+        private static bool TryPrefixWithDegree(BaseDimensions dimensions, BaseUnits baseUnits, Prefix prefix, int degree, [NotNullWhen(true)] out BaseUnits? prefixedBaseUnits)
+        {
+            prefixedBaseUnits = baseUnits.Clone();
+
+            // look for a dimension that is part of the non-zero exponents
+            if (baseUnits.N is { } baseAmountUnit && dimensions.N == degree)
```
angularsen, line 205: I took the liberty of renaming a few vars/params for my own understanding, take a look if you agree with it
99b378f9269567018357add0ce6518d534cc1b72

angularsen (changes requested):
This seems like it works as advertised, I've checked quite a few of the generated units and they all look good to me.

The feedback is mainly about me struggling to truly understand it and that we should try to document it a bit.

CodeGen/Generators/QuantityJsonFilesParser.cs
```diff
@@ -124,5 +126,246 @@ private static Localization[] GetLocalizationForPrefixUnit(IEnumerable<Localizat
...
+                if (TryPrefixWithDegree(dimensions, oldUnits, prefixInfo.Prefix, degree, out BaseUnits? prefixedUnits))
+                {
+                    return prefixedUnits;
+                }
+            }
+            
+            return null; 
+        }
+
+        private static IEnumerable<int> GetNonZeroDegrees(this BaseDimensions dimensions)
```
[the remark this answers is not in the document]
lipchev: What, you think I haven't been making shit up along the way? 😄

CodeGen/Generators/QuantityJsonFilesParser.cs
lipchev: Yes, I think I'll try to refactor this all into a pre-processing step for the time being, but generally speaking - I was thinking that, one day, we should maybe flattening out all of the json files:
- it's fairly easy to do: once everything is generated just re-write all json files (this should flatten out both the `Abbreviations` and the `BaseUnits`)
- this should make it easier for other parsers/generators to work with the json files
- make it slightly harder to add new quantities / units - but that shouldn't happen as often as before (especially if we could make it easy to extend on the client-side)

lipchev: Or we could do something like the `UnitRelations.json` which is overridden every time.

UnitsNet/GeneratedCode/Quantities/AbsorbedDoseOfIonizingRadiation.g.cs
```diff
@@ -76,22 +76,22 @@ static AbsorbedDoseOfIonizingRadiation()
...
+                    new UnitInfo<AbsorbedDoseOfIonizingRadiationUnit>(AbsorbedDoseOfIonizingRadiationUnit.Centigray, "Centigrays", new BaseUnits(length: LengthUnit.Decimeter, time: DurationUnit.Second), "AbsorbedDoseOfIonizingRadiation"),
                     new UnitInfo<AbsorbedDoseOfIonizingRadiationUnit>(AbsorbedDoseOfIonizingRadiationUnit.Femtogray, "Femtograys", BaseUnits.Undefined, "AbsorbedDoseOfIonizingRadiation"),
                     new UnitInfo<AbsorbedDoseOfIonizingRadiationUnit>(AbsorbedDoseOfIonizingRadiationUnit.Gigagray, "Gigagrays", BaseUnits.Undefined, "AbsorbedDoseOfIonizingRadiation"),
                     new UnitInfo<AbsorbedDoseOfIonizingRadiationUnit>(AbsorbedDoseOfIonizingRadiationUnit.Gray, "Grays", new BaseUnits(length: LengthUnit.Meter, time: DurationUnit.Second), "AbsorbedDoseOfIonizingRadiation"),
                     new UnitInfo<AbsorbedDoseOfIonizingRadiationUnit>(AbsorbedDoseOfIonizingRadiationUnit.Kilogray, "Kilograys", BaseUnits.Undefined, "AbsorbedDoseOfIonizingRadiation"),
```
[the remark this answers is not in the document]
lipchev: Yes this a bit of an inconvenience, but if (in the future) we added the possibility to specify the _scaling factor_ then we should be able to express ***everything*** in terms of the `BaseUnits * ScalingFactor` - which would make the `FromBase/ToBase` expressions unnecessary (although for us, _regular humans_, I bet this is likely to be harder to work with)..

CodeGen/Generators/QuantityJsonFilesParser.cs
lipchev: I haven't seen anything that suggests a particular rule, but I think the overarching objective would be to minimize the scaling factor. The current algorithm is not optimal in that regard but happens to produce near-optimal results. I don't have a concrete example, but I assume it's possible for a dimension of degree 1 to jump from `Micro` to `Giga` where the dimension with a larger exponent would have selected a _smaller jump_ such as `Micro` -> `Milli`. 

Another way to think of this is in terms of the scales of the quantities that would be produced by an operation involving the base units: _I've measured `1 g`, my flask is `1 ml`- so my density is `1 g/ml`_ (with the division producing `1 g` or `1 ml` respectively. In the absence of the ideal `g/ml` combination, then the next-best combination would be the the one that produces the smallest jump in the the `Value` of the resulting division:  `10 xx` and `0.1 xx` are clearly _better_ options then `1e9 xx` and `1e-9 xx`.
 
Figuring out the optimal solution to this problem with up to 7 dimensions and all their possible combinations isn't likely to be a trivial task (although as usual, there is probably some simple matrix multiplication that solves it in one go 😄 )

CodeGen/Generators/QuantityJsonFilesParser.cs
```diff
@@ -124,5 +126,246 @@ private static Localization[] GetLocalizationForPrefixUnit(IEnumerable<Localizat
...
+                return null;
+            }
+
+            // iterate the non-zero dimensions in the order [1, -1, 2, -2...n, -n]
+            foreach (var degree in dimensions.GetNonZeroDegrees().OrderBy(int.Abs).ThenByDescending(x => x))
+            {
+                if (TryPrefixWithDegree(dimensions, oldUnits, prefixInfo.Prefix, degree, out BaseUnits? prefixedUnits))
+                {
+                    return prefixedUnits;
+                }
```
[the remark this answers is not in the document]
lipchev: I've included these examples in the xml-docs - it now requires a table of contents to navigate through it, but doesn't matter:  the more there is for the AI to read, the better.. 😃

lipchev:
So I refactored the code, grouping most of the functionality w.r.t. to prefixed-unit-generation as its own thing (`UnitPrefixBuilder`). 

The prefixes for the base quantities are loaded first (no more hard-coding), before generating the prefixed-units for the complete list of quantities.

Nothing changed in the generated files, so I guess that's good news.. I could experiment with a different algorithm later, but I think we already have plenty to work with..

CodeGen/Helpers/PrefixBuilder/UnitPrefixBuilder.cs
```diff
@@ -0,0 +1,334 @@
...
+    ///                         ❌ There is no metric prefix we can raise to the power of 2 and get Kilo, e.g., Deca*Deca = Hecto, Kilo*Kilo = Mega, etc.
+    ///                     Trying third dimension T=-3:
+    ///                         SI base unit in time dimension is Second, same as configured base unit, so base prefix scale is 0
+    ///                         Inferred prefix is Deci: (base prefix scale 0 + requested prefix scale (-3)) / exponent -3 = -3 / -3 = 1
+    ///                         ❌ There is no Duration unit Decasecond (we could add it)
+    ///                 </code>
+    ///             </description>
+    ///         </item>
+    ///     </list>
+    /// </example>
```
angularsen, line 178: Poor soul of the next guy trying to digest this, but I can't wait for the bots to pick up my made up words for these concepts 🤣

angularsen (approved):
Valiant effort! And good news the codegen turned out the same results, it's kind of a nice sanity check in all the madness.

PR #1486, opened 2025-01-01
Update the ToString format of the BaseUnits/Dimensions

lipchev:
- `BaseUnits`: no longer using the `AbbeviationsCache`, the new format is `L=Meter, M=Kilogram, T=Second` 
- `BaseDimensions`: the exponent moved inside the dimension-brackets: `[Length][Time^-1]`
- `BaseDimensions`: minor performance improvements

As mentioned in #1452, the main motivation here is the removal of the potential side effects of accessing/loading the unit abbreviations (e.g. during the `QuantityInfo` construction)

### DIFF
--- a/UnitsNet.Tests/BaseDimensionsTests.cs
+++ b/UnitsNet.Tests/BaseDimensionsTests.cs
@@ -74,7 +74,7 @@ namespace UnitsNet.Tests
         {
             Assert.False(Acceleration.BaseDimensions.IsBaseQuantity());
         }
-
+        
         [Theory]
         [InlineData(2, 0, 0, 0, 0, 0, 0)]
         [InlineData(0, 2, 0, 0, 0, 0, 0)]
@@ -733,13 +733,13 @@ namespace UnitsNet.Tests
         [Fact]
         public void CheckToStringUsingMolarEntropy()
         {
-            Assert.Equal("[Length]^2[Mass][Time]^-2[Temperature]^-1[Amount]^-1", MolarEntropy.BaseDimensions.ToString());
+            Assert.Equal("[Length^2][Mass][Time^-2][Temperature^-1][Amount^-1]", MolarEntropy.BaseDimensions.ToString());
         }
 
         [Fact]
         public void CheckToStringUsingSpeed()
         {
-            Assert.Equal("[Length][Time]^-1", Speed.BaseDimensions.ToString());
+            Assert.Equal("[Length][Time^-1]", Speed.BaseDimensions.ToString());
         }
 
         [Fact]
@@ -782,5 +782,6 @@ namespace UnitsNet.Tests
             // Example case
             Assert.True(Level.BaseDimensions.IsDimensionless());
         }
+
     }
 }
--- a/UnitsNet.Tests/BaseUnitsTests.cs
+++ b/UnitsNet.Tests/BaseUnitsTests.cs
@@ -139,16 +139,16 @@ namespace UnitsNet.Tests
                 TemperatureUnit.Kelvin,
                 AmountOfSubstanceUnit.Mole,
                 LuminousIntensityUnit.Candela);
-            
-            var m = LengthUnit.Meter;
-            var kg = MassUnit.Kilogram;
-            var s = DurationUnit.Second;
-            var A = ElectricCurrentUnit.Ampere;
-            var K = TemperatureUnit.Kelvin;
-            var mol = AmountOfSubstanceUnit.Mole;
-            var cd = LuminousIntensityUnit.Candela;
 
-            Assert.Equal($"[Length]: {m}, [Mass]: {kg}, [Time]: {s}, [Current]: {A}, [Temperature]: {K}, [Amount]: {mol}, [LuminousIntensity]: {cd}", siBaseUnits.ToString());
+            Assert.Equal("L=Meter, M=Kilogram, T=Second, I=Ampere, Θ=Kelvin, N=Mole, J=Candela", siBaseUnits.ToString());
+        }
+
+        [Fact]
+        public void ToString_WithFewerDimensions_ContainsOnlyTheSpecifiedDimensions()
+        {
+            var siBaseUnits = new BaseUnits(length:LengthUnit.Meter, mass:MassUnit.Gram, time:DurationUnit.Second, temperature:TemperatureUnit.DegreeCelsius);
+
+            Assert.Equal("L=Meter, M=Gram, T=Second, Θ=DegreeCelsius", siBaseUnits.ToString());
         }
 
         [Fact]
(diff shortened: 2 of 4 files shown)


lipchev:
@angularsen  Happy new year, here's an easy one which we've talked about in #1452 

I wasn't 100% sure back then (and I never actually meant for it to be _optimal_), but I did a quick benchmark and the `string.Join` variant is just as fast (on net8.0) as the the `StringBuilder` with `AppendFormat` (but using much less allocations). The only variant in which the `StringBuilder` was slightly faster (but still more allocations) was this one (_AI generated, haven't actually verified the output_):

```  csharp
public string ToStringWithAppend()
{
    if (Equals(Undefined))
    {
        return "Undefined";
    }
    var sb = new StringBuilder();
    if (Length is not null)
    {
        sb.Append("[Length]: ").Append(Length).Append(", ");
    }
    if (Mass is not null)
    {
        sb.Append("[Mass]: ").Append(Mass).Append(", ");
    }
    if (Time is not null)
    {
        sb.Append("[Time]: ").Append(Time).Append(", ");
    }
    if (Current is not null)
    {
        sb.Append("[Current]: ").Append(Current).Append(", ");
    }
    if (Temperature is not null)
    {
        sb.Append("[Temperature]: ").Append(Temperature).Append(", ");
    }
    if (Amount is not null)
    {
        sb.Append("[Amount]: ").Append(Amount).Append(", ");
    }
    if (LuminousIntensity is not null)
    {
        sb.Append("[LuminousIntensity]: ").Append(LuminousIntensity).Append(", ");
    }
    if (sb.Length > 2)
    {
        sb.Length -= 2; // Remove the trailing ", "
    }
    return sb.ToString();
}

```

UnitsNet.Tests/BaseDimensionsTests.cs
```diff
@@ -733,13 +733,13 @@ public void CheckBaseDimensionMultiplicationWithForceEqualsMassTimesAcceleration
...
         {
-            Assert.Equal("[Length]^2[Mass][Time]^-2[Temperature]^-1[Amount]^-1", MolarEntropy.BaseDimensions.ToString());
+            Assert.Equal("[Length^2][Mass][Time^-2][Temperature^-1][Amount^-1]", MolarEntropy.BaseDimensions.ToString());
         }
 
         [Fact]
         public void CheckToStringUsingSpeed()
         {
-            Assert.Equal("[Length][Time]^-1", Speed.BaseDimensions.ToString());
+            Assert.Equal("[Length][Time^-1]", Speed.BaseDimensions.ToString());
```
lipchev, line 742: We haven't really discussed such change, but I personally find it slightly more readable. 
Not a strong preference, of course.. 😄

angularsen: Fine by me 👍

UnitsNet.Tests/BaseUnitsTests.cs
```diff
@@ -139,19 +139,24 @@ public void ToStringGivesExpectedResult()
...
+            
+            var m = LengthUnit.Meter;
+            var kg = MassUnit.Kilogram;
+            var s = DurationUnit.Second;
+            var A = ElectricCurrentUnit.Ampere;
+            var K = TemperatureUnit.Kelvin;
+            var mol = AmountOfSubstanceUnit.Mole;
+            var cd = LuminousIntensityUnit.Candela;
 
             Assert.Equal($"[Length]: {m}, [Mass]: {kg}, [Time]: {s}, [Current]: {A}, [Temperature]: {K}, [Amount]: {mol}, [LuminousIntensity]: {cd}", siBaseUnits.ToString());
```
angularsen, line 151: I don't really care much about this, but an alternative format to consider is `L=Meter, M=Kilogram, T=Second, `... etc.

UnitsNet/BaseDimensions.cs
```diff
@@ -30,9 +29,13 @@ public BaseDimensions(int length, int mass, int time, int current, int temperatu
...
-            bool onlyOneEqualsOne = dimensionsArray.Select(dimension => dimension is 0 or 1 ? dimension : 2).Sum() == 1;
-            return onlyOneEqualsOne;
+            return (Length == 1 && Mass == 0 && Time == 0 && Current == 0 && Temperature == 0 && Amount == 0 && LuminousIntensity == 0) ||
+                   (Length == 0 && Mass == 1 && Time == 0 && Current == 0 && Temperature == 0 && Amount == 0 && LuminousIntensity == 0) ||
+                   (Length == 0 && Mass == 0 && Time == 1 && Current == 0 && Temperature == 0 && Amount == 0 && LuminousIntensity == 0) ||
+                   (Length == 0 && Mass == 0 && Time == 0 && Current == 1 && Temperature == 0 && Amount == 0 && LuminousIntensity == 0) ||
+                   (Length == 0 && Mass == 0 && Time == 0 && Current == 0 && Temperature == 1 && Amount == 0 && LuminousIntensity == 0) ||
+                   (Length == 0 && Mass == 0 && Time == 0 && Current == 0 && Temperature == 0 && Amount == 1 && LuminousIntensity == 0) ||
+                   (Length == 0 && Mass == 0 && Time == 0 && Current == 0 && Temperature == 0 && Amount == 0 && LuminousIntensity == 1);
```
angularsen, line 38: LINQ was kinda overkill here, I agree

UnitsNet.Tests/BaseDimensionsTests.cs
```diff
@@ -733,13 +733,13 @@ public void CheckBaseDimensionMultiplicationWithForceEqualsMassTimesAcceleration
         [Fact]
         public void CheckToStringUsingMolarEntropy()
         {
-            Assert.Equal("[Length]^2[Mass][Time]^-2[Temperature]^-1[Amount]^-1", MolarEntropy.BaseDimensions.ToString());
+            Assert.Equal("[Length^2][Mass][Time^-2][Temperature^-1][Amount^-1]", MolarEntropy.BaseDimensions.ToString());
```
angularsen, line 736: I don't recall exactly where I saw this, but during debugging #1485  I saw values like `L^2MT^-2`, which was a nice and short presentation also I think.

I have no preference here.

angularsen (approved):
Looks good to me, a couple of suggestions

UnitsNet.Tests/BaseDimensionsTests.cs
lipchev: Yes, I also vaguely remember encountering something like `L^2Θ` (and thinking that I need to fix the `ToString`, but then forgot and so #1485 now has the _simpler_ format). 🤷

angularsen:
> @angularsen Happy new year, here's an easy one which we've talked about in #1452

I meant to answer and forgot, happy new year @lipchev 🎉

I'm very glad you are around to learn from, and discuss and share ideas with, I really appreciate it ❤️  
Also your energy and brilliance is exactly what this project needs to move things forward, I do not have as much time to spend on it as I would like to 🙇

PR #1488, opened 2025-01-03
Ensure that all base quantity units are BaseUnits

lipchev:
Ensure that all base quantity units are `BaseUnits`:
- introduced a test for each of the base quantities (`Length`, `Mass` etc.) that ensures that all of their units have valid `BaseUnits` definition
- fixed some incorrect `BaseUnits` definitions: `Acceleration.StandardGravity` and `Temperature.MillidegreeCelsius`
- added missing `BaseUnits` definitions: `Length`, `Temperature` and `TemperatureChangeRate` (removing 3 out of its 4 `[Skip]`-ed tests)


UnitsNet/GeneratedCode/Quantities/TemperatureChangeRate.g.cs
```diff
@@ -79,15 +79,15 @@ static TemperatureChangeRate()
...
+                    new UnitInfo<TemperatureChangeRateUnit>(TemperatureChangeRateUnit.DegreeKelvinPerMinute, "DegreesKelvinPerMinute", new BaseUnits(time: DurationUnit.Minute, temperature: TemperatureUnit.Kelvin), "TemperatureChangeRate"),
+                    new UnitInfo<TemperatureChangeRateUnit>(TemperatureChangeRateUnit.DegreeKelvinPerSecond, "DegreesKelvinPerSecond", new BaseUnits(time: DurationUnit.Second, temperature: TemperatureUnit.Kelvin), "TemperatureChangeRate"),
                     new UnitInfo<TemperatureChangeRateUnit>(TemperatureChangeRateUnit.HectodegreeCelsiusPerSecond, "HectodegreesCelsiusPerSecond", BaseUnits.Undefined, "TemperatureChangeRate"),
                     new UnitInfo<TemperatureChangeRateUnit>(TemperatureChangeRateUnit.KilodegreeCelsiusPerSecond, "KilodegreesCelsiusPerSecond", new BaseUnits(time: DurationUnit.Millisecond, temperature: TemperatureUnit.DegreeCelsius), "TemperatureChangeRate"),
                     new UnitInfo<TemperatureChangeRateUnit>(TemperatureChangeRateUnit.MicrodegreeCelsiusPerSecond, "MicrodegreesCelsiusPerSecond", BaseUnits.Undefined, "TemperatureChangeRate"),
```
lipchev, line 93: Ok check this out, I found another _limitation_ of the `UnitPrefixBuilder` - given that the `Temperature` now has a the `BaseUnit` set for `MillidegreesCelsius`, you'd think that this would get resolved here. However, the reason it doesn't is because there is nothing in `Temperature.json` that _links_ the `MillidegreesCelsius` to the prefix `Milli` (outside the name). The same applies to the `TemperatureDelta.g.cs`  where the `MillidegreesCelsius` doesn't have any `BaseUnits`, as it is prefixed, while the base quantity isn't. 

I've checked for other _conventionally defined_ prefixes in the base quantities, and found just the `Microinch` (as well as the `Mil`, if this is supposed to be `Kilo`). There isn't anything stopping us from making the `Microinch` a simple prefix unit:

```  json
    {
      "SingularName": "Inch",
      "PluralName": "Inches",
      "BaseUnits": {
        "L": "Inch"
      },
      "FromUnitToBaseFunc": "{x} * 2.54e-2",
      "FromBaseToUnitFunc": "{x} / 2.54e-2",
      "Prefixes": [ "Micro" ],
      "Localization": [
        {
          "Culture": "en-US",
          "Abbreviations": [ "in", "\\\"", "″" ],
          "AbbreviationsForPrefixes": { "Micro": "µin" }
        },
        {
          "Culture": "ru-RU",
          "Abbreviations": [ "дюйм" ],
          "AbbreviationsForPrefixes": { "Micro": "микродюйм" }
        },
        {
          "Culture": "zh-CN",
          "Abbreviations": [ "英寸" ],
          "AbbreviationsForPrefixes": { "Micro": "微英寸" }
        }
      ]
    }, 
```
I tested it locally, hoping to see some other quantity picking up on the newly available prefix, but there weren't any (I think there should be some, but are probably just missing their `BaseUnits`).

For the `Mil` yeah, not gonna happen.. And the `Temperature`, we cannot have it as `Prefix` since it is an _affine quantity_. I could of course, hard-code them in the `BaseUnitPrefixes` 😄

PR #1489, opened 2025-01-03
Changing the Dimensions and BaseUnit for the FuelEfficiency

lipchev:
Changing the `Dimensions` and `BaseUnit` for the `FuelEfficiency`:
- `BaseUnit` changed from `LiterPer100Kilometers` (defaults to `+Infinity`) to `KilometersPerLiter` (defaults to `0`)
- added `Dimensions` of `"L": -2` (was previously treated as a dimensionless quantity)
- added the `MeterPerCubicMeter` (m/m³) unit
- updated the relevant tests

### DIFF
--- a/Common/UnitDefinitions/FuelEfficiency.json
+++ b/Common/UnitDefinitions/FuelEfficiency.json
@@ -1,55 +1,84 @@
 {
-    "Name": "FuelEfficiency",
-    "BaseUnit": "LiterPer100Kilometers",
-    "XmlDocSummary": "Fuel efficiency is a form of thermal efficiency, meaning the ratio from effort to result of a process that converts chemical potential energy contained in a carrier (fuel) into kinetic energy or work. Fuel economy is stated as \"fuel consumption\" in liters per 100 kilometers (L/100 km). In countries using non-metric system, fuel economy is expressed in miles per gallon (mpg) (imperial galon or US galon).",
-    "XmlDocRemarks": "https://en.wikipedia.org/wiki/Fuel_efficiency",
-    "Units": [{
-        "SingularName": "LiterPer100Kilometers",
-        "PluralName": "LitersPer100Kilometers",
-        "FromUnitToBaseFunc": "{x}",
-        "FromBaseToUnitFunc": "{x}",
-        "Localization": [{
+  "Name": "FuelEfficiency",
+  "BaseUnit": "KilometerPerLiter",
+  "XmlDocSummary": "In the context of transport, fuel economy is the energy efficiency of a particular vehicle, given as a ratio of distance traveled per unit of fuel consumed. In most countries, using the metric system, fuel economy is stated as \"fuel consumption\" in liters per 100 kilometers (L/100 km) or kilometers per liter (km/L or kmpl). In countries using non-metric system, fuel economy is expressed in miles per gallon (mpg) (imperial galon or US galon).",
+  "XmlDocRemarks": "https://en.wikipedia.org/wiki/Fuel_efficiency",
+  "BaseDimensions": {
+    "L": -2
+  },
+  "Units": [
+    {
+      "SingularName": "LiterPer100Kilometers",
+      "PluralName": "LitersPer100Kilometers",
+      "FromUnitToBaseFunc": "100 / {x}",
+      "FromBaseToUnitFunc": "100 / {x}",
+      "Localization": [
+        {
           "Culture": "en-US",
           "Abbreviations": [
             "l/100km"
           ]
-        }]
-      },
-      {
-        "SingularName": "MilePerUsGallon",
-        "PluralName": "MilesPerUsGallon",
-        "FromUnitToBaseFunc": "(100 * 3.785411784) / (1.609344 * {x})",
-        "FromBaseToUnitFunc": "(100 * 3.785411784) / (1.609344 * {x})",
-        "Localization": [{
+        }
+      ]
+    },
+    {
+      "SingularName": "MilePerUsGallon",
+      "PluralName": "MilesPerUsGallon",
+      "FromUnitToBaseFunc": "{x} * 1.609344 / 3.785411784",
+      "FromBaseToUnitFunc": "{x} * 3.785411784 / 1.609344",
+      "Localization": [
+        {
           "Culture": "en-US",
           "Abbreviations": [
             "mpg (U.S.)"
           ]
-        }]
-      },
-      {
-        "SingularName": "MilePerUkGallon",
-        "PluralName": "MilesPerUkGallon",
-        "FromUnitToBaseFunc": "(100 * 4.54609188) / (1.609344 * {x})",
-        "FromBaseToUnitFunc": "(100 * 4.54609188) / (1.609344 * {x})",
-        "Localization": [{
+        }
+      ]
+    },
+    {
+      "SingularName": "MilePerUkGallon",
+      "PluralName": "MilesPerUkGallon",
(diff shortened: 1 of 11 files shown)


Common/UnitDefinitions/FuelEfficiency.json
```diff
@@ -1,55 +1,84 @@
...
+      "FromBaseToUnitFunc": "{x} * 1e6",
+      "Localization": [
+        {
+          "Culture": "en-US",
+          "Abbreviations": [
+            "m/m³"
+          ]
+        }
+      ]
+    }
```
lipchev, line 82: I only added this because of the `UnitSystem.SI` coverage, but I'm not really _set_ on having it: you decide.

UnitsNet.Tests/CustomCode/FuelEfficiencyTests.cs
```diff
@@ -17,25 +17,22 @@
...
-        [MemberData(nameof(UnitTypes))]
-        public override void ToUnit_FromDefaultQuantity_ReturnsQuantityWithGivenUnit(FuelEfficiencyUnit unit)
+        protected override double MetersPerCubicMeterInOneKilometerPerLiter => 1e6;
+        protected override double KilometersPerLiterInOneKilometerPerLiter => 1;
+        protected override double LitersPer100KilometersInOneKilometerPerLiter => 100;
+        protected override double MilesPerUkGallonInOneKilometerPerLiter => 2.824809363318222;
+        protected override double MilesPerUsGallonInOneKilometerPerLiter => 2.352145833333333;
+
+        [Fact(Skip = "See about changing this to MetersPerCubicMeter")]
+        public override void BaseUnit_HasSIBase()
```
lipchev, line 33: Again, it's up to you to decide if we should un-`[Skip]` it or change the skip reason...

lipchev:
I was also going to include this as part of the `UnitRelations.json`:
```  json
"Length.Kilometer = FuelEfficiency.KilometerPerLiter * Volume.Liter -- NoInferredDivision",
```
but ended up reverting the change (_not that it didn't work or anything_). Here is how this would affect the two quantities:
```  csharp
    public readonly partial struct FuelEfficiency :
        IArithmeticQuantity<FuelEfficiency, FuelEfficiencyUnit>,
#if NET7_0_OR_GREATER
        IMultiplyOperators<FuelEfficiency, Volume, Length>,
#endif
```

```  csharp
    public readonly partial struct Volume :
        IArithmeticQuantity<Volume, VolumeUnit>,
#if NET7_0_OR_GREATER
        IMultiplyOperators<Volume, Molarity, AmountOfSubstance>,
        IMultiplyOperators<Volume, ReciprocalLength, Area>,
        IDivisionOperators<Volume, Length, Area>,
        IMultiplyOperators<Volume, Length, AreaMomentOfInertia>,
        IDivisionOperators<Volume, VolumeFlow, Duration>,
        IMultiplyOperators<Volume, EnergyDensity, Energy>,
        IMultiplyOperators<Volume, FuelEfficiency, Length>,
        IMultiplyOperators<Volume, ReciprocalArea, Length>,
        IDivisionOperators<Volume, Area, Length>,
        IMultiplyOperators<Volume, Density, Mass>,
        IMultiplyOperators<Volume, MassConcentration, Mass>,
        IDivisionOperators<Volume, SpecificVolume, Mass>,
        IDivisionOperators<Volume, Mass, SpecificVolume>,
        IDivisionOperators<Volume, Duration, VolumeFlow>,
#endif
```

Here are the pros and cons to consider:
- the multiplication operator seems reasonable: we multiply some amount (volume) of fuel by some fuel efficiency and get the travel distance. Technically there isn't anything wrong with dividing some length by a given fuel efficiency, in order to get the amount of fuel- however I don't immediately think of fuel efficiency as the result of dividing a length by volume (although I can't say I have any particular _default_ quantity in mind 🤷 ).
- the downside (even with the `NoInferredDivision`) is that it _binds_ the two quantities together, making it harder (if not impossible) to separate the `Volume` and the `FuelEfficiency` later (e.g. `UnitsNet.Common` + `UnitsNet.Extended`)
- as a middle ground, we could just add (by hand) the `multiply` and `divide` operators on the side of the `FuelEfficiency`, however this might seem odd w.r.t. the multiplication..

@angularsen  A lot of important decisions for you to make in this PR 😄 

PS Of course we could just close this PR: I don't care much either way, I had the `json` file changed while testing other stuff (from months ago), so one way or another I wanted to resolve my change-set..

UnitsNet.Tests/CustomCode/FuelEfficiencyTests.cs
lipchev: Or if you want, there is another solution that doesn't require us having the `[Skip]` warning as a permanent feature: we could use a boolean property, maybe something like `SupportsSIUnitSystem` to skip the generation of these `not-gonna-be-fixed` type of tests.. 🤔

angularsen: If a unit doesn't make sense or is simply never used, then I don't think we should add it just to make some si unit system logic work, which arguably is a very niche part of UnitsNet and probably little used to begin with.

angularsen (changes requested):

UnitsNet.Tests/CustomCode/FuelEfficiencyTests.cs
lipchev: Removed the `MeterPerCubicMeter` in https://github.com/angularsen/UnitsNet/pull/1489/commits/d7169c17ec184b7a73d0817c106dd55f94b43ba7

lipchev: Oh wait, it didn't remove it from the `Units.json`.. (i have to do it by hand)

lipchev:
@angularsen  Ok, I've removed the `MeterPerCubicMeter` , skipped the tests and updated the PR summary to reflect the changes.

I've also updated the summary of the related issue: #1463, have a look at the remaining tasks..

angularsen:
>  the downside (even with the NoInferredDivision) is that it binds the two quantities together, making it harder (if not impossible) to separate the Volume and the FuelEfficiency later (e.g. UnitsNet.Common + UnitsNet.Extended)

I'm not sure I follow, how is this worse than binding to other quantities? Is it because you consider this a less-widely-used quantity?

Regarding extensibility and splitting up into packages, this is tricky with static typed stuff that we primarily deal with, but I have some ideas in https://github.com/angularsen/UnitsNet/pull/1500

> as a middle ground, we could just add (by hand) the multiply and divide operators on the side of the FuelEfficiency, however this might seem odd w.r.t. the multiplication..

If the inferred operators don't make sense, then yes, we always have the option of manually adding them.

lipchev:
> > the downside (even with the NoInferredDivision) is that it binds the two quantities together, making it harder (if not impossible) to separate the Volume and the FuelEfficiency later (e.g. UnitsNet.Common + UnitsNet.Extended)
> 
> I'm not sure I follow, how is this worse than binding to other quantities? Is it because you consider this a less-widely-used quantity?

Yes, I'd say that the _minimal quantity set_ (e.g. `UnitsNet.Common`) would have to include the SI quantities + all of their associated operators (e.g. Mass -> Density) + all of their associations (e.g. Density -> KinematicViscosity, DynamicViscosity etc..).. I haven't tested how many quantities would get covered by this graph, but If there are enough left-overs we could put them in a separate nuget.

angularsen:
Alright, I fear it will be tough to split up into two packages where one cannot depend on the other, but maye there is such a cut.

PR #1490, opened 2025-01-03
Fix typo in MassMomentOfInertia.TonneSquareMilimeter

lipchev:
Renaming `MassMomentOfInertia.TonneSquareMilimeter` to `MassMomentOfInertia.TonneSquareMillimeter` (as part of #1200 )


Common/UnitEnumValues.g.json
```diff
@@ -965,7 +965,10 @@
     "TonneSquareCentimeter": 25,
     "TonneSquareDecimeter": 26,
     "TonneSquareMeter": 27,
-    "TonneSquareMilimeter": 28
+    "TonneSquareMilimeter": 28,
+    "KilotonneSquareMillimeter": 30,
+    "MegatonneSquareMillimeter": 37,
+    "TonneSquareMillimeter": 32
```
lipchev, line 971: This maybe a little hack-y, but I think that if we changed these by hand (to the old values) - I think they would _stick_.

angularsen: For fixing typos, we can definitely tweak these by hand to match the old values 👍

angularsen (changes requested):
As you said, let's change the Common/UnitEnumValues.g.json to match the old values. I can't think of a scenario where this breaks, since it is practically the same unit with a different name.

Other than that, just go ahead and merge this when ready.

Common/UnitEnumValues.g.json
lipchev: fixed in https://github.com/angularsen/UnitsNet/pull/1490/commits/23a68bc58f0ce5c6aff630ec721368b4730f6ca7

PR #1491, opened 2025-01-03
Rename SpecificFuelConsumption.GramPerKiloNewtonSecond to GramPerKilonewtonSecond

lipchev:
Rename `SpecificFuelConsumption.GramPerKiloNewtonSecond` to `GramPerKilonewtonSecond` (part of #1200)


Common/UnitEnumValues.g.json
```diff
@@ -1334,7 +1334,9 @@
     "GramPerKiloNewtonSecond": 1,
     "KilogramPerKilogramForceHour": 2,
     "KilogramPerKiloNewtonSecond": 3,
-    "PoundMassPerPoundForceHour": 4
+    "PoundMassPerPoundForceHour": 4,
+    "GramPerKilonewtonSecond": 11,
+    "KilogramPerKilonewtonSecond": 6
```
lipchev, line 1339: Same here: 
> This maybe a little hack-y, but I think that if we changed these by hand (to the old values) - I think they would stick.

angularsen: Yup, I suggest we do the same here, match the old values.

angularsen (approved):
One suggestion, other than that ready to merge

Common/UnitEnumValues.g.json
lipchev: fixed in https://github.com/angularsen/UnitsNet/pull/1491/commits/730eab19ddb691296c3c6cadd1278ac24f87ed89

lipchev:
Hmm, I just realized that in my other project I actually have the SI coverage tests [Skipped] for this one..

As you've correctly observed in #1473 the `BaseUnits` for the `GramPerKilonewtonSecond` also simplify to `Meter, Second`

``` json
{
  "SingularName": "GramPerKilonewtonSecond",
  "PluralName": "GramsPerKilonewtonSecond",
  "BaseUnits": {
    "L": "Meter",
    "T": "Second"
  },
  "FromUnitToBaseFunc": "{x}",
  "FromBaseToUnitFunc": "{x}",
  "Prefixes": [ "Kilo" ],
  "Localization": [
    {
      "Culture": "en-US",
      "Abbreviations": [ "g/(kN�s)" ],
      "AbbreviationsForPrefixes": { "Kilo": [ "kg/(kN�s)" ] }
    }
  ]
}
```

However this showcases another limitation of using the `BaseUnits` to determine the proper SI unit for quantities having a _hidden_ dimension.

The problem is even worse, when this is the `Mass` dimension, with it's goddamn `Kilo` prefix.. Even if we removed the `BaseUnits` from the `GramPerKilonewtonSecond` and introduced a brand new unit

```  json
    {
      "SingularName": "GramPerNewtonSecond",
      "PluralName": "GramsPerNewtonSecond",
      "BaseUnits": {
        "L": "Meter",
        "T": "Second"
      },
      "FromUnitToBaseFunc": "{x} * 1000",
      "FromBaseToUnitFunc": "{x} / 1000",
      "Prefixes": [ "Kilo" ],
      "Localization": [
        {
          "Culture": "en-US",
          "Abbreviations": [ "g/(N�s)" ],
          "AbbreviationsForPrefixes": { "Kilo": [ "kg/(N�s)" ] }
        }
      ]
    }
```

This still wouldn't give us the proper SI unit, as the `Kilo` prefix would scale down the `Length` dimension, turning the `BaseUnits` to `Millimeter, Second`.

The only way this can be made _right_ is to have the `KilogramsPerNewtonSecond` as it's own non-prefixed unit, with the rest of the units being scaled accordingly (_this must be why I had the `Microsecond` in it's `BaseUnits` #1473 🤷 )_.

PS A little more of this and the doctor's diagnosis is going to be `brain-fucked`.. 🧠

lipchev:
> The problem is even worse, when this is the `Mass` dimension, with it's goddamn `Kilo` prefix.. Even if we removed the `BaseUnits` from the `GramPerKilonewtonSecond` and introduced a brand new unit
> 
> ```json
>     {
>       "SingularName": "GramPerNewtonSecond",
>       "PluralName": "GramsPerNewtonSecond",
>       "BaseUnits": {
>         "L": "Meter",
>         "T": "Second"
>       },
>       "FromUnitToBaseFunc": "{x} * 1000",
>       "FromBaseToUnitFunc": "{x} / 1000",
>       "Prefixes": [ "Kilo" ],
>       "Localization": [
>         {
>           "Culture": "en-US",
>           "Abbreviations": [ "g/(N�s)" ],
>           "AbbreviationsForPrefixes": { "Kilo": [ "kg/(N�s)" ] }
>         }
>       ]
>     }
> ```
> 
> This still wouldn't give us the proper SI unit, as the `Kilo` prefix would scale down the `Length` dimension, turning the `BaseUnits` to `Millimeter, Second`.
> 
> The only way this can be made _right_ is to have the `KilogramsPerNewtonSecond` as it's own non-prefixed unit, with the rest of the units being scaled accordingly (_this must be why I had the `Microsecond` in it's `BaseUnits` #1473 🤷 )_.
> 
> PS A little more of this and the doctor's diagnosis is going to be `brain-fucked`.. 🧠

Oh I think I made a mistake here- we won't actually need `KilogramsPerNewtonSecond` to be an independent unit, the `BaseUnits` of `Millimeter, Second` would actually scale _up_ to the `Meter, Second` for the `KilogramPerNewtonSecond`.

lipchev:
> > The problem is even worse, when this is the `Mass` dimension, with it's goddamn `Kilo` prefix.. Even if we removed the `BaseUnits` from the `GramPerKilonewtonSecond` and introduced a brand new unit
> > ```json
> >     {
> >       "SingularName": "GramPerNewtonSecond",
> >       "PluralName": "GramsPerNewtonSecond",
> >       "BaseUnits": {
> >         "L": "Meter",
> >         "T": "Second"
> >       },
> >       "FromUnitToBaseFunc": "{x} * 1000",
> >       "FromBaseToUnitFunc": "{x} / 1000",
> >       "Prefixes": [ "Kilo" ],
> >       "Localization": [
> >         {
> >           "Culture": "en-US",
> >           "Abbreviations": [ "g/(N�s)" ],
> >           "AbbreviationsForPrefixes": { "Kilo": [ "kg/(N�s)" ] }
> >         }
> >       ]
> >     }
> > ```
> > This still wouldn't give us the proper SI unit, as the `Kilo` prefix would scale down the `Length` dimension, turning the `BaseUnits` to `Millimeter, Second`.
> > The only way this can be made _right_ is to have the `KilogramsPerNewtonSecond` as it's own non-prefixed unit, with the rest of the units being scaled accordingly (_this must be why I had the `Microsecond` in it's `BaseUnits` #1473 🤷 )_.
> > PS A little more of this and the doctor's diagnosis is going to be `brain-fucked`.. 🧠
> 
> Oh I think I made a mistake here- we won't actually need `KilogramsPerNewtonSecond` to be an independent unit, the `BaseUnits` of `Millimeter, Second` would actually scale _up_ to the `Meter, Second` for the `KilogramPerNewtonSecond`.

Ok, I think I almost got it figured out- for the `GramPerNewtonSecond` we could have the `BaseUnits` defined as either `Meter, Millisecond` or `Kilometer, Second`, however the `UnitPrefixBuilder` isn't impressed by my attempts to trick it into giving me a `non-prefixed` unit.. In the first case it was able to derive the `Millimeter, Millisecond`, and the second one returning `Undefined`..

This is because the `BaseUnitPrefixes` aren't storing the non-prefixed units (corresponding to a prefix-exponent of `0`). I don't know if there are any other cases that aren't being _de-prefixed_ like that, but I'm going try to fix the `BaseUnitPrefixes` and see what comes out..

lipchev:
> Ok, I think I almost got it figured out- for the `GramPerNewtonSecond` we could have the `BaseUnits` defined as either `Meter, Millisecond` or `Kilometer, Second`, however the `UnitPrefixBuilder` isn't impressed by my attempts to trick it into giving me a `non-prefixed` unit.. In the first case it was able to derive the `Millimeter, Millisecond`, and the second one returning `Undefined`..
> 
> This is because the `BaseUnitPrefixes` aren't storing the non-prefixed units (corresponding to a prefix-exponent of `0`). I don't know if there are any other cases that aren't being _de-prefixed_ like that, but I'm going try to fix the `BaseUnitPrefixes` and see what comes out..

Good, it seems this wasn't all in vain - #1492 turned up with quite a few new options for the `BaseUnits`. I also tried the two cases that were failing before - and they now both produce `Meter, Second`.

Anyway, this PR was only about renaming the unit, so it should be good to go. For the rest of it, I'll open another one later..

PR #1492, opened 2025-01-04
Include the non-prefixed BaseUnits in the BaseUnitPrefixes

lipchev:
Include the non-prefixed `BaseUnits` in the `BaseUnitPrefixes`:
- the `BaseUnitPrefixes` now also handles the case where prefixing an already prefixed base unit results in a non-prefixed base unit
- regenerating the quantities resulted in a number of new `BaseUnits` to become available (as well as some which got their scales reduced)


UnitsNet/GeneratedCode/Quantities/Power.g.cs
```diff
@@ -109,7 +109,7 @@ static Power()
                     new UnitInfo<PowerUnit>(PowerUnit.MetricHorsepower, "MetricHorsepower", BaseUnits.Undefined, "Power"),
                     new UnitInfo<PowerUnit>(PowerUnit.Microwatt, "Microwatts", new BaseUnits(length: LengthUnit.Meter, mass: MassUnit.Milligram, time: DurationUnit.Second), "Power"),
                     new UnitInfo<PowerUnit>(PowerUnit.MillijoulePerHour, "MillijoulesPerHour", BaseUnits.Undefined, "Power"),
-                    new UnitInfo<PowerUnit>(PowerUnit.Milliwatt, "Milliwatts", BaseUnits.Undefined, "Power"),
+                    new UnitInfo<PowerUnit>(PowerUnit.Milliwatt, "Milliwatts", new BaseUnits(length: LengthUnit.Meter, mass: MassUnit.Gram, time: DurationUnit.Second), "Power"),
```
angularsen, line 112: Ok, this example I can relate to. Watt is `m*kg*s`, so Milliwatt can be `m*g*s`.

angularsen (approved):

UnitsNet/GeneratedCode/Quantities/Power.g.cs
angularsen: Seems like you keep getting more and more base units defined, nice 👏

PR #1494, opened 2025-01-04
🚸UnitParser.Parse should throw AmbiguousUnitParseException before any UnitNotFoundException (#1484)

### LINKED ISSUE #1423: Quantity.TryParse won't parse "10 MM" because "MM" is capitalized. Other abbreviations work fine.

lipchev:
Fixes #1423 for `release-v6`
Carry over from #1484

`UnitParser.Parse<LengthUnit>("MM")` fails due to matching both `Megameter` and `Millimeter` in case-insensitive matching, but matches neither of them in the case-sensitive fallback. It was confusing to get `UnitsNotFoundException` in this case, since case-insensitive usually works for most units.

### Changes
- Handle this case and throw `AmbiguousUnitParseException` instead of `UnitNotFoundException`
- Describe the case-insensitive units that matched
- Fix existing test `Parse_WithMultipleCaseInsensitiveMatchesButNoExactMatches_ThrowsUnitNotFoundException`
- Skip retrying with fallback culture if no specific `formatProvider` was given
- Avoid some of the unnecessary array allocations

### DIFF
--- a/UnitsNet.Tests/UnitParserTests.cs
+++ b/UnitsNet.Tests/UnitParserTests.cs
@@ -115,8 +115,8 @@ namespace UnitsNet.Tests
             var exception2 = Assert.Throws<AmbiguousUnitParseException>(() => Length.Parse("1 pt", CultureInfo.InvariantCulture));
 
             // Assert
-            Assert.Equal("Cannot parse \"pt\" since it could be either of these: DtpPoint, PrinterPoint", exception1.Message);
-            Assert.Equal("Cannot parse \"pt\" since it could be either of these: DtpPoint, PrinterPoint", exception2.Message);
+            Assert.Equal("""Cannot parse "pt" since it matches multiple units: DtpPoint ("pt"), PrinterPoint ("pt").""", exception1.Message);
+            Assert.Equal("""Cannot parse "pt" since it matches multiple units: DtpPoint ("pt"), PrinterPoint ("pt").""", exception2.Message);
         }
 
         [Theory]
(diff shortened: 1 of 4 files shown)


UnitsNet.Tests/UnitParserTests.cs
```diff
@@ -144,6 +144,13 @@ public void Parse_MappedCustomUnit()
             Assert.Equal(HowMuchUnit.Some, parsedUnit);
         }
         
+        [Fact]
+        public void Parse_LengthUnit_MM_ThrowsExceptionDescribingTheAmbiguity()
+        {
+            var ex = Assert.Throws<AmbiguousUnitParseException>(() => UnitsNetSetup.Default.UnitParser.Parse<LengthUnit>("MM"));
+            Assert.Equal("""Cannot parse "MM" since it matches multiple units: Megameter ("Mm"), Millimeter ("mm").""", ex.Message);
+        }
```
lipchev, line 152: Notice the new message format, the other one seemed excessively long..

UnitsNet/CustomCode/UnitParser.cs
lipchev, line 1: What I said in #1484 about the recursive call requiring additional handling was wrong: I was under the persuasion that if there are multiple matches with the first culture, then we still want to try to disambiguate with the fallback culture, but that makes no sense..

UnitsNet.Tests/UnitParserTests.cs
angularsen: I like it

PR #1496, opened 2025-01-13
Minor corrections to some unit abbreviations

Jagailo:
**AreaMomentOfInertia:**
> Use superscript (cm², m³) instead of cm^2, m^3

**MolarEntropy:**
> Use · for products (N·m instead of Nm, N*m or N.m)

**SpecificEntropy:**
> Use · for products (N·m instead of Nm, N*m or N.m)

Using parentheses for mathematically correct display; using the correct symbol for degrees Celsius

**SpecificFuelConsumption:**
It looks like some unusual multiplication sign was used

**WarpingMomentOfInertia:**
> Use superscript (cm², m³) instead of cm^2, m^3

---

For ^4 and ^6 UnitParser.NormalizeUnitString() will handle this situation


Common/UnitDefinitions/SpecificEntropy.json
```diff
@@ -55,7 +55,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "BTU/lb·°F", "BTU/lbm·°F" ]
+          "Abbreviations": [ "BTU/lb·°F", "BTU/lbm·°F", "Btu/(lb·°F)", "Btu/(lbm·°F)" ]
```
angularsen, line 58: Can we maybe normalize instead so we don't have to explicitly list variants with parantheses for all units? 

We could use regex to remove parantheses around two or more multiplied terms so `Btu/(lbm·°F)` becomes `Btu/lbm·°F`, after normalizing `*` to `·`.

Common/UnitDefinitions/SpecificFuelConsumption.json
```diff
@@ -37,8 +37,8 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "g/(kN�s)" ],
-          "AbbreviationsForPrefixes": { "Kilo": [ "kg/(kN�s)" ] }
+          "Abbreviations": [ "g/(kN·s)" ],
+          "AbbreviationsForPrefixes": { "Kilo": [ "kg/(kN·s)" ] }
```
angularsen, line 41: Good catch, must have been an encoding issue at some point 🤔

Common/UnitDefinitions/AreaMomentOfInertia.json
```diff
@@ -17,7 +17,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "m⁴", "m^4" ]
```
angularsen, line 20: Good, this is much better

Common/UnitDefinitions/SpecificEntropy.json
```diff
@@ -17,7 +17,7 @@
       "Localization": [
         {
           "Culture": "en-US",
-          "Abbreviations": [ "J/kg.K" ]
+          "Abbreviations": [ "J/kg.K", "J/(kg·K)" ]
```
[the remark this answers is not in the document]
angularsen: Also, this is probably a preference thing, but should we standardize on `J/(kg·K)` or `J/kg·K` ?

In our wiki we have kinda said the former, but I'm not sure what is "the most common" convention here.
https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#abbreviation-naming-conventions

angularsen (changes requested):
Good improvements! A couple of suggestions.

Common/UnitDefinitions/SpecificEntropy.json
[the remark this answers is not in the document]
Jagailo: I don't think that all units need to be written correctly according to mathematical rules.
After all, I don't think that there are common cases when different order of operations can cause confusion in different quantities.

But regarding specific heat capacity, [Wikipedia uses parentheses](https://en.wikipedia.org/wiki/Specific_heat_capacity#International_system) because in real life on paper such unit would be written through a horizontal line:
image

I find this quite logical and obvious. However, I only have a school-level knowledge of physics, so I might be wrong, and real physicists might say that brackets are unnecessary.

Common/UnitDefinitions/SpecificEntropy.json
Jagailo: It's up to you; are you suggesting to update the NormalizeUnitString() method with the regex?

Common/UnitDefinitions/SpecificEntropy.json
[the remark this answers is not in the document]
angularsen: I don't know, the wiki uses all sorts of formatting and does not seem to standardize on any of them. There are several examples with and without parantheses, and other forms.

image
image
image

I have no strong opinion here, we can stick with parantheses until we learn otherwise. This is already what UnitsNet wiki describes as our convention.

Common/UnitDefinitions/SpecificEntropy.json
angularsen: Yes, but if per our above discussion want to standardize on parantheses, then the normalization must remove parantheses from both the string coming from our JSON definitions and the input string to parse.

I just want to avoid having to maintain duplicate abbreviations in JSON for all our 1000+ units just to support parsing with and without parantheses.

Jagailo: Ok, looks like this is going beyond this pull request and is becoming something global.
I've reverted the original values ​​with other fixes; now only one abbreviation.

angularsen (approved):
Ok, there are a couple of minor breaking changes here, but I'll risk it.

I think very few will be upset that we no longer support parsing `*` or `.` in these unit abbreviations. I'd rather have the parser use normalization or attempt a few fallbacks to cover such alternative formats.

Common/UnitDefinitions/SpecificEntropy.json
angularsen: Yeah, good call

Jagailo:
I work in the medical domain and our users enter units by hand.
The · symbol isn't on the keyboard, but they are too lazy to press *, so we have a huge bunch of units of different quantities ​​written with a dot, because it's faster for them.
Units with a dot are just a mistake, and * is replaced in NormalizeUnitString, so I highly doubt that this fix will be inconvenient for anyone.
We shouldn't support incorrectly written units, imho.

PR #1498, opened 2025-01-19
Decomposed the rounded expressions in the unit definitions into operations with exact coefficients (non-breaking)

lipchev:
This is a port of #1393, targeting the `release-v6` consisting of the changes that didn't require any modifications to the tests. 

I'll create another PR with the rest of the (breaking) changes in a bit.


Common/UnitDefinitions/WarpingMomentOfInertia.json
```diff
@@ -87,8 +87,8 @@
       "BaseUnits": {
         "L": "Inch"
       },
-      "FromUnitToBaseFunc": "{x} * Math.Pow(2.54e-2, 6)",
-      "FromBaseToUnitFunc": "{x} / Math.Pow(2.54e-2, 6)",
+      "FromUnitToBaseFunc": "{x} * 0.000000000268535866540096",
+      "FromBaseToUnitFunc": "{x} / 0.000000000268535866540096",
```
angularsen, line 91: In a way, I kind of prefer the Math.Pow formula, since it is easier to recognize the constant 2.54 and even repeating it 6 times `2.54e-2 * 2.54e-2 * ...` could be one option if Math.Pow throws a curveball.
I have no strong opinion here, just an observation.

angularsen (approved):
I'll go ahead and merge this, just one minor comment/suggestion.

Tons of great work here 👏 
I only skimmed the new constants though and spot tested a few places, but I have not verified every single change since it seems the tests still pass and that is a good indicator things are in order.

Common/UnitDefinitions/WarpingMomentOfInertia.json
lipchev: I can work with `Math.Pow` as well, but if we (or whoever) wanted to created a generator for another language, the uses of `Math` means more hoops to deal with..

PR #1503, opened 2025-01-20
Replaced 'L' with 'l' for liter abbreviations

Jagailo:
When I was reading the [article](https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#abbreviation-naming-conventions) on adding new units I noticed this line:
> Use abbreviations defined by [SI Unit System](https://en.wikipedia.org/wiki/International_System_of_Units), such as l instead of L for liters

I ran a script and about half of the "liters" use `L` and the other half use `l`. Even within the same quantity, different liters are used:
image

In this pull request I replaced `L` with `l` according to abbreviation naming conventions. Open to any suggestions on how else this could be handled.


angularsen:
Sorry for no response on this one, I'm positive to make this change, but please retarget the PR for `release/v6` branch. It's probably not a breaking change per-se, but it feels like something that belongs in the next major version bump.

Jagailo:
Moved to another branch

angularsen:
There are some merge conflicts to handle after changing target branch. 

Depending on your comfort level with Git it may be easier to just start over and manually copy the changes and re-generate. I believe most of your changes should still work for release/v6, and the conflicts are mostly related to the generated code.

PR #1504, opened 2025-01-21
VolumeConcentration: fix the SingularName for LiterPerLiter and LiterPerMilliliter

lipchev:
fixes the `VolumeConcentration` units `LitersPerLiter` and `LitersPerMilliliter` (using the _PluralName_  for the `SingularName` property)


lipchev:
This is very likely going to cause some pain.. and whoever wrote this in the first place, is definitely going to feel it 😄

PR #1511, opened 2025-02-04
💚FluidResistance: Remove duplicate, invalid base units

angularsen:
Removing all base units except for the base unit and one more trivial variant.


angularsen:
@lipchev To fix the broken tests, I went ahead and remove most of the `BaseUnits` since it was not trivial for me to determine all of them.

Just a heads up for when we merge into v6, and if you want to, you can add the correct definitions for the BaseUnits you think we should keep.

lipchev:
> @lipchev To fix the broken tests, I went ahead and remove most of the `BaseUnits` since it was not trivial for me to determine all of them.

Same here, I've got such an acute PTSD from these that as long as they aren't wrong- I don't care.

Issue #1483, opened 2024-12-30
NumberExtensions should not use Convert for net7.0


lipchev:
I just noticed this simple performance improvement that's likely to have a huge performance impact on the `NumberExtensions`..
``` csharp
        public static Mass Centigrams<T>(this T value)
            where T : notnull
#if NET7_0_OR_GREATER
            , INumber<T>
            => Mass.FromCentigrams(double.CreateChecked(value));
#else
            => Mass.FromCentigrams(Convert.ToDouble(value));
#endif
```

Not sure if we want this in `v5`,  `v6` or I could just put it `up-for-grabs` and leave it for later..

Note that there is probably a `Convert.ToDecimal` somewhere, just waiting to popup as a merge conflict...

lipchev:
@angularsen Oh you're gonna love this: thanks to the `IConvertible` interface, on `net48`- the following syntax should work just fine: `Mass.FromCentigrams(2).Centigrams()` 😆

angularsen:
Let's not bother with v5, it can receive new units and bug fixes, but new features can mostly just go into v6 so we can focus on finalizing and shipping it before it becomes an infinite project. Scope is definitely creeping.

As for `net48`, I can sympathize being stuck on a legacy project, but come on! Get on with the times people! .NET Core 2.x has been available for 6 years already 😅

Good example though, something to consider when discussing `IConvertible`.

My current thinking is, if the argument is WPF and XAML-like bindings, then value converters are a better approach than requiring all quantities to implement `IConvertible`.

```cs
public class FooTests
{
#if NET48
        [Fact]
        public void Foo()
        {
            Mass newMass = NumberToMassExtensions.Kilograms(Mass.FromGrams(2000));
            Assert.Equal(Mass.FromKilograms(2), newMass);
            /*
            Assert.Equal() Failure: Values differ
            Expected: 2 kg
            Actual:   2000 kg
            */
        }
#endif

}
```

angularsen:
Regarding the issue description, sounds good to me 👍

Muximize:
FYI, getting rid of `IConvertible` would clean up 99 lines of code in each of 126 quantities for a total of 12.474 lines removed.

angularsen:
Move on, nothing to see here bot

andrewimcclement:
@angularsen this issue should presumably be closed?

angularsen:
Right, thanks!

Fixed in #1518

PR #1518, opened 2025-02-26
Use INumberBase<T>.CreateChecked on .NET 7+

andrewimcclement:
Implementing change suggested in #1483 .

### DIFF
--- a/CodeGen/Generators/UnitsNetGen/NumberExtensionsGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/NumberExtensionsGenerator.cs
@@ -49,8 +49,10 @@ namespace UnitsNet.NumberExtensions.NumberTo{_quantityName}
             where T : notnull
 #if NET7_0_OR_GREATER
             , INumber<T>
-#endif
+            => {_quantityName}.From{unit.PluralName}(double.CreateChecked(value));
+#else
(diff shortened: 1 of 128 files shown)


lipchev:
I don't personally use these extensions, but here are a few more remarks:
- for the old framework it's possible to restrict the type to `IConvertible` and replace the `Convert.ToDouble(value)` with `value.ToDouble(null)` (given that we're already requiring a struct we don't have to concern ourselves with nullability)
- we're not testing the old framework: https://github.com/angularsen/UnitsNet/blob/8d4fa525ce97f2fc2b640b95366bb83eff147c8c/UnitsNet.NumberExtensions.Tests/UnitsNet.NumberExtensions.Tests.csproj#L4

andrewimcclement:
Using IConvertible does avoid the boxing, which is nice.

lipchev (approved):
LGTM
Hopefully, one day we would be able to add the properties here as well..

angularsen:
Looks good to me also, thanks!

@lipchev you can also merge PRs by the way unless you want a pair of extra eyes on it. 
I'll probably look over it anyways, but no need to wait for me.

lipchev:
@angularsen I know, but I didn't know how you want to handle the release of this one, and the merging of the changes between `master` and `release-v6`..

PR #1529, opened 2025-03-29
Removing the FrequencyUnit.BUnit

lipchev:
Removing the `FrequencyUnit.BUnit`.

As discussed in https://github.com/angularsen/UnitsNet/pull/1393#issuecomment-2211588432

### DIFF
--- a/Common/UnitDefinitions/Frequency.json
+++ b/Common/UnitDefinitions/Frequency.json
@@ -108,18 +108,6 @@
           "Abbreviations": [ "с⁻¹" ]
         }
       ]
-    },
-    {
-      "SingularName": "BUnit",
-      "PluralName": "BUnits",
-      "FromUnitToBaseFunc": "Math.Sqrt({x} * 1e3)",
-      "FromBaseToUnitFunc": "{x} * {x} * 1e-3",
-      "Localization": [
-        {
-          "Culture": "en-US",
-          "Abbreviations": [ "B Units" ]
-        }
-      ]
     }
   ]
 }
--- a/UnitsNet.NanoFramework/GeneratedCode/Quantities/Frequency.g.cs
+++ b/UnitsNet.NanoFramework/GeneratedCode/Quantities/Frequency.g.cs
@@ -84,11 +84,6 @@ namespace UnitsNet
         public double BeatsPerMinute => As(FrequencyUnit.BeatPerMinute);
 
         /// <summary>
-        ///     Gets a <see cref="double"/> value of this quantity converted into <see cref="FrequencyUnit.BUnit"/>
-        /// </summary>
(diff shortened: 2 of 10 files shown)


angularsen:
Just so I understand, we are removing this unit due to using `Math.Sqrt`?

In a way, I don't like the idea of removing something that someone actually requested and added, although yes it does seem very domain specific and I can't find much about this unit online.

I can get behind removing it for the reason that it is not widely used, but I have a feeling we'll need to support functions like Sqrt, Sin, Cos etc at some point anyway. Can we easily make exceptions for units like these?

lipchev:
> Just so I understand, we are removing this unit due to using `Math.Sqrt`?
> 
> In a way, I don't like the idea of removing something that someone actually requested and added, although yes it does seem very domain specific and I can't find much about this unit online.
> 
> I can get behind removing it for the reason that it is not widely used, but I have a feeling we'll need to support functions like Sqrt, Sin, Cos etc at some point anyway. Can we easily make exceptions for units like these?

The idea is that a unit conversions should be reversible, as such they cannot use irrational functions (I'm 99% sure there is format definition about it).

I don't see where it says that we need to support (now or in the future) `any-type-of-expression` as a `valid unit for a quantity`. In fact this, along with the `FeetOfElevention` (as a unit of pressure 😏) is currently the only use of a `Math` function in the unit definition (there are some uses of `Math.PI` but as long as it's defined as rational constant that's fine).

If you want the `BUnit`, it can be a property - but there is no way that it's gonna pass the equality tests.

lipchev:
Here's another way of looking at this: having an expression such as `value^2` would be the same as defining an `AreaUnit` for the `Length` quantity (or vice-versa). These in my opinion are clearly `operations`, not `unit-conversions`.

angularsen:
Useful insight regarding operations vs unit conversions.

I don't know of any definitions regarding rules for only allowing rational unit conversions, can't find any with a quick google. Won't all logarithmic units be disallowed then?

I tried to think of examples with PI, but realize these are probably operations like you say, converting radius to area or circumference.

Anyway, we can remove this. Mostly because it is not widely used, maybe because it violates some unit definition rule, and probably because it is actually an operation and not a unit conversion.

lipchev:
> I don't know of any definitions regarding rules for only allowing rational unit conversions, can't find any with a quick google. Won't all logarithmic units be disallowed then?

The `unit-conversion` functions for the logarithmic quantities are still linear (`x*a + b`). The way we have it, this is also true for the multiplication and division, but not for the additions and subtractions.

angularsen:
[Release UnitsNet/6.0.0-pre015 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F6.0.0-pre015)

PR #1534, opened 2025-04-11
Removing the `SupportsSIUnitSystem => true` property overrides

lipchev:
# fixes https://github.com/angularsen/UnitsNet/issues/1463


lipchev:
As expected, there are only the 
- `ElectricApparentEnergy` 
- `ElectricReactiveEnergy`
- `FuelEfficiency`

which override this property (and also `[Skip]` the corresponding tests). The only other tests that are skipped are for the `BaseUnit_HasSIBase` (as noted in the linked issue).

Issue #1509, opened 2025-01-31
No fall-back culture when using UnitAbbreviationsCache.CreateEmpty


lipchev:
Have a look at this test (it's targeting the v6 but that should be reproducible in v5)
```  csharp

        [Fact]
        public void MapUnitToDefaultAbbreviation_GivenUnitAndNoCulture_SetsDefaultAbbreviationForUnitForCurrentCulture()
        {
            using var cultureScope = new CultureScope(NorwegianCultureName);
            var cache = new UnitAbbreviationsCache();

            cache.MapUnitToDefaultAbbreviation(MassUnit.Gram, "zz");

            Assert.Equal("zz", cache.GetDefaultAbbreviation(MassUnit.Gram));
            Assert.Equal("g", cache.GetDefaultAbbreviation(MassUnit.Gram, AmericanCulture));
        }
```

It fails with 

> Expected: "g"
> Actual:   ""

Everything works fine when initializing with the list of `QuantityInfo`:

``` csharp
        [Fact]
        public void MapUnitToDefaultAbbreviation_WithDefaultLookup_GivenUnitAndNoCulture_SetsDefaultAbbreviationForUnitForCurrentCulture()
        {
            using var cultureScope = new CultureScope(NorwegianCultureName);
            var cache = UnitAbbreviationsCache.CreateDefault();

            cache.MapUnitToDefaultAbbreviation(MassUnit.Gram, "zz");

            Assert.Equal("zz", cache.GetDefaultAbbreviation(MassUnit.Gram));
            Assert.Equal("g", cache.GetDefaultAbbreviation(MassUnit.Gram, AmericanCulture));
        }
```

The reason why this doesn't work is not immediately obvious- it's because when calling `MapUnitToDefaultAbbreviation`, when the `UnitInfo` is not found in the `QuantityInfoLookup` the `PerformAbbreviationMapping` creates a _dummy_ `UnitInfo` which doesn't have anything set for the `QuantityName` property.

In my opinion, the best solution would be to use these constructors:

``` csharp

        /// <summary>
        ///      Create an instance of the cache and load all the built-in quantities defined in the library.
        /// </summary>
        /// <returns>Instance for mapping any of the built-in units.</returns>
        public UnitAbbreviationsCache()
            :this(UnitsNetSetup.Default.QuantityInfoLookup)
        {
        }
        
        /// <summary>
        ///     Creates an instance of the cache using the specified set of quantities.
        /// </summary>
[...]

PR #1539, opened 2025-04-11
`UnitAbbreviationsCache`: `GetDefaultAbbreviation` should throw when no abbreviations are mapped

lipchev:
fixes the first part of #1509 
- `UnitAbbreviationsCache`: `GetDefaultAbbreviation` should throw when no abbreviations are mapped
- `Volume`: added the missing abbreviation mappings


lipchev:
This is missing a test for the _unhappy_ path:
``` csharp
        [Fact]
        public void GetDefaultAbbreviation_ForUnitWithoutAbbreviations_ThrowsInvalidOperationException()
        {
            Assert.Throws<InvalidOperationException>(() => new UnitAbbreviationsCache([HowMuch.Info]).GetDefaultAbbreviation(HowMuchUnit.AShitTon));
        }
```

But this doesn't work just yet.. I'll add it later, probably after the little refactoring I got planned for the the `UnitAbbreviationsCacheTests`.

PR #1541, opened 2025-04-11
re-organizing the QuantityIFormattableTests and UnitAbbreviationsCacheTests 

lipchev:
- moved the `ToString(format, culture)` tests from the `UnitAbbreviationsCacheTests` to the `QuantityIFormattableTests`
- re-ordered a few of the tests in the `UnitAbbreviationsCacheTests` (grouped by the method they call)


angularsen:
Reactoring PRs like this you can just go ahead and merge unless you do want a review on them.
I will probably look over the PR anyway and can comment after the fact, if I spot anything.

PR #1545, opened 2025-04-14
adding calorie per hour power unit

Dnigor:


Common/UnitDefinitions/Power.json
```diff
@@ -135,6 +135,19 @@
           "Abbreviations": [ "TR" ]
         }
       ]
+    },
+    {
+      "SingularName": "CaloriePerHour",
+      "PluralName": "CaloriesPerHour",
+      "FromUnitToBaseFunc": "{x} / 859.8452279",
+      "FromBaseToUnitFunc": "{x} * 859.8452279",
```
angularsen, line 143: @lipchev I know you fixed some other calorie units earlier, could you look over these?

lipchev: I did look over this (earlier) and frankly couldn't find any relation to the existing coefficients for the calorie and energy..
@Dnigor Could you express this in terms of the existing calorie coefficients (or explain why I cannot).. I assume there should be something `/ 60` in the expression..

Dnigor: Hi, I used https://convertlive.com/u/convert/watts/to/calories-per-hour

lipchev: Ok I figured it out, it's the inverse of what I must have been doing before:
``` json
 {
   "SingularName": "Calorie",
   "PluralName": "Calories",
   "FromUnitToBaseFunc": "{x} * 4.184",
   "FromBaseToUnitFunc": "{x} / 4.184",
   "Prefixes": [ "Kilo", "Mega" ],
   "Localization": [
     {
       "Culture": "en-US",
       "Abbreviations": [ "cal" ]
     }
   ]
 },
```

so the `859.8452279` seems to be the rounded expression of `1 / (4.184 / 3600)`, which is ~860.420. I have no idea why the difference is so large, the converter is probably using the _other_ definition of Calorie, but what I do know is that wouldn't trust an _online converter_ that don't provide me with the rational form of the conversion coefficient. 😄

Dnigor: Calorie should be an energy unit but we need a power unit

lipchev: I know, I'm asking for an exact expression- meaning that things that don't have a terminating decimal result, such as `1/60` we'd like to keep as `1/60` instead of rounding to `0.016667` (at least as far as the json files are concerned). 

At the same time, we try to be consistent in the use of the same conversion coefficients, for quantities which have historically (usually) had different values, such as the [Calorie](https://en.wikipedia.org/wiki/Calorie) or the [BTU](https://en.wikipedia.org/wiki/British_thermal_unit).

See #1393 for more details.

If haven't checked which _variant_ of this (if any) the converter is using, but was hoping that you could elaborate (given that you probably know the domain better than me).

angularsen: @lipchev We checked the calorie definitions last time and I believe your coefficients is what we should use, if nothing else to be consistent since there are many definitions of calorie. 

This PR should be adapted to our existing coefficients as you posted above.

For your reference:
859.845227859 is _international calorie_ (IT)
860.4206500956 is _thermochemical calorie_ (th)

https://www.translatorscafe.com/unit-converter/en-US/power/1-42/Vat-kalori%20(IT)/saat/

> An international calorie (IT) per hour (cal/h) is a metric unit of power.
An International calorie (IT cal) is an outdated non-SI metric unit of energy. In most fields, its use has been replaced by, the joule, the SI unit of energy. However, it remains a commonly used unit for energies in the field of chemistry. 1 IT calorie = 4.1868 J as defined by the Fifth International Conference on Properties of Steam. By definition, the small calorie or gram calorie approximates the energy needed to increase the temperature of 1 gram of water by 1 °C at standard atmospheric pressure (101.325 kPa).

> A thermochemical calorie (th) per hour (cal (th)/h) is a metric unit of power.
A thermochemical calorie (th cal) is an outdated non-SI metric unit of energy equal to 4.184 J.

Wolfram says nutritional calorie is 0.8604 (presumably kilocalories)
image

Wolfram says thermal calories are:
860.4 cal_th/h (thermochemical calories per hour)

Aqua-calc confirms @lipchev 's coefficients
> 1 W = 0.86042065
https://www.aqua-calc.com/convert/power/watt-to-calorie-per-hour
image

PR #1546, opened 2025-04-17
Add octet to Information and octet/s to Bitrate

ThomasGaboriau:
The octet is commonly used in place of the byte in French speaking countries and Romania (I'm gonna trust Wikipedia on that one), and in telecoms and internet stantards.


lipchev:
I don't know if I would classify this as a separate unit, to me this is just the _translation_ of the word _byte_.

ThomasGaboriau:
Octet and byte both exists in French and English. Octet is explicitly defined as an 8 bits byte, as the byte wasn't always 8 bits (ok, I guess it's been more than 30y thats not really the case anymore).

lipchev:
Right, it does have it's own wikipedia page (en): https://en.wikipedia.org/wiki/Octet_(computing) which doesn't say _..the french word for byte.._ So I guess it's fine.

Although I'd be surprised if the RFC standard used the `o` and `ko` abbreviations. 😄

angularsen:
I'm not entirely convinced we should add this. Byte is pretty standard internationally and Octet is practically just a different name for it. At least in today's use.

However, if you will actually make use of Octet and this is not just a theoretical exercise, then sure, we can add it for completness sake.

https://english.stackexchange.com/questions/628949/byte-or-octet

ThomasGaboriau:
Yes I made this PR because I plan to use Octet.

angularsen:
[Release UnitsNet/6.0.0-pre015 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F6.0.0-pre015)

PR #1548, opened 2025-04-18
`UnitAbbreviationsCache.CreateEmpty` should use the default `QuantityInfoLookup`

### LINKED ISSUE #1509: No fall-back culture when using UnitAbbreviationsCache.CreateEmpty

lipchev:
- `UnitAbbreviationsCache`: refactored the constructors / Create factory methods, with the Map/Get methods throwing an exception for missing units / unmapped abbreviations
- `UnitAbbreviationsCache`: replaced the `string[]` return types with `IReadOnlyList` (breaking change for `.Length` which now becomes `.Count`)
- `UnitAbbreviationsCache`: removed  the overload accepting a `UnitInfo`
- `UnitParser`: optimized the unit-parsing-with-fallback-culture methods, making the `FromUnitAbbreviation` behavior consistent with that of `ParseUnit`
- `UnitParser`:  `UnitInfo GetUnitFromAbbreviation(string unitAbbreviation, IFormatProvider? formatProvider)` : flipped the parameters and made the method public (_this is optional_)
- optimized the  `Mass.ParseUnit` calls (avoiding the `UnitInfo` lookups)
- completed the coverage of the `UnitParser` and `UnitAbbreviationsCache`

fixes #1509


lipchev:
@AndreasLeeb There are still some tests missing but I'm on my way out- will try to add them when I get back. Please have a look at the remaining `TODOs` and tell me if you agree with my comments.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -586,7 +586,7 @@ public static bool TryParseUnit([NotNullWhen(true)]string? str, out {_unitEnumNa
         /// <param name=""provider"">Format to use when parsing number and unit. Defaults to <see cref=""CultureInfo.CurrentCulture"" /> if null.</param>
         public static bool TryParseUnit([NotNullWhen(true)]string? str, IFormatProvider? provider, out {_unitEnumName} unit)
         {{
-            return UnitsNetSetup.Default.UnitParser.TryParse<{_unitEnumName}>(str, provider, out unit);
+            return UnitParser.Default.TryParse(str, Info, provider, out unit);
```
angularsen, line 589: Nitpick, but let's consistently use either Info or Info.UnitInfos overload. The latter saves an indirection.

UnitsNet.Benchmark/Initializations/UnitAbbreviationsCacheInitializationBenchmarks.cs
```diff
@@ -52,20 +52,20 @@ public string WithSpecificQuantityAndCustomMapping()
...
-        return cache.GetAbbreviations(Mass.Info.BaseUnitInfo)[0];
+        return Mass.GetAbbreviation(Mass.BaseUnit);
     }
 
     [Benchmark]
     public string WithSpecificQuantityWithoutLookup()
     {
         var cache = new UnitAbbreviationsCache([Mass.Info]);
-        return cache.GetAbbreviations(Mass.Info.BaseUnitInfo)[0];
+        return Mass.GetAbbreviation(Mass.BaseUnit);
```
angularsen, line 69: I'm not sure I understand what this is testing? `cache` is never used, so we test both constructing a new local `UnitAbbreviationsCache` instance + a lookup of the global default `UnitAbbreviationsCache`?

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
```diff
@@ -15,43 +16,47 @@ namespace UnitsNet
...
         private ConcurrentDictionary<AbbreviationMapKey, IReadOnlyList<string>> AbbreviationsMap { get; } = new();
 
         /// <summary>
-        ///     Create an empty instance of the cache, with no default abbreviations loaded.
+        ///      Create an instance of the cache and load all the built-in quantities defined in the library.
         /// </summary>
+        /// <returns>Instance for mapping any of the built-in units.</returns>
         public UnitAbbreviationsCache()
-            : this(new QuantityInfoLookup([]))
+            :this(UnitsNetSetup.Default.QuantityInfoLookup)
```
angularsen, line 59: I kind of liked how the previous default ctor created an empty instance, and in order to create a new instance with defaults loaded you'd have to call `CreateDefault()`. Most people will use the singleton instance anyway, but it just seems more conventional to me.

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
```diff
@@ -226,9 +280,13 @@ public string GetDefaultAbbreviation(UnitKey unitKey, IFormatProvider? formatPro
...
         /// <param name="formatProvider">The format provider to use for lookup. Defaults to <see cref="CultureInfo.CurrentCulture" /> if null.</param>
         /// <returns>Unit abbreviations associated with unit.</returns>
+        /// <exception cref="UnitNotFoundException">
+        ///     Thrown when no unit information is found for the specified
+        ///     <paramref name="unit" />.
+        /// </exception>
         public string[] GetUnitAbbreviations<TUnitType>(TUnitType unit, IFormatProvider? formatProvider = null) where TUnitType : struct, Enum
         {
-            return GetUnitAbbreviations(UnitKey.ForUnit(unit), formatProvider).ToArray();  // TODO can we change this to return an IReadonlyCollection (as the GetAbbreviations)?
+            return GetUnitAbbreviations(UnitKey.ForUnit(unit), formatProvider).ToArray(); // TODO can we change this to return an IReadOnlyList (as the GetAbbreviations)?
```
angularsen, line 289: Let's do it, we are more free to change the underlying type later

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
```diff
@@ -247,128 +309,147 @@ public string[] GetUnitAbbreviations(Type unitType, int unitValue, IFormatProvid
...
-                ? GetAbbreviations(unitInfo, FallbackCulture)
+            return abbreviations.Count == 0 && HasFallbackCulture(culture)
+                ? GetAbbreviationsForCulture(unitInfo, FallbackCulture)
                 : abbreviations;
         }
 
+        internal static bool HasFallbackCulture(CultureInfo culture)
+        {
+            // accounting for the fact that we're using the same abbreviations for both "en-US" and the "Invariant" culture (Name == string.Empty) 
+            return culture.Name != string.Empty && culture.Name != FallbackCulture.Name;
```
angularsen, line 413: Nit, `!Equals(culture, CultureInfo.InvariantCulture)` is more explicit.

The `culture.Name != FallbackCulture.Name` must stay though, since the name is what matters and not the culture info configuration.

angularsen (changes requested):
I think this looks very good, just a few suggestions. Just merge when ready.

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
lipchev: I thought the same thing as I was reviewing this today, but when I tried to change it- i realized that unlike the `ParseUnit` version, this would require two statements:

`if(UnitParser.Default.TryParse(.., out unitInfo)){..}`

and every line adds another `10K` or so to the file size, so I suggest we leave it as an `file-size-reduction` optimization.

UnitsNet.Benchmark/Initializations/UnitAbbreviationsCacheInitializationBenchmarks.cs
lipchev: Originally I was using this overload for testing the initialization of the `UnitAbbreviationsCache`, as it avoids the call to the `QuantityInfoLookup` but now that the `Mass.GetAbbreviation` call does it instead (using the internal overload)- I could switch it over and hopefully remove the method (see the TODO - we shouldn't accept a `UnitInfo` as an argument, as it might not be the same `UnitInfo` that was registered earlier).

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
lipchev: The goal here is to have the list of `QuantityInfo`s provided via the constructor, unlike before where a new `UnitInfo` was created when trying to get / map an abbreviation to a `TUnit` that doesn't exist in the `QuantityInfoLookup`.

The natural meaning IMO is that this creates an empty cache (meaning nothing is cached) using the default quantities (for mapping). In the 🐲  version I've got a few more constructor overloads, which should make it easier to construct / customize the default list.

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
lipchev: 👍 After changing this to `IReadOnlyList` - there shouldn't be any need for either `ToArray` or `ToList` .

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
angularsen: This can probably go into an extension method too at some point, right?

lipchev: I haven't got the preview installed- but from what I gather yes, that should work as a static extension.

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
angularsen: Okay maybe I just don't understand it, to me it seems like we're getting a cache with all built-in unit abbreviations when calling `new UnitAbbreviationsCache()` ?

lipchev: Yes, but it's an _empty cache_, as one would expect from a call to `new Cache()`

lipchev: Of course it has always been an empty cache, it's just that before it was an empty cache _and_ an empty list of quantities to map (which was problematic).

angularsen: Right, and this is more a theoretical exercise to challenge the design, but it seems nice to be able to use the pieces on their own without involving all the built-in unit definitions.

```cs
// For some reason, we don't want to initialize with HowMuch at this point
var cache = new UnitAbbreviationsCache();

// Instead add it later, in one of two ways
cache.AddQuantity(HowMuch.Info);
cache.Quantities.Add(HowMuch.Info);

// Then map the abbreviations
cache.MapUnitToDefaultAbbreviation(HowMuchUnit.Some, "sum");
```

angularsen: I guess it comes down to that it feels more intuitive to get this behavior from the default ctor, than having to call `CreateEmpty()`.

lipchev: Oh, that would be very difficult to accomplish.. Unless you want to add a message bus or something 😄. Of course we could have an assumption that the user would call `AddQuantity` before attempting to `Map` or `Get` it but again- that would make it impossible to use the `FrozenDictionary`, not to mention the concurrency issues once again, please let's not.. (_insert meme here_).

The `UnitParser` is now really lightweight, the user should be able to create one on the fly, but for the initial mapping (of the `DefaultAbrreviations`) I really hope that we could use something like `UnitsNet.Configure(c=> ... add custom stuff here)` at the app start.

angularsen: Alright, this is not an important point for me. I get that the new design forces you to either use the default ctor, with built-in quantities, or pass your own quantities, and that it becomes immutable at that point. That's fine.

We'll still need to accomodate adding custom quantities at runtime for the default singleton though, but this should be doable either by calling something early on startup as you say or swapping the immutable instance with a new one.

UnitsNet/CustomCode/UnitParser.cs
```diff
@@ -6,309 +6,609 @@
...
+    ///     This method performs a series of searches to find the unit:
+    ///     <list type="number">
+    ///         <item>Case-sensitive search using the current culture.</item>
+    ///         <item>Case-sensitive search using the fallback culture, if applicable.</item>
+    ///         <item>Case-insensitive search.</item>
+    ///     </list>
+    ///     Note that this method is not optimized for performance, as it enumerates all units and their abbreviations
+    ///     during each invocation.
+    /// </remarks>
+    public UnitInfo GetUnitFromAbbreviation(string unitAbbreviation, IFormatProvider? formatProvider)
```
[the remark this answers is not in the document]
lipchev: fixed in https://github.com/angularsen/UnitsNet/pull/1548/commits/f9a2766b959334432c531c44afca207f37c155cc

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
lipchev: fixed in https://github.com/angularsen/UnitsNet/pull/1548/commits/f9a2766b959334432c531c44afca207f37c155cc

UnitsNet.Benchmark/Initializations/UnitAbbreviationsCacheInitializationBenchmarks.cs
lipchev: the overload (and it's one test) have been removed in https://github.com/angularsen/UnitsNet/pull/1548/commits/f9a2766b959334432c531c44afca207f37c155cc

UnitsNet/CustomCode/UnitAbbreviationsCache.cs
lipchev: Sorry, forgot about this one: `culture.Equals` is a bit more expensive than simply checking the name:
``` csharp
  public override bool Equals([NotNullWhen(true)] object? value)
  {
    if (this == value)
      return true;
    return value is CultureInfo cultureInfo && this.Name.Equals(cultureInfo.Name) && this.CompareInfo.Equals((object) cultureInfo.CompareInfo);
  }
```

and I think there is a potential risk for key-duplication if the user uses a custom culture (derived from `CultureInfo.InvariantCulture`).

As for the constant- I tried it, but it looked weird without the accompanying comment (which got even more vague when the string value got hidden by the constant name 😄 ).

PS This looks weird to me too, but I don't see how to improve it..

UnitsNet.Benchmark/Initializations/UnitAbbreviationsCacheInitializationBenchmarks.cs
lipchev: >cache is never used
Oh I see, this isn't doing anything anymore.. I'll just remove them.

UnitsNet/QuantityFormatter.cs
```diff
@@ -170,7 +170,7 @@ private static string FormatUntrimmed<TUnitType>(IQuantity<TUnitType> quantity,
                 {
                     var abbreviations = UnitsNetSetup.Default.UnitAbbreviations.GetUnitAbbreviations(quantity.Unit, formatProvider);
 
-                    if (abbreviationIndex >= abbreviations.Length)
+                    if (abbreviationIndex >= abbreviations.Count)
```
lipchev, line 173: This is the unfortunate side effect of moving away from the arrays, but there is no way around it..

I'm merging it..

PR #1551, opened 2025-04-19
`QuantityFormatter`: introducing an `UnitAbbreviationsCache` instance and `Format`-ing using a generic `TQuantity`

### LINKED ISSUE #1447: Add the possibility to instantiate a QuantityFormatter
Just like the `UnitParser` and the `QuantityParser` can be created with an `UnitAbbreviationsCache` - the same goes for the `QuantityFormatter`. 

This should allow for different parts of the application to use their own "defaults".

In terms of implementation - we can keep the static `Format` methods, but they should delegate the call to the `Default` instance. 

There is a slight problem with the name collision- but my plan involves changing the instance member generic parameter to

``` csharp
    public string Format<TQuantity>(TQuantity quantity, string? format, IFormatProvider? formatProvider)
        where TQuantity : IQuantity
```

This avoids the boxing of the struct via the `IQuantity<TUnit>` interface which is currently used by the static `Format`. 

Now, obviously once this is done- the two existing overloads should be deprecated, however the question arises- whether there should by a static alternative or can we simply point to the `QuantityFormatter.Default`? Note that should we decide to have the static `TQuantity` versions, then we won't be able to have both use the same name.. 

We should probably also take / expose an instance of the `QuantityFormatter` on the `UnitsNetSetup`, but that's not a requirement.

lipchev:
- `QuantityFormatter`: introducing an `UnitAbbreviationsCache` instance and `Format`-ing using a generic `TQuantity`
- `UnitsNetSetup`: introduced an instance property for the `UnitFormatter`
- `IQuantity`: added the `UnitKey` property (implemented explicitly)
- replaced the existing usages of the `QuantityFormatter` and marked the static `Format` overloads as `[Obsolete]`

fixes #1447


lipchev:
If you're ok with the addition of the `UnitKey` to the `IQuantity` interface (as a public property), then we could compensate it by removing the `string ToString(IFormatProvider? provider)` :

``` csharp
    public static string ToString<TQuantity>(this TQuantity quantity, IFormatProvider? formatProvider)
        where TQuantity : IQuantity, IFormattable
    {
        return quantity.ToString(null, formatProvider);
    }
```

UnitsNet/QuantityFormatter.cs
```diff
@@ -206,19 +186,59 @@ private static string FormatUntrimmed<TUnitType>(IQuantity<TUnitType> quantity,
...
-        return string.Format(formatProvider, $"{{0:{format}}} {{1}}", quantity.Value, abbreviation);
+        var abbreviation = _unitAbbreviations.GetDefaultAbbreviation(quantity.UnitKey, formatProvider);
+        if (abbreviation.Length == 0)
+        {
+            return quantity.Value.ToString(format, formatProvider);
+        }
+
+#if NET
+        // TODO see about using the Span<char> overloads (net 8+)
+        return quantity.Value.ToString(format, formatProvider) + ' ' + abbreviation;
```
angularsen, line 201: No idea if it makes sense to bother with StringBuilder here, or if it's just overkill. I'm not familiar with any Span<char> overloads to concatenate strings without allocations, but maybe there exists some stuff here.

UnitsNet/QuantityFormatter.cs
```diff
@@ -206,19 +186,59 @@ private static string FormatUntrimmed<TUnitType>(IQuantity<TUnitType> quantity,
...
+#else
+        return quantity.Value.ToString(format, formatProvider) + ' ' + abbreviation;
+#endif
     }
 
-    private static string ToStringWithSignificantDigitsAfterRadix<TUnitType>(IQuantity<TUnitType> quantity, IFormatProvider formatProvider, int number)
-        where TUnitType : struct, Enum
+    private string ToStringWithSignificantDigitsAfterRadix<TQuantity>(TQuantity quantity, IFormatProvider formatProvider, int number)
```
angularsen, line 207: Why do we use generics `TQuantity` several places in `QuantityFormatter`? 

It does not seem used for anything or provide anything beyond what a regular `IQuantity` parameter provides.
The JIT has to create a copy of each generic class or class member, for each concrete type you use it with at runtime.

Usually, generics are used to preserve type info for outputs, but the string formatter only outputs strings.

angularsen (approved):
Looks good, just a few minor suggestions.

angularsen:
> If you're ok with the addition of the `UnitKey` to the `IQuantity` interface (as a public property), then we could compensate it by removing the `string ToString(IFormatProvider? provider)` :
> 
> ```cs
>     public static string ToString<TQuantity>(this TQuantity quantity, IFormatProvider? formatProvider)
>         where TQuantity : IQuantity, IFormattable
>     {
>         return quantity.ToString(null, formatProvider);
>     }
> ```

This feels a bit unnecessary, `UnitKey` and `Unit` duplicates the same information, and as far as I can tell we use `UnitKey` primarily for internal optimization purposes.

The implementation for `UnitKey` property also just calls `UnitKey.ForUnit(Unit)` anyway, so why don't we just call `UnitKey.ForUnit(myQuantity.Unit)` whenever we need the key internally? 

Or have an extension method to do `myQuantity.GetUnitKey()` for shorter syntax.

angularsen:
A couple of comments on `UnitKey` type:
1. The `ctor(Type, int)` should 
	2. Throw if passing non-enum `Type`
	3. Throw if `!Enum.IsDefined(UnitType, UnitValue)`
	4. ..or make the ctor private so you can only construct via generic factory methods, that kind of ensure these checks without perf penalty
	5. Parameters should have xmldoc, in particular it is not described that `UnitType` should be an enum type and that `UnitValue` is the enum's integer value
6. Rename `UnitType` to `UnitEnumType` or `EnumType`
7. Rename `UnitValue` to `UnitEnumValue` or `EnumValue`

UnitsNet/QuantityFormatter.cs
lipchev: it's more interesting than just the concatenation; the game is to use a `stacalloc` to create a buffer and then use `TryFormat`  from the `ISpanFormattable` interface on the number, appending the abbreviation after it- with zero intermediate strings allocations.

UnitsNet/QuantityFormatter.cs
lipchev: We want to avoid boxing the quantity using the `IQuantity` interface. 

I'm aware of the (very small increase) that the generic method generates, but here we're already offsetting this by removing another generic method down the line (on the `UnitFormatter`).

I originally monitored the size after every additional method, but I didn't bother recording it here now - it's very small.

lipchev:
> This feels a bit unnecessary, `UnitKey` and `Unit` duplicates the same information, and as far as I can tell we use `UnitKey` primarily for internal optimization purposes.
> 
> The implementation for `UnitKey` property also just calls `UnitKey.ForUnit(Unit)` anyway, so why don't we just call `UnitKey.ForUnit(myQuantity.Unit)` whenever we need the key internally?
> 
> Or have an extension method to do `myQuantity.GetUnitKey()` for shorter syntax.

The role of the `UnitKey` is as a replacement for the `Enum Unit` property of the `IQuantity` interface - it's use is redundant only when the `TUnit` is known. However, if we wanted to use `TQuantity` with a generic `TUnit` we'd have to use 2 type parameters; `<TQuantity, TUnit>`, but now the same methods can only have the `TQuantity` parameter, without having to go though boxing and unboxing the `Enum`.

UnitsNet/QuantityFormatter.cs
lipchev: By the way, the whole `QuantityFormatter` can be re-written without the use of any generics (by using something like `double value, UnitKey unit, string format` for the parameters).

lipchev:
> A couple of comments on `UnitKey` type:
> 
>     1. The `ctor(Type, int)` should
>        2. Throw if passing non-enum `Type`
>        3. Throw if `!Enum.IsDefined(UnitType, UnitValue)`
>        4. ..or make the ctor private so you can only construct via generic factory methods, that at least checks the type is an enum, but `IsDefined` check is probably useful to have for those call paths too
>        5. Parameters should have xmldoc, in particular it is not described that `UnitType` should be an enum type and that `UnitValue` is the enum's integer value

I'll add another `Create` method and make the constructor internal, however I'm not sure about the existence check - I was kinda counting on it for the _extending with custom units_ part. It's a bit of a _hacky_ way of extending the `MassUnit`, but it works.

>     2. Rename `UnitType` to `UnitEnumType` or `EnumType`
> 
>     3. Rename `UnitValue` to `UnitEnumValue` or `EnumValue`

I used the same names as we typically use for these parameters: `GetDefaultAbbreviation(Type unitType, int unitValue)` - I hope to eventually replace all these with a single `UnitKey` parameter (less primitives).

If I had to rename it, I'd probably use the `UntEnumType` and `UnitEnumValue` but honestly I don't see much reason to do it - so if you want it renamed, you pick the name. 😛

UnitsNet/QuantityFormatter.cs
```diff
@@ -206,19 +186,59 @@ private static string FormatUntrimmed<TUnitType>(IQuantity<TUnitType> quantity,
...
+    ///     The quantity to format.
+    /// </param>
+    /// <returns>
+    ///     A string representation of the quantity, formatted using the "G" format and the current culture.
+    /// </returns>
+    public string Format<TQuantity>(TQuantity quantity)
+        where TQuantity : IQuantity
+    {
+        return Format(quantity, null, null);
     }
```
lipchev, line 243: What do you think of this overload - I wasn't sure whether to have it as a separate overload or have the format of the other overload as optional: `string? format = null`?

angularsen: Default values instead of many overloads.

angularsen: > I used the same names as we typically use for these parameter

I know, but it really isn't obvious what `unitType` and `unitValue` means without having to read the xmldoc first. `unitEnumType` is more explicit, we don't accept any other types. I'll take readability over a bit longer names any day.

UnitsNet/QuantityFormatter.cs
angularsen: I guess we could just pass the concrete values instead then, or even create a separate `struct` for value+UnitKey if we feel it helps. 

I don't know, it just feels overkill to tack generics to everything just to avoid boxing of quantities by interface.

lipchev: Frankly, now that I think about it- the reason for having a quantity as the format parameter in the first place was that it was necessary for those other formats that we removed (tapping into the `QuantityInfo`).

I think this might actually be a worthwhile change to make (should I do it?).

On the question of boxing the structs- well that's probably the main reason microsoft went crazy with the `INumber` interfaces 😄

lipchev:
> > However, if we wanted to use `TQuantity` with a generic `TUnit` we'd have to use 2 type parameters; `<TQuantity, TUnit>`, but now the same methods can only have the `TQuantity` parameter, without having to go though boxing and unboxing the `Enum`.
> 
> For my understanding, can you provide an example with and without `UnitKey` added to `IQuantity`, that illustrates why it is useful or needed?

I think the `QuantityFormatter` was probably the best example, but let me try make a simplified example:

``` csharp
    internal string GetDefaultAbbreviation<TQuantity, TUnit>(TQuantity quantity)
        where TQuantity : IQuantity<TUnit>
        where TUnit : struct, Enum
    {
        return _unitAbbreviations.GetDefaultAbbreviation(quantity.Unit); // no boxing
    }

    internal string GetDefaultAbbreviationWithTypeParameter<TQuantity>(TQuantity quantity)
        where TQuantity : IQuantity
    {
        return _unitAbbreviations.GetDefaultAbbreviation(quantity.UnitKey); // no boxing
    }
```

Now imagine how we would call these two extensions with a concrete type:

``` csharp
        var a1 = GetDefaultAbbreviation<Mass, MassUnit>(Mass.Zero);
        var a2 = GetDefaultAbbreviationWithTypeParameter(Mass.Zero);
```

In the example of `a1` there is no way for the second parameter to be inferred automatically (I have to type it manually).

UnitsNet/QuantityFormatter.cs
lipchev: Ah, perhaps this is a good place to note the possibility to implement certain things on the interface itself (and the caveats). Consider this example:
``` csharp

#if NET
        internal static abstract TQuantity Create(QuantityValue value, UnitKey unit); // on an intermediate interface extending IQuantity
#else

#if NET7_0_OR_GREATER
        /// <summary>
        ///     Creates an instance of the quantity from a specified value and unit.
        /// </summary>
        /// <param name="value">The numerical value of the quantity.</param>
        /// <param name="unit">The unit of the quantity.</param>
        /// <returns>An instance of the quantity with the specified value and unit.</returns>
        static abstract TSelf From(QuantityValue value, TUnitType unit);

        static TSelf IQuantityInstance<TSelf>.Create(QuantityValue value, UnitKey unit) => TSelf.From(value, unit.ToUnit<TUnitType>());
#endif
```

The only thing that the `Mass` (or the `HowMuch`) needs to implement is the static `From(QuantityValue, MassUnit)` method (which it already does) and now we can have an extension method that takes just a [single type parameter](https://github.com/lipchev/UnitsNet/blob/fractional-quantity-value/UnitsNet/QuantityExtensions.cs#L132):

``` csharp
    internal static TQuantity ArithmeticMean<TQuantity>(this IEnumerable<TQuantity> quantities)
        where TQuantity : IQuantityInstance<TQuantity>
```

And now the resulting `TQuantity` can be created directly using:

``` csharp
return TQuantity.Create(sumOfValues / nbValues, resultUnit); // resultUnit is of type `UnitKey`
```

Since we're talking about static methods, there isn't any boxing and we get 0 allocations.

However, this would not be the case if we were to use a default implementation of an instance method such as `ToUnit(TUnit)` (this is counterintuitive, and may change in a future .NET versions but currently this involves an _interface dispatch_ or something, that ultimately makes the way like this: `Mass -> IQuantity<Mass, MassUnit> -> Mass`).

That's less of an issue when the return type is already an interface, like with the _obsolete_ versions of these methods:

``` csharp
        [Obsolete("This method will be removed from the interface in the next major update. Consider using the UnitConverter.Default.ConvertTo(quantity, unit) method instead.")]
        IQuantity ToUnit(Enum unit)
#if NET
            => UnitConverter.Default.ConvertTo(this, unit)
#endif
        ;
```

UnitsNet/QuantityFormatter.cs
lipchev: > > I used the same names as we typically use for these parameter
> 
> I know, but it really isn't obvious what `unitType` and `unitValue` means without having to read the xmldoc first. `unitEnumType` is more explicit, we don't accept any other types. I'll take readability over a bit longer names any day.

It's like asking me to rename my children but ok...  fixed in https://github.com/angularsen/UnitsNet/pull/1551/commits/3d7a7c5cecaedfdec630b873155f091bcc55dbe5

lipchev: I also added the `Create` method with an `Enum` check, but it might also make sense to actually use it (in the places where we currently have the primitives)..

lipchev: > I also added the `Create` method with an `Enum` check, but it might also make sense to actually use it (in the places where we currently have the primitives)..

fixed in https://github.com/angularsen/UnitsNet/pull/1551/commits/38e868a9295410df977302630c20e355a9fa84c3

lipchev: > Default values instead of many overloads.

fixed in https://github.com/angularsen/UnitsNet/pull/1551/commits/e829882371bd46d46918f13dd3b14de016a3517e

Does this also apply for the methods of the `IFormattable` interface? I'd be happy to remove the overloads and simply make the parameters optional (there wouldn't be any need for the `ToString` extensions). Note we currently have just one of the overloads defined on the interface, the other one (`ToString(string? format)`)  is just generated as a sort of an overload helper. 

The problem is that removing them would be breaking change:
1) for the `HowMuch` extensions: I reckon this is not gonna be missed
2) for people using `Mass.ToString(CultureInfo.InvariantCulture)` : there is probably somebody using this, but I'm not sure how useful this is..

angularsen: We're already breaking changes all over, but generally I'd like existing code to "just work" when recompiled as much as possible.

I see `IFormattable` requires `public string ToString(string? format, IFormatProvider? provider)`, and `ValueType` requires  `public override string ToString()` so maybe for ToString() we are a bit limited on how much we can use default values.

lipchev: > I see IFormattable requires public string ToString(string? format, IFormatProvider? provider), and all objects already have public string ToString(), so maybe for ToString() we are a bit limited on how much we can use default values to avoid overloads.
We can still change make the `nullable` parameters optional in the override- it's still attached to the `IFormattable` interface.

> We're already breaking changes all over, but generally I'd like existing code to "just work" when recompiled as much as possible.

Unfortunately, here this is exactly the tradeoff - if we want to not have a breaking change, we have to have both overloads (whether it's as extensions or members).

UnitsNet/QuantityFormatter.cs
lipchev: > Frankly, now that I think about it- the reason for having a quantity as the format parameter in the first place was that it was necessary for those other formats that we removed (tapping into the `QuantityInfo`).
> 
> I think this might actually be a worthwhile change to make (should I do it?).

Actually, after thinking about a little more- although totally feasible, I don't think we should do it:
1. Using the `UnitKey` as a parameter is exposing the _internals_ and should be generally be used internally (or for people who are forced to use loosely typed stuff).
2. While the `Mass.ToString` itself wouldn't change, if we wanted to use `QuantityFormatter.Default.Format(mass.Value, mass.UnitKey)` (or another instance) we couldn't have the `UnitKey` implementation as explicit.

lipchev:
@angularsen Unless you have something to add, I think this is ready to go. Tell me how you want to deal with the `ToString` stuff, and I'll open another PR.

angularsen:
Yeah let's merge this and address topics in separate PRs.

1. Ok, let's try `UnitKey` in `IQuantity` and keeping generics in `QuantityFormatter`. It took me some time to grok why we couldn't keep `UnitKey` an internal implementation detail, but I realize now the concrete type must provide the `int` and `Type` for the enum to avoid boxing of `Enum`. The benchmarks show an infinite improvement on allocation, so that is nice 👏 

2. ToString() overloads - let's discuss this separately, preferably with a PR on how you think it should work and with some bullet points on pros/cons of ways to do it. It will be easier for me to comment.

PR #1552, opened 2025-04-19
Replacing the ToString overloads with extensions

lipchev:
- introducing the `QuantityExtensions` (in the Extensions folder) with the first two extensions for the `ToString` overloads
- removed the `ToString(IFormatProvider?)` overload from the `IQuantity` interface
- removed the `ToString` overloads from the generated quantities (and `HowMuch`)


lipchev:
Ok, here is the size difference:
> before: 2.19 MB (2 301 440 bytes)
> after: 2.17 MB (2 286 080 bytes)

Since we've only got one generic parameter, the bug with Resharper does not apply.

The only breaking parts concern the cases where a given class doesn't have the `UnitsNet` namespace import (e.g. when given a type that contains a quantity, which he then access without assigning to a variable).

I'll make separate PR, with the other version which simply removes the overloads and makes the parameters optional.

angularsen:
Yeah, I don't like how default parameters for ToString() make it awkward to only pass a culture. It's just so common to have relevant ToString() overloads for this.
https://github.com/angularsen/UnitsNet/pull/1553#discussion_r2051710626

Extension methods could work though, but it does have some potential developer expirence friction as you point out regarding missing namespace imports. I expect this to be fairly uncommon though and it only affects 2 of 4 ToString() overloads, and IDEs may do a decent job of importing namespaces for you.

I guess it comes down to, is the 25 kB binary size reduction worth a potentially worse developer experience? ToString() is kind of used a lot. 

I'm kinda leaning towards maybe not enough impact to justify this change, what do you think?

angularsen:
With the perspective that we want to try and use extension methods more in general to avoid duplicating code for N quantities, maybe we'll just have to try this and see how it feels.

lipchev:
For me personally it's just the interface method that's bothering me.. It might look like I'm optimizing for size, but that's really not a much of priority (in my use case).

angularsen:
Merged as part of #1587

PR #1553, opened 2025-04-19
Making the `ToString` paramters optional (IFormattable)

lipchev:
- removed the `ToString(IFormatProvider?)` overload from the `IQuantity` interface
- removed the `ToString` overloads from the generated quantities (and `HowMuch`)
- replaced the usages of `ToString(IFormatProvider?)` in the tests

duplicate of #1552


lipchev:
Hm, I was expecting there to be a very small size difference compared to #1552,  but 
> before: 2.19 MB (2 301 440 bytes)
> after: 2.18 MB (2 287 616 bytes)
> #1552 : 2.17 MB (2 286 080 bytes) 😮 

I don't know what to make of these numbers..

Anyway, I was also a little surprised by the number of breaking changes that this caused in the tests (non-generated).. I initially found the `ToString` extensions a little awkward, but I now I think they might be the better option..

UnitsNet.Tests/CustomCode/LengthTests.cs
```diff
@@ -167,15 +167,15 @@ public void LengthTimesSpecificWeightEqualsPressure()
         public void ToStringReturnsCorrectNumberAndUnitWithDefaultUnitWhichIsMeter()
         {
             var meter = Length.FromMeters(5);
-            string meterString = meter.ToString(CultureInfo.InvariantCulture);
+            string meterString = meter.ToString(null, CultureInfo.InvariantCulture);
```
angularsen, line 170: This I don't like, c# users expect certain ToString() signatures to exist and having to use named parameters or pass `null` is awkward.

PR #1575, opened 2025-07-25
refactor: change DefaultQuantities to use property initializer

angularsen:
Fixup of #1555


angularsen:
Right, but I thought the idea was that this is the built-in provider. Any customizations happen further out, such as in `MassInfo.CreateDefault(Func customizeUnits)`: https://github.com/angularsen/UnitsNet/pull/1555#pullrequestreview-2785926445

lipchev:
I've merged #1555

PR #1576, opened 2025-07-25
refactor: rename to Quantity.DefaultProvider.Quantities

angularsen:
Fixup of #1555

### DIFF
--- a/UnitsNet/QuantityInfoLookup.cs
+++ b/UnitsNet/QuantityInfoLookup.cs
@@ -1,7 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
-using System.Linq;
+﻿using System.Linq;
 #if NET8_0_OR_GREATER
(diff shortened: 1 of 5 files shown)


UnitsNet/QuantityInfoLookup.cs
```diff
@@ -23,7 +23,7 @@ namespace UnitsNet;
 /// </remarks>
 internal class QuantityInfoLookup
 {
-    private readonly QuantityInfo[] _quantities;
+    private readonly IReadOnlyList<QuantityInfo> _quantities;
```
lipchev, line 26: there is a slight performance overhead of using the interface on the private field, but if you prefer clarity- it's fine by me..

angularsen: This change was collateral damage, I'll revert it. The idea was to accept IReadOnlyList as input and reuse instance if compatible, but I'll do the same for array instead.

angularsen:
Sounds good

PR #1578, opened 2025-07-25
✨Configure unit conversions via IQuantityInfo

angularsen:
- Add `IQuantityInfo.RegisterUnitConversions` callback to configure unit conversions
- Change `UnitConverter` to configure unit conversions from a list of `IQuantityInfo`
- Change `UnitsNetSetup` to pass the default quantity infos into its default `UnitConverter`
- Add sample conversions for `HowMuch`

### DIFF
--- a/UnitsNet/UnitConverter.cs
+++ b/UnitsNet/UnitConverter.cs
@@ -1,12 +1,8 @@
 ﻿// Licensed under MIT No Attribution, see LICENSE file at the root.
 // Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
 
-using System;
 using System.Collections.Concurrent;
-using System.Diagnostics.CodeAnalysis;
-using System.Reflection;
 using UnitsNet.InternalHelpers;
-using UnitsNet.Units;
 
 namespace UnitsNet
 {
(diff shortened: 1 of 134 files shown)


angularsen:
Maybe you already have done something like this in your dragon PR, but while reviewing it seemed natural to add something like this.

UnitsNet/UnitConverter.cs
```diff
@@ -76,19 +72,11 @@ private ConcurrentDictionary<ConversionFunctionLookupKey, ConversionFunction> Co
...
+        private static void RegisterUnitConversions(UnitConverter unitConverter, IEnumerable<QuantityInfo> quantityInfos)
         {
-            if (unitConverter is null)
-                throw new ArgumentNullException(nameof(unitConverter));
-
-            foreach (var quantity in Quantity.DefaultProvider.Quantities)
+            foreach (QuantityInfo quantityInfo in quantityInfos)
             {
-                var registerMethod = quantity.QuantityType.GetMethod(nameof(Length.RegisterDefaultConversions), BindingFlags.NonPublic | BindingFlags.Static);
-                registerMethod?.Invoke(null, new object[]{unitConverter});
```
angularsen, line 91: Good riddance to this reflection stuff

lipchev:
Yes, this is already covered these by introducing the `ConversionFromBase` and `ConversionToBase` properties (of type `ConversionExpression`) in the `UnitInfo`.

PR #1587, opened 2025-07-27
Refactored the `IQuantity` interface

lipchev:
- refactored the `IQuantity` interface: removed the `Equals` with tolerance and the `ToString` overloads
- introduced `IAffineQuantity` (see `Temperature`), `ILogarithmicQuantity` (see `AmplitudeRatio`) and `ILinearQuantity` (see `Mass`)
- introduced extension methods for `Equals` with tolerance as well as `Sum` and `Average` (as applicable)


lipchev:
@angularsen This is the most minimalist version of the interface, applicable to the current code-base (which relies on the `DefaultUnitConverter`): 

1. In the 🐲 PR there isn't the need for the (newly added) `double As(UnitKey unitKey)` method to exist on the interface (or the extending type), currently it's just there as a work-around the inability to use `UnitConverter.Default.ConvertValue(quantity, toUnit)` from the extension.
2. I've removed the _very generic_ `Equals(IQuantity, IQuantity)` (as we have agreed in an earlier discussion)
3. I've also removed the concrete `Equals` with tolerance- it is now an extension method. It is possible to have it on the interface - just not on that level (as we need to be able to specify the type of tolerance):

This definition applies only to the `ILinearQuantity` and `ILogarithmicQuantity`
```  csharp
        /// <summary>
        ///     <para>
        ///     Compare equality to <paramref name="other"/> given a <paramref name="tolerance"/> for the maximum allowed +/- difference.
        ///     </para>
        ///     <example>
        ///     In this example, the two quantities will be equal if the value of b is within 0.01 of a (0.01m or 1cm).
        ///     <code>
        ///     var a = Length.FromMeters(2.0);
        ///     var b = Length.FromMeters(2.1);
        ///     var tolerance = Length.FromCentimeters(10);
        ///     a.Equals(b, tolerance); // true, 2m equals 2.1m +/- 0.1m
        ///     </code>
        ///     </example>
        ///     <para>
        ///     It is generally advised against specifying "zero" tolerance, due to the nature of floating-point operations.
        ///     </para>
        /// </summary>
        /// <param name="other">The other quantity to compare to.</param>
        /// <param name="tolerance">The absolute tolerance value. Must be greater than or equal to zero.</param>
        /// <returns>True if the absolute difference between the two values is not greater than the specified tolerance.</returns>
        bool Equals(TSelf? other, TSelf tolerance);
```

For the `IAffineQuantity` (e.g. the `Temperature`) we need to use this definition:

``` csharp
        bool Equals(TSelf? other, TOffset tolerance);
```

UnitsNet/ExceptionHelper.cs
lipchev, line 1: Sorry, I should have added this one earlier.. Note that in the 🐲 PR this (and the exceptions) are inside an `Exceptions` folder (no namespace generation)- same as the `Extensions` folder.

UnitsNet/UnitMath.cs
```diff
@@ -9,64 +10,39 @@ namespace UnitsNet
...
+        public static TQuantity Abs<TQuantity>(this TQuantity value) where TQuantity : IQuantity
         {
-            return source.Select(selector).Sum(unitType);
+            // TODO see about constraining to IQuantityInstance<TQuantity>
+// #if NET
+//             return TQuantity.Create(QuantityValue.Abs(value.Value), value.UnitKey);
+// #else
+//             return value.QuantityInfo.Create(QuantityValue.Abs(value.Value), value.UnitKey);
+// #endif
+            return (TQuantity)value.QuantityInfo.From(Math.Abs(value.Value), value.UnitKey);
```
lipchev, line 36: Eh, one of the commented-out bits has slipped in without me noticing, but the `TODO` is valid (but that's another breaking change).

Personally, I think we should deprecate this method and make it a static method in the `Quantity.cs`: e.g. `var absMass = Quantity.Abs(mass)`.

Alternatively, this could be a static method on the concrete type- e.g. `Mass.Abs(mass)` but that would add bytes to build (we could also have it as a static extension property in `net10`).

UnitsNet/Extensions/QuantityExtensions.cs
```diff
@@ -0,0 +1,139 @@
...
+
+    /// <summary>
+    ///     Calculates the arithmetic mean of a sequence of quantities.
+    /// </summary>
+    /// <typeparam name="TQuantity">The type of the quantity.</typeparam>
+    /// <param name="quantities">The collection of quantities to average.</param>
+    /// <returns>The average of the quantities, using the unit of the first element in the sequence.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when the collection is null.</exception>
+    /// <exception cref="InvalidOperationException">Thrown when the collection is empty.</exception>
+    internal static TQuantity ArithmeticMean<TQuantity>(this IEnumerable<TQuantity> quantities)
```
angularsen, line 67: There are a lot of mathy extension methods spread around, that seem related:
- QuantityExtensions.ArithmeticMean
- UnitMath
- GenericMathExtensions

I think we should gather these into a single file, but out of scope for this PR.

We can put everything here, in `QuantityExtensions` and a partial file, or in a separate `QuantityMathExtensions` class to isolate a bit for intellisense.

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
```diff
@@ -187,9 +187,6 @@ public Angle(double value, AngleUnit unit)
         /// </summary>
         public static Angle Zero => Info.Zero;
 
-        /// <inheritdoc cref="Zero"/>
-        public static Angle AdditiveIdentity => Zero;
```
angularsen, line 191: Clever using default interface implementations to not have to do this for each.

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
```diff
@@ -224,6 +221,11 @@ public Angle(double value, AngleUnit unit)
         [DebuggerBrowsable(DebuggerBrowsableState.Never)]
         QuantityInfo<AngleUnit> IQuantity<AngleUnit>.QuantityInfo => Info;
 
+#if NETSTANDARD2_0
+        [DebuggerBrowsable(DebuggerBrowsableState.Never)]
+        IQuantityInstanceInfo<Angle> IQuantityInstance<Angle>.QuantityInfo => Info;
+#endif
```
angularsen, line 227: Out of scope, but I don't find `IQuantityInstanceInfo` and `IQuantityInstance` names intuitive. 

It says
> This is a specialization of IQuantityInfo that is used (internally) for constraining certain methods, without having to include the unit type as additional generic parameter.

1. Should they be `internal`?
2. How about naming them `ITypedQuantityInfo` and `ITypedQuantity` or similar?

angularsen (commented):
I didn't get to the end and have to run, but here are comments so far.

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
```diff
@@ -786,6 +788,15 @@ public bool Equals(Angle other)
 
         #pragma warning restore CS0809
 
+        /// <summary>
+        ///     Returns the hash code for this instance.
+        /// </summary>
+        /// <returns>A hash code for the current Angle.</returns>
+        public override int GetHashCode()
+        {
+            return new { Info.Name, Value, Unit }.GetHashCode();
```
angularsen, line 797: Except for netstandard, this can be `return HashCode.Combine(Info.Name, Value, Unit)` and probably save an allocation.

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
```diff
@@ -822,88 +833,6 @@ public int CompareTo(Angle other)
...
-        ///     Note that it is advised against specifying zero difference, due to the nature
-        ///     of floating-point operations and using double internally.
-        ///     </para>
-        /// </summary>
-        /// <param name="other">The other quantity to compare to.</param>
-        /// <param name="tolerance">The absolute or relative tolerance value. Must be greater than or equal to 0.</param>
-        /// <param name="comparisonType">The comparison type: either relative or absolute.</param>
-        /// <returns>True if the absolute difference between the two values is not greater than the specified relative or absolute tolerance.</returns>
-        [Obsolete("Use Equals(Angle other, Angle tolerance) instead, to check equality across units and to specify the max tolerance for rounding errors due to floating-point arithmetic when converting between units.")]
-        public bool Equals(Angle other, double tolerance, ComparisonType comparisonType)
```
angularsen, line 866: I love getting rid of these hacks ❤️

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
```diff
@@ -1077,27 +1012,6 @@ public override string ToString()
...
-        /// <summary>
-        /// Gets the string representation of this instance in the specified format string using <see cref="CultureInfo.CurrentCulture" />.
-        /// </summary>
-        /// <param name="format">The format string.</param>
-        /// <returns>The string representation.</returns>
-        public string ToString(string? format)
-        {
-            return ToString(format, null);
-        }
-
```
angularsen, line 1100: The idea of having these as extension methods have matured on me, I think it'll work nicely for those that need those overloads.

UnitsNet/GeneratedCode/Quantities/Temperature.g.cs
```diff
@@ -778,6 +710,12 @@ public double As(TemperatureUnit unit)
             return ToUnit(unit).Value;
         }
 
+        /// <inheritdoc cref="IQuantity.As(UnitKey)"/>
+        public double As(UnitKey unitKey)
+        {
+            return As(unitKey.ToUnit<TemperatureUnit>());
+        }
```
angularsen, line 717: I think this could just as well be added as an extension method?

A lot of the existing methods could probably also be moved to extension methods, but let's not go overboard with potential breaking changes. Something to consider for v7 with the intent to reduce binary size.

UnitsNet/IArithmeticQuantity.cs
```diff
@@ -1,20 +1,60 @@
...
+}
+
+/// <summary>
+///     An <see cref="IQuantity{TSelf, TUnitType}" /> that (in .NET 7+) implements generic math interfaces for arithmetic
+///     operations.
 /// </summary>
 /// <typeparam name="TSelf">The type itself, for the CRT pattern.</typeparam>
 /// <typeparam name="TUnitType">The underlying unit enum type.</typeparam>
-public interface IArithmeticQuantity<TSelf, TUnitType> : IQuantity<TSelf, TUnitType>
+public interface IArithmeticQuantity<TSelf, TUnitType> : IQuantity<TSelf, TUnitType>, ILinearQuantity<TSelf>
```
angularsen, line 55: I think it'll be more to clear to name it `ILinearArithmeticQuantity`, if it's arithmetic is for linear quantities.

UnitsNet/IQuantity.cs
```diff
@@ -207,6 +178,121 @@ public interface IQuantity<TSelf, TUnitType> : IQuantity<TUnitType>
...
+        ///     The logarithmic scaling factor.
+        /// </value>
+        double LogarithmicScalingFactor { get; }
+    #endif
+    }
+
+    /// <inheritdoc cref="ILogarithmicQuantity{TSelf}"/>
+    /// <typeparam name="TSelf">The type itself, for the CRT pattern.</typeparam>
+    /// <typeparam name="TUnitType">The underlying unit enum type.</typeparam>
+    public interface ILogarithmicQuantity<TSelf, TUnitType> : IQuantity<TSelf, TUnitType>, ILogarithmicQuantity<TSelf>
```
angularsen, line 286: Let's move these to separate files

angularsen (approved):
Final batch of comments

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -323,12 +342,12 @@ private void GenerateStaticProperties()
         /// </summary>
         public static {_quantity.Name} Zero => Info.Zero;
 ");
-
-            if (_quantity.GenerateArithmetic)
+
```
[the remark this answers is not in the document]
lipchev: It is configured but I don't think I've ever used the `Format/Cleanup` in the source-generator code (fearing that it would reformat it differently). 

We should really do a full solution reformat (but please not until the 🐲 is closed 😄 )

UnitsNet/UnitMath.cs
```diff
@@ -9,64 +10,39 @@ namespace UnitsNet
...
+        /// <param name="value">
+        ///     The quantity whose absolute value is to be calculated.
+        /// </param>
+        /// <returns>
+        ///     A quantity of type <typeparamref name="TQuantity" /> representing the absolute value of the input quantity.
+        /// </returns>
+        /// <exception cref="NullReferenceException">
+        ///     Thrown if the input <paramref name="value" /> is <c>null</c>.
         /// </exception>
-        public static TQuantity Sum<TSource, TQuantity, TUnitType>(this IEnumerable<TSource> source, Func<TSource, TQuantity> selector, TUnitType unitType)
```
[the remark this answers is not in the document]
lipchev: Probably not: the way they are defined, they are still valid (couldn't be used with the `Temperature` as it doesn't have the required operator), but unlike the new extensions, these can only be used on `net8+`.

Their only use case would be for some type of custom quantity which implements `IQuantity, IAdditionOperators<TQuantity, TQuantity, TQuantity>, IAdditiveIdentity<TQuantity, TQuantity>` but does not implement either of the _concrete_ interfaces (`ILinearQuantity`, `IAffineQuantity` etc).

My original thought (which I forgot to act on) was to deprecate the whole class, but if you want I can remove it right away.

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
lipchev: 1. Wish it could, but that would make it impossible to implement in the `HowMuch` (besides, it's being extended by another public interface).
2. I don't now, the thing I was trying to convey with the name (unsuccessfully)  was that when using it as a constraint (e.g. for a collection of quantities)- all quantities would be _instances_ of the the `T` parameter. What about `IQuantityOfType`?

UnitsNet/GeneratedCode/Quantities/Temperature.g.cs
lipchev: > I think this could just as well be added as an extension method?

This shouldn't exist at all- it's currently only called from one place (note the comment):
``` csharp
    /// <inheritdoc cref="IQuantity.As(UnitKey)" />
    /// <remarks>This should be using UnitConverter.Default.ConvertValue(quantity, toUnit) </remarks>
    internal static double GetValue<TQuantity>(this TQuantity quantity, UnitKey toUnit)
        where TQuantity : IQuantity
    {
        return quantity.As(toUnit);
    }
```

> A lot of the existing methods could probably also be moved to extension methods, but let's not go overboard with potential breaking changes. Something to consider for v7 with the intent to reduce binary size.

Yes, everything that isn't a calculated property (until net10), or part of an interface (like `IFormattable`).

UnitsNet/IArithmeticQuantity.cs
lipchev: Do you mean to rename the `IArithmeticQuantity` interface (which would be a breaking change) or the `ILinearQuantity` interface?

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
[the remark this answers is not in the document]
lipchev: should be fixed in https://github.com/angularsen/UnitsNet/pull/1587/commits/2ea2f0bd57c30b7baeef31a5b35cce6569d14efc (if got the right blocks to format).

CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
```diff
@@ -658,7 +692,7 @@ public static bool TryParseUnit([NotNullWhen(true)]string? str, IFormatProvider?
 
         private void GenerateArithmeticOperators()
         {
-            if (!_quantity.GenerateArithmetic) return;
+            if (_quantity.IsAffine) return;
```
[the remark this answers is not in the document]
lipchev: fixed in https://github.com/angularsen/UnitsNet/pull/1587/commits/2ea2f0bd57c30b7baeef31a5b35cce6569d14efc#diff-d61a53679273c1f953c027ba478079004e1a62063d0678939234bb3cc9fd7f21

Technically we could auto-generate the code, if we had the correct units to use, but that is not critical given that there is just the `Temperature` that fits the bill.

CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
```diff
@@ -978,35 +978,104 @@ public void Equals_Null_ReturnsFalse()
...
+            Assert.True(quantity.Equals(quantity, maxTolerance));
+            Assert.True(quantity.Equals(otherQuantity, largerTolerance));
+            Assert.False(quantity.Equals(otherQuantity, smallerTolerance));
+            // note: it's currently not possible to test this due to the rounding error from (quantity - otherQuantity) 
+            // Assert.True(quantity.Equals(otherQuantity, maxTolerance));
         }}
 
         [Fact]
-        public void EqualsReturnsFalseOnTypeMismatch()
+        public void Equals_WithNegativeTolerance_DoesNotThrowArgumentOutOfRangeException()
```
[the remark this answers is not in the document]
lipchev: Fixed in https://github.com/angularsen/UnitsNet/pull/1587/commits/2ea2f0bd57c30b7baeef31a5b35cce6569d14efc

CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
```diff
@@ -978,35 +978,104 @@ public void Equals_Null_ReturnsFalse()
...
+            }
+            else // quantities with a linear scale
+            {
+                Writer.WL($@"
+
+        [Theory]
+        [InlineData(1, 2)]
+        [InlineData(100, 110)]
+        [InlineData(100, 90)]
+        public void Equals_WithTolerance_IsImplemented(double firstValue, double secondValue)
```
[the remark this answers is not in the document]
lipchev: fixed in https://github.com/angularsen/UnitsNet/pull/1587/commits/2ea2f0bd57c30b7baeef31a5b35cce6569d14efc

UnitsNet/CustomCode/Quantities/Temperature.extra.cs
```diff
@@ -58,6 +58,7 @@ public partial struct Temperature
         /// <param name="factor">Factor to multiply by.</param>
         /// <param name="unit">Unit to perform multiplication in.</param>
         /// <returns>The resulting <see cref="Temperature" />.</returns>
+        [Obsolete("Affine quantities, such as Temperature, cannot be multiplied directly: consider using the TemperatureDelta type instead.")]
```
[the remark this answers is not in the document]
lipchev: removed in https://github.com/angularsen/UnitsNet/pull/1587/commits/2ea2f0bd57c30b7baeef31a5b35cce6569d14efc

will try to create the PR for `maintenance/v5` some time tomorrow..

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
lipchev: fixed in https://github.com/angularsen/UnitsNet/pull/1587/commits/2ea2f0bd57c30b7baeef31a5b35cce6569d14efc (u

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
lipchev: It appears that github has been invaded by Unicorns! 🦄 so I'm unable to link to the files, but have a look at the `LinearQuantityExtensionTests`, `AffineQuantityExtensionTests` (and the other one) - there are a few tests (_testing equality with tolerance around the border_) that I had to `[Skip]` . Those are passing in the 🐲 PR, but would very hard (see impossible) to fix here.

UnitsNet/UnitMath.cs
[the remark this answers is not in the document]
angularsen: Ok, so with the new interfaces in v6 we probably don't need/want to support those generic math methods for custom quantities anymore. We can just obsolute them in v6 and remove later.

UnitsNet/Extensions/QuantityExtensions.cs
angularsen: Sounds good.

As long as `Sum()` and `Average()` still work for untyped `IQuantity`, without consumers having to know about linear vs affine vs logarithmic, I think we're good. 
If affine or logarithmic needs a different implementation or we can optimize in some way, we'll just delegate that internally.

UnitsNet/GeneratedCode/Quantities/Angle.g.cs
angularsen: I like `IQuantityOfType<T>`

UnitsNet/Extensions/AffineQuantityExtensions.cs
```diff
@@ -0,0 +1,187 @@
...
+        where TQuantity : IAffineQuantity<TQuantity, TOffset>, ISubtractionOperators<TQuantity, TQuantity, TOffset>
+        where TOffset : IQuantityInstance<TOffset>, IAdditiveIdentity<TOffset, TOffset>
+    {
+        if (other is not TQuantity otherInstance)
+        {
+            return false;
+        }
+
+        // // TODO see about this (I think the exception should take precedence to the null check, but the QuantityTests disagree)
+        // if (tolerance is not TOffset toleranceQuantity)
```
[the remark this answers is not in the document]
lipchev: I'm not sure I understand- `Equals(TSelf? other, TSelf tolerance)` this overload is a given, my 1) 2) 3) were alternatives for the _very generic overload_ `Equals(IQuantity? other, IQuantity tolerance)`. Do you mean you pick option 2) as in `(TQuantity, IQuantity, TQuantity)` for the non-affine and `(TQuantity, IQuantity, TOffset)` for the temperature (which is already the case).

Regarding 4) - so you mean you'd like me to make the EqualsAbsolute internal (it's being used by the two other overloads)?

UnitsNet/IArithmeticQuantity.cs
angularsen: Gotcha, no I meant renaming `IArithmeticQuantity` to `ILinearArithmeticQuantity` since it's now scoped to only linear quantities, where as before it covered all quantities.

Or is it just that we haven't implemented for affine/logarithmic yet, but this may come later?

angularsen:
LGTM and ready to merge, if you want to address any of the open comments before you merge just do so.

UnitsNet/Extensions/AffineQuantityExtensions.cs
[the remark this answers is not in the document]
angularsen: Okay I misread, `IQuantity` and `TQuantity` reads very similar.

1. For tolerance equality, I'd generally expect the fully typed `TQuantity` experience, for both `other` and `tolerance` to match `this`.
2. But I guess, why not also allow untyped `IQuantity` and do type checking internally. It may be useful in certain cases and it costs us little to add. The only thing I'm unsure of, is whether it can trip up intellisense, auto-complete and AI coding assistance. 
4. Rename `EqualsAbsolute` to `Equals`, if it works without issues, or make it `private` if it's only intended to be used by the other extension methods.

UnitsNet/Extensions/QuantityExtensions.cs
angularsen: > And am pretty sure we couldn't have it both ways, as the extensions would become ambiguous.

Huh yeah, I hadn't thought of that.

I guess, the only way to support extensions on both untyped and typed quantity interfaces are:
- Different signatures: usually not viable
- Different namespaces: please god no
- Different names, like `SumUntyped`: 🤮 

I'd rather throw untyped `IQuantity` extensions under the bus to get typed extensions if there is no obvious workaround to support both.

UnitsNet/UnitMath.cs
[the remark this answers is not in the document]
lipchev: Fixed in https://github.com/angularsen/UnitsNet/pull/1587/commits/a6fffc79e252d16816f452bed964fb09f6fd0aab : I've made the `GenericMathExtensions` obsolete (with some additional remarks), I'll open another issue/PR regarding the stuff that's remaining in the `UnitMath`.

UnitsNet/Extensions/AffineQuantityExtensions.cs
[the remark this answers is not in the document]
lipchev: fixed in https://github.com/angularsen/UnitsNet/pull/1587/commits/bdf04d54a62054c91fb6e8aac637d7cea5eaacbf - please double-check the signatures of the `Equals` extensions.

lipchev:
> LGTM and ready to merge, if you want to address any of the open comments before you merge just do so.

I've addressed most of the issues, with the exception of the `UnitMath` for which I'll open a separate issue/PR. Please double-check the signatures of the `Equals` extensions and if everything looks good, then merge away..

lipchev:
Not that anyone cares about size, but in case you're curious- here's the _before_ and _after_:

> Before (netstandard2.0):
> 2.46 MB (2 580 480 bytes)
> 
> After (netstandard2.0):
> 2.42 MB (2 544 128 bytes)
> 
> Before (net9.0):
> 2.55 MB (2 682 368 bytes)
> 
> After (net9.0):
> 2.49 MB (2 616 832 bytes)

UnitsNet/Extensions/AffineQuantityExtensions.cs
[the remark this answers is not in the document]
angularsen: Ref no 2, see comment:
https://github.com/angularsen/UnitsNet/commit/bdf04d54a62054c91fb6e8aac637d7cea5eaacbf#r163347229

angularsen:
I'll just merge this for now.

UnitsNet/Extensions/AffineQuantityExtensions.cs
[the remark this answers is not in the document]
lipchev: The link doesn't work, but I read it in the notification:
> public static bool Equals(this IQuantity quantity, IQuantity? other, IQuantity tolerance)

This is only going to work if only one of the extensions is visible- `(IQuantity, IQuantity, IQuantity)` would be ambiguous  with `(TQuantity, TQuantity, TQuantity)`..

PR #1589, opened 2025-07-30
Deprecating the Multiply/Divide methods for the Temperature

lipchev:
As discussed in #1587, these are supposed to go away in `v6`


lipchev:
@angularsen Should I also try to deal with the `Equals` method? It's part of the generated code, and if we were to introduce the overload (with `TemperatureDelta` as tolerance), we'd have to probably also handle it in the _very generic_ overload..

angularsen:
> @angularsen Should I also try to deal with the `Equals` method? It's part of the generated code, and if we were to introduce the overload (with `TemperatureDelta` as tolerance), we'd have to probably also handle it in the _very generic_ overload..

I don't have a good overview, but if you see a good way to obsolete equals methods in v5 where better overloads already exist in v5 and not only introduced in v6, then sure.

angularsen:
[Release UnitsNet/5.75.1 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.75.1)

PR #1592, opened 2025-08-03
Clean-up some commented-out bits from QuantityInfo and IQuantityInfo

lipchev:
- cleaned up some commented-out bits from `QuantityInfo` and `IQuantityInfo`
- extended the test for `XXX_QuantityInfo_ReturnsQuantityInfoDescribingQuantity` with the checks for the explicit interface implementations

### DIFF
--- a/UnitsNet/IQuantityInfo.cs
+++ b/UnitsNet/IQuantityInfo.cs
@@ -1,11 +1,7 @@
 ﻿// Licensed under MIT No Attribution, see LICENSE file at the root.
 // Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
 
-using System;
-using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
 using System.Resources;
-using UnitsNet.Units;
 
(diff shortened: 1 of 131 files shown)


UnitsNet/IQuantityInfo.cs
lipchev, line 1: In an earlier version of the `QuantityInfo`, I used to rely heavily on this interface- so I had it split-up the same way as the implementation, with every single property and overload possible. As things started to mature (and members got dropped from the `IQuantity` interface) this was no longer required. In fact this interface is now only used by `netstandard`, so I don't see a reason to not make it internal for `net8+`.

If you want to, we could exclude it completely from the build for `net8+`,  but then we'd have to duplicate the comments in the implementation, as there wouldn't be anything for the `<inheritdoc/>`.

angularsen: Why do we need `IQuantityInfo` for netstandard2.0 though? 
Can't we just use the abstract `QuantityInfo` and it's number of generic typed descendants?

As for inheritdoc, you can use `<inheritdoc cref="" />` syntax to inherit from any type/member so I wouldn't keep it around just for that. Also, duplication of xmldoc is not a big deal.

PR #1593, opened 2025-08-03
Making the methods from UnitMath that exist in System.Linq as [Obsolete]

lipchev:
Making the  `Min` / `Max` of an `IEnumerable` and their selector overloads `[Obsolete]` in `v5` (to be removed in `v6`).

### DIFF
--- a/UnitsNet/UnitMath.cs
+++ b/UnitsNet/UnitMath.cs
@@ -79,6 +79,7 @@ namespace UnitsNet
         /// <exception cref="ArgumentException">
         ///     <paramref name="source">source</paramref> contains quantity types different from <paramref name="unitType" />.
         /// </exception>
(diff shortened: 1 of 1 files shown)


lipchev:
@angularsen Note that extensions in `System.Linq` aren't directly applicable to a collection of `IQuantity` as it doesn't implement the `IComparable` interface (and even if it did- the result of sorting a mix of different quantities wouldn't be obvious).

If you agree that we don't want them any more, I can simply remove these in #1587 and we'll have less to talk about next time..

angularsen:
[Release UnitsNet/5.75.1 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F5.75.1)

PR #1594, opened 2025-08-03
Replaced the NotImplementedException which is thrown from the ToUnit method with UnitNotFoundException

lipchev:
- moved all exceptions into a new solution folder
- replaced the `NotImplementedException` which is thrown from the `ToUnit` method with `UnitNotFoundException`


lipchev:
@angularsen Well, technically this would be a breaking change for someone who's actually catching for the `NotImplementedException` - but given how not a single test failed when I changed it, I guess that's not a big deal.

PR #1596, opened 2025-08-03
Refactored test: From_ValueAndUnit_ReturnsQuantityWithSameValueAndUnit

lipchev:
Replaced the in-lined iteration with `Assert.All` and replaced the use of the named property with the `Value`


angularsen:
Nice improvement, but feel free to merge improvements to tests like these on your own

PR #1597, opened 2025-08-03
Making the QuantityInfoLookup public

lipchev:
- making the `QuantityInfoLookup` type public
- exposing the `Quantities` associated with a given `UnitParser` and `UnitAbbreviationsCache`
- exposing the `QuantityInfoLookup` property of the `UnitsNetSetup`

### DIFF
--- a/UnitsNet/CustomCode/UnitsNetSetup.cs
+++ b/UnitsNet/CustomCode/UnitsNetSetup.cs
@@ -1,11 +1,6 @@
 ﻿// Licensed under MIT No Attribution, see LICENSE file at the root.
 // Copyright 2013 Andreas Gullberg Larsen (andreas.larsen84@gmail.com). Maintained at https://github.com/angularsen/UnitsNet.
 
-using System;
-using System.Collections.Generic;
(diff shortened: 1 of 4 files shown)


lipchev:
@angularsen I don't see any reason to hold this one internal, but other than the _utility_ functions it provides to the user, there isn't any particular requirement for it to be public.

UnitsNet/CustomCode/UnitsNetSetup.cs
```diff
@@ -88,8 +83,5 @@ public UnitsNetSetup(IEnumerable<QuantityInfo> quantityInfos, UnitConverter unit
     /// <summary>
     ///     The quantities and units that are loaded.
     /// </summary>
-    /// <remarks>
-    ///     Access type is <c>internal</c> until this class is matured and ready for external use.
-    /// </remarks>
-    internal QuantityInfoLookup QuantityInfoLookup { get; }
```
lipchev, line 94: I haven't got any burning need to rename this, but it would perhaps make it more consistent if we used `Quantities` as the property name.

PR #1598, opened 2025-08-03
✨Add Equals() extension methods for untyped IQuantity

angularsen:
Make it easier to work with untyped `IQuantity`, and favor typed extensions methods when working with concrete types.

- Remove `Equals` overloads taking untyped `IQuantity other` from `LinearQuantityExtensions`, `AffineQuantityExtensions`, `LogarithmicQuantityExtensions`
- Remove unnecessary generic type parameters on linear/affine/logarithmic methods
- Add `QuantityExtensions.Equals()` with only untyped `IQuantity` parameters
- Move/copy relevant tests over to `QuantityExtensionsTests`

## TODO
10 test cases currently break due to having a generic implementation, need to delegate to the implementations for linear, affine and logarithmic.

image

### DIFF
--- a/UnitsNet/Extensions/QuantityExtensions.cs
+++ b/UnitsNet/Extensions/QuantityExtensions.cs
@@ -23,7 +23,7 @@ public static class QuantityExtensions
         if (tolerance is null) throw new ArgumentNullException(nameof(tolerance), "Tolerance cannot be null. Use a zero quantity instead.");
         if (other is null) return false;
 
-        // TODO Defer to implementations in LinearQuantityExtensions, AffineQuantityExtensions, LogdecimalQuantityExtensions, etc. Several tests in QuantityExtensionsTests break on this default implementation.
+        // TODO Delegate to implementations in LinearQuantityExtensions, AffineQuantityExtensions, LogdecimalQuantityExtensions, etc. Several tests in QuantityExtensionsTests break on this default implementation.
         UnitKey quantityUnit = quantity.UnitKey;
         return Comparison.EqualsAbsolute(quantity.Value, other.GetValue(quantityUnit), tolerance.GetValue(quantityUnit));
     }


angularsen:
@lipchev Here is an attempt at having untyped `IQuantity` extension methods, it doesn't seem to create any ambiguity errors.

There is a TODO to fix the implementation and delegate to the linear/affine/logarithmic equals, but due to them taking generic type params more work is needed and I ran out of time. If you like this direction, please help out making it work. I'll probably be less active the next few days again, work and life is resuming tomorrow.

UnitsNet/Extensions/QuantityExtensions.cs
```diff
@@ -10,14 +10,31 @@ namespace UnitsNet;
...
 public static class QuantityExtensions
 {
+    /// <summary>
+    ///
+    /// </summary>
+    /// <param name="quantity"></param>
+    /// <param name="other"></param>
+    /// <param name="tolerance"></param>
+    /// <returns></returns>
+    public static bool Equals(this IQuantity quantity, IQuantity? other, IQuantity tolerance)
```
lipchev, line 20: I'll have to double-check which cases are hitting this, but even if it works as expected - I really don't see how anyone could ever use this.. If you don't know what type of quantity you're working with, how can you select the appropriate tolerance?

I was hoping you'd pick option 3, instead you picked option -1 🤣

angularsen: OK you have me convinced.

There are consumers of the untyped `IQuantity`, so it just seemed to me like a natural thing to support equality checks for, since we do offer this interface. Mostly for completeness sake.

But I totally get your point that passing an untyped tolerance seems very fringe. A custom equality comparer to reuse for different quantities is the best I can come up with, but even then I'm sure a generic typed to either linear/affine/logarithmic would be better.

With the fraction PR and accurate equality, I suppose tolerance equality becomes less of a thing. 

I do think there are some improvements to the type parameters here, to have fewer of them, but it still can be reduced yet more. Let me give it another go and update this PR.

PR #1599, opened 2025-08-09
Moved the ExceptionHelper to the Exceptions folder and updated the Skip reason for the unknown unit tests

lipchev:
- moved the `ExceptionHelper` to the Exceptions folder
- replaced the Skip reason for the tests that cause a stack overflow


lipchev:
@angularsen I was hoping that simply un-skipping the tests would be enough (now that we're throwing the `UnitNotFoundException` instead of `NotImplementedException`) - but unfortunately that wasn't the case:
```
        var quantity = Length.FromMeters(1);
        var invalidQuantity = new Length(1, (LengthUnit)(-1));
        Assert.Throws<UnitNotFoundException>(() => invalidQuantity.Equals(quantity, quantity));
        Assert.Throws<UnitNotFoundException>(() => quantity.Equals(quantity, invalidQuantity));
```
The first assert works as expected, but the second one enters an infinite recursion. The issue is also present in `v5` (but not in the 🐲 PR).

angularsen:
Ok, as long as a fix is already in place in the bigger PR we'll ignore it for now

PR #1602, opened 2025-08-09
🧪Fix test on Norwegian culture, order by StringComparer.Ordinal

angularsen:
`aa` is sorted as `å` in Norwegian culture with the default comparer it seems, so explicitly specify a string comparer.


angularsen:
@lipchev FYI

PR #1611, opened 2025-08-19
making the TQuantity Create(double, UnitKey) function public 

lipchev:
making the `TQuantity Create(double, UnitKey)` function public (`IQuantityOfType` / `IQuantityInstanceInfo`)


lipchev:
@angularsen I tried to keep these internal, but now I think they are simply too useful: they would allow me (as a user) to have the same types of neat _single-generic-type-parameter-extensions_ like the ones we have for `Sum`/`Average`.

PR #1633, opened 2025-10-13
feat(Volume): add CubicAngstrom unit

Laendor:
- Extended UnitsNet.Volume with CubicAngstrom
- Updated tests and documentation accordingly


Common/UnitDefinitions/Volume.json
```diff
@@ -245,6 +245,21 @@
...
     },
+    {
+      "SingularName": "CubicAngstrom",
+      "PluralName": "CubicAngstroms",
+      "BaseUnits": {
+        "L": "Angstrom"
+      },
+      "FromUnitToBaseFunc": "{x} * 1E-30",
+      "FromBaseToUnitFunc": "{x} / 1E-30",
+      "Localization": [
```
angularsen, line 256: Looks good, but you are missing implementing the test cases for the new units, so the build currently fails.

See https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit

angularsen (changes requested):

PR #1644, opened 2026-01-14
Add NominalPower

upcu:
Based on [New Unit added: Nominal Power](https://github.com/angularsen/UnitsNet/pull/736) by @McGum


CodeGen/Generators/NanoFrameworkGen/QuantityGenerator.cs
```diff
@@ -167,67 +167,67 @@ private void GenerateStaticFactoryMethods()
         private void GenerateConversionMethods()
         {
             Writer.WL($@"
-                #region Conversion Methods
```
angularsen, line 170: Please don't change generator code unless intentional

Common/UnitDefinitions/NominalPower.json
```diff
@@ -0,0 +1,25 @@
+{
+  "Name": "NominalPower",
```
angularsen, line 2: Should we rename this to disambiguate this from other kinds of nominal power? 
It seems maybe a bit ambiguous: https://en.wikipedia.org/wiki/Nominal_power

For example, `NominalPowerPhotovoltaic`

angularsen (changes requested):
Thanks for the contribution! A couple of things need to be fixed in `NominalPower.json`:

### 1. Missing `BaseDimensions`

NominalPower is dimensionally equivalent to Power (W = kg·m²/s³). The quantity definition needs `BaseDimensions`, same as [Power.json](https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/Power.json#L5-L9):

```json
"BaseDimensions": {
  "L": 2,
  "M": 1,
  "T": -3
}
```

### 2. Missing `BaseUnits` on `WattPeak`

The WattPeak unit needs `BaseUnits` to map to SI base units, same as [Power.json's Watt](https://github.com/angularsen/UnitsNet/blob/master/Common/UnitDefinitions/Power.json#L14-L18):

```json
"BaseUnits": {
  "L": "Meter",
  "M": "Kilogram",
  "T": "Second"
}
```

Please add these and the PR should be good to go!

Issue #1643, opened 2026-01-05
Strategy to use new preview version of .NET nanoFramework nugets


josesimoes:
Hi @angularsen !

We now have (in preview) support for generics in nanoFramework.
Because this is still in preview, it would be nice to be able to have UnitsNet nuget previews using these to keep things separated and clear for developers.

Assuming you are OK with it, I wanted to discuss with you what would be the best approach for this.  Thank you.

josesimoes:
@angularsen pinging you about this.

josesimoes:
Spamming everyone... 😅  @angularsen @lipchev @tmilnthorp

lipchev:
My personal opinion is that `nanoFramework` is better suited for a completely separate branch, having it's own code-base / generator and build / release pipeline. Frankly, I don't see any intersections between the two projects other than the `json` files - which are public and available to anyone who wants to use them with their own code generator. 

I'm sorry, but other than building the project, I have currently no way of verifying that anything works as expected- let alone manage this in the build pipeline.

josesimoes:
With the introduction of generics from our end, apart from the need to reference different mscorlib nugets and sys.math, it will close the gap between both. As far as I understand the release of these it's a matter of some "magic" at build time. Andreas has kindly taking care of this until now. 

Building and publishing it's all we're asking from your end. No need to worry about tests and validation, which we can handle from our end.

My initial questions was motivated by the need of (ideally) being able to release these new nugets with preview.  
Last version is [5.75.0](https://www.nuget.org/packages/UnitsNet.nanoFramework.Temperature/5.75.0) which would require that a new versions would be released as 5.NN.0-preview.MM (or similar).

OK for me to submit a PR with the changes in the refs and you can make the build happen?

angularsen:
Hi all, sorry for neglecting to respond, just really busy.

@josesimoes Sure, just bring the PR, I'll get it deployed.
We're already on pre-release for v6, and nanoFramework packages are already prerelease versions: https://www.nuget.org/packages/UnitsNet.nanoFramework.Length/6.0.0-pre019

So in that sense, no better time than now to add preview dependencies to nanoFramework. Currently, UnitsNet and UnitsNet.nanoFramework* nugets share versioning. But this is primarily how I run some scripts to bump versions for me, we can easily separate those if need be to have completely separate versioning for them, similar to how we version the JSON nugets.

angularsen:
[...]

PR #1646, opened 2026-03-02
Updates on nanoFramework build components

josesimoes:
- Add optional parameter to choose preview builds (and nuget packages).
- Update nuspec (and code generator) to use offical nano1.0 Target Framework Moniker.
- Remove unnused declaration properties generator.
- Remove unnecessary property in reference project entry.
- Remove all checks and installer for nanoFramework VS extension, as all that is handled in AZDO step.
- Remove pre-check of nuget and msbuild as builds are now running in ADZO which include these in the VM image.
- Add input to install nanoFramework build components to allow specifying preview build. (settable from pipeline var UseNanoPreview)
- Not using nuget from tools folder anymore as ADZO includes this in the VM image.

Addresses  #1643

### DIFF
--- a/Build/build-functions.psm1
+++ b/Build/build-functions.psm1
@@ -51,19 +51,6 @@ function Start-Build {
 function Start-BuildNanoFramework {
   write-host -foreground blue "Start-BuildNanoFramework (MSBuild)...`n---"
 
-  # Check prerequisites
-  if (-not $msbuildx64 -or -not (Test-Path $msbuildx64)) {
-    write-host -foreground red "ERROR: Cannot build .NET nanoFramework - MSBuild not found."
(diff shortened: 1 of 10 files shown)


josesimoes:
To give this a bit more context: found the need to improve this as I was handling the update to use the preview versions with the generics support. Bumped on several places with notes about removing things when moving to AZDO (which has happened in the meantime) and the code to install the VS extension which was duplicating the existing step in AZDO yaml.

As I'm not _that_ familiar with the build system, I may have overlooked some things. Happy to fix. Also feel free to update the code directly, if that more efficient.

On the usage of a pipeline variable to control the use of the preview build, I thought that would be a nice approach to deal with it without having to touch the yaml and configs. Again, not sure if you're happy with it. Again, happy to change as you see fit.

josesimoes:
Following #1643 this is not need anymore.

PR #1648, opened 2026-03-31
Add Pressure.MilligramForcePerSquareMeter/Foot

TS-Andritz:
Adding two units used in our domain for small pressure values in two different systems of measurement (metric and US customary).


angularsen (changes requested):
Hi and thanks for contributing, although these units are probably a bit niche it seems useful enough to include.


## Use exact coefficients for MilligramForcePerSquareFoot

Per our [conversion function guidelines](https://github.com/angularsen/UnitsNet/wiki/Adding-a-New-Unit#conversion-function-conventions), prefer exact constituent constants over pre-computed decimals so the physical constants remain visible and no precision is lost.

The `MilligramForcePerSquareFoot` conversion uses a pre-computed decimal. 

Please change from:

```json
"FromUnitToBaseFunc": "{x} * 1.0555790208802639466e-4",
"FromBaseToUnitFunc": "{x} / 1.0555790208802639466e-4",
```

To:

```json
"FromUnitToBaseFunc": "{x} * 9.80665e-6 / 9.290304e-2",
"FromBaseToUnitFunc": "{x} / 9.80665e-6 * 9.290304e-2",
```

This keeps the exact definition constants visible (`9.80665` = standard gravity, `9.290304e-2` = ft² in m²) and avoids precision loss from pre-computation.


## Otherwise LGTM

Naming, localization, and test values all look correct.

TS-Andritz:
Hi, thanks for your reply.

I just made the change and pushed the commit.

Regarding the units, we were ourself a bit surprised by this combination of metric and imperial units, but it seems to be used by metalworking companies, specifically in the "US Customary" system of measurements.

angularsen:
[Release UnitsNet/6.0.0-pre020 · angularsen/UnitsNet](https://github.com/angularsen/UnitsNet/releases/tag/UnitsNet%2F6.0.0-pre020)

Issue #1067, opened 2022-03-20
Map and look up unit abbreviations by strings instead of enums


angularsen:
`QuantityType` enum is already deprecated and will be removed in #982 in order to better support third-party quantities and units.

The same could be done for the 100+ unit enums, like `LengthUnit`

Instead of
```cs
string abbreviation = UnitAbbreviationsCache.Default.GetDefaultAbbreviation(unitInfo.GetType(), (int)unitInfo);
```

We could support, without breaking change:
```cs
string abbreviation = UnitAbbreviationsCache.Default.GetDefaultAbbreviation("Length", "Millimeter");
```

Optionally, we could deprecate the enum-based overloads and remove them in #982 or future major version bump.

Related to discussion:
https://github.com/angularsen/UnitsNet/issues/1062#issuecomment-1073219229

bdbonazz:
@angularsen 
Do you think this way of implementing the solution is acceptable?

https://github.com/bdbonazz/UnitsNet/commit/daa79425453dffefa40930506b20938337fe13bd

If so, I'll add the necessary tests and comments, and submit a PR.

angularsen:
Hi, I think the proposed implementation is heavier than necessary.

  ### What already exists

  The building blocks for string-based lookup are already in place:

  - `QuantityInfoLookup.GetUnitByName("Length", "Meter")` → returns `UnitInfo`
  - `UnitInfo` has a `.UnitKey` property
  - `UnitAbbreviationsCache.GetDefaultAbbreviation(UnitKey)` already works

  So today you can already do:

  ```csharp
  var unitInfo = Quantity.ByName["Length"].UnitInfos.First(u => u.Name == "Meter");
  var abbreviation = UnitAbbreviationsCache.Default.GetDefaultAbbreviation(unitInfo.UnitKey);
  ```

  ### My suggestion

  I haven't verified this works, but something like this should. 
  Rather than generating a `GetUnitKey(string)` switch method on all 130+ quantity types, we can add a simple convenience overload on `UnitAbbreviationsCache` that delegates to the existing infrastructure:

  ```csharp
  public string GetDefaultAbbreviation(string quantityName, string unitName, IFormatProvider? formatProvider = null)
  {
      var unitInfo = Quantity.DefaultProvider.GetUnitByName(quantityName, unitName);
      return GetDefaultAbbreviation(unitInfo.UnitKey, formatProvider);
  }
  ```

[...]

PR #1649, opened 2026-04-06
Add IQuantity.UnitInfo convenience property

angularsen:
## Summary
- Adds `IQuantity.UnitInfo` default interface member on .NET 5+ targets, so consumers can write `quantity.UnitInfo` instead of `quantity.QuantityInfo[quantity.UnitKey]`
- Adds typed `IQuantity<TUnitType>.UnitInfo` returning `UnitInfo<TUnitType>` on .NET 5+
- Adds `GetUnitInfo()` and `GetUnitInfo<TUnit>()` extension methods on `netstandard2.0` for the same functionality

## Motivation
Getting `UnitInfo` from a quantity instance currently requires the verbose `quantity.QuantityInfo[quantity.UnitKey]`. This is a common operation, especially when chaining with APIs like `UnitAbbreviationsCache` that benefit from having `UnitInfo` directly (see #1067).

### TFM fragmentation
The extension methods are gated behind `#if !NET` so that .NET 5+ consumers only see the cleaner property syntax in IntelliSense. Library authors who multi-target can use `quantity.QuantityInfo[quantity.UnitKey]` which works on all targets, or write a one-line wrapper. If this turns out to be a real pain point, removing the `#if` guard is a non-breaking one-line change.

### Extension method tests
The test project targets `net8.0;net9.0;net10.0` only, so `#if !NET` code is never compiled in tests. The extension methods are trivial one-liner wrappers over the same indexer call that the default interface members use, so they are effectively covered by the property tests.

## Test plan
- [x] `dotnet build UnitsNet/UnitsNet.csproj` — 0 errors across all 4 target frameworks
- [x] New tests pass on net8.0, net9.0, net10.0:
  - `UnitInfo_ReturnsUnitInfoForQuantityUnit` — non-base unit via `IQuantity`
  - `UnitInfo_Zero_ReturnsBaseUnitInfo` — `.Zero` instance
  - `UnitInfo_TypedQuantity_ReturnsTypedUnitInfo` — typed `IQuantity<LengthUnit>`
  - `UnitInfo_MatchesUnit` — `Assert.All` across all quantities

### DIFF
--- a/UnitsNet.Tests/CustomCode/IQuantityTests.cs
+++ b/UnitsNet.Tests/CustomCode/IQuantityTests.cs
@@ -60,6 +60,28 @@ public partial class IQuantityTests
     }
 
     [Fact]
+    public void UnitInfo_ReturnsUnitInfoForQuantityUnit()
+    {
+        var length = new Length(3.0, LengthUnit.Centimeter);
+        IQuantity quantity = length;
+
+        UnitInfo unitInfo = quantity.UnitInfo;
+
+        Assert.Equal(nameof(LengthUnit.Centimeter), unitInfo.Name);
+        Assert.Equal(quantity.UnitKey, unitInfo.UnitKey);
+    }
+
+    [Fact]
+    public void UnitInfo_ReturnsBaseUnitInfo_WhenDefaultConstructed()
+    {
+        IQuantity quantity = new Length(1.0, Length.BaseUnit);
+
+        UnitInfo unitInfo = quantity.UnitInfo;
+
+        Assert.Equal(Length.Info.BaseUnitInfo.UnitKey, unitInfo.UnitKey);
+    }
+
+    [Fact]
     public void ToUnit_UnitSystem_ThrowsArgumentExceptionIfNotSupported()
     {
         var unsupportedUnitSystem = new UnitSystem(new BaseUnits(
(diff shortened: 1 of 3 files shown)


lipchev:
@angularsen I though you wanted a lean interface.. Introducing two interface members that are nothing more than getters from an existing property (which **is explicitly implemented**) feels like going in the direction of having a `Dimensions` property on the interface.

My second argument against such a property on the interface, is that this implies that _if my extended `IQuantity` implements the `UnitInfo` properties in some other way, something will behave differently_. This wouldn't be the case- all types such as the `UnitAbbreviationsCache` the `UnitConverter` etc, are all oblivious of the existence of such a property (and in fact _expect_ that the `UnitInfo` is inseparable from the `QuantityInfo` that was used in the constructor).

My last argument against this (although I haven't verified it) is that by not implementing the property on the structs, using something like `Mass.Zero.UnitInfo` is likely going to incur an allocation cost (because of the interface dispatch of the struct).

lipchev:
@angularsen Frankly, I would have probably removed the `QuantityInfo` instance members as well, were they not already part of the v5 interface. Their presence implies that we actually keep the object as an instance member, which we don't.

PS That's not a _crazy idea_, but a lot needs to be done if we _want to get there_ (not counting the serialization).

angularsen:
Yeah I agree. I just noticed `QuantityInfo` property was already sitting there and this felt like a natural convenience, but in hindsight I think extension methods is better overall and to also deprecate `QuantityInfo` as an instance member of `IQuantity`, and instead maybe promote it to a static abstract interface member for .NET5+ with some back compat to netstandard2.0.

I reverted the new interface members and reused extension methods here as a first step:
https://github.com/angularsen/UnitsNet/pull/1657

Issue #158, opened 2016-04-11
FeetInches always shows inches


MusicMonkey5555:
When creating a FeetInches object if you pass it a whole value like 4ft it turns into 3ft 12 inch. It would be nice if there was an option to tell it how to handle this.

anjdreas:
Hi, sorry for the delay. Sounds like a rounding problem to me?
I ran the following in LinqPad on UnitsNet 3.33.0:
```
Length.FromFeet(4).FeetInches
// Feet	4
// Inches	7.1054273576010019E-15

Length.FromFeet(4).FeetInches.ToString() // 4 ft 0 in

new FeetInches(4, 0)
// Feet	4
// Inches	0

new FeetInches(4, 0).ToString() // 4 ft 0 in

```

So I can't reproduce, however I do see that there is a very small negative value on Inches, so without proper rounding it would give 3 ft 12 in. Can you paste me an example code that produces this error?

anjdreas:
Did that example make sense to you?

anjdreas:
Closing, will reopen if there are any further discussions.

PR #1656, opened 2026-04-21
fix handling of rounding to 12 inches in FeetInches

Applesauce314:
updates the string outputs of `FeetInches` to propagate the rounding of inches up to feet if the inches are rounded to 12.
i.e 3.9999 ft in feet inches is 3 feet 11.9988 inches.
`ToString` will round the 11.9988 to 12 and output 3 ft 12 in, which is incorrect.
`ToArchitecturalString` will do the same if the denominator is is not small enough to result is a fractional part that is not rounded to 1. 

this PR updates these methods to propagate the rounding of these functions up to the feet part while resetting the inches to 0.

i.e. `3 ft 11.9988 in` becomes `4 ft 0 in` instead of `3 ft 12 in`.

the current behavior is unexpected for a mixed radix system, and this PR implements the expected default behavior.
(for an example of a similar situation, if rounding  09:59:50 to the nearest minute, you would get 10:00, not 09:60)

this issue had been previously discussed in #158, but does not provide a way to configure the behavior as mentioned there.


Applesauce314:
it looks like this pipeline failure is unrelated to my change, anything i need to do?

angularsen:
Needs test cases per bot review.

Applesauce314:
been super busy, will get those unit tests added as soon as i have the time

Applesauce314:
@angularsen 
I think this plus unit tests is ready to go, i don't know what the pipeline failure is, looks like it is probably a transient failure.

angularsen:
@claude review this

angularsen:
Seems the review caught a bug, plus some minor stuff. 

1. Create a test case for the missing sign it found.
2. The warning about compile doesn't seem like a problem, PR pipeline ran tests just fine.
3. Look over the minor issues in review and consider fixing

angularsen (changes requested):
Please see bot review and my comment for changes needed.

UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -233,12 +277,19 @@ int GreatestCommonDivisor(int a, int b)
 
             inchPart.Append('"');
 
-            if (Feet == 0)
+            if (feet == 0)
             {
                 return inchPart.ToString();
             }
```
Applesauce314, line 288:
```suggestion
{
if (isNegative)
{
    //negate inches so we output the correct sign. 
    inchPart = -inchPart;
}
return inchPart.ToString();
}
```

UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -190,17 +217,33 @@ public string ToArchitecturalString(int fractionDenominator)
...
+            //if negative value we record this and invert the values, at the end we add a negative sign as necessary, but all the calculations are done positive so rounding behavior is the same.
+            var isNegative = Feet < 0 || Inches < 0;
+            if (isNegative)
+            {
+                feet = -feet;
+                inches = -inches;
+            }
+
+
+            var inchTrunc = (int)Math.Truncate(inches);
```
Applesauce314, line 227:
```suggestion
            }

            var inchTrunc = (int)Math.Truncate(inches);
```

Applesauce314 (commented):
fixing errors

UnitsNet/CustomCode/Quantities/Length.extra.cs
```diff
@@ -46,7 +44,8 @@ public static Length FromFeetInches(QuantityValue feet, QuantityValue inches)
         /// <returns>Parsed length.</returns>
         public static Length ParseFeetInches(string str, IFormatProvider? formatProvider = null)
         {
-            if (str == null) throw new ArgumentNullException(nameof(str));
+            if (str == null)
+                throw new ArgumentNullException(nameof(str));
```
Applesauce314, line 48:
```suggestion
            if (str == null) throw new ArgumentNullException(nameof(str));
```

Applesauce314:
when fixing i discovered there is a behavior difference of `feet = -feet` on netcore vs netframework for the value 0 (newer runtimes support negative 0 older ones do not), causing an issue with how `ToString` worked as well.
added a bunch more tests, let see how this goes.

Applesauce314:
give me a minute to fix this guy, apparently some errors from merging and trying to use the github interface as an IDE...

Applesauce314:
should be good to go now.
added a bunch of test cases, fixed my build errors (don't try to use github as an IDE :/ )

PR #1657, opened 2026-05-01
Revert IQuantity.UnitInfo DIM, expose GetUnitInfo() extension on all TFMs

angularsen:
## Summary

Reverts the default interface members added in #1649 and exposes `GetUnitInfo()` as extension methods on all target frameworks instead. The typed overload returns the most specific `UnitInfo<TQuantity, TUnit>` when overload resolution can infer both type parameters from the receiver.

## Rationale

Per @lipchev's [review feedback on #1649](https://github.com/angularsen/UnitsNet/pull/1649#issuecomment-4236027376):

- **Lean interface.** The DIM is just a getter over `QuantityInfo[UnitKey]` — no new state, sets a precedent for further interface bloat (`Dimensions` etc.).
- **Semantic trap.** A custom `IQuantity` implementor overriding `UnitInfo` would silently diverge from internal lookups (`UnitConverter`, `UnitAbbreviationsCache`, `QuantityFormatter`) which all go through `QuantityInfo`, not the new property.
- **Boxing on structs.** Calling `Mass.Zero.UnitInfo` through `IQuantity` on a struct quantity boxes the receiver per call; concrete quantities don't override the property so the DIM path is the only one available.

The extension method gives the same call-site ergonomics (`quantity.GetUnitInfo()`) on every TFM with none of these issues.

## Overload resolution

| Caller has         | Resolved overload | Return type |
|--------------------|-------------------|-------------|
| `Length q;`           | `GetUnitInfo<TQuantity, TUnit>(IQuantity<TQuantity, TUnit>)` | `UnitInfo<Length, LengthUnit>` |
| `IQuantity q;`        | `GetUnitInfo(IQuantity)` | `UnitInfo` |
| `IQuantity<TUnit> q;` | `GetUnitInfo(IQuantity)` (fallback) | `UnitInfo` |

The `IQuantity<TUnit>` reference case loses its typed return — `IQuantity<TUnit>` does not satisfy the `IQuantity<TQuantity, TUnit>` constraint, so resolution falls back to the non-generic overload. That scenario is uncommon compared to the concretely-typed receiver case, which is now strictly better than before (returns `UnitInfo<TQuantity, TUnit>` instead of `UnitInfo<TUnit>`).

## Follow-up

PR #1658 stacks on top of this branch: adds `static abstract Info` on `IQuantityOfType<T>` / `IQuantity<TSelf, TUnitType>`, obsoletes the instance `QuantityInfo` member, and provides matching `GetQuantityInfo()` extensions. Out of scope for this PR.

### DIFF
--- a/UnitsNet.Tests/CustomCode/IQuantityTests.cs
+++ b/UnitsNet.Tests/CustomCode/IQuantityTests.cs
@@ -60,44 +60,44 @@ public partial class IQuantityTests
     }
 
     [Fact]
-    public void UnitInfo_ReturnsUnitInfoForQuantityUnit()
+    public void GetUnitInfo_ReturnsUnitInfoForQuantityUnit()
     {
         var length = new Length(3.0, LengthUnit.Centimeter);
         IQuantity quantity = length;
 
-        UnitInfo unitInfo = quantity.UnitInfo;
+        UnitInfo unitInfo = quantity.GetUnitInfo();
 
         Assert.Equal(nameof(LengthUnit.Centimeter), unitInfo.Name);
         Assert.Equal(quantity.UnitKey, unitInfo.UnitKey);
     }
 
     [Fact]
-    public void UnitInfo_Zero_ReturnsBaseUnitInfo()
+    public void GetUnitInfo_Zero_ReturnsBaseUnitInfo()
     {
         IQuantity quantity = Length.Info.Zero;
 
-        UnitInfo unitInfo = quantity.UnitInfo;
+        UnitInfo unitInfo = quantity.GetUnitInfo();
 
         Assert.Equal(Length.Info.BaseUnitInfo.UnitKey, unitInfo.UnitKey);
     }
 
     [Fact]
-    public void UnitInfo_TypedQuantity_ReturnsTypedUnitInfo()
+    public void GetUnitInfo_TypedQuantity_ReturnsTypedUnitInfo()
     {
         IQuantity<LengthUnit> quantity = new Length(3.0, LengthUnit.Centimeter);
 
-        UnitInfo<LengthUnit> unitInfo = quantity.UnitInfo;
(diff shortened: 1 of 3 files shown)


lipchev:
@angularsen This is surely better than having them on the interface, but here are my thoughts:
- the generic version uses `IQuantity<TUnit>` resulting in boxing when used with something like `Mass.Zero` (while still producing `TUnit`-number of concrete implementations)
- I haven't tested it, but I'm pretty sure both extension members would show up in the `Mass.Zero.` auto-completion list
- We could potentially ship these (and others like them) in separate nuggets, like the number extensions- using something like 'CSharp14' as a suffix where they could live as regular extension properties (note, I'm 99% certain, those can still target `netstandard2.0`)

Bottom line- I don't mind the addition, but if I were to ever need to use one of these properties in my own projects- I'd probably write my own `extension` block.

angularsen:
Autocomplete shows both in Rider, but I think that is fine:

image

It favors the typed one:
image


Regarding boxing, this only applies when calling `Mass.Zero.GetUnitInfo()`, and this is not something I expect anyone to do in a hot path so I really think that's fine.

At the end of the day, these are convenience additions that also improve on discoverability - but at least now it's reusing the same global configuration of custom quantities instead of whatever the instance implementation decides to deliver. I don't think this is worth shoving into a separate nuget, since it adds very little bloat.

lipchev:
I'm fine either way, but still can't see any good uses for having a convenience method for the `UnitInfo<TUnit>`  - if it were the `UnitInfo<TQuantity, TUnit>` I guess I could see it being used as an instance factory or something, but this just feel like another set of _breaking changes to avoid in the future_.

angularsen:
I changed it to return `UnitInfo<TQuantity, TUnit>` instead of `UnitInfo<TUnit>`.

PR #1658, opened 2026-05-01
Add static abstract Info and obsolete instance QuantityInfo on .NET 5+

angularsen:
## Summary

Builds on top of #1657 (revert of the `IQuantity.UnitInfo` DIM) with the bigger ergonomics + perf shift @lipchev called for in his review. **Base this PR on #1657 and rebase once it merges.**

- **Adds a static abstract `Info` member** on `IQuantityOfType<TQuantity>` and `IQuantity<TSelf, TUnitType>` (under `#if NET`).
- **Marks the existing instance `QuantityInfo` property as `[Obsolete]`** on .NET 5+ across all three interfaces (`IQuantity`, `IQuantity<TUnitType>`, `IQuantity<TSelf, TUnitType>`). Kept un-obsoleted on netstandard2.0 to preserve back-compat there.
- **Adds `GetQuantityInfo()` extension methods** (non-generic + typed) on `QuantityExtensions` so callers have a single discoverable API on every TFM.
- **Migrates internal callers** of the obsolete instance member to either `TSelf.Info` (where the generic constraint allows) or `quantity.GetQuantityInfo()` (where it doesn't).

## Why

The instance `QuantityInfo` property invariably returns a per-type static value. Exposing it as an instance member:

1. Implies it can vary per instance, which it cannot.
2. Costs interface dispatch and boxes the receiver on struct quantities for every call.
3. Encourages custom `IQuantity` implementors to override it in ways that silently diverge from the rest of the library, which only ever consults `QuantityInfo` indirectly through the static `TSelf.Info` (via UnitsNetSetup, etc.).

The static abstract member lets generic algorithms reach the info with `TSelf.Info` directly — no boxing, no virtual call. The extension method covers the common `IQuantity` / `IQuantity<TUnit>` reference case via a `UnitsNetSetup.Default.Quantities` lookup.

## Implementation notes

- **Generated quantities already expose** `public static QuantityInfo<TSelf, TUnitType> Info { get; }`. This directly satisfies the typed static abstract — no codegen change required.
- **The non-generic `IQuantityOfType<TSelf>.Info`** is satisfied by a default static implementation in the typed interface:
  ```csharp
  static QuantityInfo IQuantityOfType<TSelf>.Info => TSelf.Info;
  ```
- **Bridges** like `QuantityInfo IQuantity.QuantityInfo => QuantityInfo;` inside the interface itself are wrapped in `#pragma warning disable CS0618` so the bridge implementation does not warn about consuming the obsolete member it bridges.
- **Callers that must work for custom quantities not registered in `UnitsNetSetup.Default`** (JsonNet serialization, the debugger proxy, `QuantityTypeConverter`) keep using the instance member with a `#pragma warning disable CS0618` and a comment explaining why. The lookup-based extension method would throw `QuantityNotFoundException` for these scenarios.
- **`HowMuch` test custom quantity** had `public static readonly QuantityInfo Info = ...` (a field). Static abstract property requires a property — changed to `public static QuantityInfo Info { get; }`.

## Migration path

| Caller has | .NET 5+ | netstandard2.0 |
|------------|---------|----------------|
| `IQuantity` | `q.GetQuantityInfo()` | `q.GetQuantityInfo()` (delegates to instance) |
| `IQuantity<TUnit>` | `q.GetQuantityInfo()` (typed return) | `q.GetQuantityInfo()` |
| `TQuantity : IQuantityOfType<T>` | `TQuantity.Info` (no boxing) | `q.GetQuantityInfo()` |
| `TSelf : IQuantity<TSelf, TUnit>` | `TSelf.Info` (typed, no boxing) | `q.QuantityInfo` |

The instance member stays warning-only this release. Promote to error / remove once netstandard2.0 is dropped — possibly backport the obsolete to v5 if we decide to fully remove in v6.

## Tests

- `GetQuantityInfo_NonGeneric_ReturnsRegisteredQuantityInfo` — `Same(Mass.Info, ((IQuantity)mass).GetQuantityInfo())`
- `GetQuantityInfo_Typed_ReturnsRegisteredQuantityInfo` — typed variant returns `QuantityInfo<MassUnit>`
- `StaticAbstract_Info_ReturnsSameAsTypedInfo` — `TSelf.Info` via `IQuantity<TSelf, TUnit>` static abstract
- `StaticAbstract_Info_NonGeneric_ReturnsSameAsTypedInfo` — `TQuantity.Info` via `IQuantityOfType<TQuantity>` static abstract (covariant return through default impl bridge)

## Test plan

- [x] `dotnet build UnitsNet.slnx` — 0 errors across all TFMs
- [x] `dotnet test UnitsNet.Tests -f net8.0` — 42303 passed (4 new), 0 failed
- [x] `dotnet test UnitsNet.Serialization.JsonNet.Tests -f net8.0` — 113 passed, 0 failed
- [x] No `IQuantity.QuantityInfo` CS0618 warnings remain in `UnitsNet/`, `UnitsNet.Serialization.JsonNet/`, `UnitsNet.Benchmark/`, `Samples/` (suppressions are explicit + commented)

## Out of scope

- Migrating the test project `.QuantityInfo` call sites (suppressed at csproj level with `<NoWarn>CS0618</NoWarn>`).
- A struct-friendly `GetQuantityInfo<TSelf, TUnit>(this TSelf q) where TSelf : IQuantity<TSelf, TUnit>` overload that would JIT-specialize per struct. Easy to add later if a perf-sensitive use case shows up.

### DIFF
--- a/Samples/MvvmSample.Wpf/MvvmSample.Wpf/Converters/UnitToStringConverter.cs
+++ b/Samples/MvvmSample.Wpf/MvvmSample.Wpf/Converters/UnitToStringConverter.cs
@@ -31,7 +31,7 @@ namespace WpfMVVMSample.Converters
             if (!(value is IQuantity quantity))
                 throw new ArgumentException("Expected value of type UnitsNet.IQuantity.", nameof(value));
 
-            Enum unitEnumValue = _settings.GetDefaultUnit(quantity.QuantityInfo.UnitType);
+            Enum unitEnumValue = _settings.GetDefaultUnit(quantity.GetQuantityInfo().UnitType);
             int significantDigits = _settings.SignificantDigits;
 
             IQuantity quantityInUnit = quantity.ToUnit(unitEnumValue);
--- a/UnitsNet.Benchmark/Conversions/ToUnit/QuantityConversionBenchmarks.cs
+++ b/UnitsNet.Benchmark/Conversions/ToUnit/QuantityConversionBenchmarks.cs
@@ -19,7 +19,7 @@ public class QuantityConversionBenchmarks
         double result = 0;
         foreach (IQuantity quantity in Quantities)
         {
-            foreach (UnitInfo unitInfo in quantity.QuantityInfo.UnitInfos)
+            foreach (UnitInfo unitInfo in quantity.GetQuantityInfo().UnitInfos)
             {
                 result = quantity.As(unitInfo.Value);
             }
--- a/UnitsNet.Serialization.JsonNet/AbbreviatedUnitsConverter.cs
+++ b/UnitsNet.Serialization.JsonNet/AbbreviatedUnitsConverter.cs
@@ -99,7 +99,9 @@ namespace UnitsNet.Serialization.JsonNet
         /// <returns>The string representation associated with the given quantity</returns>
         protected string GetQuantityType(IQuantity quantity)
         {
+#pragma warning disable CS0618 // IQuantity.QuantityInfo: serialization must work for custom quantities not registered in UnitsNetSetup.Default.
             return _quantities[quantity.QuantityInfo.Name].Name;
+#pragma warning restore CS0618
         }
 
         /// <inheritdoc />
--- a/UnitsNet.Serialization.JsonNet/UnitsNetBaseJsonConverter.cs
+++ b/UnitsNet.Serialization.JsonNet/UnitsNetBaseJsonConverter.cs
@@ -164,7 +164,9 @@ namespace UnitsNet.Serialization.JsonNet
         {
             quantity = quantity ?? throw new ArgumentNullException(nameof(quantity));
 
+#pragma warning disable CS0618 // IQuantity.QuantityInfo: serialization must work for custom quantities not registered in UnitsNetSetup.Default.
             return new ValueUnit {Value = (double)quantity.Value, Unit = $"{quantity.QuantityInfo.UnitType.Name}.{quantity.Unit}"};
+#pragma warning restore CS0618
         }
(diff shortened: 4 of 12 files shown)


lipchev:
I'm not saying this is the "right path forward" (I honestly don't know) - but back when I was "rationalizing the existence" of the `QuantityInfo.From` methods (other than for the netstandard2.0 use-case) I was toying with the idea of actually storing the `QuantityInfo` as an instance member of the quantity itself (I'm sure you've thought of it as well). 

This is similar in spirit to the `MoneyContext` used by `Node.Money`. 

Obviously, such a property couldn't go into the standard json format (the two sides would need to use compatible quantity-configurations).

angularsen:
These changes are straight from the bot, but I just wanted to get your eyes on the direction.

I am also pondering the removal of the `IQuantity.QuantityInfo` instance member.

On one hand, it feels more correct to enforce a single QuantityInfo instance per quantity type. I don't see a good use case for two Mass values having different QuantityInfo instances. It's just simpler.

On the other hand, instance members work with netstandard and a reference pointer is cheap enough.

What are the benefits you see of keeping it an instance member?

lipchev:
1. No breaking changes (that was main motivator at the time)
2. Adding custom units / customizing existing ones - right now this is still in the _hacky_ stage, but _technically possible_. The downside is that unless you do it globally (on the default config), you will get in trouble when using something like `myCustomMass + Mass.Zero` (when `myCustomMass` is using a `UnitKey` that isn't in the default config).

angularsen:
Superseded by #1675 (branch agl/static-info).

PR #1662, opened 2026-05-04
Unify generic arithmetic interfaces across quantity types

Muximize:
## Motivation

The current quantity interface hierarchy duplicates generic-math contracts and couples arithmetic
capabilities to interfaces that also carry the unit enum type.

This results in verbose generic constraints such as:

```csharp
where TQuantity : ILinearQuantity<TQuantity>, IQuantity<TQuantity, TUnit>
```

Linear and logarithmic quantities also declare many of the same arithmetic operator interfaces
independently, making the hierarchy harder to understand and extend.

This PR separates the arithmetic capability from unit typing and composes the more specific quantity
interfaces from those reusable contracts.

## Changes

- Introduce `IArithmeticQuantity<TSelf>` as the unit-independent arithmetic capability.
- Make `IArithmeticQuantity<TSelf, TUnit>` combine:
  - `IQuantity<TSelf, TUnit>`
  - `IArithmeticQuantity<TSelf>`
- Introduce `ILinearQuantity<TSelf, TUnit>` as the strongly typed linear quantity contract.
- Reuse `IArithmeticQuantity<TSelf>` from both linear and logarithmic quantities.
- Move affine addition/subtraction contracts to the unit-independent
  `IAffineQuantity<TSelf, TOffset>` interface.
- Generate linear quantities as `ILinearQuantity<TSelf, TUnit>`.
- Simplify `Sum()` and `Average()` constraints to:

```csharp
where TQuantity : ILinearQuantity<TQuantity, TUnit>
```

- Regenerate affected quantity declarations.

## Result

Generic APIs can express the intended quantity capability with one constraint:

```csharp
static TQuantity Sum<TQuantity, TUnit>(
    IEnumerable<TQuantity> quantities,
    TUnit unit)
    where TQuantity : ILinearQuantity<TQuantity, TUnit>
    where TUnit : struct, Enum
{
    return quantities.Sum(unit);
}
```

The hierarchy also gives linear and logarithmic quantities a common arithmetic capability without
requiring a unit enum when the generic algorithm does not use units.

This is intended as an interface-organization change, not a behavior change. Existing arithmetic
interfaces remain transitively implemented by the generated quantity types.

## Compatibility considerations

This changes the declared interface hierarchy of generated quantity structs, so the public API and
binary compatibility implications should be reviewed carefully even though assignability through
the existing base interfaces is preserved.


angularsen:
@claude Review idea in PR description and current code change

Muximize:
@angularsen If it helps for reviewing I can split up this PR, keeping the simplifying here and make a separate PR for adding `IComparisonOperators`.

lipchev:
As I was refactoring the interfaces last time, I actually did pretty much the same thing - but in the end decided to revert it due to the following consideration: the `INumber` interface (even our partial implementation) is quite large. While I personally would probably always go all-the-way in implementing all possible interfaces in my custom quantities, this maybe more of a headache for others - who just care about the unit conversion / type safety.

Muximize:
I removed the `IComparisonOperators` stuff from this PR, so it is now purely about simplifying `ILinearQuantity`, `ILogarithmicQuantity` and `IAffineQuantity`, removing 1 layer of inheritance.

Muximize:
@lipchev Yes indeed, moving interfaces up will make things harder for custom quantities. I was struggling a bit with `HowMuch`, but I don't use them myself. What is you usecase for custom quantities?

angularsen:
@claude review

UnitsNet/Extensions/LogarithmicQuantityExtensions.cs
```diff
@@ -228,7 +228,7 @@ public static TQuantity Sum<TQuantity, TUnit>(this IEnumerable<TQuantity> quanti
     ///     logarithmic space.
     /// </remarks>
     public static TQuantity Sum<TSource, TQuantity, TUnit>(this IEnumerable<TSource> source, Func<TSource, TQuantity> selector, TUnit targetUnit)
-        where TQuantity : ILogarithmicQuantity<TQuantity, TUnit>
+        where TQuantity : ILogarithmicQuantity<TQuantity>, IQuantity<TQuantity, TUnit>
```
angularsen: I suggest keeping the two-parameter quantity interfaces, if viable.

I think this is simpler for consumers: they can choose between the unit-agnostic and unit-specific interfaces, instead of repeating both constraints in their own APIs:

```csharp
where TQuantity : ILogarithmicQuantity<TQuantity>, IQuantity<TQuantity, TUnit>
```

A composite interface captures that relationship directly:

```csharp
public interface ILogarithmicQuantity<TSelf, TUnit> :
    IQuantity<TSelf, TUnit>,
    ILogarithmicQuantity<TSelf>
    where TSelf : ILogarithmicQuantity<TSelf, TUnit>
    where TUnit : struct, Enum
{
}
```

This matches the existing shape of the quantity interfaces (before the PR):

```csharp
IArithmeticQuantity<TSelf, TUnit>
IAffineQuantity<TSelf, TUnit, TOffset>
ILogarithmicQuantity<TSelf, TUnit>
```

To me, this represents a useful concept: “a logarithmic quantity with a known unit enum”, while avoiding repeated multi-constraint combinations in public APIs.

angularsen (changes requested):

UnitsNet/Extensions/LogarithmicQuantityExtensions.cs
angularsen: Actually, this goes for all the interfaces; affine, logarithmic, arithmetic. 

Can we preserve the two-parameter composite interfaces to avoid consumers having to specify multiple generic constraints?

angularsen: I pushed fix that I think works, take a look if you agree. I tried to minimize diff in PR also, we can refactor later.

angularsen:
I merged this to unblock myself, but let me know if you wanted any changes to this.

Muximize:
Nice, this is an improvement by itself 👍 

There is a small cleanup missing that was in my original PR. I made a new one for that: #1694

PR #1674, opened 2026-06-03
Add structural engineering unit relations

JustASquid:
## Motivation

Some generated unit relations that are useful in structural engineering were missing. Adding them makes it possible to express common derived relationships directly with UnitsNet quantities instead of manually converting through raw numeric values.

## Changes

- Add relation definitions for structural-engineering workflows, including density/length to area density, area/area to area moment of inertia, specific weight/volume to force, and acceleration/area density to pressure.
- Regenerate the affected relation operators and tests.

## Validation

- GitHub PR Build: passing.
- Azure UnitsNet PR: passing.
- Local maintainer check: merged with current `master`, reran code generation with no extra diff, and ran the affected relation test classes on net9 successfully.
- The only red check is Claude code review, which should not block this PR.


angularsen (approved):
This looks good to me. The added relations are dimensionally consistent, useful for structural engineering, and fit the existing relation-generation pattern.

From my side this is ready to merge.

Issue #849, opened 2020-10-14
Unit for "Amount of times"


selmaohneh:
**Is your feature request related to a problem? Please describe.**
I have a property of type `IQuantity` that users of my framework can fill as they desire. Often the value represents a technical number with a corresponding unit. 
But sometimes the value should represents a simple count of how often something was done. This quantity would be unit-less although being numerical.
Therefore the users have no way to generate the quantity, since the unit is a required argument of i.e. `Quantity.From(...)`.

**Describe the solution you'd like**
A `QuantityType` called "None" would fit my needs. It should contain a single unit called "Times", e.g. [1].

angularsen:
Can you make do with Ratio or Frequency?

selmaohneh:
Frequency does not work, since it has a time component in it, i.e. times per second.
Ratio would work, since it is unit-less. Still it is not that intuitive, since a ratio implies a relationship between **two** numbers. 

`Temperature.From...`, `Pressure.From...` and similar methods force the user to think about what type of quantity they want to create - that's good API design. It's easy, concise and leads the user to use UnitsNet correctly (pit of success). Telling the user to use `Ratio` for a "count"/"number of times" is misleading and not self-explaining (e.g. I had to open this issue first :D).

In my opinion the QunatityType "None" with unit "Times" would therefore be a good enhancement, even if just for API purposes.

What do you think?

angularsen:
You are not the first to ask, and the answer has typically been to use `double`, `Ratio` or `Frequency`.
Is there a particular usecase where having this quantity would be more useful to you than `double` or `Ratio`?

Since people keep asking about it, I suppose we could add it.
How about something like this:
- Quantity `Amount` with units:
	- `Count` (ideally `long`, but probably `decimal` like `Information`)
	- Thousand
	- Million
    - Billion
	- etc..

```c#
Amount.FromMillons(1.5).Count; // 1.5e6
Amount.FromCount(1500).Thousands; // 1.5
```

selmaohneh:
I am taking the self-explanatory API position again:

var numberOfSheep = Amount.FromCount(42);

reads very good. This would not be the case when the word Ratio is somewhere in that line... So I'd say you should go for the new type and unit.

I guess .FromCount is enough. The other creation methods might lead to missusages. C# already has enough syntax sugar for creating big numbers, i.e. 42e6.

[...]

PR #1680, opened 2026-07-17
Remove Scalar quantity

angularsen:
## Motivation

`Scalar` was added in #889 for #849 as a way to wrap plain user-entered counts or values in an `IQuantity` shape. The historical use case was essentially "double/float/etc in an `IQuantity` format", not a dimensional quantity with real units.

Recent discussion in #1679 re-raised whether `Scalar` should exist at all. Looking at the current codebase, `Scalar` has not grown meaningful semantic usage: it only appears as its own generated quantity, generated number extensions, generated tests, and aggregate registration entries. It is not used by quantity relations or handwritten domain code.

Removing it avoids keeping a unitless value wrapper in the quantity model where APIs such as `scalar.As(ScalarUnit.Amount)` and `Scalar / Scalar -> double` are awkward. For callers that need plain numeric values, `double`/`decimal`/`QuantityValue` are clearer; for dimensionless physical relationships, existing quantities such as `Ratio` remain available when they are semantically appropriate.

## Changes

- Remove `Common/UnitDefinitions/Scalar.json`.
- Remove generated `Scalar`, `ScalarUnit`, resource, tests, and number-extension artifacts.
- Regenerate aggregate quantity registration and `IQuantity` test lists without `Scalar`.
- Remove the stale `Scalar` entry from `Common/UnitEnumValues.g.json`.

## Breaking change

This removes the public `Scalar` quantity and `ScalarUnit.Amount`, including factory/accessor APIs such as `Scalar.FromAmount(...)`, `.Amount`, parsing/formatting APIs, and number extensions such as `2.Amount()` / `2.Amount`.

Marked obsolete in v5: #1681 

## Validation

- `rg -n "\bScalar\b" Common UnitsNet UnitsNet.Tests UnitsNet.NumberExtensions UnitsNet.NumberExtensions.Tests UnitsNet.NumberExtensions.CS14 UnitsNet.NumberExtensions.CS14.Tests`
- `dotnet build UnitsNet.slnx`
- `dotnet test UnitsNet.Tests --no-build`
- `dotnet test UnitsNet.NumberExtensions.Tests --no-build`
- `dotnet test UnitsNet.NumberExtensions.CS14.Tests --no-build`

Refs #849, #1200, #1679.


angularsen:
@Muximize @lipchev Would like a second opinion here. 

Scalar is largely unused in our code base, and as discussed in https://github.com/angularsen/UnitsNet/pull/1679#pullrequestreview-4723302162 I don't believe Scalar is an improvement over `double` in the UnitsNet API for scalar results.

I'd rather move towards `QuantityValue` as part of #1544 if it helps improve precision beyond what `double` can do.

However, looking at the original motivation for adding Scalar https://github.com/angularsen/UnitsNet/issues/849, I can see some minor convenience in offering it and it may just cause more harm removing it than keeping it.

Thoughts?

Muximize:
I've read the motivations in #849 and I don't find them very convincing.

> I have a property of type IQuantity

But if the value is not always a quantity, why was that type chosen? It looks like a design problem in that framework which was worked around by adding a misfeature to UnitsNet.

I'm in favor of removing `Scalar`, also to avoid confusion about whether to use it where `double` etc. would suffice (I guess I was a victim of that in #1679 😅).

angularsen:
Thanks, yes I also lean towards removing it because it semantically isn't a quantity and adds confusion rather than value.
Also, anyone can just implement their own Scalar if actually needed.

angularsen:
Another use case for Scalar: https://github.com/angularsen/UnitsNet/discussions/1678

Applesauce314:
I use `Scalar` and would have issues if it disappears
i have types with a quantity column, the quantity can be any type, feet, inches, gallons, pounds, etc.
but it can also be "Each" or count.
i.e. 
- 3ft of wire - `Length`
- 1 oz glue - `Mass`
- 3 (EA) wire lugs - `Scalar`

without the scalar type, this property can no longer be a units net `IQuantity` as it is now, as I will have to revert to having untyped data, or reimplement the Scalar myself.

this pattern is common in Bill of Material scenarios.
i think there has been no usage of the type in unitsnet code because its a simple guy, its only purpose is like the above and it requires no modifications otherwise.

PR #1685, opened 2026-07-20
Add ounce-force torque units

MH-ZShearer:
Includes the units:
- ounce-force inches (ozf·in)
- ounce-force feet (ozf-ft)

I wasn't sure if I should have included the abbreviated unit abbreviations. More specifically, lbf·ft is the technical abbreviation for pound-force foot, but lb·ft is a common shorthand, along with most non-Newton based torque units.

On the other hand, oz·in and g·mm are the conventional units of `static unbalance` in rotor/propeller balancing. From a brief search, static unbalance seems to be a rather niche industry and may not qualify for Units.NET's requirements. It would prevent possible conflicts for someone making a custom quantity.


angularsen (changes requested):
Thanks, I think these units are reasonable for `Torque`.

Let's keep the primary abbreviations as `ozf·in` and `ozf·ft`, since those are the technically precise force-unit forms. Add the common shorthand aliases as secondary abbreviations. I found `oz·in`, `oz-in`, and `in-oz` used for small torque, and `oz·ft`/`oz-ft` used in torque converters. I did not find enough support for `ft-oz`, so I would leave that out for now.

I suggest this shape:

`OunceForceFoot`:

```json
"Abbreviations": [ "ozf·ft", "oz·ft", "ozf-ft", "oz-ft" ]
```

`OunceForceInch`:

```json
"Abbreviations": [ "ozf·in", "oz·in", "ozf-in", "oz-in", "in-oz" ]
```

Since this PR already touches `Torque`, let's also add the same common aliases for the existing pound-force torque units:

`PoundForceFoot`:

```json
"Abbreviations": [ "lbf·ft", "lb·ft", "lbf-ft", "lb-ft", "ft-lb" ]
```

`PoundForceInch`:

```json
"Abbreviations": [ "lbf·in", "lb·in", "lbf-in", "lb-in", "in-lb" ]
```

`ft-lb` could also be added to `EnergyUnit.FootPound` later, but I think that is out of scope for this PR.

One request before merging: please update the custom test constants to use values checked against a verifiable source, so the tests don't just mirror the conversion formula. The adding-unit docs call this out here: https://github.com/angularsen/UnitsNet/blob/master/Docs/adding-a-new-unit.md#4-fix-generated-test-stubs-to-resolve-compile-errors

NIST SP 811 Appendix B.9 lists `1 ozf·in = 7.061552e-3 N·m`, so for example:

```csharp
// NIST SP 811 Appendix B.9: 1 ozf·in = 7.061552e-3 N·m.
protected override double OunceForceInchesInOneNewtonMeter => 141.611928935735;

// Derived from the same NIST factor and 12 in/ft.
protected override double OunceForceFeetInOneNewtonMeter => 11.800994077978;
```

MH-ZShearer:
Not sure how the commits got all messed up, but I updated my branch to correctly match `master` now. I'm also not sure how I calculated those incorrect conversion factors, but I admittedly didn't spend enough time searching.

I'll get the conversion factor and additional unit abbreviations added tomorrow. I double checked our internal data, and it looks like we've only ever performed 4 ozf-in calibrations, 0 ozf-ft calibrations, and only about 500-600 torque calibrations in total across the last decade. Granted, we're first and foremost a force calibration laboratory, but apparently some of our customers have a need for those small loads.

MH-ZShearer:
I did some backtracking and it looks like the information on Wikipedia may be more precise than the information provided in NIST's SP 811. While SP 811 is likely sufficient for almost all cases, I believe there is room to improve. More specifically, I don't believe SP 811 to be wrong, per se, but I believe the values are rounded for display purposes.

Source | lbf·ft | N·m
---|---|---
[Wikipedia](https://en.wikipedia.org/wiki/Pound-foot) | 1 | 1.355 817 948 331 4004
[NIST SP 811](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication811e2008.pdf) | 1 | 1.355 818

Ultimately, this is your call and will update the coefficients as you see fit. If you'd prefer, I can do what I admittedly should have done originally and done the math by hand using known, exact values. Or another option is I can try and reach out to one of my NIST contacts to get their exact values. Fair warning: it may take a while to hear back (government and all that).

angularsen:
You’re right, the values in the SP 811 conversion table are rounded for presentation. SP 811 footnote 23 gives the exact pound-force conversion as 1 lbf = 4.4482216152605 N, and the international foot is exactly 0.3048 m.

Using those definitions, the conversion formulas and your existing test values are correct. Let’s keep them as-is.
 
The abbreviation aliases from my earlier review still need to be added, followed by regenerating the code.

PR #1694, opened 2026-07-23
Simplify IArithmeticQuantity and put ILinearQuantity in its own file

Muximize:


angularsen (changes requested):
Moving `ILinearQuantity` to its own file is good, but I think `IArithmeticQuantity<TSelf, TUnitType>` was accidentally removed.

I prefer keeping both `IArithmeticQuantity` interfaces on all target frameworks, even though they are empty on `netstandard2.0`. UnitsNet interfaces should not appear or disappear depending on the target framework. The previous shape handles this by making only the `System.Numerics` inheritance conditional.

Can you restore `IArithmeticQuantity<TSelf, TUnitType>` and keep the arithmetic inheritance from `ILinearQuantity` and `ILogarithmicQuantity` unconditional? Then this remains a straightforward file move without changing the public interface hierarchy introduced in #1662.

PR #1697, opened 2026-07-24
Make unit relation codegen ordering culture-invariant too + small improvements

Muximize:


angularsen:
@claude please review the latest changes in this PR.

angularsen (changes requested):
Thanks, the conflict is fixed now and the generated diffs look like pure reordering of existing relation operators. 

I agree with keeping relation ordering deterministic, but I don't quite understand the current `Ordinal` -> `OrdinalIgnoreCase` change. The PR title says culture-invariant ordering, but `StringComparison.Ordinal` was already culture-invariant. `OrdinalIgnoreCase` changes case sensitivity, not culture sensitivity.

1. Can we keep this as `StringComparison.Ordinal` or was there a specific reason relation ordering should ignore casing?

2. What is the reason for removing separators in `SortString`? It can theoretically make two different relations produce the same key.

Issue #797, opened 2020-05-28
Remove unnecessary format strings


tmilnthorp:
Remove:
- "V" and "v"
- "S" and "s".

Instead of "v", use `IQuantity.Value.ToString()` to get the value string.
Instead of "s", use "f2", "n4" or any of the standard numeric formats.

See https://github.com/angularsen/UnitsNet/wiki/String-Formatting

angularsen:
Fixed in v5 branch.

PR #1721, opened 2026-08-01
Standardize quantity formatting on .NET numeric formats

angularsen:
## Motivation

UnitsNet v6 has already removed the proprietary `U`, `V`, and `Q` quantity format strings. This completes that cleanup by removing the remaining `A` and `S` formats, so quantity format parameters consistently accept .NET numeric format strings and can be described accurately with `StringSyntaxAttribute.NumericFormat`.

The historical rationale supports the cleanup:

- `A` only returns unit abbreviations, which are already available explicitly through generated `GetAbbreviation()` APIs and the configurable `UnitAbbreviationsCache`.
- #797 intended to remove `S`, and commit 721287d removed its documentation and tests, but the formatter implementation remained in v5. This PR completes that cleanup. The discussion in #1450 also noted that `S` can hide precision and that callers should choose an explicit numeric format.

## Changes

- Remove `A`/`An` and `S`/`Sn` formatting from `QuantityFormatter` and `QuantityValue`.
- Throw focused `FormatException` messages for those formerly recognized formats with migration guidance.
- Keep standard and custom .NET numeric formats, such as `G3`, `F2`, `N2`, `E2`, and `0.##`.
- Annotate generated quantity formatting, `QuantityFormatter`, `QuantityValue`, quantity extensions, and type-converter formatting with `StringSyntaxAttribute.NumericFormat`.
- Add an internal `StringSyntaxAttribute` compatibility definition for the `netstandard2.0` target.
- Regenerate quantities and tests, and update the string-formatting guide and v6 upgrade guide.

## Migration

| Removed | Replacement |
|---|---|
| `A`, `A0`, `A1`, ... | `Length.GetAbbreviation(unit)` or `UnitAbbreviationsCache.GetUnitAbbreviations(unit)` |
| `S`, `S2`, ... | An explicit standard/custom numeric format such as `G3`, `F2`, `N2`, `E2`, or `0.##` |
| `U` | `quantity.Unit` |
| `V` | `quantity.Value` |
| `Q` | Static metadata such as `Length.Info.Name` |

Currency (`C`) and percent (`P`) formats remain intentionally unsupported for physical quantities.

Refs #1200.

## Validation

- `generate-code.bat`
- `dotnet build UnitsNet/UnitsNet.csproj --no-restore` (`netstandard2.0`, `net8.0`, `net9.0`, `net10.0`)
- `dotnet test UnitsNet.Tests --no-restore -f net10.0` (52,506 passed, 16 skipped)
- `dotnet test UnitsNet.Tests --no-restore -f net48` (45,321 passed, 16 skipped)

### DIFF
--- a/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/QuantityGenerator.cs
@@ -1032,7 +1032,9 @@ namespace UnitsNet
         /// <summary>
         /// Gets the string representation of this instance in the specified format string using the specified format provider, or <see cref=""CultureInfo.CurrentCulture"" /> if null.
         /// </summary>
-        public string ToString(string? format, IFormatProvider? provider)
+        public string ToString(
+            [StringSyntax(StringSyntaxAttribute.NumericFormat)] string? format,
+            IFormatProvider? provider)
         {{
             return QuantityFormatter.Default.Format(this, format, provider);
         }}
--- a/CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
+++ b/CodeGen/Generators/UnitsNetGen/UnitTestBaseClassGenerator.cs
@@ -1255,26 +1255,6 @@ namespace UnitsNet.Tests
             Writer.WL($@"
         }}
 
-        [Fact]
-        public void ToString_SFormat_FormatsNumberWithGivenDigitsAfterRadixForCurrentCulture()
-        {{
-            var _ = new CultureScope(CultureInfo.InvariantCulture);
-            Assert.Equal(""0.1{_baseUnitEnglishAbbreviation}"", new {_quantity.Name}(0.123456, {_baseUnitFullName}).ToString(""s1""));
-            Assert.Equal(""0.12{_baseUnitEnglishAbbreviation}"", new {_quantity.Name}(0.123456, {_baseUnitFullName}).ToString(""s2""));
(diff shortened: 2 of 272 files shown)


lipchev:
@angularsen I'm not against the removal, but I want to bring your attention to the fact that the current `v5` still has those extra formats:

https://github.com/angularsen/UnitsNet/blob/e3b6adfd8cd77185787a1197ca5ea734ebed3515/UnitsNet/QuantityFormatter.cs#L132

I don't remember if they were removed in `v6` or we just talked about it..

Anyway, given how recent the the `pre21` is - I doubt anyone has had the time to start using `.Value.ToString("S")` - we could simply let the `QuantityValue` format fallback to the generic exception (`_ => ..`).
Or if you want to be safe, we could do it before the release of the official version.

Finally, I just want to point out, that removing the `Ax` format doesn't mean we shouldn't address the ordering of the abbreviations of the same unit- those are still *not stable*.

FYI I just updated my projects from the forked branch to the `pre21` release, and everything looks good. 🥳

angularsen:
> @angularsen I'm not against the removal, but I want to bring your attention to the fact that the current `v5` still has those extra formats:

Huh, yeah you're right. The intent was to remove them, but clearly that slipped.

We should still remove them, I'll update the PR description.
v6 is still yolo, and anyone trying to use `S` or `A` formats will now get an explanatory exception.

> Finally, I just want to point out, that removing the `Ax` format doesn't mean we shouldn't address the ordering of the abbreviations of the same unit- those are still _not stable_.

Ok, good to know. I suspect very few rely on the `Ax` indexing though, and we don't really guarantee the ordering will never change.
 
> FYI I just updated my projects from the forked branch to the `pre21` release, and everything looks good. 🥳

Very nice!

Issue #1569, opened 2025-07-16
Calculations with db are wrong


BernhardJosef:
**Describe the bug**
The result of adding 2 levels is wrong
The result of subtracting 2 levels is wrong

**To Reproduce**
Level level1 = new Level(1.6, Level.ParseUnit("dB"));
Level level2 = new Level(0.7, Level.ParseUnit("dB"));

Level level3 = level1 - level2;
Level level4 = level1 + level2;

**Expected behavior**
db additions or subtractions should be executed without any transformations

**Additional context**
see https://sengpielaudio.com/calculator-db.htm

angularsen:
Hi, thanks for reporting. 

If you are able to, could you please create a pull request with a unit test that highlights the error and what the result should be? Huge bonus points if you also provide a fix for the implementation. 

It will help move this along much faster, I have limited time to look into it.

angularsen:
I think the core issue here is that `Level + Level` and `Level - Level` currently encode one specific interpretation of dB arithmetic, but users can reasonably expect another.

Today, logarithmic quantities implement `+`/`-` as linear-space aggregation:

```csharp
Level total = level1 + level2;
```

This means roughly:

```csharp
10 * log10(10^(level1 / 10) + 10^(level2 / 10))
```

So this does not produce direct value arithmetic:

```csharp
1.6 dB + 0.7 dB != 2.3 dB
```

That behavior is useful when adding quantities represented in dB, but it is surprising when the user wants to add/subtract dB values as gains/ratios.

For v6, I think we should make this explicit and avoid `+`/`-` for logarithmic quantities. My current preference is to provide named APIs for the two meanings:

```csharp
// Convert to linear space, operate, convert back.
Level total = level1.LinearAdd(level2);
Level total = levels.LinearSum();
Level mean = levels.LinearMean();

// Operate directly on the logarithmic values.
Level gain = level1.LogValueAdd(level2);
Level delta = level1.LogValueSubtract(level2);
Level gain = Level.FromDecibels(level1.Decibels + level2.Decibels);
```

The exact names can be discussed, but I think the important part is this split:

- `Linear*` means convert to linear space, do the operation, convert back.
- `LogValue*` means operate directly on the logarithmic unit values.
- `+`/`-` should probably be removed for logarithmic quantities, because both interpretations are common enough that the operators are ambiguous.

How does this match your expectation?

BernhardJosef:
[...]

PR #1725, opened 2026-08-06
Pin the logarithmic subtraction edge cases for Level

Rafael-SOWNet:
## Motivation

While reproducing #1569 I found that `Level`'s logarithmic operators are exercised by
`LevelTestsBase.ArithmeticOperators` for the ordinary case, but not where the linear result
stops being positive. Two such cases return silently today:

```
Level.FromDecibels(40) - Level.FromDecibels(40)    ->  -Infinity dB
Level.FromDecibels(0.7) - Level.FromDecibels(1.6)  ->  NaN dB
```

The second is inconsistent with the rest of the quantity. `Level(double quantity, double
reference)` rejects a non-positive ratio with `ArgumentOutOfRangeException` and the message
*"The base-10 logarithm of a number ≤ 0 is undefined"*, while `operator-` performs that same
logarithm and yields `NaN` without signalling anything.

## Changes

Three tests in `UnitsNet.Tests/CustomCode/LevelTests.cs`:

- `LogarithmicSubtraction_OfEqualLevels_ReturnsNegativeInfinity`
- `LogarithmicSubtraction_WhenSubtrahendIsLarger_ReturnsNaN`
- `LogarithmicAddition_OfTwoLevels_CombinesThemInLinearSpace`, using the values from #1569 and
  asserting against `10 * log10(10^0.16 + 10^0.07)` rather than a bare literal, so the test
  states why the answer is what it is.

## What this does not do

**It changes no behaviour, and none of these tests fail without a source change.** They are
characterisation tests: they pin what the operators do today so that any later change to it is
a visible decision rather than a silent one. I have deliberately not made `operator-` throw —
that is a breaking change and your call, not something to slip in under a test PR. If you would
like it to throw, say so and I will open that separately with these tests updated.

Nothing is asserted about `AmplitudeRatio` or `PowerRatio`, which have the same logarithmic
machinery with a scaling factor of 20 and 10 respectively. I stopped at `Level` because that is
what #1569 is about; happy to extend if useful.

## Relationship to the v6 plan for logarithmic operators

I missed this when opening the PR. In https://github.com/angularsen/UnitsNet/issues/1569#issuecomment-3140306604
the plan for v6 is to **remove `+`/`-` on logarithmic quantities** in favour of named
`Linear*` and `LogValue*` methods. That splits these three tests in two:

- `LogarithmicSubtraction_OfEqualLevels_ReturnsNegativeInfinity` and
  `LogarithmicSubtraction_WhenSubtrahendIsLarger_ReturnsNaN` are **not** affected by the
  spelling. A linear-space subtraction has to answer for a non-positive linear result however
  it is named, and today it answers silently while the constructor rejects the same logarithm
  with `ArgumentOutOfRangeException`. These carry over to `LinearSubtract` unchanged.
- `LogarithmicAddition_OfTwoLevels_CombinesThemInLinearSpace` **does** pin operator behaviour
  that is slated for removal. It is a record of today rather than something to preserve. Say
  the word and I will drop it, or rewrite it against whichever named API lands.

## On #1569 itself

The arithmetic is not wrong, though "not a bug" was too strong of me — the maintainer had
already identified a real design problem there before I commented. `Level` is
logarithmic, so `+` combines two levels in linear power space; `1.6 dB + 0.7 dB` is
`4.18357203248652 dB`, which matches `10·log10(10^0.16 + 10^0.07)` exactly. The reporter wanted
plain arithmetic on the decibel numbers, which is the right operation for a gain rather than a
level, and already works via `Level.FromDecibels(a.Decibels + b.Decibels)`.

## Validation

```
dotnet test UnitsNet.Tests/UnitsNet.Tests.csproj -f net10.0 --filter "FullyQualifiedName~LevelTests"
Passed!  - Failed: 0, Passed: 82, Skipped: 0, Total: 82
```

Local run on .NET 10 only; I have not run the full multi-target matrix.

### DIFF
--- a/UnitsNet.Tests/CustomCode/LevelTests.cs
+++ b/UnitsNet.Tests/CustomCode/LevelTests.cs
@@ -41,5 +41,41 @@ namespace UnitsNet.Tests
             // reference can't be zero or less than zero if quantity is postive.
             Assert.Throws<ArgumentOutOfRangeException>(() => new Level(quantity, reference));
         }
+
+        [Fact]
+        public void LogarithmicAddition_OfTwoLevels_CombinesThemInLinearSpace()
+        {
+            // The values from https://github.com/angularsen/UnitsNet/issues/1569, which read
+            // 4.18 dB as wrong on the assumption that adding levels adds their decibel numbers.
+            // Adding two levels combines them in linear power space, so the result is
+            // 10 * log10(10^0.16 + 10^0.07).
+            Level sum = Level.FromDecibels(1.6) + Level.FromDecibels(0.7);
+
+            var expected = 10 * Math.Log10(Math.Pow(10, 0.16) + Math.Pow(10, 0.07));
+            AssertEx.EqualTolerance(expected, sum.Decibels, DecibelsTolerance);
+            AssertEx.EqualTolerance(4.18357203248652, sum.Decibels, DecibelsTolerance);
+        }
+
+        [Fact]
+        public void LogarithmicSubtraction_OfEqualLevels_ReturnsNegativeInfinity()
+        {
+            // Removing a level from itself leaves no power at all, and log10(0) is -infinity.
+            // Pinned because it is returned silently rather than signalled.
+            Level v = Level.FromDecibels(40);
+
+            Assert.Equal(double.NegativeInfinity, (double)(v - v).Decibels);
+        }
+
+        [Fact]
+        public void LogarithmicSubtraction_WhenSubtrahendIsLarger_ReturnsNaN()
+        {
+            // The linear difference is negative here, and log10 of a negative number is undefined.
+            // The constructor rejects that case with ArgumentOutOfRangeException; this operator
+            // returns NaN silently instead. Pinned as current behaviour, not endorsed as correct.
+            Level smaller = Level.FromDecibels(0.7);
+            Level larger = Level.FromDecibels(1.6);
+
+            Assert.Equal(double.NaN, (double)(smaller - larger).Decibels);
+        }
     }
 }


Rafael-SOWNet:
CI evidence, since this PR shows no checks (first-time-contributor gate).

Full matrix run on my fork: https://github.com/Rafael-SOWNet/UnitsNet/actions/runs/31136648228 — **success**.

```
Passed! - Failed: 0, Passed: 52563, Skipped: 16, Total: 52579   UnitsNet.Tests
Passed! - Failed: 0, Passed:  1731, Skipped:  0, Total:  1731   UnitsNet.NumberExtensions.Tests
Passed! - Failed: 0, Passed:  1731, Skipped:  0, Total:  1731   UnitsNet.NumberExtensions.CS14.Tests
Passed! - Failed: 0, Passed:   213, Skipped:  0, Total:   213   UnitsNet.Serialization.JsonNet.Tests
Passed! - Failed: 0, Passed:   241, Skipped:  0, Total:   241   UnitsNet.Serialization.SystemTextJson.Tests
```

The same workflow on the same base without this PR's commit passes **52560** in `UnitsNet.Tests`
(https://github.com/Rafael-SOWNet/UnitsNet/actions/runs/31136388644), so the delta is exactly the
three tests added here and nothing else moved.

The run is from `ci-check/level-logarithmic`, which is this PR's commit plus one extra commit
touching only `.github/workflows/ci.yml` — that is #1727, needed because the dispatch inputs are
currently broken and dispatching is the only way to get a matrix run while checks are gated. The
diff between that branch and this PR's head is that workflow file alone; `UnitsNet/` and
`UnitsNet.Tests/` are byte-identical.

